PR #1, opened 2015-07-08
removed spring-xd-dirt dependencies

dturanski:
This removes direct dependencies on spring-xd-dirt and provides the minimum classses needed for the tests to pass. Classes copied from dirt are kept in o.s.xd.dirt... to be easily identified. Most of these are in spring-xd-runner. A couple are in spring-cloud-streams to avoid circular dependencies.  A modified codec.xml is provided which removes the TupleCodec. This could be added back to support Tuple payloads, currently not used in the tests. This would require a dependency on spring-xd-tuple. Codec config will need to be redone soon anyway. Also, dirt provides a number of ModuleOptionsMetadata classes needed to use common modules that are now missing. Module options is another area to revisit.


spring-cloud-streams/src/main/resources/META-INF/spring-bus/codec.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+	   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:util="http://www.springframework.org/schema/util"
```
markfisher: nitpick, but I don't think we need the util namespace for this file

pom.xml
```diff
@@ -10,12 +10,13 @@
...
-      <spring-boot.version>1.3.0.BUILD-SNAPSHOT</spring-boot.version>
-      <spring-cloud.version>Brixton.BUILD-SNAPSHOT</spring-cloud.version>
-      <spring-xd.version>2.0.0.BUILD-SNAPSHOT</spring-xd.version>
-    </properties>
+	<properties>
+		<java.version>1.7</java.version>
+		<spring-boot.version>1.3.0.BUILD-SNAPSHOT</spring-boot.version>
+		<spring-cloud.version>Brixton.BUILD-SNAPSHOT</spring-cloud.version>
+		<spring-xd.version>2.0.0.BUILD-SNAPSHOT</spring-xd.version>
+		<spring.framework.version>4.2.0.RC2</spring.framework.version>
```
markfisher: why not use spring-framework (instead of the dot delimiter) for consistency with the others?

spring-cloud-streams/src/main/java/org/springframework/cloud/streams/config/ChannelBindingAdapterConfiguration.java
```diff
@@ -49,7 +49,7 @@
  *
  */
 @Configuration
-@ImportResource("classpath*:/META-INF/spring-xd/bus/codec.xml")
+@ImportResource("classpath*:/META-INF/spring-bus/codec.xml")
```
markfisher: note for later: we should change the spring-bus directory name here

spring-xd-runner/pom.xml
```diff
@@ -8,7 +8,7 @@
 	<version>1.0.0.BUILD-SNAPSHOT</version>
 	<packaging>jar</packaging>
 
-	<name>spring-xd-runner</name>
+	<name>spring-xd-module-runner</name>
```
markfisher: should we change the name of the containing directory to match?

pom.xml
dturanski: ok

spring-xd-runner/src/main/java/org/springframework/xd/dirt/modules/metadata/LogSinkOptionsMetadata.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.xd.module.options.spi.ModulePlaceholders;
+
+
+/**
+ * Captures options for the {@code log} sink module.
+ * 
+ * @author Eric Bottard
+ * @author Gary Russell
+ */
+public class LogSinkOptionsMetadata {
```
markfisher: would it be possible to include these optionsmetadata classes only within the sample modules? (either way, they will disappear once we have pure boot config property support in place)

spring-xd-runner/src/main/java/org/springframework/xd/dirt/plugins/job/JobPluginMetadataResolver.java
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * A {@link ModuleOptionsMetadataResolver} that will dynamically add the module options (such as {@code makeUnique})
+ * supported by job modules.
+ *
+ * @see JobPlugin
+ * @author Eric Bottard
+ * @author Ilayaperumal Gopinathan
+ */
+public class JobPluginMetadataResolver implements ModuleOptionsMetadataResolver {
```
markfisher: interested in what @mminella might have to add, but we could probably do without any job-related metadata support at this point

spring-cloud-streams/src/main/java/org/springframework/cloud/streams/config/ChannelBindingAdapterConfiguration.java
dturanski: I think the XML config goes away. It's been removed from XD already.  The Rabbit and Redis MB configurations are XML that have <constructor-arg ref="codec"/>  I would like to get rid of the magic bean name and Autowire by type instead. Then it should be a matter of adding a jar that provides an implementation of MultitypeCodec.

spring-xd-runner/pom.xml
dturanski: sorry, I had to do a manual merge due to the repo and name changes. I should revert.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-streams/commit/3218c39d477af7e71cb01bd76491b2b6f6f434d4

spring-cloud-streams/src/main/java/org/springframework/cloud/streams/config/ChannelBindingAdapterConfiguration.java
dsyer: I made `spring-bus -> spring-cloud-streams`. The XML should probably go away as well.

PR #3, opened 2015-07-09
XD-3242 created spring-cloud-streams-codec

dturanski:
This is the first step in moving spring-xd-codec into SCS.  Note the package names in spring-cloud-streams-codec cannot be changed until the messagebus code is ported, since they depend on these types.  Also added 'spring-cloud-streams-common'.  Note that the spring-xd-version is changed to 1.2.1.BUILD-SNAPSHOT.  This is required to bring in the latest codec stuff which was recently merged into 1.2.x.  I think this makes sense, since we should never have any XD 2.0 dependencies anyway.


dturanski:
hold off until I fix the problems with .gitignore

PR #4, opened 2015-07-09
Remove current SearchStrategy for ChannelBindingProperties bean

ilayaperumalg:
- Since the ModuleProperties (which is of type ChannelBindingProperties) might
have been initialized by the initilaizer, the search strategy for the conditional
bean needs to look at the hierarchy.


markfisher:
This should be a temporary change. Once the ModuleOptions support is pure boot and we remove the initializer, we should be able to set this back to CURRENT. Thanks @ilayaperumalg for adding a TODO to that effect. I'll merge it now so the build passes.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-streams/commit/a1ca7e8963afd5fa36785ce57c75427a8c9a7d3a

PR #6, opened 2015-07-09
Rename vanilla-samples

ilayaperumalg:
- Change the name to spring-cloud-streams-samples which is already used in pom.xml


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-streams/commit/e3b3bd82fca0df99c0729c738648f831993d6562

PR #8, opened 2015-07-10
Xd 3244  Initial port of XD MessageBus

dturanski:
This brings over XD MessageBus code basically as-is:
* Some job related functionality that depends on JobPlugin, etc. was deleted along with corresponding tests  
* I may have another commit on this if not already merged (see below) , but I don't recommend waiting. 
* KafkaMessageBusTests is failing trying to connect to ZK. (@Ignored) @mbogoevici can have a look, but kafka is not a high priority.
 
The test code had to be rearranged to bring in relevant MB test packages from DIRT and some of the Test harness needed refactoring to sort out dependencies. Currently everything is under _spring-cloud-streams-bindings_.  There is a common test project _spring-cloud-streams-binding-test_. This should take another iteration in a future PR to separate out transport specific things as much as possible. spring-xd-dirt supported testing all transports and had some common base classes, etc. these were dealt out to the binding-* components to get to this point.


mbogoevici:
I wouldn't block this b/c of Kafka.

spring-cloud-streams-bindings/spring-cloud-streams-binding-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/WindowingOffsetManager.java
```diff
@@ -0,0 +1,265 @@
...
+/**
+ * An {@link OffsetManager} that aggregates writes over a time or count window, using an underlying delegate to
+ * do the actual operations. Its purpose is to reduce the performance impact of writing operations
+ * wherever this is desirable.
+ *
+ * Either a time window or a number of writes can be specified, but not both.
+ *
+ * @author Marius Bogoevici
+ */
+public class WindowingOffsetManager implements OffsetManager, InitializingBean, DisposableBean {
```
mbogoevici: This class is shared with the Kafka Source so we should ideally move it to Spring Integration Kafka eventually. Beyond the scope of this PR, but needs a reminder.

dturanski: I'll add a TODO

garyrussell:
There's something funky with this PR and STS; looks like there's a transitive dep on xd-codec someplace - I am seeing unused imports (`kryo.*`, `MutiTypeCodec`) in `ChannelBindingAdapterConfiguration`.

When I fix imports (in STS) it pulls in 2 imports for each). In the CP, spring-xd-codec is there.

spring-cloud-streams/src/main/java/org/springframework/cloud/streams/config/ChannelBindingAdapterConfiguration.java
```diff
@@ -169,15 +175,22 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
+
 	protected static class CodecConfiguration {
 		@Autowired
 		ApplicationContext applicationContext;
 
 		@Bean
 		@ConditionalOnMissingBean(name = "codec")
-		public MultiTypeCodec<?> codec() {
-			Map<String, KryoRegistrar> kryoRegistrarMap = this.applicationContext.getBeansOfType(KryoRegistrar
+		public MultiTypeCodec codec() {
```
garyrussell: raw type, needs `<?>`

dturanski: I'll check it out

garyrussell:
@dturanski re #9 you can just add this commit to your PR if you like...

https://github.com/garyrussell/spring-cloud-streams/commit/d41745f0c0be800463db5de4024c790a3d571ade

garyrussell:
Maybe it's not such a great idea to use the same packages. It can cause all kinds of problems if people have cr*p on their classpath.

spring-cloud-streams/src/main/java/org/springframework/cloud/streams/config/ChannelBindingAdapterConfiguration.java
dturanski: should be better now

garyrussell:
Here's an updated commit for #9 

https://github.com/garyrussell/spring-cloud-streams/commit/e653d3a7272dd87643b6323d4bf7d1a59597829f

garyrussell:
rebased my #9 commit

https://github.com/garyrussell/spring-cloud-streams/commit/dd0173095656612491a7add0dee9e40bf73cc01f

markpollack:
Added commits for changing to JDK7 build and move transport specific packages in binding-test to specific transport test projects.

Squashed and merged as 5377eaf258f6f06b0798d4d8554220ea8b0f38be

Issue #14, opened 2015-07-13
TimeSource sample enhancements


markpollack:
Show alternative implementations

* return a simple String on the `@InboundChannelAdapter` method instead of `MessageSource<String>`
* Use SI Java DSL

Change application.yml to reference `outputChannelName` instead of `inputChannelName`

sabbyanandan:
This is not applicable anymore.

PR #15, opened 2015-07-13
TimeSource sample enhancements

markpollack:
PR to discuss the alternative implementations.

May want to just apply the changes to application.yml, changing the
output channel name.

fixes gh-14

### DIFF
--- a/spring-cloud-streams-samples/source/src/main/resources/application.yml
+++ b/spring-cloud-streams-samples/source/src/main/resources/application.yml
@@ -4,7 +4,7 @@ fixedDelay: 5000
 spring:
   cloud:
     channels:
-      inputChannelName: testtock
+      outputChannelName: testtock
       # uncomment below to use the last digit of the seconds as a partition key
       # hashcode(key) % N is then applied with N being the partitionCount value
       # thus, even seconds should go to the 0 queue, odd seconds to the 1 queue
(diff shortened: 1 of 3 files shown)


ghillert:
Please also make sure the sample compiles with JDK 7. The `pom.xml` still contains:

```
<java.version>1.8</java.version>
```

spring-cloud-streams-samples/source/src/main/resources/application.yml
```diff
@@ -4,7 +4,7 @@ fixedDelay: 5000
 spring:
   cloud:
     channels:
-      inputChannelName: testtock
+      outputChannelName: testtock
```
mbogoevici: This looks like a slip. We haven't been supporting this configuration model since some time.

mbogoevici: In any case, it should have stayed `input`

mbogoevici: Ah, it's not a new PR. Scrap the comments.

ericbottard:
I think we can close this PR, right?

markfisher:
closing this - at some point we should add examples to a reference manual for various techniques, e.g. the Spring Integration DSL

PR #25, opened 2015-07-16
Config update

cppwfs:


cppwfs:
Will close for now, so we can focus on S1.

PR #31, opened 2015-07-20
Minor suggestions

ilayaperumalg:
- Use `ApplicationRunner` instead of `CommandLineRunner` for adapter lifecycle start
 - Remove `Configuration` annotation for `LifecycleConfiguration` and `EnableModuleConfiguration` as these aren't configuration classes
 - Rename the above classes after their functionality

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAdapterRunner.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAdapterRunner.java
@@ -17,15 +17,14 @@
 package org.springframework.cloud.stream.config;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.ApplicationArguments;
-import org.springframework.boot.ApplicationRunner;
+import org.springframework.boot.CommandLineRunner;
 import org.springframework.cloud.stream.adapter.ChannelBindingAdapter;
 
 /**
  * @author Dave Syer
  *
  */
-public class ChannelBindingAdapterRunner implements ApplicationRunner {
+public class ChannelBindingAdapterRunner implements CommandLineRunner {
 
 	@Autowired
 	private ChannelBindingProperties module;
@@ -34,7 +33,7 @@ public class ChannelBindingAdapterRunner implements ApplicationRunner {
 	private ChannelBindingAdapter adapter;
 
 	@Override
-	public void run(ApplicationArguments args) throws Exception {
+	public void run(String... args) throws Exception {
 		if (!adapter.isRunning() && module.isAutoStartup()) {
 			adapter.start();
 		}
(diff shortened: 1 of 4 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAdapterRunner.java
```diff
@@ -17,16 +17,15 @@
...
 import org.springframework.cloud.stream.adapter.ChannelBindingAdapter;
-import org.springframework.context.annotation.Configuration;
 
 /**
  * @author Dave Syer
  *
  */
-@Configuration
-public class LifecycleConfiguration implements CommandLineRunner {
+public class ChannelBindingAdapterRunner implements ApplicationRunner {
```
dturanski: Should be @Component at least? Sounds like it for http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-command-line-runner but not clear from the docs re. ApplicationRunner

ilayaperumalg: My reasoning for not removing `@Configuration` (even `@Component`) in this case is that this specific class isn't a configuration bean and this isn't component scanned as it is imported in EnableModule annotation. I was even hoping to change EnableModule to use `@ImportResource` for non-configuration beans but kept it intact.

ilayaperumalg: The reason to remove `CommandLineRunner` is that, this specific class isn't exactly called from cmdline. The `ApplicationRunner` seems to fit well here as its javadoc suggests: 
> Interface used to indicate that a bean should run when it is contained within a SpringApplication.

dsyer: If `CommandLineRunner` isn't working isn't that a bug in Spring Boot?

ilayaperumalg: Yeah, `CommandLineRunner` works fine. I just thought `ApplicationRunner` could be a better fit as this class isn't run via cmdLine directly ;)

dsyer: It is. And further more `ApplicationRunner` is called wherever `CommandLineRunner` is. Actually if you aren't using the arguments you might as well use `CommandLineRunner`?

ilayaperumalg: Yeah, both works fine in this context. It is just that name `CommandLine` made me think otherwise. I can revert the changes to use `CommandLineRunner`. Shall I?

markfisher:
rebased, squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/47c880d9fc13668a03603ba2b04e8c8171e5ab2f

PR #38, opened 2015-07-22
Explicit module annotations

ilayaperumalg:
- Introduce explicit module annotations like `EnableSourceModule`,
`EnableSinkModule` and `EnableProcessorModule` and have them use
`EnableModule` with the class value specified.
    - The reasoning here is that the user doesn't need to know about the
module class (value) when enabling module. This will also simplify the
annotation based on the module kind.
 - Some code cleanup

### DIFF
--- a/spring-cloud-stream-samples/double/src/main/java/config/SinkModuleDefinition.java
+++ b/spring-cloud-stream-samples/double/src/main/java/config/SinkModuleDefinition.java
@@ -18,7 +18,8 @@ package config;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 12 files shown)


cppwfs:
Isn't this a bit XD specific?

markpollack:
There are already source/sink/processor interfaces in SCS....

ilayaperumalg:
Closing this PR as there could be more channel providers (apart from Source, Sink and Processors) that need to be bound when enabling module. For instance:  `@EnableModule({Source.class, Metrics.class})`. Hence, explicit annotations for each channel providers don't make sense.

PR #41, opened 2015-07-23
Note on how to work around redis local binding.

cppwfs:


markfisher:
thank you Glenn; merged

PR #42, opened 2015-07-23
Set channel resolver for binding adapter

ilayaperumalg:
- Set `ChannelBindingAdapter`'s channel resolver to use
`BinderAwareChannelResolver`.
 - Previously the runtime `Binder` singleton object was lazily
instantiated. But the instantiation actually happened at the constructor
of BinderAwareRouterBeanPostProcessor itself. Hence, removed the lazy
instantiation of the Binder object. Also, I don't see any impact of
creating the `Binder` object at the time of BeanPostProcessor
creation.


mbogoevici:
Rebased, merged https://github.com/spring-cloud/spring-cloud-stream/commit/219d790910af8203efeec6c0c3ef1290eae9ff63

PR #48, opened 2015-07-29
Do not set launcher CL as parent

mbogoevici:


spring-cloud-stream-module-launcher/src/main/java/org/springframework/boot/loader/ModuleJarLauncher.java
```diff
@@ -45,8 +54,51 @@ protected void postProcessClassPathArchives(List<Archive> archives) throws Excep
...
+					copy.add(url);
+				}
+			}
+		}
+		Collections.addAll(copy, urls);
+		return new LaunchedURLClassLoader(copy.toArray(new URL[copy.size()]), null);
+	}
+
+
+	private boolean addDefaultClassloaderUrl(URL[] urls, URL url) {
```
markfisher: maybe rename: shouldAdd...?

mbogoevici: possibly. Those were copied from the base class directly so I didn't change the code. we probably want to refactor this in the future.

markfisher: if that's copied, I'll keep it as-is... hopefully this subclass will be temporary

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/8d7ebe536ea5d6616451120563c7c48ba197671a

PR #57, opened 2015-07-31
Updated Docker file to set export and permissions

cppwfs:
Updated Docker file to set export and permissions to the local repo
Removed dockerize and updated README with the path to repo


markfisher:
This is most likely not necessary now that https://github.com/spring-cloud/spring-cloud-stream/commit/5b5b31d1783f3200556644a2c51721f081f13de6 is merged.

Also, do we really want to remove `dockerize.sh`? It would still be useful to create the Docker image locally.

cppwfs:
Agreed.  Closing as not necessary

PR #58, opened 2015-07-31
create local repo dir if it does not exist

markfisher:


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
```diff
@@ -67,7 +68,10 @@
 	 * may be null or empty if the local repository is off line.
 	 */
 	public AetherModuleResolver(File localRepository, Map<String, String> remoteRepositories) {
-		Assert.isTrue(localRepository.exists(), "File " + localRepository + " does not exist.");
+		if (!localRepository.exists()) {
+			Assert.isTrue(localRepository.mkdir(),
```
mbogoevici: We should use `mkdirs()`  to create the entire path if necessary.

mbogoevici:
Merged w/minor changes as per comments https://github.com/spring-cloud/spring-cloud-stream/commit/5b5b31d1783f3200556644a2c51721f081f13de6

PR #59, opened 2015-07-31
No longer need to do a mkdir for local repo

cppwfs:
dockerfile is even smaller ;-)


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/5a74f88b228112185ea23164cd82cadd272369eb

PR #64, opened 2015-08-06
Refactored Binder and Codec configuration

dturanski:
Addresses XD-3259, XD-3351, XD-3352

* Replace local Codec classes with SI Codec (4.2.BUILD-SNAPSHOT)
* Replace `*-bus.xml` with `@Configuration` MessageChannelBinderConfiguration
* Rename binder properties to use `spring.cloud.stream.binder` prefix
* Create `@ConfigurationProperties` classes for Binder implementations
* Remove Codec as a constructor arg on the Binder implementation, since they simply delegate to `MessageChannelBinderSupport`.setCodec() they have no direct dependency. 
* Remove CodecConfiguration from `@EnableModule`, moved to Binder configuration where needed
* `spring-cloud-stream-codec` provides the `CodecConfiguration` class referenced in the Binder configurations. An alternate implementation would have to exclude this dependency from the dependent binders (except Local) and provide an implementation of this class along with required Codec implementation.


spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/SpringRabbitMQProperties.java
```diff
@@ -0,0 +1,106 @@
...
+package org.springframework.cloud.stream.binder.rabbit.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.core.io.Resource;
+
+/**
+ * @author David Turanski
+ */
+@ConfigurationProperties(prefix = "spring.rabbitmq")
+class SpringRabbitMQProperties {
```
ilayaperumalg: Can we re-use `org.springframework.boot.autoconfigure.amqp.RabbitProperties` here?

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/config/RedisMessageChannelBinderConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.integration.codec.Codec;
+
+/**
+ * @author David Turanski
+ */
+@Configuration
+@Import(CodecConfiguration.class)
+@EnableConfigurationProperties(RedisBinderConfigurationProperties.class)
+@ConfigurationProperties(prefix = "spring.cloud.stream.binder.redis")
```
ilayaperumalg: Is it necessary for this `Configuration` class to act as `ConfigurationProperties` as well just for the `headers` property? Can we move this specific property to `RedisBinderConfigurationProperties` ?

dturanski: There are different prefixes used, `spring.cloud.stream.binder.redis` and `spring.cloud.stream.binder.redis.default` and the pattern is generally repeated for all Binders. so I preserved that. I could split out into a separate ConfigurationProperties, but I'm not sure if it's worth it.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/SpringRabbitMQProperties.java
dturanski: Maybe. I'll look into it.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/RabbitServiceConfiguration.java
```diff
@@ -34,11 +35,12 @@
...
  * @author Dave Syer
  * @author Glenn Renfro
+ * @author David Turanski
  */
 @Configuration
 @ConditionalOnClass(RabbitMessageChannelBinder.class)
 @ConditionalOnMissingBean(RabbitMessageChannelBinder.class)
-@ImportResource("classpath*:/META-INF/spring-cloud-stream/binder/rabbit-binder.xml")
+@Import(RabbitMessageChannelBinderConfiguration.class)
 @PropertySource("classpath:/META-INF/spring-cloud-stream/rabbit-binder.properties")
```
ilayaperumalg: With `spring-boot` specific configuration properties, is this the right time to move all these default *-binder.properties into a centralized boot aware yaml file (src/main/resources/application.yml) ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableModule.java
```diff
@@ -47,7 +46,7 @@
 @Configuration
 @MessageEndpoint
 @Import({RedisServiceConfiguration.class, RabbitServiceConfiguration.class,
-	ChannelBindingAdapterConfiguration.class, CodecConfiguration.class, ChannelBindingAdapterRunner.class,
```
mbogoevici: Now that Rabbit and Redis have `@Configuration` support, I think that we should move `RedisServiceConfiguration` as well as `RabbitServiceConfiguration` into the binder packages. There's no reason for them to be in SCS core.

dturanski: How should these be referenced or resolved? We should import what ever binder configuration is on the class path. There are several ways to do this. I'm not sure what's easiest in this case.

mbogoevici: Maybe auto configuration via {{spring.factories}} ? I'm a little wary about that one because of what it does when multiple binder implementations are on the CP, but it's no different than this, with the added advantage of not decluttering {{@EnableModule}}.

mbogoevici: Anyway, that may be something we can address in a separate PR - this one's pretty big as it is.

dturanski: spring.factories would also work for CodecConfiguration. Currently this class is explicitly imported in the Binder configurations as described in the PR description. An alternate implementation would have to provide a CodecConfiguration.  spring.factories is more flexible but it would be harder to find where the Codec is configured.

mbogoevici: Yeah, I think we should strike a balance between explicit and auto configuration. Maybe use @EnableBinderAutoconfiguration/@EnableCodecAutoconfiguration as a hint (just a thought). I feel it very wrong that for adding Kafka support we have to expand @EnableModule again (and with the discussion of adding different bus implementations it becomes unwieldy)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/RabbitServiceConfiguration.java
dturanski: Yes, but a separate PR

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
```diff
@@ -63,12 +72,6 @@
 			<groupId>org.springframework.xd</groupId>
 			<artifactId>spring-xd-tuple</artifactId>
 			<version>${spring-xd.version}</version>
-			<exclusions>
```
ericbottard: Not sure I get this (haven't looked in detail), but if we're now relying on SI for codec, doesn't that exclusion make *even* more sense now?

dturanski: spring-xd-tuple still depends on spring-xd-codec. Previously, we had the same classes (and package names) in s-c-s-codec since they were also referenced in the binders. Now there is no longer a conflict. The plan is to eventually break out tuple as spring project.

spring-cloud-stream-binders/spring-cloud-stream-binder-local/src/main/java/org/springframework/cloud/stream/binder/local/config/LocalMessageChannelBinderConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @author David Turanski
+ */
+@Configuration
+@ConfigurationProperties(prefix = "spring.cloud.stream.binder.local")
+@EnableConfigurationProperties(LocalExecutorConfigurationProperties.class)
+public class LocalMessageChannelBinderConfiguration {
+	
+	private int queueSize = Integer.MAX_VALUE;
```
ericbottard: Can we move those to the `LocalExecutorConfigurationProperties` class?
I'm not a big fan of `ConfigurationProperties` on the `@Configuration` class itself, yet alone on the mixed approach.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitMessageChannelBinderConfiguration.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.integration.codec.Codec;
+
+@Configuration
+@Import(CodecConfiguration.class)
+@EnableConfigurationProperties({RabbitBinderConfigurationProperties.class, SpringRabbitMQProperties.class})
+public class RabbitMessageChannelBinderConfiguration {
+	@Autowired Codec codec;
```
ericbottard: any particular reason for using package access in these, both for fields and `@Bean`?

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/PropertiesConfigTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+
+/**
+ * @author David Turanski
+ */
+public class PropertiesConfigTest {
+	@Test
+	public void test() {
+
+		SpringApplication application = new SpringApplicationBuilder(MyConfig.class)
```
ericbottard: I think `@IntegrationTest("foo = bar")` is exactly tailored for this kind of setup

ericbottard: What is this class testing anyway? Doesn't seem to be related to this PR, is it?

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/config/RedisMessageChannelBinderConfiguration.java
ericbottard: Ah, gotcha (related to my comment above)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/RabbitServiceConfiguration.java
mbogoevici: I think that centralizing configuration for all types of binders would create unnecessary coupling. They should be able to evolve independently.

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/config/RedisMessageChannelBinderConfiguration.java
mbogoevici: I think we should consider whether we still need the 'default'/'non-default' split, given that we don't have global bus anymore.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/RabbitServiceConfiguration.java
mbogoevici: Just as a clarification: I meant it in what concerns the project structure. We can definitely create a centralized configuration for use with spring-cloud-config for example.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitMessageChannelBinderConfiguration.java
mbogoevici: Will address during merge.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaTestBinder.java
```diff
@@ -46,7 +52,7 @@ public KafkaTestBinder(KafkaTestSupport kafkaTestSupport, MultiTypeCodec<Object>
 			zookeeperConnect.setZkConnect(kafkaTestSupport.getZkConnectString());
 			KafkaMessageChannelBinder binder = new KafkaMessageChannelBinder(zookeeperConnect,
 					kafkaTestSupport.getBrokerAddress(),
-					kafkaTestSupport.getZkConnectString(), codec);
```
mbogoevici: We should inject the Codec via setter, now that it's not set via constructor.

spring-cloud-stream-binders/spring-cloud-stream-binder-local/src/main/java/org/springframework/cloud/stream/binder/local/config/LocalMessageChannelBinderConfiguration.java
mbogoevici: Probably should - we need to address this in the general context of having multiple ConfigurationProperties classes (and namespaces). Will merge as-is but make a note.

mbogoevici:
LGTM. The following points may be addressed later
- reuse of `org.springframework.boot.autoconfigure.amqp.RabbitProperties`;
- polishing `ConfigurationProperties` and/or namespaces (see https://github.com/spring-cloud/spring-cloud-stream/pull/64#discussion_r36469466 and https://github.com/spring-cloud/spring-cloud-stream/pull/64#discussion_r36577199);
- autoconfiguration for codec/binder (see https://github.com/spring-cloud/spring-cloud-stream/pull/64#discussion_r36519136) ;

mbogoevici:
Squashed, merged with minor polishing: https://github.com/spring-cloud/spring-cloud-stream/commit/0a84afd6ea9e6bcf4aa8781824c82748e2894dc6

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableModule.java
ericbottard: I created XD-3380 to track this, giving it a try as it is the basis for some test support I want to provide for module authors (XD-3381)

PR #67, opened 2015-08-10
XD-3380: Use AutoConfiguration for Binder Configurations

ericbottard:
Also, fix rabbit configuration props


spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitBinderConfigurationProperties.java
```diff
@@ -31,15 +31,15 @@
...
 
-	private int backOffMultiplier;
+	private double backOffMultiplier;
 
 	private boolean transacted;
 
 	private boolean concurrency;
 
-	private MessageDeliveryMode defaultDeliveryMode;
+	private MessageDeliveryMode deliveryMode;
```
dturanski: The accessors are set/getDefaultDefaultDeliveryMode(). I checked this with Gary. This name was intentional for consistency

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/resources/META-INF/spring-cloud-stream/rabbit-binder.properties
```diff
@@ -18,6 +18,6 @@ spring.cloud.stream.binder.rabbit.default.prefetch:  1
 spring.cloud.stream.binder.rabbit.default.replyHeaderPatterns:       STANDARD_REPLY_HEADERS,*
 spring.cloud.stream.binder.rabbit.default.republishToDLQ:    false
 spring.cloud.stream.binder.rabbit.default.requestHeaderPatterns:     STANDARD_REQUEST_HEADERS,*
-spring.cloud.stream.binder.rabbit.default.requeue:   true
+spring.cloud.stream.binder.rabbit.default.requeueRejected:   true
 spring.cloud.stream.binder.rabbit.default.transacted:false
-spring.cloud.stream.binder.rabbit.default.txSize:    1
+spring.cloud.stream.binder.rabbit.default.txSize:    1
```
dturanski: new line

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration:\
+org.springframework.cloud.stream.binder.rabbit.config.RabbitServiceAutoConfiguration
```
dturanski: new line

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/resources/META-INF/spring-cloud-stream/redis-binder.properties
```diff
@@ -2,4 +2,4 @@ spring.cloud.stream.binder.redis.default.backOffInitialInterval:    1000
 spring.cloud.stream.binder.redis.default.backOffMaxInterval:        10000
 spring.cloud.stream.binder.redis.default.backOffMultiplier:         2.0
 spring.cloud.stream.binder.redis.default.concurrency:               1
-spring.cloud.stream.binder.redis.default.maxAttempts:               3
+spring.cloud.stream.binder.redis.default.maxAttempts:               3
```
dturanski: new line

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration:\
+org.springframework.cloud.stream.binder.redis.config.RedisServiceAutoConfiguration
```
dturanski: new line

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitBinderConfigurationProperties.java
ericbottard: Ah, was wondering and hesitated to ask. Will revert (the properties file in master is hosed anyway)

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/resources/META-INF/spring-cloud-stream/rabbit-binder.properties
dturanski: while we're in here, should rename the default prefix 'xdbus.' -> 'binder.' or 'binder.rabbit.'

spring-cloud-stream/pom.xml
```diff
@@ -50,6 +50,7 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-stream-binder-redis</artifactId>
+			<optional>true</optional>
```
mbogoevici: If we do this (which is sane), we need to change our examples - the build currently fails building them.

https://travis-ci.org/spring-cloud/spring-cloud-stream/builds/74914672

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitBinderConfigurationProperties.java
markfisher: should we reconsider the layering of defaults via property names now that we have the layering of properties via boot? cc @garyrussell

garyrussell: I am not fully tuned in for s-c-s, but in the XD case, they are bus-level defaults, which can be overridden at the stream level. In the case of `getDefaultDefaultDeliveryMode` it's strictly correct because there's a default at the bus level, default at the stream level (overrides bus) and delivery mode at the message level (overrides both).

If s-c-s only has one output channel then I suppose, we can dump the default.

However, I thought the intention is to support multiple output channels (and we also have taps etc), so they really are bus defaults that might be overridden by the stream definition.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/resources/META-INF/spring-cloud-stream/rabbit-binder.properties
markfisher: I would vote for just "binder" (and use that consistently across binder implementations) since this property maps to a name of a resource that is created within rabbit itself, thus having "rabbit" in the name also, seems redundant. The main purpose is to provide a namespace for the queues/topics we create.

dturanski:
LGTM; rebased squashed and merged.

PR #68, opened 2015-08-11
Use a non-default boot prefix for redis connection factory

markpollack:
* Note, did not change the use of the word 'default' in the prefix. To discuss.
* Note, did not move redis-binder.properties into
  RedisBinderConfigurationProperties. To discuss.

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/config/RedisAutoConfiguration.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/config/RedisAutoConfiguration.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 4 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/config/RedisAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.redis.connection.jedis.JedisConnection;
+import org.springframework.data.redis.connection.jedis.JedisConnectionFactory;
+import org.springframework.data.redis.core.RedisOperations;
+import redis.clients.jedis.Jedis;
+
+import java.net.UnknownHostException;
+
+/**
+ * Creates a dedicated RedisConnectionFactory different from the one spring-boot autoconfigure
```
ericbottard: Bogus javadoc

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/test/java/org/springframework/cloud/stream/binder/redis/config/RedisAutoConfigurationTests.java
```diff
@@ -0,0 +1,68 @@
...
+    public void close() {
+        if (this.context != null) {
+            this.context.close();
+        }
+    }
+
+    @Test
+    public void testDefaultRedisConfiguration() throws Exception {
+        load();
+        assertNotNull(BeanFactoryAnnotationUtils.qualifiedBeanOfType(this.context.getBeanFactory(),
```
ericbottard: Would be nice to have a test that triggers both this `@Configuration` and the standard boot one, and validate that they both live side by side IMO

ericbottard:
While working on the counter sink, realized this approach can't work: If I'm a module author and want redis (from boot), adding `RedisAutoConfiguration` won't work, as `@ConditionalOnMissingBean` there is done _by type_.

markfisher:
closing this PR since the underlying issue will fall within the scope of multiple binder support

cc @mbogoevici - might glean some ideas from this though

PR #69, opened 2015-08-11
Align dockerization process with spring.io guides

dsyer:
(And with existing Spring Cloud projects.) Uses a maven plugin to
build the image. It can be pushed manually, or else using
-Ddocker.image.prefix=<path_to_repo>. Removes some of the hard-coded
host names and stuff that came with the old version.


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherApplication.java
```diff
@@ -16,33 +16,23 @@
...
+ * provided at launch.
  *
- * @see ModuleLauncher#launch(String[], String[]) for module and argument structure and format
+ * @see ModuleLauncher#launch(String[], String[]) for module and argument structure and
+ * format
  *
  * @author Marius Bogoevici
  */
 @SpringBootApplication
-@ComponentScan(basePackageClasses = ModuleLauncherApplication.class)
```
mbogoevici: not sure why don't we need the ComponentScan here

mbogoevici: Ah, sorry. `@SpringBootAnnotation` is `@ComponentScan`

markfisher:
LGTM; merged

PR #70, opened 2015-08-11
Add auto-config for Local binder

ilayaperumalg:
- Make `local` binder compile dependency as optional
 - Add `local` binder auto configuration
    - This imports local binder properties and binder
 - Add `spring.factories` to enable auto configuration when this archive
is on the classpath


ericbottard:
Rebased and merged as 5559188

PR #72, opened 2015-08-12
XD-3381: Provide test support for modules

ericbottard:
Example usage is here: https://github.com/ericbottard/spring-cloud-stream-modules/commit/3ad04862f8f844e23c51b09a52c5adfb58000d10


pom.xml
```diff
@@ -34,7 +34,8 @@
 		<module>spring-cloud-stream-module-launcher</module>
 		<module>spring-cloud-stream-samples</module>
 		<module>docs</module>
-	</modules>
+    <module>spring-cloud-stream-test-support</module>
```
dturanski: indent

pom.xml
```diff
@@ -147,4 +148,4 @@
 			</snapshots>
 		</pluginRepository>
 	</pluginRepositories>
-</project>
+</project>
```
dturanski: new line

spring-cloud-stream-test-support/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-stream-parent</artifactId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+	<groupId>org.springframework.cloud</groupId>
+	<artifactId>spring-cloud-stream-test-support</artifactId>
+	<name>spring-cloud-stream-test-support</name>
+	<url>http://maven.apache.org</url>
```
dturanski: remove url

spring-cloud-stream-test-support/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-binder-spi</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-codec</artifactId>
+		</dependency>
```
dturanski: shouldn't need codec

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	/**
+	 * Bridges a QueueChannel to the module output channel, so that it can
+	 * be easily queried by {@link org.springframework.cloud.stream.test.matcher.MessageChannelMatcher}.
+	 */
+	@Override
+	public void bindProducer(String name, MessageChannel outboundBindTarget, Properties properties) {
+		BridgeHandler bridge = new BridgeHandler();
+		QueueChannel queue = new QueueChannel();
```
dturanski: not sure why we need this to be asynch

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinderAutoConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Installs the {@link TestSupportBinder}.
+ * Note that this auto-configuration has higher priority than regular binders, so adding
+ * this on the classpath in test scope is sufficient to have support kick in.
+ *
+ * @author Eric Bottard
+ */
+@Configuration
+@ConditionalOnMissingBean(Binder.class)
+@AutoConfigureOrder(Ordered.LOWEST_PRECEDENCE - 10)
```
dturanski: I think this should be HIGHEST_PRECEDENCE since we don't ever want another Binder implementation if this is on the classpath. and no ConditionalOnMissingBean

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
ericbottard: I'm afraid I don't understand what you mean. I think, on the contrary, that we prefer to be synchronous/blocking when writing tests.

dturanski:
Need to add some tests

spring-cloud-stream-test-support/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration:\
+org.springframework.cloud.stream.test.binder.TestSupportBinderAutoConfiguration
```
dturanski: new line

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinderAutoConfiguration.java
dturanski: I did it a bit different here:
@Configuration
@Order(Ordered.HIGHEST_PRECEDENCE)
public class TestSupportBinderAutoConfiguration {

	@Bean
	public MessageCache messageCache() {
		return new MessageCache();
	}

	@Bean
	public Binder testSupportBinder() {
		return new TestSupportBinder(messageCache());
	}
}

So tests can @Autowire the MessageCache and use that instead of the Binder

ericbottard: Interesting. I assume the MessageCache has a mapping between Channels and .. something?

But then, how do you use it?

ericbottard: I guess one could do
```
assertThat(messageCache.for(Processor.output()), receivedMessageThat(...))
```

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
dturanski: Why not just 
   
       (SubscribableChannel)outputBindTarget).subscribe(...)

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinderAutoConfiguration.java
ericbottard: A bit more verbose, but less static. I like it. Not a big fan of the "cache" name, though

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
ericbottard: Good point. At first, I thought the `MessageChannelBinderSupport.Binding` concept would help me, and thus I stuck with it. Not sure how much like a "real" binder we want this to feel

dturanski: Yeah. I'm not sure at this point.

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * A minimal binder that<ul>
+ *     <li>does nothing about binding consumers, leaving the channel as-is, so that a test author can interact with it directly,</li>
+ *     <li>registers a queue channel on the producer side, so that it is easy to assert what is received.</li>
+ * </ul>
+ *
+ * @author Eric Bottard
+ * @see org.springframework.cloud.stream.test.matcher.MessageChannelMatcher
+ */
+public class TestSupportBinder extends MessageChannelBinderSupport {
```
mbogoevici: I would call this `TestBinder` (it's easier to read)

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
```diff
@@ -0,0 +1,103 @@
...
+
+	public static Map<MessageChannel, QueueChannel> channelToBindings =
```
mbogoevici: Perhaps this doesn't need to be static, and could be injectable as a bean - thus saving the need to reference directly the binder and enforcing isolation between tests?

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinderAutoConfiguration.java
mbogoevici: +1

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
ericbottard: It all comes down to how it reads in _e.g._ a Hamcrest matcher (see discussion below).
At first, I really wanted to be able to say `assertThat(theChannel, ...)`

mbogoevici: Right, I think that the discussion below provides a very good alternative :)

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
dturanski: Yes but we have https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractTestBinder.java  so may be confusing

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinderAutoConfiguration.java
ericbottard: Should we go the extra mile and provide support for injecting the message results _per channel_ (given that most modules will have only one output channel), eg
```
@Autowired
@ResultsFor(Processor.OUTPUT)
private Queue<Message<?>> result;

...

assertThat(result, receivesPayloadThat(equalTo("hellofoo")).within(10));
```

dturanski: I think so

dturanski: How about  s/MessageCache/TestMessageConsumer ?

ericbottard:
Updated the PR to use injection rather than static access.
Added extensive tests.

Yet, I have not gone through my proposed idea to inject _per-channel_, because I think it is hard to read in tests. I think it is important to see 
```
doSomethingWith(input())
// and just after
assertThat(someReferenceTo(output()), ...)
```

Annotating with _e.g._ `Processor.OUTPUT` has two drawbacks IMO:
* the reference is at the injection point, not the assertion point
* it is not obvious that `Processor.OUTPUT` relates to `output()`

WDYT?

markfisher:
Is it worth considering a Gateway as the component that could be injected into test code? That would provide `sendAndReceive` from the perspective of the test *as a client*.

ericbottard:
@markfisher could you elaborate (if I understand what you're saying, I believe it only makes sense for Processor, correct?). And would not cover cases like 1 input for 2 outputs, _etc._

markfisher:
actually the spring-integration Gateway supports: send, receive, and sendAndReceive, so it could be generally useful as the test "client" cc @garyrussell

ericbottard:
Still not sure I get it. Could you provide an example usage (from the perspective of test writer) of say, testing a routing module that has one input and 2 outputs ("odd" and "even"):
```
public interface MyChannels {
   @Input("input")
   MessageChannel input();

   @Ouput("odd")
   MessageChannel odd();

   @Output("even")
   MessageChannel even();
}
```

BTW, aren't the types in Sink & Source currently backwards (assuming we want/need the down-typing to `SubscribableChannel`)

markfisher:
They are not backwards. The "direction" of those channels is from the perspective of the module. That is why I was suggesting looking at the test from a "client" perspective (outside of the black-box). The test code would send to the module's input, and receive from the module's output.

mbogoevici:
@ericbottard Personally, I think that relying on the interface methods - i.e.  `input()` vs INPUT is much better, too. Personally, I wouldn't register channel beans at all, weren't for a total pain to use things like `@ServiceActivator` that rely on finding channels by name. 

I would agree that we could provide a test Gateway, but can we do it as a future enhancement ? Also, can't see how we can use it for single-channel modules (e.g. sources and sinks). I think it's a valuable idea, but not having support is not a blocker for solving the immediate problem of having a test binder?

markfisher:
one thing I was discussing with @markpollack is that we could inject the "other side" into the test... for example if we are testing a Source module (which sends messages to its own "output" channel), the test itself could be injected with an `@Input`. Essentially the test is acting as if it's a direct-bound Sink in that case (and vice versa of course).

mbogoevici:
@markfisher that is taking things in a completely different direction, and I am afraid it both complicates things and makes them very spring-cloud-data centric. In the current approach, a tester will get access to all the 'bound' interfaces of a module (which they know very well) and can act and test them directly. In the approach where we bind and bridge, they would have to deal with an additional set of channels, and it is not very clear where are they coming from. Additionally, for multi-inputs/multi-outputs, you'd have to be fairly descriptive about the binding.

markpollack:
For source, it would seem a more traditional approach of the test executing the stimulus the source expects, http,jms message, rabbit etc, would be very straightforward.  See https://github.com/spring-projects/spring-xd/blob/master/extensions/spring-xd-extension-http/src/test/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapterTests.java

I like the Gateway approach for processors.

pom.xml
```diff
@@ -34,7 +34,8 @@
 		<module>spring-cloud-stream-module-launcher</module>
 		<module>spring-cloud-stream-samples</module>
 		<module>docs</module>
-	</modules>
+		<module>spring-cloud-stream-test-support</module>
+  </modules>
```
markfisher: fix indentation

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
```diff
@@ -0,0 +1,50 @@
...
+		LinkedBlockingDeque<Message<?>> result = new LinkedBlockingDeque<>();
+		Assert.isNull(results.put(channel, result));
+		return result;
+	}
+
+	/*package*/ void unregister(MessageChannel channel) {
+		Assert.notNull(results.remove(channel));
+	}
+
+	public BlockingQueue<Message<?>> forChannel(MessageChannel channel) {
```
markfisher: should this call register if not already registered?

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @author Eric Bottard
+ */
+public class MessageCollector {
+
+	private Map<MessageChannel, BlockingQueue<Message<?>>> results = new HashMap<>();
+
+	/*package*/ BlockingQueue register(MessageChannel channel) {
+		LinkedBlockingDeque<Message<?>> result = new LinkedBlockingDeque<>();
+		Assert.isNull(results.put(channel, result));
```
markfisher: this assertion is too late - the new empty queue has already been *put* into the map, but that is what needs to be prevented (also the Assert should provide a useful message).

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
```diff
@@ -0,0 +1,50 @@
...
+		return result;
+	}
+
+	/*package*/ void unregister(MessageChannel channel) {
+		Assert.notNull(results.remove(channel));
```
markfisher: does this really need to Assert? maybe return a boolean instead, but I don't see why it can't be considered harmless to unregister something that does not exist

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
```diff
@@ -0,0 +1,129 @@
...
+	private final MessageCollector messageCollector;
+
+	public TestSupportBinder(MessageCollector messageCollector) {
+		this.messageCollector = messageCollector;
+	}
+
+
+	@Override
+	public void bindConsumer(String name, MessageChannel inboundBindTarget, Properties properties) {
+		// Do nothing. A module author can grab hold forChannel the input channel
```
markfisher: comment grammar (should it be at javadoc level?)

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/matcher/MessageQueueMatcher.java
```diff
@@ -0,0 +1,160 @@
...
+import org.hamcrest.SelfDescribing;
+
+import org.springframework.cloud.stream.test.binder.TestSupportBinder;
+import org.springframework.integration.util.Function;
+import org.springframework.messaging.Message;
+
+/**
+ * A Hamcrest Matcher meant to be used in conjunction with {@link TestSupportBinder}.
+ *
+ * <p>Expected usage is forChannel the form (with appropriate static imports):
```
markfisher: this doesn't make sense

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
ericbottard: My reasoning is that this is being called from the TestSupportBus only, it should never try to unregister something it did not register itself in the first place

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
ericbottard: Same reason as above, we may even want to fail if not found, as this would happen from a misuse of the test support IMO

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
ericbottard: Hmm, weird sentence indeed. May be IDEA refactoring from text in comments. The idea is already expressed in the class javadoc anyhow. Will remove

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
markfisher: JavaDoc should be added including a `@throws` description (same for `register`), but I still don't see why an exception needs to be thrown in the case of unregistering a channel that was not registered.

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
markfisher: also this needs javadoc

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java
```diff
@@ -0,0 +1,129 @@
...
+		// and interact with it directly
+	}
+
+	@Override
+	public void bindPubSubConsumer(String name, MessageChannel inboundBindTarget, Properties properties) {
+
+	}
+
+	/**
+	 * Bridges a QueueChannel to the module output channel, so that it can
```
ericbottard: Outdated javadoc, will remove

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/matcher/MessageQueueMatcher.java
ericbottard: Hmm, same "forChannel" occurence. Damn IDEA. Will revisit

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
markfisher: it should either fail or register on-demand (latter might be better), but returning null here seems dangerous

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
ericbottard: `put` returns what was there _before_. I want to fail if there was something already registered.

markfisher: What I was saying is that you are not *preventing* the change in the map this way. Sure, the assert fails, but the damage has already been done.

ericbottard: Yes, I know. But as you say, there is will be an exception just after. Can refactor so that it does not mutate.

markfisher: My point was that a caller could handle the exception and not be stuck with an invalid MessageCollector, whereas the current code is making that MessageCollector invalid regardless of how the exception is handled.

You can use `putIfAbsent` so otherwise it would be the same (although that is only added to the core Map in 1.8, so would need to use ConcurrentMap, which would be fine).

ericbottard: This class should really only be used by the TestSupportBinder (which you
can tell from the two package protected methods).
We may even want to extract an interface for the `forChannel()` method and
move this impl. to be private to the binder

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/matcher/MessageQueueMatcher.java
```diff
@@ -0,0 +1,160 @@
...
+ *        processor.input().send(new GenericMessage{@literal <}Object>("hello"));
+ *        assertThat(messageCollector.forChannel(processor.output()), receivesPayloadThat(is("hellofoo")).within(10));
+ *    }
+ *
+ * }</pre>
+ * </p>
+ *
+ * @author Eric Bottard
+ */
+public class MessageQueueMatcher<T> extends BaseMatcher<BlockingQueue<Message<?>>> {
```
markfisher: This class has a lot of warnings, such as "rawtypes" (showing up in the IDE). Also, is there any way to avoid some of the "unchecked" suppressions via parameterization?

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/MessageCollector.java
markfisher: If that's really true, can it be encapsulated completely into the binder? Even so, I don't see why that precludes making the assertion proactive instead of reactive etc. That makes the code easier to understand if nothing else.

ericbottard:
Rebased, force pushed

ericbottard:
Bumped. Would be nice to have some closure on this (I can merge myself if everyone is ok with its current state)

dturanski:
squashed and merged as e2497ef

PR #75, opened 2015-08-12
ModuleConfiguration properties

ilayaperumalg:
- Add `prefix` based module configuration properties
 - Add default properties in `application.yml`
 - Move the properties into separate `ConfigurationProperties` bean rather than being inside `Configuration` bean


mbogoevici:
Squashed, rebased, merged https://github.com/spring-cloud/spring-cloud-stream/commit/e078e29db4c3c69a68866f8bca43c04e38e6b1c3

PR #76, opened 2015-08-13
ModuleRunner enhancements

ilayaperumalg:
- Move the launcher property `modules` to `ModuleLauncherProperties`
   - This will make the `ModuleRunner` to be `Component` bean only; Also, it only needs to enable the configuration properties of `ModuleLauncherProperties`

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
@@ -18,44 +18,38 @@ package org.springframework.cloud.stream.module.launcher;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.springframework.beans.factory.InitializingBean;
+
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.ApplicationArguments;
 import org.springframework.boot.ApplicationRunner;
-import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Component;
-import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
+ * Spring boot {@link ApplicationRunner} that triggers {@link ModuleLauncher} to launch the modules.
+ *
  * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
  */
 @Component
-@ConfigurationProperties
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
```diff
@@ -18,45 +18,45 @@
...
-	public void setModules(String[] modules) {
-		this.modules = modules;
-	}
+	@Autowired
+	private ModuleLauncher moduleLauncher;
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-		Assert.notEmpty(this.modules, "A list of modules must be specified");
+		Assert.notEmpty(moduleLauncherProperties.getModules(), "A list of modules must be specified");
```
ericbottard: This should move as jsr303 validation in the ModuleLauncherProperties class

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/ModuleResolverProperties.java
```diff
@@ -0,0 +1,57 @@
...
+
+	/**
+	 * Local maven repository to resolve the module location.
+	 */
+	private File localRepository;
+
+	/**
+	 * Remote maven repository to resolve the module location.
+	 */
+	private String remoteRepository;
```
ericbottard: Unrelated to this PR, but do we want `String[]` instead of `String` (or even `URL[]`)?

markfisher: +1 and that's being addressed here: https://github.com/spring-cloud/spring-cloud-stream/pull/83/files#diff-d0fe905e9297c6c526220607041efc5aR41

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
ilayaperumalg: That would be nice. But, do we really want to have `spring-boot-starter-validation` dependency for `spring-cloud-stream` or at least for `module-launcher`? I see `spring-boot-starter-validation` defaults to `hibernate-validator`.

ilayaperumalg:
Rebased to pull the recent changes around module launcher/resolver properties.

ericbottard:
Taking this over, as I need this as a base to an ever further generalization for args passing, for which I'll issue a PR shortly.
Ilaya's changes are sound by themselves, so merging as a first commit.

ericbottard:
Merged as 7663372

PR #78, opened 2015-08-13
Remove application.yml and initialize properties within ModuleLaunche…

mbogoevici:
…rProperties


ericbottard:
Merged as b729670

PR #79, opened 2015-08-13
Fix dependencies

dsyer:
In order to join the Spring Cloud release train we have to align the
dependencies and (more importantly) the dependency management strategy.
The strategy is to not put version  ids in `<dependencies>` (only
in `<dependencyManagement>`) where possible. And more importantly
we cannot use the releae train itself (creates a cycle).

This change achieves most of that (and mkae it possible for client
apps to use spring-cloud-starter-parent:Brixton.BUILD-SNAPSHOT)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,6 +34,13 @@
 	<dependencyManagement>
 		<dependencies>
 			<dependency>
+				<groupId>org.springframework.integration</groupId>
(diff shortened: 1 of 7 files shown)


mbogoevici:
Squashed, rebased. https://github.com/spring-cloud/spring-cloud-stream/commit/dafa0fdf4000e69e1e813001a1ff5db4c6758565

PR #80, opened 2015-08-13
Fix launcher ClassLoader parent

ericbottard:


mbogoevici:
I agree, in the sense that my original change set it to `null` :). 

Unfortunately, that gives us the bootstrap CL but not the extension CL of the JVM, and that causes modules launched via module-launcher to fail on things normally found in extensions, such as Xalan. So your PR is correct in the sense that we should try to use the extension class loader (which is normally the parent of the application class loader - that has the CP jars). 

However, having a hierarchy of loaders is not mandated by the Java Spec - so it could be that there is just a single system classloader, and we should try to use that one then - at the risk of adding the additional class path JARs. Both are corner cases, but I think we should err in favour of ensuring that the JVM jars (bootstrap and extension) are present. I don't think that we generally expect to have additional jars, since the module launcher is an uberjar.

mbogoevici:
Merged with discussed changes https://github.com/spring-cloud/spring-cloud-stream/commit/47d84a8dc48e4986b00e79c774a920c89b1b8a24

PR #83, opened 2015-08-13
Support comma separated remote repos

ilayaperumalg:
- Change `remoteRepository` option to `remoteRepositories` that accepts comma separated string values.
 - AetherModuleResolver would expect set of remote repositories and create `RemoteRepository` list
for the given set.

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
@@ -38,13 +38,13 @@ public class ModuleLauncherProperties {
 	/**
 	 * Location of comma separated remote maven repositories from which modules will be downloaded, if not available locally.
 	 */
-	private String remoteRepositories = "https://repo.spring.io/libs-snapshot";
+	private String[] remoteRepositories = new String[] {"https://repo.spring.io/libs-snapshot"};
 
-	public void setRemoteRepositories(String remoteRepositories) {
+	public void setRemoteRepositories(String[] remoteRepositories) {
 		this.remoteRepositories = remoteRepositories;
 	}
 
-	protected String getRemoteRepositories() {
+	protected String[] getRemoteRepositories() {
 		return remoteRepositories;
 	}
 
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
@@ -36,16 +36,16 @@ public class ModuleLauncherProperties {
 			+ File.separator + ".m2" + File.separator + "repository");
 
 	/**
-	 * Location of a remote maven repository from which modules will be downloaded, if not available locally.
+	 * Location of comma separated remote maven repositories from which modules will be downloaded, if not available locally.
 	 */
-	private String remoteRepository = "https://repo.spring.io/libs-snapshot";
+	private String remoteRepositories = "https://repo.spring.io/libs-snapshot";
 
-	public void setRemoteRepository(String remoteRepository) {
-		this.remoteRepository = remoteRepository;
+	public void setRemoteRepositories(String remoteRepositories) {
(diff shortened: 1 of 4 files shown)


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
```diff
@@ -88,10 +89,10 @@ public AetherModuleResolver(File localRepository, Map<String, String> remoteRepo
 		}
 		this.localRepository = localRepository;
 		this.remoteRepositories = new LinkedList<>();
-		if (!CollectionUtils.isEmpty(remoteRepositories)) {
-			for (Map.Entry<String, String> remoteRepo : remoteRepositories.entrySet()) {
-				RemoteRepository remoteRepository = new RemoteRepository.Builder(remoteRepo.getKey(),
-						DEFAULT_CONTENT_TYPE, remoteRepo.getValue()).build();
+		if (!CollectionUtils.isEmpty(remoteRepositoriesSet)) {
+			for (String remoteRepo : remoteRepositoriesSet) {
+				RemoteRepository remoteRepository = new RemoteRepository.Builder("",
```
mbogoevici: I think that we should generate a unique id per repo here. I don't think it will have a huge functional impact right now, but it may affect things like adding credentials.

ilayaperumalg: It is still possible to add support to have `remoteRepositories` specified as `id1:repo1,id2:repo2`. But I don't see any significance of having the ID for the RemoteRepository.

ilayaperumalg: I just started adding comment as you started it seems. My above comment isn't a reply to yours :-) I noticed Aether always set the ID to `""` if it doesn't find. we can generate a unique ID to be on the safer side. I will fix it.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
```diff
@@ -36,16 +36,16 @@
 			+ File.separator + ".m2" + File.separator + "repository");
 
 	/**
-	 * Location of a remote maven repository from which modules will be downloaded, if not available locally.
+	 * Location of comma separated remote maven repositories from which modules will be downloaded, if not available locally.
 	 */
-	private String remoteRepository = "https://repo.spring.io/libs-snapshot";
+	private String remoteRepositories = "https://repo.spring.io/libs-snapshot";
```
markfisher: We should be able to use an array (or List) type for the remoteRepositories field rather than having to handle the comma-delimited String conversion ourselves.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
markfisher: Please see my comment above. I think we can avoid a lot of complexity by using a Collection type directly in the ConfigurationProperties class. Based on this discussion about "id", it should probably be a Map.

ilayaperumalg: Sure, I will fix with the Collection type approach.

mbogoevici: Wouldn't it be simpler for the end-user if we allowed a comma-separated list of values that gets converted to a collection/array and we built the map ourselves (using default/generated ids)? Passing properties for binding to a map making this slightly more complicated?

ilayaperumalg: I just pushed some changes that generates IDs and passing the map to the resolver. Does that answer your concern?

mbogoevici:
LGTM, merging

mbogoevici:
Merged with minor changes: https://github.com/spring-cloud/spring-cloud-stream/commit/4ab01b6e7b9c3af4fe9f6827541b9dbdbef29909

PR #84, opened 2015-08-13
Re-instate lazy lookup for Binder in DestinationResolver

dsyer:
It's important not to depend on something heavyweight like a Binder in a
BeanPostProcessor. The usual trick (as implemented here) is to do a lazy
lookup of the thing that is eventually injected in the post processes
beans.

Rebased ontop of gh-79

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,6 +34,13 @@
 	<dependencyManagement>
 		<dependencies>
 			<dependency>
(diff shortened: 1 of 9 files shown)


mbogoevici:
Cherry-picked and merged. https://github.com/spring-cloud/spring-cloud-stream/commit/15e3c6066f585417d28cc93033d28fdf390d88b7

PR #86, opened 2015-08-17
remove .idea/copyright

dturanski:


ericbottard:
why ?

markpollack:
Would like to leave this and also check in an eclipse directory for code formatting...

markpollack:
oops, didn't see that idea itself overrides it - ignore my comment

PR #88, opened 2015-08-18
Make immediate receive the default

ericbottard:
Makes more sense IMO, and avoids tests that run indefinitely by mistake.

### DIFF
--- a/spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/matcher/MessageQueueMatcher.java
+++ b/spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/matcher/MessageQueueMatcher.java
@@ -113,6 +113,14 @@ public class MessageQueueMatcher<T> extends BaseMatcher<BlockingQueue<Message<?>
 		return new MessageQueueMatcher<>(this.delegate, timeout, unit, this.extractor);
 	}
 
+	public MessageQueueMatcher<T> immediately() {
+		return new MessageQueueMatcher<>(this.delegate, 0, null, this.extractor);
+	}
+
+	public MessageQueueMatcher<T> indefinitely() {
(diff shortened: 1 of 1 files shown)


markfisher:
I agree that indefinite blocking is not a good default, since that can lead to hanging tests in some situations. However, it's very common that the test code will be ready to wait for a reply before whatever processing is being tested has completed. It might be worth a compromise that blocks with a (default) timeout value? (yet in other cases, such as expecting null after already receiving all the expected messages, a 0 timeout is appropriate, but for those cases the test writer can explicitly use the immediate return).

ericbottard:
So I guess now we need to come up with that sensible default value...
1s ?

markfisher:
1s is probably too short for many cases if we want to avoid false negatives, especially if the code under test requires initialization before processing begins. Since the value is the *max* wait time (e.g. anything that returns in a second or less will still return in a second or less regardless of the timeout value), I would opt for a higher value. We just don't want indefinite blocking. AND we should provide guidance (JavaDoc and reference manual for test-support) for using the 0-timeout (immediate) option in cases where processing has ended and the expectation is "no more messages", e.g. the equivalent of `assertNull(receiveNoWait())`

ericbottard:
Note that the way this test-support is currently used, the context and everything is already built, etc.
So, apart from remote-calling tests, I'm not sure we really need more.
Look, we know we'll end up using 5s, so I'll propose 10, so you can make a counter offer. WDYT?

markfisher:
7500 ;)

(seriously, I think 5s is reasonable)

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/645b49129d01cf668193b3af89abe3bd5c8e0dcc

along with this commit that changed the default timeout to 5 seconds: https://github.com/spring-cloud/spring-cloud-stream/commit/a20d151217c1141bbfc91d9f5cb4e7140d130acb

PR #89, opened 2015-08-18
Add cloudfoundry cloud-connector

markfisher:


spring-cloud-stream-binders/spring-cloud-stream-binder-redis/pom.xml
```diff
@@ -44,6 +44,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-cloudfoundry-connector</artifactId>
```
ilayaperumalg: I believe this cf-connector dependency needs to be added to rabbit binder (not sure if we have kafka as cf service) as well. May be, this can be moved up to the binder parent?

ericbottard: In general I'd refrain from "factoring out" dependencies. I think it makes sense for s-c-s-modules up to the point where every pulled up dep is 100% sure to be used/useful.
In the particular case you describe, the cf stuff may well be useless to future binders, so I'd -1

markfisher: Agreed. Having this explicit in each binder that is supported by cloud-connector/service-binding is fine. That said, we should go ahead and update the rabbit binder POM as we're doing this for redis.

ericbottard:
And so, dumb question, but ... what's the point of this? Does that artifact contain an autoconf that would kick in?
I'd be happy to merge this (and add to rabbit as well), but I'm not sure how I am supposed to test

mbogoevici:
It adds support for discovering services on Cloud Foundry - it provides support for CF when the cloud profile is activated by the build pack.

@markfisher and I have tested this particular change on Tuesday for Redis (but refrained from merging so that the docker image remains unaffected - had a demo at the same time). Testing means deploying time source and log sink in CF (e.g. PEZ), so the test for rabbit could be symmetric.  

One thing to particularly test locally is that the change doesn't break lattice (shouldn't).

ericbottard:
Verified this works on CF (after a small hiccup). If you don't hear from me in the next couple of hours and really want this in, then the remaining part is lattice

ilayaperumalg:
This can now be closed as the `cloud-connector` dependencies are removed from `s-c-s` and added to `s-c-s-modules`: https://github.com/spring-cloud/spring-cloud-stream/pull/104

PR #91, opened 2015-08-19
Add SmartLifecycle to ChannelBindingAdapter

ilayaperumalg:
- Also set Integer.MIN_VALUE as the phase value for CBA that enables this lifecycle bean to start
before any other message producing lifecycle beans


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/config/ChannelBindingAdapterConfigurationTests.java
```diff
@@ -82,6 +82,7 @@ private void refresh() {
 			channel.setTapped(true);
 		}
 		this.adapter.setOutputChannels(channels);
+		this.adapter.stop();
```
ericbottard: Shouldn't this be just one line above, before `setOutputChannels() ` ?

ilayaperumalg: Actually, it doesn't matter here. All we need is to let the binding happens (when we restart the adapter) so that the appropriate binding names are set for the output channel.

markfisher: True, Ilaya, but I agree with Eric if for nothing but readability.

ilayaperumalg: Sure, I updated the PR :-)

PR #92, opened 2015-08-19
Remove build log noise, take 2

ericbottard:
Re-issuing against s-c-s as this did not make it to s-c-build.
Set default to go to file, override in travis.yml, as travis does not
allow to browse the build directory. Feel free to not commit that bit
if I got it wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,11 @@ before_install:
   - echo "https://$GH_TOKEN:@github.com" > .git/credentials
   - gem install asciidoctor
 install:
-- ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=true
+- ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=false
 - ./docs/src/main/asciidoc/ghpages.sh
 script:
-- '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw --settings .settings.xml deploy -P release -nsu -Dmaven.test.redirectTestOutputToFile=true'
-- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] || ./mvnw install -nsu -Dmaven.test.redirectTestOutputToFile=true'
+- '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw --settings .settings.xml deploy -P release -nsu -Dmaven.test.redirectTestOutputToFile=false'
+- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] || ./mvnw install -nsu -Dmaven.test.redirectTestOutputToFile=false'
 env:
   global:
   - GIT_NAME="Dave Syer"
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,26 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
+	<build>
+		<pluginManagement>
+			<plugins>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-javadoc-plugin</artifactId>
+					<configuration>
+						<quiet>true</quiet>
+					</configuration>
+				</plugin>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -14,11 +14,11 @@ before_install:
   - echo "https://$GH_TOKEN:@github.com" > .git/credentials
   - gem install asciidoctor
 install:
-- ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=true
+- ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=false
```
markpollack: I'm not sure I understand how changing this to `false` is going to reduce log noise, since redirecting test output to a file vs. stdout would seem to be the correct settings.

.travis.yml
```diff
@@ -14,11 +14,11 @@ before_install:
...
 - ./docs/src/main/asciidoc/ghpages.sh
 script:
-- '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw --settings .settings.xml deploy -P release -nsu -Dmaven.test.redirectTestOutputToFile=true'
-- '[ "${TRAVIS_PULL_REQUEST}" = "false" ] || ./mvnw install -nsu -Dmaven.test.redirectTestOutputToFile=true'
+- '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw --settings .settings.xml deploy -P release -nsu -Dmaven.test.redirectTestOutputToFile=false'
```
markpollack: I think this PR is about changing true->false, but https://github.com/spring-cloud/spring-cloud-stream/commit/397ef6f5571b877ae5a4b12e6f1a55bf66d57677#diff-354f30a63fb0907d4ad57269548329e3 was done to change the maven target not to publish (as we have a bamboo job for that....)   So once we get the true/false value cleared up, I can make the appropriate changes.

ericbottard:
The aim is to reduce the noise when building on a terminal/laptop (hence redirecting to file by default).
But then logs would be lost on travis, as there is no way to access files created by the build there, hence switching back to false (ie dump on stdout) fro travis build.
Hope this makes sense

markpollack:
Sorry, getting back to this after a long delay.  As we are building/releasing s-c-s with Bamboo  ( https://build.spring.io/browse/SCS-SMJLE ) I do not think we need the `'[ "${TRAVIS_PULL_REQUEST}" = "false" ]` condition added. 

I think we need to 

1) change the existing [line](https://github.com/spring-cloud/spring-cloud-stream/blob/master/.travis.yml#L20) to
```
 '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw package -nsu -Dmaven.test.redirectTestOutputToFile=true
```

2) keep the changes you made to add javadoc/surefire plugins.

3) Remove the spring profile from pom.xml  as we won't release using that profile.

This would give us the integration with travis to run a build against a PR but let the snapshot and release builds be managed via Bamboo.

.travis.yml
ericbottard: The intent of this PR is to reduce noise when you're looking at it (ie laptop, stdout).
But one can't look at other files on travis, so this needs to be on stdout for travis (although too long files seem to be a problem as well there).

ericbottard:
> I think we need to
>1) change the existing line to

Yep, can even remove the PR/notPR switch all together, as we won't be deploying at all from travis. This to use travis as a build success indicator only.

ericbottard:
Actually, I think we can simplify to this (which I submitted to s-c-d as well):
https://github.com/ericbottard/spring-cloud-dataflow/blob/Travis/.travis.yml

```
sudo: false
cache:
  directories:
    - $HOME/.m2
language: java
jdk:
  - oraclejdk8
services:
  - redis-server
install: true
script:
  - ./mvnw -s .settings.xml package -nsu
```

One thing I don't quite get anyhow, is how the asciidoc / github pages thing is protected to only publish what's on master (and not on branches)

markpollack:
Yes, seems we can remove even more, but wasn't the inital point to get the logging set to `-Dmaven.test.redirectTestOutputToFile=false` so that it appears in the stdout when browsing travis on the web?

ericbottard:
The **very** initial point was to have redirectTestOutputToFile=true for the common case. Then added `=false` for travis

PR #93, opened 2015-08-19
Add a converter String->Expession

ericbottard:
Modules that want to use this just have to import the configuration

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpelExpressionConverterConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpelExpressionConverterConfiguration.java
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.stream.support;
+package org.springframework.cloud.stream.config;
 
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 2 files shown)


markfisher:
Can this go in the 'config' package rather than creating a new one?

ericbottard:
Sure thing, did not want to pollute the thing related to pure binding.

markfisher:
Ultimately, I think this should go into the default converters used by the BeanFactory. Maybe issue a PR for that?

ericbottard:
Will do. I don't see how it could hurt indeed

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpelExpressionConverterConfiguration.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2015 the original author or authors.
```
markfisher: need an empty line after the "Copyright..."

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/2ab2d5c1a904a99929ec5f3d4f44cd6b369677f1

PR #94, opened 2015-08-20
Core tests

mbogoevici:

PR #96, opened 2015-08-20
Add binding tests

mbogoevici:
An evolution of https://github.com/spring-cloud/spring-cloud-stream/pull/94

- Adds binding tests for Source, Sink, Processor and arbitrary interfaces
- while quite repetitive, the three interfaces above will be used quite extensively so deserve individual testing;
- Fixes an issue where using @Input/@Output without a value caused a bean registration failure;
- Adds test for channel interception;

### DIFF
--- a/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/ArbitraryInterfaceBindingTestsWithBindingTargets.java
+++ b/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/ArbitraryInterfaceBindingTestsWithBindingTargets.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.binder;
+
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+import java.util.Properties;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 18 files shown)


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/FooChannels.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * @author Marius Bogoevici
+ */
+public interface FooChannels {
+
+	@Input
+	MessageChannel foo();
+
+	@Input//(value = "bar")
```
ericbottard: Remove. Or else, why are those comments here and not on the two others?

mbogoevici: Whoops, slip.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/ArbitraryInterfaceBindingTestsWithBindingTargets.java
```diff
@@ -0,0 +1,70 @@
...
+	public void testArbitraryInterfaceChannelsBound() {
+		verify(binder).bindConsumer(eq("someQueue.0"), eq(fooChannels.foo()), Mockito.<Properties>any());
+		verify(binder).bindConsumer(eq("someQueue.1"), eq(fooChannels.bar()), Mockito.<Properties>any());
+		verify(binder).bindProducer(eq("someQueue.2"), eq(fooChannels.baz()), Mockito.<Properties>any());
+		verify(binder).bindProducer(eq("someQueue.3"), eq(fooChannels.qux()), Mockito.<Properties>any());
+		verifyNoMoreInteractions(binder);
+	}
+
+	@EnableModule(FooChannels.class)
+	@EnableAutoConfiguration
```
ilayaperumalg: I believe the changes in this [PR] (https://github.com/spring-cloud/spring-cloud-stream/pull/98) would remove the need to do `EnableAutoConfiguration` here.

mbogoevici: Well, I was actually trying to reproduce the behaviour of `@SpringBootApplication` - which cannot be used because it scans the entire package.

mbogoevici: If we can do without the AutoConfiguration, all the better.

ericbottard: I think your first comment is more to the point @mbogoevici : we should mimic what a real module would do, as this is really an integration test. Merging as is, please shout if you disagree.

ericbottard:
Merged as squashed 2e4fd84

PR #97, opened 2015-08-20
Fix copyrights

ericbottard:


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/IntegerEncoderDecoder.java
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
```
markfisher: should always be one space after the header, before the package declaration

can you please regenerate this PR? (also we've had some additions with more violators since this was submitted)

ericbottard: Hmmm, turns out the license plugin I used does not add a new line, even if I put it there. What a shame. gonna do some `sed`ding

ericbottard:
Fixed linefeeds, also removed added headers on `package-info.java`

markfisher:
thanks Eric! merged

PR #98, opened 2015-08-20
Annotate channelbinding properties instead of autoconfig

ilayaperumalg:
- Autowire `ChannelBindingProperties` into `ChannelBindingConfiguration` via
`@EnableConfigurationProperties` support
  - This change removes explicit `@AutoConfiguration` for ChannelBindingProperties


ericbottard:
This *does* change the semantics a little, although I'm not sure that level of support was needed (and pretty sure we don't have it in similar situations): your PR always creates a properties bean unless one with the same (default) name already exists, while previously it would add it unless one (_by type_) was not already present. 

In a nutshell, to override:
* before your PR: just add a bean, whatever its name
* after your PR: must add with exact default name, which is not a "public" feature, as far as I know

cc @dsyer

dsyer:
I would go with the more usual idiom (i.e. accept the PR) and see what could break. The naming convention for the bean name is a necessary evil I believe in general (although individual beans can break it if they really need).

ericbottard:
Will merge then

ericbottard:
Merged as b0beec8

PR #99, opened 2015-08-20
Use @ as maven resource filtering token

ericbottard:
Trivia: what is the name of the binding in this module:
```
delay: 5
name: foo
spring:
  cloud:
    stream:
      bindings:
        input: ${name}
```

You got it wrong, it's not `foo`, it's whatever the `${project.name}` of the maven project is, because of maven filtering. This PR stops this non-sense by forcing the use of `@name@` if we actually want maven filtering

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,24 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
+	<build>
+		<pluginManagement>
+			<plugins>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-resources-plugin</artifactId>
+					<configuration>
+						<!-- Prevents ${foo} placeholders from being interpolated by maven in application.yml-->
+						<delimiters>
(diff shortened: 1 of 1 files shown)


markfisher:
I would prefer to reach consensus with boot.

dsyer:
Boot makes a recommendation that you use `@` as a delimiter, but doesn't enforce it anywhere unless you are using the `spring-boot-starter-parent` (which libraries like s-c-stream should not use). I think it makes perfect sense to copy the resource configuration from the boot parent into the stream parent.

ericbottard:
OK to close, but
> Boot makes a recommendation that you use @ as a delimiter

is not enough IMO, because the problem is more about keys that you *did not* expect to be replaced (_i.e._ you think boot interpolation at runtime, but you get accidental maven interpolation at buildtime)

markfisher:
closing for now, since we don't have key conflicts at the moment and we've removed all the `application.yml` files (and will avoid them)

PR #102, opened 2015-08-21
Use the required autoconfiguration in test

ilayaperumalg:
- Instead of using `@EnableAutoConfiguration` use `PropertyPlaceholderAutoConfiguration` which is all required for the property resolution


ilayaperumalg:
Merged as 4109949a9258ed7e352db67a6433128ffeccb47d

PR #104, opened 2015-08-22
Remove cloud connector dependencies from S-C-S

ilayaperumalg:
- Remove lattice and cf connector dependencies from core S-C-S and binder
  - The idea is to have these dependencies defined in s-c-s-modules so that they
are specified in one place


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/1e8989a45abc789119d6326a911908b18332549b

PR #107, opened 2015-08-25
XD-3414: @Rule RedisAvailable & friends

ericbottard:
Rename s-c-s-test-support -> s-c-s-modules-t-s
Created s-c-s-test-support to include Redis, Rabbit & Kafka JUnit rules
Remove similar classes which were in binder test projects
Replace with new ones


ericbottard:
@pperalta Once this is merged in, we can apply to s-c-data:RedisStreamDefRepoTests as well as to some modules

spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/junit/AbstractExternalResourceTestSupport.java
```diff
@@ -0,0 +1,133 @@
...
+ * Abstract base class for JUnit {@link Rule}s that detect the presence of some external resource. If the resource is
+ * indeed present, it will be available during the test lifecycle through {@link #getResource()}. If it is not, tests
+ * will either fail or be skipped, depending on the value of system property {@value #XD_EXTERNAL_SERVERS_REQUIRED}.
+ *
+ * @author Eric Bottard
+ * @author Gary Russell
+ */
+public abstract class AbstractExternalResourceTestSupport<R> implements TestRule {
+
+    public static final String XD_EXTERNAL_SERVERS_REQUIRED = "XD_EXTERNAL_SERVERS_REQUIRED";
```
markfisher: should avoid "XD"

markfisher:
I don't like the name `spring-cloud-stream-modules-test-support` within this project, because it should not know anything about `spring-cloud-stream-modules` which is a higher-level project that builds on this. I know it's intended to be a general name to differentiate from the "internal" test support for `spring-cloud-stream`'s own testing, but perhaps we can find better name(s) to differentiate. Ideally I'd preserve `spring-cloud-stream-test-support` for the one developers use when building apps based on `spring-cloud-stream` since that would be most intuitive. So, can we find a better name for the internal test support? (could possibly use "internal' in fact, but I bet there's something better still).

ericbottard:
> could possibly use "internal' in fact, but I bet there's something better still)

Especially given it's not meant to be *that* internal (_i.e._ it can benefit module authors as well)


`-junit`? (meh)

markfisher:
If it's beneficial to developers as well, then what is the rationale for 2 separate test-support sub-projects?

By the way, I think we should refrain from thinking of "module" at the `spring-cloud-stream` level at all since simply using `spring-cloud-stream` does not necessary imply the result will be used as a "module", e.g. in the `spring-cloud-data` sense.

ericbottard:
Because the 'module' support part is not for this project, nor is it for
s-c-d.

ericbottard:
I hope my mast comment made sense (sorry, quickly sent from my AZERTY phone). What I meant is that stuff like MessageCollector is of no use to s-c-s itself, nor is it to s-c-data.
Obviously, we _could_ merge those two together (especially given they end up under the broad spring-cloud groupId, but you get the idea)

pperalta:
Once this is merged, will we include `spring-cloud-stream-test-support-internal` as a test dependency of `spring-cloud-data`? Should this be part of SCS or should it be its own top level project?

markpollack:
Just to mention wrt to the moving of stuff around to other projects - looks like we can't put the abstract base class into spring-test, so we have to live with some duplication across SI, spring data redis, spring amqp etc.  So we should address issues wrt to naming etc, but otherwise merge this code into the s-c-s repo

pperalta:
Merged as 7023e161967da795ab5f94fde6b4a5964d8c1aa2

PR #108, opened 2015-08-25
Pre-register default bindings for channels

mbogoevici:
Pre-setting defaults via a PropertySource also allows them to be properly overridden by environment variables

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
@@ -18,11 +18,17 @@ package org.springframework.cloud.stream.config;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Properties;
 
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
 import org.springframework.cloud.stream.annotation.EnableModule;
 import org.springframework.cloud.stream.utils.MessageChannelBeanDefinitionRegistryUtils;
+import org.springframework.context.EnvironmentAware;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.ImportBeanDefinitionRegistrar;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.PropertiesPropertySource;
 import org.springframework.core.type.AnnotationMetadata;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.MultiValueMap;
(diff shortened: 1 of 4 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
```diff
@@ -31,19 +37,41 @@
...
-			MessageChannelBeanDefinitionRegistryUtils.registerChannelBeanDefinitions(type, registry);
+			registeredChannelNames.addAll(MessageChannelBeanDefinitionRegistryUtils.registerChannelBeanDefinitions(type, registry));
 			MessageChannelBeanDefinitionRegistryUtils.registerChannelsQualifiedBeanDefinitions(
 					ClassUtils.resolveClassName(metadata.getClassName(), null), type,
 					registry);
 		}
+		Properties defaultChannelNameProperties = new Properties();
+		for (String registeredChannelName : registeredChannelNames) {
+			defaultChannelNameProperties.put(SPRING_CLOUD_STREAM_BINDINGS_PREFIX + "." + registeredChannelName,
+					"${spring.application.name:spring.cloud.stream}" + "." + registeredChannelName);
```
ilayaperumalg: Now that `default` binding names are always registered, can we remove the default name handling from `ChannelBindingProperties`'s getBindingPath(String channelName) ?

mbogoevici: Yeah, we should.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
```diff
@@ -31,19 +37,41 @@
...
+		}
+		if (environment != null) {
```
ilayaperumalg: This `null` check make me believe the `default` binding names are not **always** set. I believe we always have a `ConfigurableEnviornment` from spring-boot and that assumption would probably remove the null check here and the the environment setter.

ilayaperumalg: Based on [this] (https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L305), I think we can get rid of the null check and directly assign the environment to `ConfigurableEnvironment`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
mbogoevici: I've kept it around because we also had ChannelBindingAdapterConfigurationTests, which doesn't participate in the normal module lifecycle.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
mbogoevici: Yeah, makes sense.

mbogoevici: Actually, I wonder if we should assume that this is always running in Boot, or a more defensive approach (which is what I initially had in mind) would be better.

ilayaperumalg: I think it's a valid assumption to be running `s-c-s` in Boot as we have a very tight coupling to it (e.g: usage of ConfigurationProperties). That also means we expect the module developer writing a spring boot application for the module. Even if the module author writes a raw spring application, [I see that would also set `ConfigurableEnvironment`] (https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java#L303)

mbogoevici: Ok, that makes sense.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
```diff
@@ -31,19 +37,37 @@
  * @author Marius Bogoevici
  * @author Dave Syer
  */
-public class ModuleRegistrar implements ImportBeanDefinitionRegistrar {
+@Configuration
```
ilayaperumalg: This class doesn't necessarily need to be be a `Configuration` class. If the intention to make this a `Configuration` class is to invoke aware methods, then I don't think that is needed as the `ImportDefinitionRegistrar`'s *aware methods are invoked [when the registrar is imported] (https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/main/java/org/springframework/context/annotation/ConfigurationClassParser.java#L501).

ilayaperumalg:
LGTM except the one comment about `@Configuration` on ModuleRegistrar.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ModuleRegistrar.java
mbogoevici: Correct - that's a slip from an earlier attempt.

ilayaperumalg:
Merged 83123c3b97ee731e448d99df26b0f0bf8e517c81 after removing `@Configuration` from ModuleRegistrar.

PR #110, opened 2015-08-26
Use the full Environment for module args

ericbottard:
This PR does a couple of things:
* Use the full environment for module arguments lookup, as some environments
such as CF don't allow passing (lengthy) command line args
* Extract the logic for namespacing those arguments to a dedicated component
with 2 methods that are the inverse of each other


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleWithArguments.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * Encapsulates a reference to a module (as maven coordinates) and a set of "arguments" that must be passed to it.
+ * Those arguments will eventually be passed to the module by the launcher, using any way appropriate.
+ *
+ * @author Eric Bottard
+ */
+public class ModuleWithArguments {
```
markfisher: should rename this (in general avoid *prepositions* and *conjunctions* in class names)

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ArgumentsNamespacingStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+ * Encapsulates the operation of isolating arguments for different modules. Provides two operations that are the
+ * inverse of each other. That is, the following pseudo invariants must hold:
+ * <pre>
+ *     props == unqualify("foo", qualify("foo", props))
+ *     unqualify("foo", qualify("bar", props)) == empty
+ * </pre>
+ *
+ * @author Eric Bottard
+ */
+public interface ArgumentsNamespacingStrategy {
```
markfisher: Is a strategy interface necessary? What other implementation do you envision? It seems like a core part of the contract that would not be changeable once in use.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/DefaultArgumentsNamespacingStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.boot.bind.PropertiesConfigurationFactory;
+import org.springframework.core.env.PropertySources;
+
+/**
+ * A simple implementation of {@link ArgumentsNamespacingStrategy} that simply prefixes each module
+ * argument with '{@literal <module key>.}'.
+ *
+ * <p>As an example, this is how one could specify the {@literal format} argument to the canonical time source:
+ * <pre>
+ *     org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT.format=hh:mm:ss
```
markfisher: I was hoping we could get away from the full module coordinate string being included. Also, it wouldn't be sufficient in the case of something like `"http | filter | filter | file"`. For any given set of modules being launched together, we probably need to instead qualify by the simple artifactId *and* index.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
```diff
@@ -34,6 +47,21 @@
 	 */
 	private String[] modules;
 
+	private List<ModuleWithArguments> modulesWithArguments = new ArrayList<>();
+
+	private PropertySources propertySources;
```
markfisher: something doesn't seem right about having PropertySources themselves in ConfigurationProperties

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
```diff
@@ -43,4 +71,16 @@ public void setModules(String[] modules) {
 		return modules;
 	}
 
+	@PostConstruct
+	public void deduceModuleArguments() throws Exception {
```
markfisher: I'd expect this logic to exist in the ModuleLauncher itself.

spring-cloud-stream-module-launcher/src/test/java/org/springframework/cloud/stream/module/launcher/DefaultArgumentsNamespacingStrategyTest.java
```diff
@@ -0,0 +1,62 @@
...
+		String module = "org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT";
+		String otherModule = "org.springframework.cloud.stream.module:log-sink:1.0.0.BUILD-SNAPSHOT";
+
+		Map<String, ?> qualified = strategy.qualify(module, Collections.singletonMap("foo", "bar"));
+		MutablePropertySources sources = new MutablePropertySources();
+		sources.addFirst(new MapPropertySource("props", (Map<String, Object>) qualified));
+		assertThat(strategy.unqualify(otherModule, sources), is(Collections.<String, String>emptyMap()));
+	}
+
+}
```
markfisher: newline

markfisher:
I would rename `ModuleWithArguments` to something like `ModuleLaunchRequest`. Then I would remove the strategy interface altogether, and push the logic for qualifying and unqualifying into either the `ModuleLauncher` or the `ModuleLaunchRequest`. The `ModuleLauncherProperties` would then revert to being a relatively simple holder of properties.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ArgumentsNamespacingStrategy.java
ericbottard: Yes, I realized that once I got it in. We _may_ want something about escaping the colons (but from separate comment, they may go altogether anyway), or turning things uppercase.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/DefaultArgumentsNamespacingStrategy.java
ericbottard: Seems what we'd want is something that would directly be the "label" from XD/s-c-d when used in that context.
What about turning modules into a `Map<label, mavenCoordinates>` and arguments into `Map<label, Map<String, String>>` ?

spring-cloud-stream-module-launcher/src/test/java/org/springframework/cloud/stream/module/launcher/DefaultArgumentsNamespacingStrategyTest.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class DefaultArgumentsNamespacingStrategyTest {
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testInverse() {
+		DefaultArgumentsNamespacingStrategy strategy = new DefaultArgumentsNamespacingStrategy();
+		String module = "org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT";
+
+		Map<String, ?> qualified = strategy.qualify(module, Collections.singletonMap("foo", "bar"));
```
mbogoevici: We should test for the expected format of namespaced properties too (right now there's no way of telling whether any qualification does does actually take place).

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
ericbottard: I put it here (and so this also justifies this class seeing PropertySources) because then this class is fully responsible for converting from Env to modules + args. IF `@ConfigurationProperties` somehow magically could bind dynamic properties from Maps of which I don't know the keys until runtime, this would "simply" be a `@ConfigurationProperties` class. Moving the `deduceModuleArguments()` logic to the launcher feels like we're leaving the `modules` bit behind. Even more so if it becomes a `Map<label, coordinates>` as expressed in another comment, don't you think?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
```diff
@@ -45,15 +48,16 @@
 	private ModuleLauncher moduleLauncher;
 
 	@Override
-	public void run(ApplicationArguments applicationArguments) throws Exception {
-		String[] launchedModules = moduleLauncherProperties.getModules();
+	public void run(String... unused) throws Exception {
```
mbogoevici: Probably shouldn't call it `unused`.

ericbottard: My point was exactly to showcase the fact that we're not relying on just the command-line args. I'd settle for something that doesn't even have them if I was sure about the lifecycle (notice how I downgraded from ApplicationRunner already).
I can replace this with a comment if that's ok with you

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
```diff
@@ -45,15 +48,16 @@
...
+		List<ModuleWithArguments> modulesToRun = moduleLauncherProperties.modulesWithArguments();
 		if (log.isInfoEnabled()) {
-			log.info("Launching: "
-					+ StringUtils.arrayToCommaDelimitedString(launchedModules)
-					+ " with arguments: "
-					+ StringUtils.arrayToCommaDelimitedString(applicationArguments
-							.getSourceArgs()));
+			StringBuilder sb = new StringBuilder(modulesToRun.size() * 100);
```
mbogoevici: We can leave the argument out - it improves performance a little bit, but it makes the code harder to read.

spring-cloud-stream-module-launcher/src/test/java/org/springframework/cloud/stream/module/launcher/DefaultArgumentsNamespacingStrategyTest.java
ericbottard: Actually, I think the combination of the 2 tests is enough to prove the behavior, from a black box point of view. We should not care about the details of how the namespacing happens (especially if this becomes a non-strategy).

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleWithArguments.java
mbogoevici: Would it be possible/desirable to use a `MultiValueMap` instead?

ericbottard:
> Then I would remove the strategy interface altogether, and push the logic for qualifying and unqualifying into either the ModuleLauncher or the ModuleLaunchRequest

The first motivation for this PR is that I want to use the thing that does the qualifying/unqualifying in the CloudfoundryDeployer (so I'd rather have it separate from the launcher itself)

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
mbogoevici: That's not entirely correct, though. They're not used in the context of this method. The application arguments would still be parsed into the Environment as far as I can tell (and that's something we want). It's just that right now we process them in a different way.

spring-cloud-stream-module-launcher/src/test/java/org/springframework/cloud/stream/module/launcher/DefaultArgumentsNamespacingStrategyTest.java
mbogoevici: Well, since anyone can invoke the module launcher independently, I think that having the namespacing format explained and illustrated clearly would help with usage.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleWithArguments.java
ericbottard: I think by the time we get here, argument values that are themselves multi-valued have already been "stringfied".

markfisher:
We do not want spring-cloud-data to depend directly on spring-cloud-stream, so reusing this in deployers should not be a goal. We can certainly have reciprocal changes there though.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
ericbottard: Actually, what if boot could do what we want?
What about mapping the equivalent of this yaml form:
```yaml
modules
  foo:    // module "label"
    coordinates: org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT
    arguments:
      foo: bar
      quizz: fix
```

markfisher: very interesting

ericbottard:
It turns out the CF deployer __will__ depend on the launcher (for other reasons, as this is what is actually deployed). But if we don't want code dependency (why?) and rely on "documented conventions" on how the launcher behaves, then we might as well skim most of this PR (as the inverse operation `qualify()` is not needed here)

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
ericbottard: I think this makes sense, but I'm afraid we may hit the "map with environment variables" problem (which ought to be fixed, IIUC)

mbogoevici: One thing about the current `modules` property is ordering - which for becomes very important for direct binding. 
How about we do this: 
`MODULES=[foo]coordinates1,[bar]coordinates2,[baz]coordinates3` with labeling being optional and defaulted to artifact/position (e.g `time-source_0`)
And then arguments can come in the form 
`FOO_FOO=bar` or `TIME_SOURCE_0`.

markfisher:
why? clean separation and a well-defined contract with the smallest possible surface area of shared knowledge

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
markfisher: Actually, I think we don't want the concept of label to propagate down to the launcher. That was the rationale behind 2 corresponding arrays. The order is important, and relying on it, no label is necessary.

ericbottard: What about the equivalent of yaml above + an explicit `modulesKeys` which is a list of labels:

```yaml
modules
  log:
    coordinates: org.springframework.cloud.stream.module:log-sink:1.0.0.BUILD-SNAPSHOT
    arguments:
      thing: 1
  time:    // module "label"
    coordinates: org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT
    arguments:
      quizz: fix
```

```MODULES_KEYS=time,log```

Comment (ericbottard) [inline spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java]: > Actually, I think we don't want the concept of label to propagate down to the launcher. That was the rationale behind 2 corresponding arrays. The order is important, and relying on it, no label is necessary.

We need some pivot key of some kind, and the coordinates are not enough (because in `http | filter | f2: filter | log` the expression `org.springframework.cloud.stream.module:filter-processor:1.0.0.BUILD-SNAPSHOT.expression=payload.length()>3` is ambiguous. So I'm not saying that the label "concept" should perspire down here, just that it happens to be a valid pivot key. These keys have no purpose/meaning other that precising which arguments go where

Comment (ericbottard) [inline spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java]: Note that the index alone is a valid pivot key.
Moreover, if we want to rely on boot to help us here, I reckon we want some known prefix for the arguments.

What about 
```
MODULES=coordinates:of:log,coordinates:of:filter,coordinates:of:time  // (notice the reversed order, which may be counter intuitive btw)
ARGUMENTS_0_LOGGER_NAME=foo
ARGUMENTS_1_EXPRESSION=true
ARGUMENTS_2_PATTERN=hh:ss
ARGUMENTS_2_FIXED_DELAY=4
```

I reckon this would bind to

```
private String[] modules;

private Map<String, Map<String, String>>> arguments;
```

Comment (mbogoevici) [inline spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java]: That would make a lot of sense and greatly simplify things - not sure why the order reversal?

Comment (ericbottard) [inline spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java]: Didn't you say order was important? I assumed this was for consumer _then_ producer reasons. Note that the launcher, even it its current form, just iterates on that array. Or has it been reversed elsewhere already?

Comment (mbogoevici) [inline spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java]: The `ModuleLauncher` can iterate the array in any order it wants. For an external user, I would suppose it would be more intuitive to say
`MODULES=time,transformer,transformer` than `MODULES=transformer,transformer,time`.
It just reads better that way, that is all.

Comment (ericbottard) [inline spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java]: I agree, but __you__ said
> One thing about the current modules property is ordering - which becomes very important for direct binding. 

which, on second thought I actually don't understand (relation to direct binding). 

Also, sad news, it seems my example above does not bind properly (again, I think because of the _ issue)

Comment (mbogoevici) [inline spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java]: If we know the specific order of the modules, we can handle the basic single-flow Source/Processor(s)/Sink easily - we can do that as a separate discussion though :). Let's just assume that we want to preserve ordering :)

Summary of a discussion with @ericbottard: it may be possible to rely on `[]` or `.` in environment variables for CF, as the format restrictions apply to shell only (see http://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names).

**Comment by mbogoevici** [comment]
LTGM in principle - this will require a change in spring-cloud-data in the way the module properties are sent.

File: spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java [hunk 10]
```diff
@@ -43,4 +75,21 @@ public void setModules(String[] modules) {
...
 
+	public void setArguments(Map<Integer, Map<String, String>> arguments) {
+		this.arguments = arguments;
+	}
+
+	public Map<Integer, Map<String, String>> getArguments() {
+		return arguments;
+	}
+
+	public List<ModuleLaunchRequest> asModuleLaunchRequests() {
```
markfisher: I still feel like this logic should be moved out, probably into the ModuleLauncher.

ericbottard: I disagree, IMO this class alone should know about the details.
Moreover, the launcher typically deals with one module at a time, so would not know about the module index.

If we follow your recommendation, the the launcher is dealing with low level maps, and we can get rid of ModuleLaunchRequest

markfisher: Sorry, `ModuleLauncher` would not be the right place; I meant *`ModuleLauncherRunner`*. I agree that `ModuleLauncher` should take the `ModuleLaunchRequest` objects, but I think this class should be for *properties*, not construction of those launch request objects.

ericbottard:
That makes more sense. Enough for me to accept the compromise :). I updated the PR

ericbottard:
> this will require a change in spring-cloud-data in the way the module properties are sent.

This is why I wanted to introduce a _code_ dependency to this contract. If that's ok with you guys, I think I will at least introduce a helper in s-c-d/deployer-spi that can be used by all interested parties.

mbogoevici:
I think a helper would be nice - I agree with @markfisher about not having code dependencies, but a helper wouldn't introduce any of that (just centralize the contract, IMO)

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -62,23 +62,27 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
...
-		for (String module : modules) {
-			List<String> moduleArgs = new ArrayList<>();
-			for (String arg : args) {
-				if (arg.startsWith("--" + module + ".")) {
-					moduleArgs.add("--" + arg.substring(module.length() + 3));
-				}
-			}
-			moduleArgs.add("--spring.jmx.default-domain=" + module.replace("/", ".").replace(":", "."));
-			launchModule(module, moduleArgs.toArray(new String[moduleArgs.size()]));
+	private String[] toProgramArguments(Map<String, String> args) {
```
markfisher: I know this is a private method, but I would probably rename "toArgArray" (the term "program" feels out of context).

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
```diff
@@ -16,24 +16,48 @@
...
+ * properties, environment variables, program arguments, <i>etc.</i> _):<ul>
+ *     <li>{@literal modules = <list>}: an ordered list of maven coordinates of modules to launch</li>
+ *     <li>{@literal arguments[<index>][<key>] = <value>}: key/value pairs that will become module arguments,
+ *     where {@literal <index>} is the 0-based index of the module in the list above</li>
+ * </ul>
+ *
+ * As an example, this is how one would launch {@literal time --fixedDelay=4 | log} canonical example:
+ * <pre>
+ *     modules = org.springframework.cloud.modules:time-source:1.0.0-SNAPSHOT,org.springframework.cloud.modules:log-sink:1.0.0-SNAPSHOT
+ *     arguments.0.fixedDelay=4
```
markfisher: maybe we can shorten "arguments" to "args" here?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
```diff
@@ -45,15 +47,26 @@
 	private ModuleLauncher moduleLauncher;
 
 	@Override
-	public void run(ApplicationArguments applicationArguments) throws Exception {
-		String[] launchedModules = moduleLauncherProperties.getModules();
+	public void run(String... args) throws Exception {
+		List<ModuleLaunchRequest> modulesToRun = toModuleLaunchRequests(moduleLauncherProperties);
```
markfisher: would change "modulesToRun" to something like "launchRequests" rather than introducing another term (run) while we're using launch for everything else

markfisher:
Agreed a helper in the shared SPI dependency would be fine. My view is that as soon as things get complicated enough that we need to share code between spring-cloud-data and *spring-cloud-stream*, then we know we're too coupled. Having some common reusable code for the various SPI implementation does make sense though.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -62,23 +62,27 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
...
 	 * <code>&lt;groupId&gt;:&lt;artifactId&gt;[:&lt;extension&gt;[:&lt;classifier&gt;]]:&lt;version&gt;</code>
 	 *
-	 * To pass arguments to a module, prefix with the module name and a dot. The arg name will be de-qualified and passed along.
-	 * For example: <code>--org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT.bar=123</code> becomes <code>--bar=123</code> and is only passed to the 'org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT' module.
-	 *
-	 * @param modules a list of modules
-	 * @param args a list of arguments, prefixed with the module name
+	 * @param moduleLaunchRequests a list of modules with their (unqualified) arguments
+	 */
+	public void launch(Iterable<ModuleLaunchRequest> moduleLaunchRequests) {
```
markfisher: We need to decide on the contract with regards to ordering of the modules. Deployment order should be starting with the sink and ending with the source. However, I don't think we are sending in that order, at least not consistently. The other question is if it's more intuitive to make the contract so that this expects the natural order instead (source in the first position, sink in the last). That is more intuitive for the user. It would mean, however, that we need to deploy what is passed in reverse order, and thus `Iterable` might not be the most convenient type for this method parameter.

ericbottard: Good point. Changing to list

ericbottard: And documenting the expected order (user-friendly) + iterating in the reverse order

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -59,26 +62,35 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
 	/**
 	 * Launches one or more modules, with the corresponding arguments, if any.
 	 *
+	 * Modules must be passed in "natural" left to right order, and will be deployed in reverse order
```
markfisher: Should not mention deploying in reverse order here - at this level, that is an implementation detail. ModuleLauncher will encapsulate these modules as a single deployment unit.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
```diff
@@ -45,15 +48,27 @@
...
 		}
-		this.moduleLauncher.launch(launchedModules, applicationArguments.getSourceArgs());
+		this.moduleLauncher.launch(launchRequests);
+	}
+
+	private List<ModuleLaunchRequest> toModuleLaunchRequests(ModuleLauncherProperties moduleLauncherProperties) {
+		List<ModuleLaunchRequest> requests = new ArrayList<>();
+		String[] modules = moduleLauncherProperties.getModules();
+		Map<Integer, Map<String, String>> arguments = moduleLauncherProperties.getArgs();
+		for (int i = 0; i < moduleLauncherProperties.getModules().length; i++) {
```
markfisher: why not just use modules.length here?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
markfisher: Also, looking at the launcher in its current state, it does *not* do anything in reverse order (but that's ok, at least for now). Should just keep this comment focused on the contract of what's expected.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/74d85cd9d2a0823eec0ebaf79461c523e01839a7

PR #111, opened 2015-08-27
[Not ready for merge yet] Remove binding defaults

ilayaperumalg:
- With spring-boot support unresolved property names with `_`,
we don't need explicit property source that registers
the default values for binding channel names.


ilayaperumalg:
This PR requires changes from spring boot PR https://github.com/spring-projects/spring-boot/pull/3836

ilayaperumalg:
Closing this for now until the original issue to to support ConfigurationProperties of type Map<String,Map> with the underscore is fixed.

PR #113, opened 2015-08-28
Channel proxies direct binding

mbogoevici:
- Remove DirectChannelProxyFactory;
- Proxies create channels as necessary, including support for PollableChannels;
- Channel beans are using the bean factory methods for the proxies;
- Coordinate bind/unbind across a context through the internal Bindable interface;
- Add support for embedding modules as independent child contexts of the main module, wrapped in a Bindable interface;
- Add support for channel name namespacing;
- Use a shared channel registry (consulted by the proxies) to create direct bindings;

Test by installing spring-cloud-stream and spring-cloud-stream-modules and then running similar to

java -jar spring-cloud-stream-module-launcher/target/spring-cloud-stream-module-launcher-1.0.0.BUILD-SNAPSHOT.jar --modules=org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT,org.springframework.cloud.stream.module:filter-processor:1.0.0.BUILD-SNAPSHOT,org.springframework.cloud.stream.module:filter-processor:1.0.0.BUILD-SNAPSHOT --args.0.fixedDelay=7 --args.1.expression='payload.contains("6")' --aggregate=true --spring.cloud.stream.bindings.output=filtered

Note that the target queue should be 'filtered' and that the module parameters are applied correctly


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
```diff
@@ -71,4 +71,15 @@ public void run(String... args) throws Exception {
 		}
 		return requests;
 	}
+
+	private String[] filterSpringProperties(String[] args) {
```
ilayaperumalg: This one is yet to be used ;)

mbogoevici: Oops.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -92,6 +119,73 @@ public void launch(List<ModuleLaunchRequest> moduleLaunchRequests) {
...
+					} catch (Exception e) {
+						log.error("Cannot start module group ", e);
+						throw new RuntimeException(e);
+					}
+				}
+			};
+
+			Thread moduleAggregatorRunnerThread = new Thread(moduleAggregatorRunner);
+			moduleAggregatorRunnerThread.setContextClassLoader(classLoader);
+			moduleAggregatorRunnerThread.setName(Thread.currentThread().getName());
```
ilayaperumalg: Would be nice to have a thread name as `ModuleAggregatorRunner`. Currently this will simply be `main`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+				= new SpringApplicationBuilder();
+		aggregatorParentConfiguration.sources(AggregatorParentConfiguration.class)
+				.web(false)
+				.headless(true)
+				.properties("spring.jmx.default-domain=" + UUID.randomUUID().toString().replace("-", ""))
+				.listeners(new UnbindOnCloseApplicationListener());
+		return aggregatorParentConfiguration.run(args);
+	}
+
+	private static LinkedHashMap<String, ConfigurableApplicationContext> createChildContexts(ConfigurableApplicationContext applicationContext, Class<?>[] modules,
```
ilayaperumalg: For better clarity, can we make the `applicationContext` to `parentContext` here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+																																										 String args[][]) {
```
ilayaperumalg: This line can be better formatted.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+			children.put(getNamespace(moduleClassName,i), childContext);
+		}
+		return children;
+	}
+
+	private static String getNamespace(String moduleClassName, int index) {
+		return moduleClassName + "_" + index;
+	}
+
+	private static SpringApplicationBuilder embedModule(ConfigurableApplicationContext applicationContext, String namespace,
```
ilayaperumalg: Same here: rename `applicationContext` to `parentContext` for readability.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+				.parent(applicationContext);
+	}
+
+	private static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] modules) {
+		DirectChannel sharedChannel = null;
+		for (int i = 0; i < modules.length; i++) {
+			Class<?> module = modules[i];
+			String moduleClassName = module.getName();
+			if (i > 0) {
+				sharedChannelRegistry.putSharedChannel(getNamespace(moduleClassName,i) + ".input", sharedChannel);
```
ilayaperumalg: Can we move the `input/output` to constants?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class SharedChannelRegistry {
+
+	private Map<String, MessageChannel> sharedChannels = new ConcurrentSkipListMap<>(String.CASE_INSENSITIVE_ORDER);
+
+	public SharedChannelRegistry() {
+		this(new ConcurrentHashMap<String, MessageChannel>());
+	}
+
+	private SharedChannelRegistry(Map<String, MessageChannel> sharedChannels) {
```
ilayaperumalg: I think this needs to be removed as we already initialize the `sharedChannels`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.messaging.MessageChannel;
+
+/**
+ * A registry for channels that can be shared between modules, used for module aggregation.
+ *
+ * @author Marius Bogoevici
+ */
+public class SharedChannelRegistry {
+
+	private Map<String, MessageChannel> sharedChannels = new ConcurrentSkipListMap<>(String.CASE_INSENSITIVE_ORDER);
```
ilayaperumalg: I think we can make this `final` and get rid of the constructors.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/UnbindOnCloseApplicationListener.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @author Marius Bogoevici
+ */
+public class UnbindOnCloseApplicationListener implements ApplicationListener<ContextClosedEvent>, ApplicationContextAware, Ordered {
+
+	private ConfigurableApplicationContext applicationContext;
+
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+
```
ilayaperumalg: extra line.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingBeanDefinitionRegistryUtils.java
```diff
@@ -48,47 +39,36 @@
...
+				channelInterfaceMethodName, registry);
 	}
 
-	public static void registerOutputChannelBeanDefinition(String name,
-			BeanDefinitionRegistry registry) {
-		registerOutputChannelBeanDefinition(DEFAULT_OUTPUT_QUALIFIER_VALUE, name, registry);
-	}
-	public static void registerOutputChannelBeanDefinition(String qualifierValue, String name,BeanDefinitionRegistry registry) {
-		registerChannelBeanDefinition(Output.class, qualifierValue, name, registry);
+	public static void registerOutputChannelBeanDefinition(String qualifierValue, String name,
```
ilayaperumalg: format issue

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableContextWrapper.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.cloud.stream.annotation.EnableModule;
+import org.springframework.context.ConfigurableApplicationContext;
+
+/**
+ * Exposes a context as a {@link Bindable} component, allowing for embedding {@link EnableModule}
+ * units as child contexts into an application.
+ *
+ * @author Marius Bogoevici
+ */
+public class BindableContextWrapper implements Bindable {
```
ilayaperumalg: I don't think we need this class anymore. Given the bindings happen via `smart-lifecycle`, I believe the BindingUtils is enough for the bindings to happen. What do you think?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+	 * @param modules a list of classes, representing root context definitions for modules
+	 * @param args arguments for the modules
+	 */
+	public static void runEmbedded(ConfigurableApplicationContext parentContext, Class<?>[] modules,
+																 String[][] args) {
+		SharedChannelRegistry bean = parentContext.getBean(SharedChannelRegistry.class);
+		prepareSharedChannelRegistry(bean, modules);
+		// create child contexts first
+		LinkedHashMap<String, ConfigurableApplicationContext> children = createChildContexts(parentContext, modules, args);
+		for (Map.Entry<String, ConfigurableApplicationContext> childContextEntry : children.entrySet()) {
```
ilayaperumalg: Given the `BindingUtils` is triggered via `ChannelBindingLifecycle` and based on the [comment] (https://github.com/spring-cloud/spring-cloud-stream/pull/113/files#r38256747), we can remove the explicit registration of context wrapper here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+				sharedChannelRegistry.putSharedChannel(getNamespace(moduleClassName,i) + ".output", sharedChannel);
+			}
+		}
+	}
+
+	/**
+	 * Basic configuration for a parent
+	 */
+	@EnableAutoConfiguration
+	@EnableModule
```
ilayaperumalg: Related to the discussion we had [before] (https://github.com/mbogoevici/spring-cloud-stream/commit/f6bc3f1ae156f60f08cfb4cc80e76e8dac76d925#commitcomment-12935029), having `@EnableModule` would also make the `ChannelBindingLifecycle` bean to be part of the aggregator parentContext. This also means, every time the parent and child contexts (for the individual modules) are refreshed, we would have this bean triggered. This is harmless as of now because:
1) Parent context will not have any `bindable` beans
2) Each child context will have it's own bindable beans

I still think we can avoid this behavior of starting the binding lifecycle from the parent.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingUtils.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.context.ConfigurableApplicationContext;
+
+/**
+ * Utility class for handling the {@link Bindable}s in a given context.
+ *
+ * @author Marius Bogoevici
+ */
+public class BindingUtils {
+
+	public static void bindAll(ConfigurableApplicationContext context) {
```
ilayaperumalg: It would be nice to set all these static methods' accessor to `package` after moving the class `UnbindOnCloseApplicationListener` to `binding` package.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingUtils.java
```diff
@@ -0,0 +1,66 @@
...
+		bindInputs(context);
+		bindOutputs(context);
+	}
+
+	public static void bindInputs(ConfigurableApplicationContext context) {
+		Map<String, Bindable> bindables = context.getBeansOfType(Bindable.class);
+		for (Bindable bindable : bindables.values()) {
+			bindable.bindInputs(context.getBean(ChannelBindingAdapter.class));
```
ilayaperumalg: Do you think we can have `assertion` to make sure this `bean` exists before proceeding with any bindable operations.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	@Override
+	public boolean isSingleton() {
+		return true;
+	}
+
+	@Override
+	public void bindInputs(ChannelBindingAdapter channelBindingAdapter) {
+		for (Map.Entry<String, ChannelHolder> channelHolderEntry : inputs.entrySet()) {
```
ilayaperumalg: We would probably need `null` check or `assertion` on channelBindingAdapter.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -0,0 +1,316 @@
...
+				.getSharedChannel(getNamespacePrefixedChannelName(name)) : null;
+	}
+
+	private String getNamespacePrefixedChannelName(String name) {
+		return channelNamespace + "." + name;
+	}
+
+	private void bridgeSubscribableToPollableChannel(SubscribableChannel sharedChannel,
+			MessageChannel inputChannel) {
+		sharedChannel.subscribe(new MessageChannelBinderSupport.DirectHandler(
```
ilayaperumalg: Can we create the same  `DirectHandler` in this class? may be that's in the next step when we remove the local binder :-)

ilayaperumalg: MessageChannelBinderSupport is in fact from the SPI. nevermind.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -0,0 +1,316 @@
...
+		consumerEndpointFactoryBean.setPollerMetadata(pollerMetadata);
+		consumerEndpointFactoryBean
+				.setHandler(new MessageChannelBinderSupport.DirectHandler(
+						subscribableChannel));
+		consumerEndpointFactoryBean.setBeanFactory(beanFactory);
+		try {
+			consumerEndpointFactoryBean.afterPropertiesSet();
+		}
+		catch (Exception e) {
+			e.printStackTrace();
```
ilayaperumalg: May be `IllegalStateException(e)` here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: This is what would happen if BindableContextWrapper goes away, right?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableContextWrapper.java
mbogoevici: I think this is something that we may want to come back to  in the future - i.e. the idea that embedded contexts are `Bindable` and managed by the lifecycle of the embedding (i.e. parent) context. Since we aren't using it as such, this may be an unnecessary complication, at this point. I'd like to revisit this at some point.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
ilayaperumalg: Actually I was convinced on having `@EnableModule` :-) My intention in this comment was to avoid calling the lifecycle beans prematurely to avoid any surprises in future. I was thinking to have some kind of exclude (based on some conditions) specifically on this lifecycle bean.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
ilayaperumalg: Moreover, refreshing child context in the above line would call the appropriate channel binding smart lifecycle beans and subsequently binding (via BindingUtils -> BindableProxyFactory). Unless I miss something, the registration of context wrapper is doing nothing here.

mbogoevici: It allows UnboundOnCloseListener to delegate to child contexts for unbinding. But since the child contexts already have a `ChannelBindingListener`, that's most likely not necessary, and can be simplified too.

mbogoevici: Anyway, if we remove the context wrapper as we discussed here, then of course we must get rid of its registration. https://github.com/spring-cloud/spring-cloud-stream/pull/113/files#r38257924

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: OK, got it.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: Actually, this is not as simple as just relying on the `ChannelBindingListeners`, since inputs must be unbound before outputs across the entire aggregate (and this is where the `BindableContextWrapper` helps). I think this needs to be revised though, it's a little too complicated.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
```diff
@@ -58,7 +58,7 @@ public void run(String... args) throws Exception {
 			}
 			log.info(sb.toString());
 		}
-		this.moduleLauncher.launch(launchRequests);
+		this.moduleLauncher.launch(launchRequests, moduleLauncherProperties.isAggregate(), filterSpringProperties(args));
```
ericbottard: I don't understand. Why would `spring.` props get special treatment?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/utils/ClassloaderUtils.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.springframework.cloud.stream.module.utils;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class ClassloaderUtils {
+
+	public static ClassLoader getExtensionClassloader() {
```
ericbottard: Method name is a bit misleading given it may be a different CL returned. Maybe at least document the fallback behavior

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * @author Marius Bogoevici
+ */
+public class ModuleAggregationUtils {
+
+	private static final Log log = LogFactory.getLog(ModuleAggregationUtils.class);
+
+	/**
+	 * Supports the aggregation of {@link Source}, {@link Sink} and {@link Processor} modules by instantiating and
```
ericbottard: Why cite those types (Source, Sink, etc) when this would work with more general inputs/outputs (unless I missed something)?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Embeds a group of modules into an existing parent context
+	 *
+	 * @param parentContext the parent context
+	 * @param modules a list of classes, representing root context definitions for modules
+	 * @param args arguments for the modules
+	 */
+	public static void runEmbedded(ConfigurableApplicationContext parentContext, Class<?>[] modules,
+																 String[][] args) {
```
ericbottard: funky indent

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: In the future, it should work with more general inputs/outputs, but right now it works with only those interfaces. (i.e. it assumes channels named 'input'/'output' - maybe that's what we should say). It would be fair to claim support for more than that,

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
ericbottard: Please document what the map key refers to.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
mbogoevici: I think this is up for debate. What Ilaya has pointed out was that if we pass all arguments to the parent context, then, due to the hierarchical nature of environments they may become visible to the children. `spring.*` props can help with configuring the binding, which is performed at the aggregate (parent) level.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableContextWrapper.java
```diff
@@ -0,0 +1,81 @@
...
+
+	private ConfigurableApplicationContext context;
```
ericbottard: can make final

mbogoevici: I think this class will go away altogether for now.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: In summary, trying to exclude that bean seems to be more trouble than it's worth, unless we explicitly exclude it through a property (which I find more confusing). If someone has a good idea, I'm happy to listen. For now, it doesn't do anything harmful, anyway - and in the future, if there would be any `Bindable`s registered with a context (e.g. in a programmatic embedding scenario), then we'd probably want them to be started as well.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -0,0 +1,316 @@
...
+						method);
+				return this.outputs.get(name).getMessageChannel();
+			}
+		}
+		// ignore
+		return null;
+	}
+
+	@Override
+	public synchronized Object getObject() throws Exception {
```
ericbottard: Would it make sense to separate the Proxy implementation from the `FactoryBean` implementation?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -0,0 +1,316 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Holds information about the channels exposed by the interface proxy, as well as
+	 * their status.
+	 *
+	 */
+	class ChannelHolder {
```
ericbottard: needs to be static

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingAdapter.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * Handles the binding of input/output channels by delegating to an underlying {@link Binder}.
+ *
+ * @author Mark Fisher
+ * @author Dave Syer
+ * @author Marius Bogoevici
+ */
+public class ChannelBindingAdapter  {
```
ericbottard: An `Adapter` usually adapts from one contract A to another contract B (implementing B). This is not the case here. Maybe a better name is needed.

mbogoevici:
Pushed a few changes, removing `BindableContextWrapper` mainly. I think we want to revisit the idea once we get to embedding contexts - I believe that there's something compelling about having an entire context behave like a `Bindable` component with inputs and outputs, but currently this just complicates things.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingLifecycle.java
```diff
@@ -0,0 +1,95 @@
...
+
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext)
+			throws BeansException {
+		this.applicationContext = (ConfigurableApplicationContext) applicationContext;
+	}
+
+	@Override
+	public void start() {
+		if (!running) {
```
ericbottard: I think running needs to be volatile for this to be safe

mbogoevici: Oops.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
mbogoevici: I don't think it would change anything, but I'm happy to make it so.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
mbogoevici: Maybe that's a future improvement. Right now the fact that the proxy delegates to the `FactoryBean` works pretty well.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingAdapter.java
mbogoevici: It's been like this since the beginning, but I'm happy to change it to `ChannelBindingService`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingUtils.java
mbogoevici: Actually, I think that the class should go away.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAdapterConfiguration.java
```diff
@@ -132,10 +85,12 @@ public BinderAwareRouterBeanPostProcessor binderAwareRouterBeanPostProcessor(
 						public MessageChannel resolveDestination(String name)
```
ericbottard: weird indent

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
ericbottard: This would at least save memory (reference to enclosing class), as well as protect from potential leaks. And it's just the right thing to do.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
ericbottard: yes, but they don't have to be the same class. This is rather confusing. This class implements no less than 3 non trivial interfaces.

mbogoevici: Ok, legibility is important. Let's see if there's a better approach.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -68,17 +86,28 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
...
-	public void launch(List<ModuleLaunchRequest> moduleLaunchRequests) {
+	public void launch(List<ModuleLaunchRequest> moduleLaunchRequests, boolean aggregate, String parentArgs[]) {
 		List<ModuleLaunchRequest> reversed = new ArrayList<>(moduleLaunchRequests);
 		Collections.reverse(reversed);
-		for (ModuleLaunchRequest moduleLaunchRequest : reversed) {
-			String module = moduleLaunchRequest.getModule();
-			moduleLaunchRequest.addArgument("spring.jmx.default-domain", module.replace("/", ".").replace(":", "."));
-			launchModule(module, toArgArray(moduleLaunchRequest.getArguments()));
+		if (moduleLaunchRequests.size() == 1 || !aggregate) {
+			launchModulesIndividually(moduleLaunchRequests);
```
markfisher: would be nice if these methods were named consistently, e.g. `launchIndividualModules` and `launchAggregatedModules`

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherRunner.java
markfisher: "filter" is a confusing name here; my first impression was that it would *exclude*

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,152 @@
...
+	public static final String OUTPUT_CHANNEL_NAME = "output";
+
+	/**
+	 * Supports the aggregation of {@link Source}, {@link Sink} and {@link Processor}
+	 * modules by instantiating and binding them directly
+	 *
+	 * @param parentArgs arguments for the parent (prefixed with '--')
+	 * @param modules a list module classes to be aggregated
+	 * @param moduleArgs arguments for the modules (prefixed with '--")
+	 * @return
```
markfisher: need to complete the return comment

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,152 @@
...
+import org.springframework.cloud.stream.annotation.Source;
+import org.springframework.cloud.stream.binding.BindableProxyFactory;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.integration.channel.DirectChannel;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class ModuleAggregationUtils {
```
markfisher: this seems like more than a "Utils" class... it's also not just for aggregation but embedding. Perhaps the name could be something like `ModuleApplication`? (as a specialized alternative to `SpringApplication` in a sense).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingLifecycle.java
markfisher: or a final AtomicBoolean

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAdapterConfiguration.java
```diff
@@ -55,64 +46,23 @@
 @EnableConfigurationProperties(ChannelBindingProperties.class)
 public class ChannelBindingAdapterConfiguration {
```
ilayaperumalg: Given the `ChannelBindingAdapter` class is renamed, can we rename this as well?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAdapterConfiguration.java
```diff
@@ -55,64 +46,23 @@
...
-	public ChannelBindingAdapter bindingAdapter() {
-		ChannelBindingAdapter adapter = new ChannelBindingAdapter(this.module,
-				this.binder);
-		adapter.setChannelLocator(new DefaultChannelLocator(this.module));
-		adapter.setOutputChannels(getOutputChannels());
-		adapter.setInputChannels(getInputChannels());
-		adapter.setChannelResolver(binderAwareChannelResolver());
-		return adapter;
+	@ConditionalOnMissingBean(ChannelBindingService.class)
+	public ChannelBindingService bindingAdapter(ChannelBindingProperties module,
```
ilayaperumalg: the bean method name can be renamed as well for clarity.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	private boolean isChannelPubSub(String channelName) {
+		Assert.isTrue(StringUtils.hasText(channelName),
+				"Channel name should not be empty/null.");
+		return channelName.startsWith("topic:");
+	}
+
+	public void unbindConsumers(String inputChannelName) {
+		this.binder.unbindProducers(inputChannelName);
```
markfisher: shouldn't this be calling unbindConsumers?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -0,0 +1,86 @@
...
+	private ChannelBindingProperties channelBindingProperties;
+
+	public ChannelBindingService(ChannelBindingProperties channelBindingProperties,
+															 Binder<MessageChannel> binder) {
+		this.channelBindingProperties = channelBindingProperties;
+		this.binder = binder;
+	}
+
+
+	public void bindMessageConsumer(MessageChannel inputChannel, String inputChannelName) {
```
markfisher: can these method names be just `bindConsumer

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -92,6 +121,73 @@ public void launch(List<ModuleLaunchRequest> moduleLaunchRequests) {
...
+				mainClassNames.add(jarFileArchive.getMainClass());
+				arguments.add(toArgArray(moduleLaunchRequest.getArguments()));
+			}
+			final ClassLoader classLoader = new LaunchedURLClassLoader(jarURLs.toArray(new URL[jarURLs.size()]),
+					ClassloaderUtils.getExtensionClassloader());
+			final List<Class<?>> mainClasses = new ArrayList<>();
+			for (String mainClass : mainClassNames) {
+				mainClasses.add(ClassUtils.forName(mainClass, classLoader));
+			}
+			Runnable moduleAggregatorRunner = new Runnable() {
```
markfisher: factoring the Runnable to an inner class would improve readability

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -92,6 +121,73 @@ public void launch(List<ModuleLaunchRequest> moduleLaunchRequests) {
...
+						// we expect the class and method to be found on the module classpath
+						Class<?> moduleAggregatorClass = ClassUtils.forName(MODULE_AGGREGATOR_CLASS, classLoader);
+						Method aggregateMethod = ReflectionUtils.findMethod(moduleAggregatorClass,
+								MODULE_AGGREGATOR_METHOD, String[].class, Class[].class, String[][].class);
+						aggregateMethod.invoke(null,
+								parentArgs,
+								mainClasses.toArray(new Class<?>[mainClasses.size()]) ,
+								arguments.toArray(new String[][]{}));
+					} catch (Exception e) {
+						log.error("Cannot start module group ", e);
```
markfisher: would change error message to "failed to start aggregated modules" (rather than use the word "group" here)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: Should.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,152 @@
...
+	/**
+	 * Supports the aggregation of {@link Source}, {@link Sink} and {@link Processor}
+	 * modules by instantiating and binding them directly
+	 *
+	 * @param parentArgs arguments for the parent (prefixed with '--')
+	 * @param modules a list module classes to be aggregated
+	 * @param moduleArgs arguments for the modules (prefixed with '--")
+	 * @return
+	 */
+	public static ConfigurableApplicationContext runAggregated(String[] parentArgs,
```
markfisher: seems like modules should be the first param (especially since there's a version of the method that only has modules as a single param)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingLifecycle.java
mbogoevici: Well, it's volatile already.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,152 @@
...
+
+	private static ConfigurableApplicationContext createParentContext(String[] args) {
+		SpringApplicationBuilder aggregatorParentConfiguration = new SpringApplicationBuilder();
+		aggregatorParentConfiguration
+				.sources(AggregatorParentConfiguration.class)
+				.web(false)
+				.headless(true)
+				.properties(
+						"spring.jmx.default-domain="
+								+ UUID.randomUUID().toString().replace("-", ""));
```
markfisher: is a random domain ok or will we need to provide access? should be other ways to achieve uniqueness here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
```diff
@@ -0,0 +1,152 @@
...
+	/**
+	 * Basic configuration for a parent
+	 */
+	@EnableAutoConfiguration
+	@EnableModule
+	public static class AggregatorParentConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean(SharedChannelRegistry.class)
+		public SharedChannelRegistry sharedChannelRegistry() {
```
markfisher: might be a crazy idea but could the parent context itself (i.e. BeanFactory) *be* the shared channel registry? It's just a Map after all.... or will the parent also contain other beans?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author Marius Bogoevici
+ */
+public class SharedChannelRegistry {
+
+	/**
+	 * A {@link Map} of channels, indexed by name. A channel's name may be prefixed by a namespace.
+	 */
+	private Map<String, MessageChannel> sharedChannels = new ConcurrentSkipListMap<>(String.CASE_INSENSITIVE_ORDER);
+
+	public MessageChannel getSharedChannel(String id) {
```
markfisher: I think having "SharedChannel" in these method names is redundant. Also, as a "registry", should the "put" be "register"? (the get could either remain as-is or be replaced with lookup/find)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -0,0 +1,349 @@
...
+public class BindableProxyFactory implements MethodInterceptor, FactoryBean<Object>,
+		BeanFactoryAware, Bindable, InitializingBean {
+
+	private static Log log = LogFactory.getLog(BindableProxyFactory.class);
+
+	public static final String SPRING_CLOUD_STREAM_INTERNAL_PREFIX = "spring.cloud.stream.internal";
+
+	public static final String CHANNEL_NAMESPACE_PROPERTY_NAME = SPRING_CLOUD_STREAM_INTERNAL_PREFIX + ".channelNamespace";
+
+	public static final String POLLABLE_BRIDGE_FREQUENCY_PROPERTY_NAME = SPRING_CLOUD_STREAM_INTERNAL_PREFIX + ".pollableBridge.frequency";
```
markfisher: maybe "interval" in place of "frequency" (which makes me think cycles-per-N, e.g. hz)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -0,0 +1,349 @@
...
+ * @see org.springframework.cloud.stream.annotation.EnableModule
+ */
+public class BindableProxyFactory implements MethodInterceptor, FactoryBean<Object>,
+		BeanFactoryAware, Bindable, InitializingBean {
+
+	private static Log log = LogFactory.getLog(BindableProxyFactory.class);
+
+	public static final String SPRING_CLOUD_STREAM_INTERNAL_PREFIX = "spring.cloud.stream.internal";
+
+	public static final String CHANNEL_NAMESPACE_PROPERTY_NAME = SPRING_CLOUD_STREAM_INTERNAL_PREFIX + ".channelNamespace";
```
markfisher: seems odd to have the word "namespace" in the actual namespace; maybe just ".channels" ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: Not that crazy - I thought about it, I just liked to have some separation between the actual channel beans installed by the interfaces and the 'shared' channels.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
mbogoevici: Not sure what you mean - that's the name of the property.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
mbogoevici: Yeah ... even that operation is redundant - seeing as in this scenario the shared channel list is immutable.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
markfisher: sorry - nevermind

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -87,5 +87,5 @@ else if (binding instanceof Map) {
 	public String getTapChannelName(String channelName) {
 		return "tap:" + getBindingPath(channelName);
 	}
-
+
```
markfisher: unnecessary whitespace

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/Bindable.java
```diff
@@ -0,0 +1,48 @@
...
+package org.springframework.cloud.stream.binding;
+
+/**
+ * Marker interface for instances that can bind/unbind groups of inputs and outputs.
+ *
+ * Intended for internal use.
+ *
+ * @author Marius Bogoevici
+ */
+public interface Bindable {
```
markfisher: will this evolve when we support more than MessageChannels? (e.g. Rx Observables/Streams)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * Handles the binding of input/output channels by delegating to an underlying {@link Binder}.
+ *
+ * @author Mark Fisher
+ * @author Dave Syer
+ * @author Marius Bogoevici
+ */
+public class ChannelBindingService {
```
markfisher: I understand the rationale for changing this from "Adapter", but I'm not crazy about "Service" here either. This seems to be simply a higher level over the `Binder` since this class is also aware of the properties and thus can expand the name to a path and can determine whether pub/sub. Will it have more responsibilities when other types of `Binder` (besides `MessageChannel`) are added, or will there be a need for other similar higher level objects for dealing with those `Binder`s?

mbogoevici: Maybe that should be pushed down into the binding SPI eventually?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/Bindable.java
mbogoevici: Yes, most definitely.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
markfisher: You're saying no put is needed at this level? Could a ChannelResolver implementation be used instead? (similar to BeanFactoryChannelResolver but more opinionated/convention-driven)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: I'm not very crazy about Service either, but to sure what else to use, seeing as Binder was taken already. I think that in the future this could have more responsibilities, such as  handling non-MessageChannel bindings and dispatching to the appropriate `Binder<T>` and/or adapting to Observable/etc

markfisher: Good point. If we were to change the `Binder` methods to be just consumer/producer (no pubSub distinction) and then have the base `MessageChannelBinderSupport` class take the `ChannelBindingProperties` as a constructor arg, then perhaps we don't need this delegating "service" level at all?

mbogoevici: Yeah, I'd be happy to get rid of it. Right now it's just a man-in-the-middle.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
mbogoevici: How would that work? I guess we'd still need to register shared channels as beans ...

markfisher: Well, you could have a `SharedChannelResover` that knows exactly what to look for... even if there's a single bean that is the Map. I guess I was confused by your comment that the "operation is redundant" - assume you were talking about "put"?

mbogoevici: We could have a different abstraction here, sure. A `DestinationResolver` through wouldn't replace the default one, though - it would just be used for locating shared channels?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: sure, but not sure what value to use here - in the single-module case we use coordinates ... maybe 'aggregate' or something.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
mbogoevici: `AggregatedApplication` and change the method to `run` perhaps ?

markfisher: maybe just `AggregateApplication` and +1 for `run()`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/ModuleAggregationUtils.java
markfisher: we would need to avoid collisions across "aggregate" apps too (and by the way, for single-module cases we probably should not use the entire coordinate string)

mbogoevici: If only modules were labeled ;)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/SharedChannelRegistry.java
mbogoevici: Another option would be to forgo the shared parent altogether and launch each module as an independent context (enhanced with their own shared channels of interest) - this way we can also control the order in which they are shut down.

mbogoevici:
Currently, the following possible improvements have been identified, to be addressed as part of future work:

- splitting BindableProxyFactory;
- Elliminating SharedChannelRegistry and relying on the shared channels being registered directly;
- Splitting the parent context and launching modules as independent contexts - allowing for more gradual control of lifecycle, binding, startup/shutdown order;

markfisher:
rebased, squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/bd47560bd5af25b07c2bdba1f6ad7e33ee907da3

PR #115, opened 2015-09-01
[Do not merge, Discussion Only] WIP Classpath enhancement

mbogoevici:
Provides support for adding JARs to a module's class path at launch

To test:

Build/install `spring-cloud-stream` 

To use (in this case, let's pretend the `time-source` module has been rebuilt without the redis binder):
```
java -jar spring-cloud-stream-module-launcher/target/spring-cloud-stream-module-launcher-1.0.0.BUILD-SNAPSHOT.jar \
--modules=org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT \
--args.0.includes=org.springframework.cloud:spring-cloud-stream-binder-redis:1.0.0.BUILD-SNAPSHOT
```

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
@@ -47,6 +47,7 @@ import org.eclipse.aether.spi.connector.transport.TransporterFactory;
 import org.eclipse.aether.transport.file.FileTransporterFactory;
 import org.eclipse.aether.transport.http.HttpTransporterFactory;
 import org.eclipse.aether.util.artifact.JavaScopes;
+import org.eclipse.aether.util.filter.PatternExclusionsDependencyFilter;
 
 import org.springframework.core.io.FileSystemResource;
 import org.springframework.core.io.Resource;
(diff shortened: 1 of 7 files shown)


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
```diff
@@ -162,4 +147,59 @@ public void serviceCreationFailed(Class<?> type, Class<?> impl, Throwable except
 		});
 		return locator.getService(RepositorySystem.class);
 	}
+
+	@Override
+	public Resource[] resolve(Coordinates root, Coordinates[] includes, CoordinatesFilter[] excludes) {
+		Assert.hasText(root.getGroupId(), "'groupId' cannot be blank.");
```
ericbottard: We'll need to change these assertions message to clarify it pertains to `root`.

And iterate over the includes, and assert

mbogoevici: Actually, we should do assertions for an entire set of coordinates, I think.

ericbottard: Ah, correct. For some reason I was under the impression that the includes did not require the same props as the root, but they do.

mbogoevici:
Some notes:
- log artifact download (on bigger POMs there's just a huge time gap when nothing happens);
- log resulting CP;

mbogoevici:
Closed in favour of https://github.com/spring-cloud/spring-cloud-stream/pull/132

PR #116, opened 2015-09-02
Support RxJava module

ilayaperumalg:
- Add a way to enable `RxJava` module
   - Add `EnableRxJavaModule` annotation that inherits `EnableModule` for `Processor` type.
      - It also registers the custom `MessageHandler` that sets up RxJava `Observable/Subscriber` to process the incoming message reactively.
      - Introduce `RxJavaProcessor` that is expected to be implemented by the module author with the input/output being `Observable`. This processor is autowired into the message handler registered above.
 - Add moving average sample


spring-cloud-stream-samples/rxjava/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+	<description>Demo project for RxJava module</description>
+
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-stream-samples</artifactId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ericbottard: Should be inherited

spring-cloud-stream-samples/rxjava/src/main/java/demo/RxJavaApplication.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.ComponentScan;
+
+import rxjava.RxJavaTransformer;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@SpringBootApplication
+@ComponentScan(basePackageClasses= RxJavaTransformer.class)
```
ericbottard: Why don't we use the same package standards in samples as in OOTB modules? why not put everything in the same package?

spring-cloud-stream-samples/rxjava/src/main/java/rxjava/RxJavaTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+	public RxJavaProcessor processor() {
+		return new RxJavaProcessor<String, String>() {
+
+			@Override
+			public Observable<String> process(Observable<String> inputStream) {
+				return inputStream.map(data -> {
+					logger.info("Got data = " + data);
+					return data;
+				})
+						.buffer(5)
```
ericbottard: weird indent

spring-cloud-stream-samples/pom.xml
```diff
@@ -21,6 +21,7 @@
 		<module>tap</module>
 		<module>double</module>
 		<module>extended</module>
+		<module>rxjava</module>
```
mbogoevici: Let's move this to Spring Cloud Stream Modules and follow the same strategy as for XD: provide an OOTB library for building RxJava modules based on `@EnableModule` support. On the long run we'll want to have this provided via `@Input Observable` etc, but that's a lofty goal for now.

mbogoevici: Also, let's call it `rxjava-processor`. In the future (when we generalize) we might want to have sources and sinks with the same functionality.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/rxjava/EnableRxJavaModule.java
```diff
@@ -0,0 +1,63 @@
...
+ * to provide a bean that implements {@link RxJavaProcessor}.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@EnableModule(Processor.class)
+@Import(EnableRxJavaModule.RxJavaMessageHandler.class)
```
mbogoevici: This would look better split into `@EnableRxJavaModule` and an accompanying `@Configuration`.  I think it reads better that way.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/rxjava/RxJavaProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Marker interface that RxJava module uses to provide the implementation bean.
+ *
+ * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
+ */
+public interface RxJavaProcessor<I, O> {
+
+	Observable<O> process(Observable<I> input);
```
mbogoevici: I wonder if we shouldn't change our approach a bit here and supply both the input and output as arguments. There's something nice about the transformation that results in an observable being returned, but I wonder if we shouldn't already introduce the notion that both the input and output observables are passed as arguments.

ericbottard: I think it feels more idiomatic to be able to do stuff like `input.map()` _etc._

What exact signature (and example implementation of a `Processor`) would you foresee?

spring-cloud-stream-samples/rxjava/src/main/java/demo/RxJavaApplication.java
ilayaperumalg: I followed the existing samples template. Here is the PR that cleans up things related to this: https://github.com/spring-cloud/spring-cloud-stream/pull/118

spring-cloud-stream-samples/pom.xml
ilayaperumalg: I am not sure if we want to have this `sample` module into SCSM repo. Even in XD case, we only have the moving average sample in XD [samples repo] (https://github.com/spring-projects/spring-xd-samples)

mbogoevici: Sure, but the XD sample reuses the general-purpose RxJava support from Spring XD - and I was suggesting something similar (with extracting the common behaviour as a reusable library - we surely don't want every new implementation to copy everything). Maybe SCSM is not the best location for this common library and we should have a `spring-cloud-stream-rxjava` project?

ilayaperumalg: I see what you mean. I can move this in a separate project. Infact, I was going down that route initially and decided to use the current approach as we just add rxjava as a `provided` dependency.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/rxjava/RxJavaProcessor.java
mbogoevici: @ericbottard I still think that the Rx programming style should apply, but on a second thought, let's leave it as it is. I'll try to answer that question in a more elaborate way. It was related to the idea of providing bindable `@Input`s and `@Output`s that are Observable(s) and, respectively, Subject(s) - case in which they would be created by the framework so a processor could be

   void process(Observable<I> input, Subject<?,O> output)  {
     input.map(something).subscribe(output);
   } 

which is still fairly idiomatic, as far as I can tell. Obviously, the last `subscribe` is boilerplate in the 1:1 transformation case of the Processor, but would be applicable for multiple outputs, etc. Anyway, I think this warrants a separate discussion.

markfisher: +1 for both input and output being components that are bindable and managed by the infrastructure (as with MessageChannels), thus not instantiated directly

mbogoevici: +1 let's try to get this in as is, though, since it brings Rx support at par with XD, and we can refactor later.

ilayaperumalg: Sure, I pushed the changes that addresses other review comments along with the rebase from the latest master.

spring-cloud-stream-samples/rxjava-processor/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+
+	<dependencies>
+		<dependency>
+			<groupId>io.reactivex</groupId>
+			<artifactId>rxjava</artifactId>
+			<version>1.0.0</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-rxjava</artifactId>
```
ilayaperumalg: I believe it is ok to have the rxjava module depend on `spring-cloud-stream-rxjava` which extends `spring-cloud-stream`

pom.xml
```diff
@@ -25,6 +25,7 @@
 	</properties>
 	<modules>
 		<module>spring-cloud-stream</module>
+		<module>spring-cloud-stream-rxjava</module>
```
ilayaperumalg: It would be nice to find a separate repo for these kind of projects. May be `spring-cloud-stream-extensions` ?

mbogoevici: Possibly, but even projects like SI host stable extensions as part of the main project. Since Rx support is something we definitely want as part of core functionality, I think it's ok to have it here.

markfisher: Agreed. It should be in the main repo, but it does need to evolve so that Observable et al are bindable in the same way as MessageChannels.

mbogoevici: Agreed

spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/RxJavaConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+public class RxJavaConfiguration {
+
+	@Autowired
+	RxJavaProcessor processor;
+
+	@ServiceActivator(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT)
+	@Bean
+	public MessageHandler subjectMessageHandler() {
+		SubjectMessageHandler messageHandler = new SubjectMessageHandler(processor);
+		messageHandler.setOutputChannelName(Processor.OUTPUT);
```
mbogoevici: Is this still necessary if after setting it on the annotation?

ilayaperumalg: yeah, it [doesn't set] (https://github.com/spring-projects/spring-integration/blob/master/spring-integration-core/src/main/java/org/springframework/integration/config/annotation/ServiceActivatorAnnotationPostProcessor.java#L64) the `outputChannelName` when the explicit message handler is registered as a bean for the `ServiceActivator`. Not sure if it's a bug or it is expected. cc @garyrussell @artembilan

artembilan: No, it's not a bug. The annotation value is ignored me in case of MessageHandler. You really must specify all MH options directly. It is more organic because of Java power . See more info here: http://docs.spring.io/spring-integration/docs/latest-ga/reference/html/configuration.html#d4e15193.
(From iPhone)

mbogoevici: Well, then it's still a little confusing. Anyway, then we should probably remove the outputChannel option on the `@ServiceActivator` channel then ..

spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/EnableRxJavaModule.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@EnableModule(Processor.class)
+@Import(RxJavaConfiguration.class)
+public @interface EnableRxJavaModule {
```
mbogoevici: Maybe we should call this `EnableRxJavaProcessor`  instead?

spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/RxJavaConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.integration.annotation.ServiceActivator;
+import org.springframework.messaging.MessageHandler;
+
+/**
+ * Configuration class for RxJava module support.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Configuration
+public class RxJavaConfiguration {
```
mbogoevici: Let's rename this to `@RxJavaProcessorConfiguration`, partly in accordance with https://github.com/spring-cloud/spring-cloud-stream/pull/116/files#r38593219, but also because it indicates better what do we enable with respect to RxJava.

spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/EnableRxJavaModule.java
ilayaperumalg: +1

spring-cloud-stream-samples/rxjava-processor/src/main/java/rxjava/RxJavaTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+import rx.Observable;
+
+@EnableRxJavaProcessor
+public class RxJavaTransformer {
+
+	private static Logger logger = LoggerFactory.getLogger(RxJavaTransformer.class);
+
+	@Bean
+	public RxJavaProcessor processor() {
+		return new RxJavaProcessor<String, String>() {
```
artembilan: Generics warning on the `@Bean` method return type.
Or `@SuppressWarning` or specify them. 

From other side this can converted to the Lambda as well:
````java
@Bean
public RxJavaProcessor<String, String> processor() {
     return inputStream -> inputStream.map(...)
                            ...
}
````
And no need to worry about Lambda argument type - just because generics on the return type do the stuff!

File: spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/RxJavaConfiguration.java (hunk 9 above)
Comment (artembilan) [inline spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/RxJavaConfiguration.java]: Well, we thought about some new annotations without `outputChannel` and other `MessageHandler` options especially for the `Endpoint`, but finally we end up with the fact just ignore all them and follow with the standard Java way to define `@Bean`. 
We need all those `MH` options on `@ServiceActivator` if the last one is specified on the non-`@Bean` method when we generate `ServiceActivatingHandler` for the method invocation.
In case of `@Bean` for `MH` we don't generate it, just because we are already `MH`. So, we rely there on the user definition and don't override anything specified by the **setters** on that `MH`.

File: spring-cloud-stream-rxjava/pom.xml [hunk 13]
```diff
@@ -0,0 +1,34 @@
...
+			<artifactId>spring-cloud-stream</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>io.reactivex</groupId>
+			<artifactId>rxjava</artifactId>
+			<version>1.0.0</version>
```
mbogoevici: We will want to move this as a property of the root POM. Also, I think it should be acceptable to make this a req'd dependency. It can always be overridden by users.

spring-cloud-stream-samples/rxjava-processor/pom.xml
mbogoevici: Yes, I think this is what we want a user to do.

spring-cloud-stream-samples/rxjava-processor/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+	</parent>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<start-class>demo.RxJavaApplication</start-class>
+		<java.version>1.8</java.version>
+	</properties>
+
+	<dependencies>
+		<dependency>
```
mbogoevici: I think that this shouldn't be required if scs-rx pulls it down.

spring-cloud-stream-samples/rxjava-processor/src/main/java/demo/RxJavaApplication.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.ComponentScan;
+
+import rxjava.RxJavaTransformer;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@SpringBootApplication
+@ComponentScan(basePackageClasses= RxJavaTransformer.class)
```
mbogoevici: Should follow the same pattern as the other demos - no scanning.

mbogoevici:
LGTM + a few comments that can be potentially addressed on merge.

spring-cloud-stream-samples/rxjava-processor/pom.xml
ilayaperumalg: Currently, the `scs-rx` is set to use the `provided` dependency so that whatever the module author provides as a dependency from the module would be considered.

mbogoevici: I know, but they shouldn't be required to do that. Rx is a required dependency and can be easily overridden via Maven.

mbogoevici:
Squashed, rebased, merged w/discussed changes. https://github.com/spring-cloud/spring-cloud-stream/commit/2b1ccc4c7c6b3c8a92081a2100e640f2121267df

PR #117, opened 2015-09-02
Copy parent URLs when creating CL

ericbottard:
Changing the way the CL for a module is built.

This all started to support the buildpack when running on CF. The buildpack runs our payload as an exploded dir and adds its thing as a `java -cp` argument, which wasn't "seen" previously.
Turns out the `LaunchedURLClassLoader` does not consult its parent when doing `getResources()` which prevented [spring.factories](https://github.com/cloudfoundry/java-buildpack-auto-reconfiguration/blob/08e47eb1b01f60ba7562f025575db982fdc63596/auto-reconfiguration/src/main/resources/META-INF/spring.factories) from the buildback from being seen and having effect (more details on [slack](https://pivotal.slack.com/archives/spring-boot/p1441187889000343) if needed).

SO, I tried the [sensible thing](https://github.com/ericbottard/spring-cloud-stream/commit/f6f52e10940feb6129eda84d20302b1f3b63b288), to no avail.

This change is somewhat ugly IMO, but looking for a new pair of eyes.

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/boot/loader/ModuleJarLauncher.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/boot/loader/ModuleJarLauncher.java
@@ -18,6 +18,7 @@ package org.springframework.boot.loader;
 
 import java.lang.reflect.Method;
 import java.net.URL;
+import java.net.URLClassLoader;
 import java.util.List;
 
 import org.springframework.boot.loader.archive.Archive;
@@ -80,12 +81,15 @@ public class ModuleJarLauncher extends ExecutableArchiveLauncher {
 	@Override
 	protected ClassLoader createClassLoader(URL[] urls) throws Exception {
 		ClassLoader systemClassLoader = ClassLoader.getSystemClassLoader();
-		// try to retrieve the extension classloader
-		ClassLoader extensionClassLoader = systemClassLoader != null ? systemClassLoader.getParent() : null;
-		// set the classloader for the module as the extension classloader if available
-		// fall back to the system classloader (which can also be null) if not available
-		ClassLoader classLoaderForModule = extensionClassLoader != null ? extensionClassLoader : systemClassLoader;
-		return new LaunchedURLClassLoader(urls, classLoaderForModule);
+		if (systemClassLoader instanceof URLClassLoader) {
+			URLClassLoader classLoader = (URLClassLoader) systemClassLoader;
+			URL[] toUse = new URL[urls.length + classLoader.getURLs().length];
+			System.arraycopy(urls, 0, toUse, 0, urls.length);
+			System.arraycopy(classLoader.getURLs(), 0, toUse, urls.length, classLoader.getURLs().length);
+			return new LaunchedURLClassLoader(toUse, null);
+		} else {
+			return new LaunchedURLClassLoader(urls, systemClassLoader);
+		}
 	}
 
 }


ericbottard:
For the record, this is what I get when trying my alternate version:
```
Caused by: java.lang.IllegalArgumentException: Cannot instantiate interface org.springframework.context.ApplicationContextInitializer : org.cloudfoundry.reconfiguration.spring.CloudPropertySourceApplicationContextInitializer
	at org.springframework.boot.SpringApplication.getSpringFactoriesInstances(SpringApplication.java:376)
	at org.springframework.boot.SpringApplication.getSpringFactoriesInstances(SpringApplication.java:353)
	at org.springframework.boot.SpringApplication.initialize(SpringApplication.java:230)
	at org.springframework.boot.SpringApplication.<init>(SpringApplication.java:206)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:972)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:961)
	at org.springframework.cloud.stream.module.time.TimeSourceApplication.main(TimeSourceApplication.java:26)
	... 6 more
Caused by: java.lang.NoClassDefFoundError: org/springframework/context/ApplicationContextInitializer
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:760)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:411)
	at org.springframework.boot.loader.LaunchedURLClassLoader.doLoadClass(LaunchedURLClassLoader.java:153)
	at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:119)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.springframework.util.ClassUtils.forName(ClassUtils.java:250)
	at org.springframework.boot.SpringApplication.getSpringFactoriesInstances(SpringApplication.java:369)
	... 12 more
Caused by: java.lang.ClassNotFoundException: org.springframework.context.ApplicationContextInitializer
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 29 more
```

which seems to come from the fact that the thing that loaded the reconfiguration listener (the system CL), does not see ApplicationContextInitializer

spring-cloud-stream-module-launcher/src/main/java/org/springframework/boot/loader/ModuleJarLauncher.java
```diff
@@ -80,12 +81,15 @@ protected void launch(String[] args, String mainClass, ClassLoader classLoader)
...
-		// set the classloader for the module as the extension classloader if available
-		// fall back to the system classloader (which can also be null) if not available
-		ClassLoader classLoaderForModule = extensionClassLoader != null ? extensionClassLoader : systemClassLoader;
-		return new LaunchedURLClassLoader(urls, classLoaderForModule);
+		if (systemClassLoader instanceof URLClassLoader) {
+			URLClassLoader classLoader = (URLClassLoader) systemClassLoader;
+			URL[] toUse = new URL[urls.length + classLoader.getURLs().length];
+			System.arraycopy(urls, 0, toUse, 0, urls.length);
+			System.arraycopy(classLoader.getURLs(), 0, toUse, urls.length, classLoader.getURLs().length);
+			return new LaunchedURLClassLoader(toUse, null);
```
mbogoevici: Just did a quick test (dumping the added URLs) and this definitely skips extensions from the launched module's class loader. So we either need to set the extension class loader as parent here, or to copy the extension class loader URLs. I would be happier with the former, as it supersedes the behaviour of the JVM in the least (and I don't think we need to extend resource loading to that class loader).

ericbottard: Yep, fine by me

mbogoevici: Ok, I'll make that a separate change before merging.

mbogoevici:
Additional changes here https://github.com/mbogoevici/spring-cloud-stream/commit/80391fa8767d75dd71dc232a2f41728e5db26a2a

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/3561ac4aa212aa8ca8f31666f0baee67c1efbc0a 

Additional changes https://github.com/spring-cloud/spring-cloud-stream/commit/80391fa8767d75dd71dc232a2f41728e5db26a2a

PR #119, opened 2015-09-02
Rename `@EnableModule` to `@EnableBinding`

mbogoevici:
Once this is applied, https://github.com/spring-cloud/spring-cloud-stream-modules/pull/30 must be applied as well.


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
```diff
@@ -42,10 +43,15 @@
 @Documented
 @Inherited
 @Configuration
-@Import({ChannelBindingServiceConfiguration.class, AggregateBuilderConfiguration.class, BindingBeansRegistrar.class})
+@Import({ ChannelBindingServiceConfiguration.class, AggregateBuilderConfiguration.class,
+		BindingBeansRegistrar.class })
```
ilayaperumalg: I believe this line can still fit in the above within the 120 limit.

mbogoevici: Agreed. At this juncture, my destiny is controlled by the code formatter. ;)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingBeansRegistrar.java
```diff
@@ -37,11 +37,11 @@
...
 			BeanDefinitionRegistry registry) {
-		MultiValueMap<String, Object> attributes = metadata.getAllAnnotationAttributes(
-				EnableModule.class.getName(), false);
-		List<String> registeredChannelNames = new ArrayList<>();
+		MultiValueMap<String, Object> attributes = metadata.getAllAnnotationAttributes(EnableBinding.class.getName(),
+				false);
 		for (Class<?> type : collectClasses(attributes.get("value"))) {
-			BindingBeanDefinitionRegistryUtils.registerChannelBeanDefinitions(type, type.getName(), registry);
+			BindingBeanDefinitionRegistryUtils.registerChannelBeanDefinitions(type,
+					type.getName(), registry);
```
ilayaperumalg: Something is weird with your IDE formatter ;) I notice the code format changes in few places...

ilayaperumalg:
At this context does it make sense to rename the `ModuleChannels` annotation to `BindingChannels`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingBeansRegistrar.java
mbogoevici: Possibly.

mbogoevici:
Possibly. Why not just `@Binding` or `@Bindings` though?

ilayaperumalg:
+1 for `@Bindings`

markfisher:
if that same annotation would eventually apply to other bindable components (e.g. Observable), then +1 for `@Bindings`

mbogoevici:
I think it should ... `@Bindings` is better than `@ModuleChannels` anyway.

markfisher:
rebased, squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/86e457fac41061e75e876779b97e7380c7e5bd88

PR #121, opened 2015-09-03
Support underscore delimiter in module args

ilayaperumalg:
- If the module argument key (--args.<index>.<option>) is delimited by underscore, then spring-boot doesn't resolve for the `ConfigurationProperties` of type `Map<?,Map<?,?>>`.
 - To overcome this issue on module launcher, this PR introduces `ModuleOptionKey` which is used as the key for module launcher args map property.
   - Add converters that converts the program arguments into appropriate `ModuleOptionKey`

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLaunchRequest.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLaunchRequest.java
@@ -33,7 +33,7 @@ public class ModuleLaunchRequest {
 
 	public ModuleLaunchRequest(String module, Map<String, String> arguments) {
 		this.module = module;
-		this.arguments = arguments != null ? new HashMap<>(arguments) : new HashMap<String, String>();
+		this.arguments = arguments != null ? arguments : new HashMap<String, String>();
 	}
 
 	public String getModule() {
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLaunchRequest.java
```diff
@@ -33,7 +33,7 @@
 
 	public ModuleLaunchRequest(String module, Map<String, String> arguments) {
 		this.module = module;
-		this.arguments = arguments != null ? new HashMap<>(arguments) : new HashMap<String, String>();
+		this.arguments = arguments != null ? arguments : new HashMap<String, String>();
```
ericbottard: The copy was intentional to prevent change by reference

ericbottard:
Doesn't your PR on boot (https://github.com/spring-projects/spring-boot/pull/3836) solve the underlying problem here?
We should not have to resort to doing this

markfisher:
Ilaya's PR for https://github.com/spring-projects/spring-boot/pull/3836 was not merged, but Dave closed the issue with a different commit. That said, I have the same concern. If that issue is not address the underlying problem, it should be reopened. We should *not* be doing any of this.

ilayaperumalg:
Closing this PR as few other options are considered to parse ARGS as environment variable.

PR #123, opened 2015-09-03
Pubsub fix

mbogoevici:


spring-cloud-stream-samples/source/src/main/resources/application.yml
```diff
@@ -5,7 +5,7 @@ spring:
   cloud:
     stream:
       bindings:
-        output: testtock
+        output: topic:testtock
```
markfisher: did you leave this intentionally?

mbogoevici: No. Needs to be reverted.

mbogoevici: We need actual binding tests for this, though.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/b8ffa630f73621b3e1ae72b0111e1019565adfb5

PR #125, opened 2015-09-03
Add support for Kafka in Cloud Foundry

mbogoevici:
- use the VCAP properties to find brokers and ports

How to test: run time log by pushing jars into CF after rebuilding SCSM with the Kafka binder.

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
@@ -18,7 +18,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<kafka.version>0.8.2.1</kafka.version>
 		<spring-integration-kafka.version>1.2.0.RELEASE</spring-integration-kafka.version>
-		<rxjava.version>1.0.0</rxjava.version>
+		<rxjava-math.version>1.0.0</rxjava-math.version>
 	</properties>
 
 	<dependencies>
@@ -67,12 +67,12 @@
 		<dependency>
 			<groupId>io.reactivex</groupId>
 			<artifactId>rxjava</artifactId>
-			<version>${rxjava.version}</version>
+			<version>${rx-java.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>io.reactivex</groupId>
 			<artifactId>rxjava-math</artifactId>
-			<version>${rxjava.version}</version>
+			<version>${rxjava-math.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.xd</groupId>
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
@@ -67,7 +67,6 @@
 		<dependency>
 			<groupId>io.reactivex</groupId>
 			<artifactId>rxjava</artifactId>
-			<version>${rx-java.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>io.reactivex</groupId>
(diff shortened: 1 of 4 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
```diff
@@ -18,7 +18,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<kafka.version>0.8.2.1</kafka.version>
 		<spring-integration-kafka.version>1.2.0.RELEASE</spring-integration-kafka.version>
-		<rxjava.version>1.0.0</rxjava.version>
+		<rxjava-math.version>1.0.0</rxjava-math.version>
```
markfisher: should this be "rx-java-math.version" to match the "rx-java.version"? (I assume we are now inheriting the latter)

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaMessageChannelBinderConfiguration.java
```diff
@@ -33,9 +34,13 @@
 @ConfigurationProperties(prefix = "spring.cloud.stream.binder.kafka")
 public class KafkaMessageChannelBinderConfiguration {
 
-	private String zkAddress;
+	private String[] zkAddress;
```
markfisher: should this be plural?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
mbogoevici: yes. This all needs to be cleaned up and moved in the parent.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaMessageChannelBinderConfiguration.java
mbogoevici: Let's make it zkNodes. I've tried to keep the old values from xd, but at this point it makes very little sense.

mbogoevici: also zkDefaultPort

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
mbogoevici: actually I've changed `rx-java` to `rxjava` since that's how the artifacts are called.

markfisher: looks good

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaMessageChannelBinderConfiguration.java
```diff
@@ -33,9 +34,13 @@
...
 
-	private String zkAddress;
+	private String[] zkNodes;
 
-	private String brokers;
+	private String zkDefaultPort;
+
+	private String[] brokers;
+
+	private String brokersDefaultPort;
```
markfisher: one final nitpick: I think I would change this to `defaultBrokerPort` and the corresponding ZK property name to `defaultZkPort`... to me that reads better and avoids the singular vs. plural situation with "brokers" currently

mbogoevici: sounds good

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/e82bfe6c912a4840aaab32c3509a64b26a9cadf8 along with https://github.com/spring-cloud/spring-cloud-stream/commit/b39c609fb03902ec77e98ffb298241b52e84f2f5

PR #126, opened 2015-09-05
Use 'aggregateArgs.*' for aggregate parameters

mbogoevici:


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -88,19 +88,19 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
...
 		}
 		else {
-			launchAggregatedModules(moduleLaunchRequests, parentArgs);
+			launchAggregatedModules(moduleLaunchRequests, toArgArray(parentArgs));
 		}
 	}
 
 	public void launch(List<ModuleLaunchRequest> moduleLaunchRequests, boolean aggregate) {
-		this.launch(moduleLaunchRequests, aggregate, new String[0]);
+		this.launch(moduleLaunchRequests, aggregate, null);
```
ericbottard: Can we use empty map instead ?

mbogoevici: We surely can. The current reasoning was that it would be nice for toArgArray to be null-safe anyway, but we can pass an empty map here, as well.

mbogoevici:
Another approach (suggested by @markfisher) could rely on `args.aggregate` instead (aligned with changes in https://github.com/spring-cloud/spring-cloud-stream/pull/128). On one hand, it reduces the complexity of `ModuleLauncherProperties`. On the other hand, clearly separating module and aggregate arguments has its benefits.

mbogoevici:
Rebased and changed to `args.aggregate`

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
```diff
@@ -64,9 +75,10 @@
 	 */
 	private Map<String, Map<String, String>> args = new HashMap<>();
 
-	public void setModules(String[] modules) {
-		this.modules = modules;
-	}
+	/**
+	 * Map of arguments to be passed to the aggregate (if aggregate is true)
+	 */
+	private Map<String, String> aggregateArgs;
```
markfisher: this is no longer needed, right?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncherProperties.java
```diff
@@ -31,14 +31,25 @@
  * properties, environment variables, program arguments, <i>etc.</i> _):<ul>
  *     <li>{@literal modules = <list>}: an ordered list of maven coordinates of modules to launch</li>
  *     <li>{@literal args[<index>][<key>] = <value>}: key/value pairs that will become module arguments,
- *     where {@literal <index>} is the 0-based index of the module in the list above</li>
+ *     where {@literal <index>} is the 0-based index of the module in the list above and '*' can be used for passing
+ *     arguments to all modules</li>
+ *     <li>{@literal aggregate = true | false}</li>: whether multiple modules launched together should be aggregated,
+ *     case in which they will be launched as a single individual unit, and {@literal args['aggregate'[<key>] = <value>}
```
markfisher: looks like one additional closing brace is needed before the equals sign

mbogoevici: No, I missed pushing a commit.  I'll remove it, sorry.

markfisher:
LGTM; merging

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/745a4bd2dd90308db467ad3f68d40e8715a3c7ce

PR #127, opened 2015-09-06
Add option for the resolver and module-launcher to operate in offline mode

mbogoevici:


spring-cloud-stream-module-launcher/src/test/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolverTests.java
```diff
@@ -82,8 +90,8 @@ public void testResolveRemote() throws IOException {
 
 	@Test
 	public void testResolveMockRemote() throws IOException {
-		ClassPathResource cpr = new ClassPathResource("local-repo");
-		File localRepository = cpr.getFile();
+		File localRepository = Files.createTempDirectory(UUID.randomUUID().toString()).toFile();
```
mbogoevici: This ensures that the remote resolution test does not use files cached by previous runs.

markfisher:
LGTM; merging

PR #130, opened 2015-09-09
Add per-binding configuration for partitioning

mbogoevici:
- consumer and producer partitioning properties are defined per-binding;
- the binding configuration will use an expanded form (where `destination` indicates the binding target, missing in the short form `spring.cloud.stream.binding.input=tessttock`;
- `partitioned` indicates that an input is partitioned (must be used in conjunction with an `instanceIndex` setting - either via arguments or via environment. `instanceCount` must be defined for Kafka.)
- `partitionKeyExpression` or `partitionKeyExtractorClass` indicate that a producer is partitioned (must be used in conjunction with supplying a `partitionCount` property)
- unit tests
- Kafka binding fixes for partitioning
- the current Binder properties need to be renamed, but due to the use of hardcoded values I left them alone;

Usage example:

java -jar time-source/target/time-source-1.0.0.BUILD-SNAPSHOT-exec.jar --spring.cloud.stream.bindings.output.destination=parttock --spring.cloud.stream.bindings.output.partitionKeyExpression=payload --spring.cloud.stream.bindings.output.partitionCount=3

java -jar log-sink/target/log-sink-1.0.0.BUILD-SNAPSHOT-exec.jar --spring.cloud.stream.bindings.input.partitioned=true --spring.cloud.stream.bindings.input.destination=parttock --spring.cloud.stream.instanceCount=3 --spring.cloud.stream.instanceIndex=0 --server.port=0

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
@@ -44,8 +44,7 @@ public class ChannelBindingProperties {
 	@Value("${INSTANCE_INDEX:${CF_INSTANCE_INDEX:0}}")
 	private int instanceIndex;
 
-	@Value("${INSTANCE_COUNT:1}")
-	private int instanceCount;
+	private int instanceCount = 1;
 
 	private Properties consumerProperties = new Properties();
 
@@ -171,9 +170,7 @@ public class ChannelBindingProperties {
 	public Properties getConsumerProperties(String inputChannelName) {
 		if (isPartitionedConsumer(inputChannelName)) {
 			Properties channelConsumerProperties = new Properties();
-			if (consumerProperties != null) {
-				channelConsumerProperties.putAll(consumerProperties);
-			}
+			channelConsumerProperties.putAll(consumerProperties);
 			channelConsumerProperties.setProperty(BinderProperties.COUNT,
 					Integer.toString(getInstanceCount()));
 			channelConsumerProperties.setProperty(BinderProperties.PARTITION_INDEX,
@@ -195,11 +192,7 @@ public class ChannelBindingProperties {
 	public Properties getProducerProperties(String outputChannelName) {
 		if (isPartitionedProducer(outputChannelName)) {
 			Properties channelProducerProperties = new Properties();
-			if (this.producerProperties != null) {
-				channelProducerProperties.putAll(this.producerProperties);
-			}
-			channelProducerProperties.put(BinderProperties.MIN_PARTITION_COUNT,
-					Integer.toString(getPartitionCount(outputChannelName)));
+			channelProducerProperties.putAll(this.producerProperties);
 			channelProducerProperties.setProperty(BinderProperties.NEXT_MODULE_COUNT,
 					Integer.toString(getPartitionCount(outputChannelName)));
 			copyChannelBindingProperty(outputChannelName, channelProducerProperties,
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
@@ -20,7 +20,9 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.TreeMap;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.cloud.stream.binder.BinderProperties;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
@@ -33,7 +35,17 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include;
 @JsonInclude(Include.NON_DEFAULT)
 public class ChannelBindingProperties {
 
-	public static final String PATH = "path";
+	public static final String TARGET = "target";
+
+	public static final String PARTITION_COUNT = "partitionCount";
+
+	public static final String PARTITIONED = "partitioned";
+
+	@Value("${INSTANCE_INDEX:${CF_INSTANCE_INDEX:0}}")
+	private int instanceIndex;
+
+	@Value("${INSTANCE_COUNT:1}")
+	private int instanceCount;
 
 	private Properties consumerProperties = new Properties();
 
@@ -41,7 +53,7 @@ public class ChannelBindingProperties {
 
 	private Map<String,Object> bindings = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
 
-	public Properties getConsumerProperties() {
+	private Properties getConsumerProperties() {
 		return this.consumerProperties;
 	}
 
@@ -65,7 +77,24 @@ public class ChannelBindingProperties {
 		this.bindings = bindings;
 	}
 
-	public String getBindingPath(String channelName) {
+	public int getInstanceIndex() {
+		return instanceIndex;
+	}
+
+	public void setInstanceIndex(int instanceIndex) {
+		this.instanceIndex = instanceIndex;
+	}
+
+	public int getInstanceCount() {
+		return instanceCount;
+	}
+
+	public void setInstanceCount(int instanceCount) {
+		this.instanceCount = instanceCount;
+	}
+
+
+	public String getBindingTarget(String channelName) {
 		Object binding = bindings.get(channelName);
 		// we may shortcut directly to the path
 		if (binding != null) {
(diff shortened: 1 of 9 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -33,7 +35,17 @@
 @JsonInclude(Include.NON_DEFAULT)
 public class ChannelBindingProperties {
 
-	public static final String PATH = "path";
+	public static final String TARGET = "target";
```
mbogoevici: will rename to DESTINATION/"destination"

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -84,8 +113,120 @@ else if (binding instanceof Map) {
...
+
+	public boolean isPartitionedConsumer(String channelName) {
+		Object binding = bindings.get(channelName);
+		// if the setting is just a target shortcut
+		if (binding == null || binding instanceof String) {
+			return false;
+		}
+		else if (binding instanceof Map) {
+			Map<?, ?> bindingProperties = (Map<?, ?>) binding;
+			Object bindingPath = bindingProperties.get(PARTITIONED);
```
ilayaperumalg: Is it ok to assume the `consumer` module can always be partitioned when the `instanceCount` is greater than 1? It would also mean, we configure the producer module with the default `partitioningExpression` (as `payload`) when the consumer module instance count is > 1. In that case, we may not need this property field `partitioned`.

mbogoevici: Multiple module instances could be competing consumers (for distributing load) - without being partitioned.

mbogoevici: Also, we should not assume anything about the consumer/producer module knowing much about each other (from within SCS). Spring Cloud Dataflow can coordinate these properties.

ilayaperumalg: ok, I thought default way of `partitioning` would help distributing the load. But I see it can't help always. Competing consumers would help in some cases.

ilayaperumalg: > Also, we should not assume anything about the consumer/producer module knowing much about each other (from within SCS). Spring Cloud Dataflow can coordinate these properties.

yeah, that clarifies my question. thanks!

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -84,8 +113,120 @@ else if (binding instanceof Map) {
...
+	 * Merge general properties provided by 'spring.cloud.stream.consumerProperties.*' with individual binding
+	 * properties supplied via binders.
+	 *
+	 * @param inputChannelName the input channel name
+	 * @return merged consumer properties
+	 */
+	public Properties getConsumerProperties(String inputChannelName) {
+		if (isPartitionedConsumer(inputChannelName)) {
+			Properties channelConsumerProperties = new Properties();
+			if (consumerProperties != null) {
```
ilayaperumalg: Given `consumerProperties` is always non-null in this class, may be we can do `!consumerProperties.isEmpty()` here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -84,8 +113,120 @@ else if (binding instanceof Map) {
...
+	 * Merge general properties provided by 'spring.cloud.stream.producerProperties.*' with individual binding
+	 * properties supplied via binders.
+	 *
+	 * @param outputChannelName the output channel name
+	 * @return merged producer properties
+	 */
+	public Properties getProducerProperties(String outputChannelName) {
+		if (isPartitionedProducer(outputChannelName)) {
+			Properties channelProducerProperties = new Properties();
+			if (this.producerProperties != null) {
```
ilayaperumalg: Same here for doing empty check instead of non null check.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -84,8 +113,120 @@ else if (binding instanceof Map) {
...
+				channelProducerProperties.putAll(this.producerProperties);
+			}
+			channelProducerProperties.put(BinderProperties.MIN_PARTITION_COUNT,
+					Integer.toString(getPartitionCount(outputChannelName)));
+			channelProducerProperties.setProperty(BinderProperties.NEXT_MODULE_COUNT,
+					Integer.toString(getPartitionCount(outputChannelName)));
```
ilayaperumalg: Nitpicky: we can probably avoid calling `getPartitionCount(String ChannelName)` twice as it is same for both `minPartitionCount` and `nextModuleCount`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -33,7 +35,17 @@
...
+	public static final String DESTINATION = "destination";
+
+	public static final String PARTITION_COUNT = "partitionCount";
+
+	public static final String PARTITIONED = "partitioned";
+
+	@Value("${INSTANCE_INDEX:${CF_INSTANCE_INDEX:0}}")
+	private int instanceIndex;
+
+	@Value("${INSTANCE_COUNT:1}")
```
ilayaperumalg: Given `INSTANCE_INDEX` and `CF_INSTANCE_INDEX` would be available in lattice/CF environments, not sure the reasoning behind adding `INSTANCE_COUNT` as an environment variable here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
mbogoevici: Right - I forgot that the properties object is initialized above -  we can actually skip the check altogether then.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
mbogoevici: Yeah - the initial reasoning was that we could set it in the environment for symmetry with `instanceIndex`, but that doesn't make a lot of sense and duplicates what we can do via `spring.cloud.stream.instanceIndex` already.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingProperties.java
```diff
@@ -84,8 +113,120 @@ else if (binding instanceof Map) {
...
+	 * @param inputChannelName the input channel name
+	 * @return merged consumer properties
+	 */
+	public Properties getConsumerProperties(String inputChannelName) {
+		if (isPartitionedConsumer(inputChannelName)) {
+			Properties channelConsumerProperties = new Properties();
+			if (consumerProperties != null) {
+				channelConsumerProperties.putAll(consumerProperties);
+			}
+			channelConsumerProperties.setProperty(BinderProperties.COUNT,
```
ilayaperumalg: I notice that the Kafka binder also checks for the `MIN_PARTITION_COUNT` property when determining the partitions for the Kafka consumer do you think we need to add that here?

mbogoevici: Actually, that makes sense. The value would be `COUNT`.

ilayaperumalg: I think we need to make the distinction between `minPartitionCount` and `partitionCount` at some point of time (especially for Kafka). Otherwise, the min partition logic may not be used at all. What do you think?

mbogoevici: Yes. The problem is that Kafka ignores `count` (as it uses the effective partition count of the topic), but uses `minPartitionCount` as a hint to create the number of requested partitions, whereas Rabbit and Redis use `count` effectively (which is why both values are currently populated). I think that the entire system of properties used by the binders needs to be reworked eventually.

mbogoevici: Note that `partitionCount` is only relevant for producers.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as f4f6cb509d53d6f802f436295ac0a9f1d5889b9c.

PR #132, opened 2015-09-10
Support for includes/excludes in ModuleLauncher

mbogoevici:
PR depends on https://github.com/spring-cloud/spring-cloud-stream/pull/126 (Will rebase after merge)


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -88,19 +93,22 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
...
+	public void launch(List<ModuleLaunchRequest> moduleLaunchRequests, boolean aggregate, Map<String,String> parentArgs) {
 		List<ModuleLaunchRequest> reversed = new ArrayList<>(moduleLaunchRequests);
 		Collections.reverse(reversed);
 		if (moduleLaunchRequests.size() == 1 || !aggregate) {
 			launchIndividualModules(moduleLaunchRequests);
 		}
 		else {
-			launchAggregatedModules(moduleLaunchRequests, parentArgs);
+			//TODO add support for CP enhancement for aggregated modules
+			launchAggregatedModules(moduleLaunchRequests, parentArgs != null ? parentArgs : Collections.EMPTY_MAP);
```
ericbottard: Can use `emptyMap()` with generics and avoid the `@SuppressWarnings` I believe

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -88,19 +93,22 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
...
 		}
 	}
 
+	@SuppressWarnings("unchecked")
 	public void launch(List<ModuleLaunchRequest> moduleLaunchRequests, boolean aggregate) {
-		this.launch(moduleLaunchRequests, aggregate, new String[0]);
+		this.launch(moduleLaunchRequests, aggregate, Collections.EMPTY_MAP);
```
ericbottard: idem

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -112,38 +120,72 @@ public void launch(List<ModuleLaunchRequest> moduleLaunchRequests) {
 	 * {@literal --foo=bar} form.
 	 */
 	private String[] toArgArray(Map<String, String> args) {
-		String[] result = new String[args.size()];
-		int i = 0;
-		for (Map.Entry<String, String> kv : args.entrySet()) {
-			result[i++] = String.format("--%s=%s", kv.getKey(), kv.getValue());
+		if (args != null) {
```
ericbottard: this is a private method, we can make sure that `args` is an empty map instead of `null`, which reads better IMHO

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/Coordinates.java
```diff
@@ -0,0 +1,71 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.module.resolver;
+
+/**
+ * Encapsulates a group of Maven coordinates.
```
ericbottard: "a group" may be misleading in this sentence

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
```diff
@@ -169,4 +154,100 @@ public void serviceCreationFailed(Class<?> type, Class<?> impl, Throwable except
...
+		for (Coordinates coordinate : coordinates) {
+			validateCoordinates(coordinate);
+		}
+		if (!ObjectUtils.isEmpty(includes)) {
+			for (Coordinates include : includes) {
+				validateCoordinates(include);
+			}
+		}
+		List<Resource> result = new ArrayList<>();
+		// if we only have one artifact to resolve, it will be set as root
```
ericbottard: Do we have a usecase with `null` `root` and several `coordinates` ?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -182,20 +236,50 @@ private void launchModule(String module, String[] args) {
 		}
 	}
 
-	private Resource resolveModule(String coordinates) {
+	private void launchModuleWithDependencies(String module, String[] args, String[] includes, String[] excludes) {
+		try {
+			Resource[] libraries = this.moduleResolver.resolve(
+					new Coordinates[]{toCoordinates(module, DEFAULT_MODULE_CLASSIFIER)},
```
ericbottard: this should not be the `-exec` classifier, but default, correct?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
mbogoevici: That seemed to be best way to handle aggregation of multiple modules.

ericbottard: ah, aggregation. gotcha

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/MultiArchiveLauncher.java
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.boot.loader.LaunchedURLClassLoader;
+import org.springframework.boot.loader.Launcher;
+import org.springframework.boot.loader.archive.Archive;
+import org.springframework.cloud.stream.module.utils.ClassloaderUtils;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.ReflectionUtils;
+
+/**
+ * A {@link Launcher} that provides a public {@link #launch(String[])} method.
```
ericbottard: Should better explain why this Launcher exists in the first place IMO

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
mbogoevici: No, it is actually `-exec`. That's the default classifier we set for module artifacts.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -112,38 +120,72 @@ public void launch(List<ModuleLaunchRequest> moduleLaunchRequests) {
...
-						seenArchives.add(jarNameWithoutExtension);
-						jarURLs.add(archive.getUrl());
+			if (!(parentArgs.containsKey(EXCLUDE_DEPENDENCIES_ARG) ||
+					parentArgs.containsKey(INCLUDE_DEPENDENCIES_ARG))) {
+				for (ModuleLaunchRequest moduleLaunchRequest : moduleLaunchRequests) {
+					Resource resource = resolveModule(moduleLaunchRequest.getModule());
+					JarFileArchive jarFileArchive = new JarFileArchive(resource.getFile());
+					jarURLs.add(jarFileArchive.getUrl());
+					for (Archive archive : jarFileArchive.getNestedArchives(ArchiveMatchingEntryFilter.FILTER)) {
+						// avoid duplication based on unique JAR names
```
ericbottard: Not part of this PR, but couldn't we let Aether do the resolution for us (ie resolve standard jars here)?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
ericbottard: yes, but we don't want uber jars here, as the uber jars deps (which are inside it at `/lib/*.jar`) will *also* be resolved as dependencies of it

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
ericbottard: Actually, are you referring to https://github.com/spring-cloud/spring-cloud-stream/pull/132/files#diff-3ca29f22b67f5c13875e045bd003bd6fR183 ? This uses `null` as first arg and several includes, which is not the same thing. Or am I missing something ?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
mbogoevici: Was thinking of it, but I'd like to restrict the use of Aether resolution to the include/exclude use cases and have that as a future enhancement.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
mbogoevici: I know. Currently we're ignoring the contents of `/lib/*.jar`, which is suboptimal. At least for a first step, I'd prefer to keep things as close as possible to the original and have this as a future enhancement.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
mbogoevici: Yes, that's handling the include/exclude case with aggregation. The list of includes passed as argument is a merger of the module list and the original includes.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
mbogoevici: OTOH, it would be simply a matter of calling `resolve(null, modules, new Coordindates[0], new String[0])` tbh, so may be as well doing it ...

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
mbogoevici: To note, I will change the aggregation handling to use the first module as root and the next ones as  direct dependencies. While not entirely identical to having all as direct dependencies, in practice the end result of the resolution process should be the same (and if there are inherent conflicts then they should be resolved as part of the exclusion process).

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
mbogoevici: Will punt on this right now, making this a future enhancement.

mbogoevici:
Rebased, squashed, added fixes for aggregated includes/excludes

mbogoevici:
Following examples should work:

java -jar ./spring-cloud-stream-module-launcher/target/spring-cloud-stream-module-launcher-1.0.0.BUILD-SNAPSHOT.jar --modules=org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT,org.springframework.cloud.stream.module:filter-processor:1.0.0.BUILD-SNAPSHOT,org.springframework.cloud.stream.module:filter-processor:1.0.0.BUILD-SNAPSHOT --args.0.fixedDelay=7 --args.1.expression='payload.contains("10")' --aggregate=true
(regression test)

java -jar ./spring-cloud-stream-module-launcher/target/spring-cloud-stream-module-launcher-1.0.0.BUILD-SNAPSHOT.jar --modules=org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT,org.springframework.cloud.stream.module:filter-processor:1.0.0.BUILD-SNAPSHOT,org.springframework.cloud.stream.module:filter-processor:1.0.0.BUILD-SNAPSHOT --args.0.fixedDelay=7 --args.1.expression='payload.contains("6")' --aggregate=true --args.aggregate.spring.cloud.stream.bindings.output=filtered-kafka-10 --args.aggregate.includes=org.springframework.cloud:spring-cloud-stream-binder-kafka:1.0.0.BUILD-SNAPSHOT --args.aggregate.excludes=:*-binder-redis


java -jar ./spring-cloud-stream-module-launcher/target/spring-cloud-stream-module-launcher-1.0.0.BUILD-SNAPSHOT.jar --modules=org.springframework.cloud.stream.module:time-source:1.0.0.BUILD-SNAPSHOT --args.0.fixedDelay=7 --args.0.spring.cloud.stream.bindings.output=filtered-kavka-10 --args.0.includes=org.springframework.cloud:spring-cloud-stream-binder-kafka:1.0.0.BUILD-SNAPSHOT --args.0.excludes=:*-binder-redis

TODO: document the options for the module launcher (separate PR, further enhancements, as discussed)

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLauncher.java
```diff
@@ -87,6 +94,7 @@ public ModuleLauncher(ModuleResolver moduleResolver) {
 	 * @param aggregate whether the modules should be aggregated at launch
 	 * @param parentArgs a list of arguments for the whole aggregate
 	 */
+	@SuppressWarnings("unchecked")
 	public void launch(List<ModuleLaunchRequest> moduleLaunchRequests, boolean aggregate, Map<String,String> parentArgs) {
 		List<ModuleLaunchRequest> reversed = new ArrayList<>(moduleLaunchRequests);
 		Collections.reverse(reversed);
```
ilayaperumalg: The `reversed` ModuleLaunchRequests aren't used at all. I see this [change] (https://github.com/spring-cloud/spring-cloud-stream/commit/bd47560bd5af25b07c2bdba1f6ad7e33ee907da3#diff-3ca29f22b67f5c13875e045bd003bd6fL75) removed it. I think we need to use the `reversed` ModuleLaunchRequests when launching modules.  There is a catch here though: for the`direct-binding`, the `AggregateApplication` does the reverse when creating the module contexts.

ilayaperumalg: I can fix this while merging.

mbogoevici: Thanks!

ilayaperumalg: Since the reversal is done when the modules launched individually as well as when launched via AggregateApplication, I will remove the reversal logic from here.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/MultiArchiveLauncher.java
```diff
@@ -0,0 +1,101 @@
...
+import org.springframework.boot.loader.Launcher;
+import org.springframework.boot.loader.archive.Archive;
+import org.springframework.cloud.stream.module.utils.ClassloaderUtils;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.ReflectionUtils;
+
+/**
+ * A {@link Launcher} for multiple independent JAR archives (which aren't nested in an uberjar). This class
+ * supports module aggregation and direct binding.
```
ilayaperumalg: I believe this class is used only for individual module launcher.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as d9c4f949c0ecde12bc0a66dd42f4f9c134618a22 along with a minor cleanup bb326ec32f450672cefda5af45b60ee5cec73d17

PR #136, opened 2015-09-19
Issue #137: Some docs enhancements

kbastani:
Fixed docs syntax error with ModuleApplicationTests

### DIFF
--- a/docs/src/main/asciidoc/intro.adoc
+++ b/docs/src/main/asciidoc/intro.adoc
@@ -51,7 +51,7 @@ The `@Output` annotation is used to identify output channels (messages leaving t
 public class ModuleApplicationTests {
 
 	@Autowired
-	private Source source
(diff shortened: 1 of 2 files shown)


kbastani:
There was a syntax error in the docs in the example code for the TimerSource unit test class

kbastani:
Added an emit log in the Source sample so you can see what the payload is that is being emitted on the interval.

markfisher:
I don't think the logging on the source side is a good idea actually. What we really want for an initial "time | log" experience is to see the messages on the log sink side. The source side messages can be confusing.

I will push the semicolon addition though. Thanks for catching that!

markfisher:
added missing semicolon here: https://github.com/spring-cloud/spring-cloud-stream/commit/cc71d66445c2d794d8800ee77dffbf1d7d21a4e6

PR #141, opened 2015-09-24
Create BindingProperties and Converter to replace the current Map-based logic

mbogoevici:


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -33,16 +34,10 @@
...
 @ConfigurationProperties("spring.cloud.stream")
 @JsonInclude(Include.NON_DEFAULT)
-public class ChannelBindingProperties {
-
-	public static final String DESTINATION = "destination";
-
-	public static final String PARTITION_COUNT = "partitionCount";
-
-	public static final String PARTITIONED = "partitioned";
+public class ChannelBindingServiceProperties {
```
mbogoevici: Renamed so that it doesn't overlap with `BindingProperties`. Also, it reflects better the relationship with `ChannelBindingService` (which needs to refactored and renamed, just not in this PR)

markfisher: So would this ultimately be *BinderProperties*?

mbogoevici: When ChannelBinderServices gets merged with Binder, sure. The difficulty to naming it as such just now is that each binder has their own configuration so we must cull the classes here a little, I think.

markfisher: Makes sense.

ilayaperumalg:
LGTM, squashed and merged as 10c1f5051cd427ddd76a6968b0aa81eecf26e859

PR #143, opened 2015-09-29
Allow configuration of Kafka embedded headers in the bus

mbogoevici:


mbogoevici:
In principle, header embedding/headerless mode should be configurable at binding level, and be a generic feature of the Binder(s), but this allows at least to use what's already there.

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/deecae93a68af209d890d8e73957f7d34fde758e

PR #145, opened 2015-09-30
Support custom message converters

ilayaperumalg:
- User provided custom message converters gets the precedence over existing message converters.
 - Module author would need to provide the collection of converters with the bean name `customMessageConverters`
 - Refactored message converter configuration

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConvertersConfigurer.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConvertersConfigurer.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.binding;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+import org.springframework.aop.framework.Advised;
+import org.springframework.aop.support.AopUtils;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cloud.stream.config.BindingProperties;
+import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
+import org.springframework.cloud.stream.converter.AbstractFromMessageConverter;
+import org.springframework.cloud.stream.converter.ByteArrayToStringMessageConverter;
+import org.springframework.cloud.stream.converter.CompositeMessageConverterFactory;
+import org.springframework.cloud.stream.converter.JavaToSerializedMessageConverter;
+import org.springframework.cloud.stream.converter.JsonToPojoMessageConverter;
+import org.springframework.cloud.stream.converter.JsonToTupleMessageConverter;
+import org.springframework.cloud.stream.converter.MessageConverterUtils;
+import org.springframework.cloud.stream.converter.PojoToJsonMessageConverter;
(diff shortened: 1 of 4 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConvertersConfigurer.java
```diff
@@ -0,0 +1,121 @@
...
+	private ApplicationContext applicationContext;
+
+	private CompositeMessageConverterFactory messageConverterFactory;
+
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		Collection<AbstractFromMessageConverter> messageConverters = new ArrayList<AbstractFromMessageConverter>();
+		if (applicationContext.containsBean(CUSTOM_MSG_CONVERTERS_BEAN_NAME)) {
+			messageConverters.addAll(
+					(Collection<AbstractFromMessageConverter>) applicationContext.getBean(CUSTOM_MSG_CONVERTERS_BEAN_NAME));
```
artembilan: That's error. 
You end up here with the `early BeanFactory access` hell.
Consider to move the logic to the `SmartInitializingSingleton`.

artembilan:
No test...
Is it OK already?..

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConvertersConfigurer.java
ilayaperumalg: @artembilan my understand was that the `beanFactory` lookup should end up creating the `singleton` as long as the `BeanDefinition` for that name exists. Though I see your point, I would like to take this opportunity to understand the side effects of early instantiation of the `singleton` this way. If you have any pointers please share. Meanwhile, I will switch to use `SmartInitializingSingleton`.

ilayaperumalg:
Sorry, it is definitely not ok without tests. So far I have been running this manually by adding a `customMessageConverters` bean in the s-c-s module.

```
@Bean
	public List<AbstractFromMessageConverter> customMessageConverters() {
		List<AbstractFromMessageConverter> converters = new ArrayList<>();
		converters.add(new AbstractFromMessageConverter(targetMimeTypes) {
			@Override
			protected Class<?>[] supportedTargetTypes() {
				return new Class<?>[] { byte[].class };
			}

			@Override
			protected Class<?>[] supportedPayloadTypes() {
				return new Class<?>[] { String.class };
			}

		});
		return converters;
	}
```

We have a long list of tests that are yet to be ported from XD to here. We will have them addressed soon.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConvertersConfigurer.java
```diff
@@ -0,0 +1,121 @@
...
+	private ChannelBindingServiceProperties channelBindingServiceProperties;
+
+	private ApplicationContext applicationContext;
+
+	private CompositeMessageConverterFactory messageConverterFactory;
+
+	@Override
+	public void afterSingletonsInstantiated() {
+		Collection<AbstractFromMessageConverter> messageConverters = new ArrayList<AbstractFromMessageConverter>();
+		if (applicationContext.containsBean(CUSTOM_MSG_CONVERTERS_BEAN_NAME)) {
```
markfisher: could we just support by-type autowiring instead?, e.g.:

```
@Autowored
List<AbstractFromMessageConverter>
```

...which could even be in the `ChannelBindingService

ilayaperumalg: That's a good idea. I will add it. The only caveat with this approach is that the `messageConverter` beans need to be specified separately rather than as a bean that returns `List`.

ilayaperumalg: My intention to keep the `messageTypeConversion` logic in this class rather than in `ChannelBindingService` is to delegate the conversion logic to the `MessageConvertersConfigurer` from `ChannelBindingService`. What do you think?

ericbottard: I like the automatic injection as well, but ordering may be an issue too here...

ilayaperumalg:
Closing this PR for now as we want to provide a better approach that uses annotation 

- to specify `override` the custom message converters over the existing converters 
- to specify `ordering` of the converters.

The module author would specify something like this:
```
@MessageConverters(override= true/false, first=true/false)
```

PR #146, opened 2015-10-05
XD-3338 Prevent optional dependencies to be automatically added as includes

mbogoevici:


markpollack:
squashed and merged as 28d1bee0dc3a732b165d22055004e714d3f166ff

PR #148, opened 2015-10-07
XD-3597 Separate Input, Output and Context start

mbogoevici:
Separates the functionality currently provided by the ChannelBindingLifecycle,
binding inputs and outputs separately at the latest, and earliest phases,
respectively - allowing for SmartLifecycle beans that subscribe to inputs at start
to subscribe before inputs start producing data.

The current functionality of automatically starting the context
on the ChannelBindingLifecycle auto-startup is deferred to a refresh listener.

Removes the call to application.stop() within the ChannelBindingLifecycle, which
is redundant (the ChannelBindingLifecycle would be stopped when the context itself
is stopped anyway).


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ContextStartAfterRefreshListener.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Automatically starts the context after a refresh.
+ *
+ * @author Marius Bogoevici
+ */
+public class ContextStartAfterRefreshListener implements ApplicationListener<ContextRefreshedEvent> {
+
+	@Override
+	public void onApplicationEvent(ContextRefreshedEvent event) {
+		ConfigurableApplicationContext source = (ConfigurableApplicationContext) event.getSource();
```
garyrussell: Strictly speaking, you should make this `ApplicationContextAware` and check the context in the event matches. We don't want to react to events from other contexts in the hierarchy - maybe not relevant here, but best practice.

mbogoevici: Ah, good point. Maybe use a `SmartInitializingSingleton` then?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
```diff
@@ -0,0 +1,116 @@
...
+	@Override
+	public void stop(Runnable callback) {
+		stop();
+		if (callback != null) {
+			callback.run();
+		}
+	}
+
+	/**
+	 * Return the lowest value to start this bean before any message producing lifecycle
```
garyrussell: ... highest value to start this bean after any consumer Lifecycle as been started.

garyrussell:
I wonder if we should leave some room at the extremities just in case a user wants to start a bean earlier/later than the bindings.

e.g. `MAX_VALUE - 1000` and `MIN_VALUE + 1000`.

mbogoevici:
We could do that probably. They're all arbitrary numbers anyway, so it wouldn't hurt - but I'm not sure how big a requirement this is, also being hard to document. 

Maybe we should have events - e.g. `BeforeBinding`,`AfterBinding`, and so on ... just thinking of a more deterministic way of handling this.  (Future work, for sure)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
mbogoevici: Gak.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ContextStartAfterRefreshListener.java
garyrussell: I think this is ok; you just need to ignore events from other contexts.

garyrussell:
Squashed and merged as d992e0b8526f54f2caf0234cd12a4d87d182bd36

PR #150, opened 2015-10-12
XD-3355, XD-3582: Support options metadata

ericbottard:
This adds support for gathering metadata about a module,
to be leveraged at the dataflow layer mostly (PR there on its way)

### DIFF
--- a/spring-cloud-stream-configuration-metadata/src/main/java/org/springframework/cloud/stream/configuration/metadata/ModuleConfigurationMetadataResolver.java
+++ b/spring-cloud-stream-configuration-metadata/src/main/java/org/springframework/cloud/stream/configuration/metadata/ModuleConfigurationMetadataResolver.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.configuration.metadata;
+
+import java.io.Closeable;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
+import org.springframework.boot.configurationmetadata.ConfigurationMetadataRepositoryJsonBuilder;
+import org.springframework.boot.loader.archive.Archive;
+import org.springframework.boot.loader.archive.ExplodedArchive;
+import org.springframework.boot.loader.archive.JarFileArchive;
+import org.springframework.cloud.stream.module.launcher.ModuleJarLauncher;
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
+import org.springframework.core.io.support.ResourcePatternResolver;
+
+/**
+ * Used to retrieve metadata about the configuration properties that can alter a module behavior.
+ *
+ * @author Eric Bottard
+ */
+public class ModuleConfigurationMetadataResolver {
+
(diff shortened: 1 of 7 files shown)


spring-cloud-stream-configuration-metadata/src/main/java/org/springframework/cloud/stream/configuration/metadata/ModuleConfigurationMetadataResolver.java
```diff
@@ -0,0 +1,77 @@
...
+import org.springframework.core.io.support.ResourcePatternResolver;
+
+/**
+ * Used to retrieve metadata about the configuration properties that can alter a module behavior.
+ *
+ * @author Eric Bottard
+ */
+public class ModuleConfigurationMetadataResolver {
+
+	public List<ConfigurationMetadataProperty> listProperties(Resource module) {
```
pperalta: Since this is a public method, perhaps some JavaDoc?

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleJarLauncher.java
```diff
@@ -47,6 +49,16 @@ public void launch(String[] args) {
 		super.launch(args);
 	}
 
+	/**
+	 * Create and return a new ClassLoader suitable for running the module, or accessing resources
+	 * in it. Disposing of the ClassLoader is the responsibility of the caller.
+	 */
+	public ClassLoader createClassLoader() throws Exception {
```
pperalta: I understand that this is a convenient place for this method because of the underlying infrastructure in place for launching jars. However I believe this method is out of place here. Instances of this class are supposed to be created for launching a jar, not for creating classloaders (hence the name of the class `ModuleJarLauncher`. Perhaps the parts that are required in `ModuleConfigurationMetadataResolver` could be extracted and placed there? I'll make a suggestion for us to pick at above.

spring-cloud-stream-configuration-metadata/src/main/java/org/springframework/cloud/stream/configuration/metadata/ModuleConfigurationMetadataResolver.java
```diff
@@ -0,0 +1,77 @@
...
+				}
+				catch (IOException e) {
+					// ignore
+				}
+			}
+		}
+		return result;
+	}
+
+	protected ClassLoader createClassLoader(Archive archive) throws Exception {
```
pperalta: Suggestion for this method so that it does not rely on `ModuleJarLauncher` - thoughts?

```java
	private static final AsciiBytes LIB = new AsciiBytes("lib/");

	protected ClassLoader createClassLoader(Archive archive) throws Exception {
		JarFile.registerUrlProtocolHandler();
		List<Archive> archives = new ArrayList<Archive>(
				archive.getNestedArchives(new Archive.EntryFilter() {
					@Override
					public boolean matches(Archive.Entry entry) {
						return !entry.isDirectory() && entry.getName().startsWith(LIB);
					}
				}));

		URL[] urls = new URL[archives.size()];
		int i = 0;
		for (Archive a : archives) {
			urls[i++] = a.getUrl();
		}

		return new LaunchedURLClassLoader(urls, getClass().getClassLoader());
	}
```

ericbottard:
About the whole ClassLoader discussion: my concern is that I don't want the logic for creating a module ClassLoader duplicated. If it's not in the launcher (which I agree is not ideal), then we need something that accepts a jar (location) and gives us a CL. The other thing in favor of putting it in the Launcher class is that it alone knows how to select subentries (ie things in `lib/`)

pperalta:
Would you consider this compromise? Instead of modifying `ModuleJarLauncher`, what about a private extension of `JarLauncher` within `ModuleConfigurationMetadataResolver`?

```java
	protected ClassLoader createClassLoader(Archive archive) throws Exception {
		return new ClassLoaderExposingJarLauncher(archive).createClassLoader();
	}

	private static class ClassLoaderExposingJarLauncher extends JarLauncher {

		public ClassLoaderExposingJarLauncher(Archive archive) {
			super(archive);
		}

		/**
		 * Create and return a new ClassLoader suitable for running the module,
		 * or accessing resources in it. Disposing of the ClassLoader is the
		 * responsibility of the caller.
		 */
		protected ClassLoader createClassLoader() throws Exception {
			JarFile.registerUrlProtocolHandler();
			List<Archive> classPathArchives = getClassPathArchives();
			return createClassLoader(classPathArchives);
		}
	}
```

pperalta:
Merged as 9b1e41841161e1f9115423995ae4537800c56ccb

PR #153, opened 2015-10-16
allow id & timestamp keys when converting to Tuple

markfisher:
also avoid possible NPE in the DefaultTuple toString() method


trisberg:
LGTM, tested with new jdbc sink

trisberg:
Merged via commit 4bec17ca1c1fb6ca4fc351f19c0a770c37a85700

PR #155, opened 2015-10-19
Refactor message converter config

ilayaperumalg:
- Move message converter configuration into a separate class `MessageConverterConfigurer`

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
@@ -16,32 +16,11 @@
 
 package org.springframework.cloud.stream.binding;
 
-import java.util.HashSet;
-import java.util.Set;
-
-import org.springframework.aop.framework.Advised;
-import org.springframework.aop.support.AopUtils;
-import org.springframework.beans.factory.InitializingBean;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.stream.binder.Binder;
-import org.springframework.cloud.stream.config.BindingProperties;
 import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
-import org.springframework.cloud.stream.converter.AbstractFromMessageConverter;
-import org.springframework.cloud.stream.converter.ByteArrayToStringMessageConverter;
-import org.springframework.cloud.stream.converter.CompositeMessageConverterFactory;
-import org.springframework.cloud.stream.converter.JavaToSerializedMessageConverter;
-import org.springframework.cloud.stream.converter.JsonToPojoMessageConverter;
-import org.springframework.cloud.stream.converter.JsonToTupleMessageConverter;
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -53,33 +32,16 @@
...
 
+	@Autowired
 	private Binder<MessageChannel> binder;
 
+	@Autowired
 	private ChannelBindingServiceProperties channelBindingServiceProperties;
 
-	private CompositeMessageConverterFactory messageConverterFactory;
-
-	public ChannelBindingService(ChannelBindingServiceProperties channelBindingServiceProperties, Binder<MessageChannel> binder) {
```
mbogoevici: Kind of a nit, but I think we should keep the `Binder` and `ChannelBindingServiceProperties` configured via constructor   to keep things clear (for that matter we could make them final too). The mix of no-arg construction in `ChannelBindingServiceConfiguration` and field autowiring makes this a little hard to follow, IMO.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.java
```diff
@@ -64,10 +65,10 @@
 	// it is used to detect a ChannelBindingService in the parent context (which we know
 	// already exists).
 	@ConditionalOnMissingBean(ChannelBindingService.class)
-	public ChannelBindingService bindingService(
-			ChannelBindingServiceProperties channelBindingServiceProperties,
+	public ChannelBindingService bindingService(ChannelBindingServiceProperties channelBindingServiceProperties,
 			Binder<MessageChannel> binder) {
-		return new ChannelBindingService(channelBindingServiceProperties, binder);
+		return new ChannelBindingService(channelBindingServiceProperties, binder,
+				messageConvertersConfigurer(channelBindingServiceProperties));
```
mbogoevici: I think we can rather pass the instance as an argument to the `bindingService()` method instead.

mbogoevici:
Looking at this, I have the feeling that we can take this a step further and extract the entire channel creation logic from `BindableProxyFactory` in a `ChannelFactory`, that the  `BindableProxyFactory`  could delegate to for creating channels (including adding converters) - then we wouldn't need the separate `MessageConverterConfigurer` and repeated calls to `ChannelBindingService`. (In theory, we could handle channel creation as part of `ChannelBindingService`, but it would be nice to have those responsibilities separated.

ilayaperumalg:
That's a good idea. But I believe it is not straightforward to have `ChannelBindingServiceProperties` in proxy factory class to do message converter configuration.

mbogoevici:
@ilayaperumalg that wouldn't be necessary. `ChannelBindingServiceProperties` would be injected in the `ChannelFactory`

ilayaperumalg:
Closing this PR to issue a new one that uses `ChannelFactory`.

PR #157, opened 2015-10-20
XD-3601 Add unique Id for ModuleLaunchRequest

ilayaperumalg:
- Add `requestId` to ModuleLaunchRequest which uniquely identifies the ModuleLaunchRequest
   - This is useful when there are multiple module launch requests for the same module and avoid any naming clash with the module launch requests
 - One other possible solution is to append `server.port` to the ModuleLaunchRequest
   - Since s-c-s module always has an embedded servlet container this might be a simple fix. But having unique requestId would make it easy to understand (for instance we can have streamName + moduleLabel as the requestId)

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLaunchRequest.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/launcher/ModuleLaunchRequest.java
@@ -24,16 +24,28 @@ import java.util.Map;
  * Those arguments will be passed to the module by the launcher.
  *
  * @author Eric Bottard
+ * @author Ilayaperumal Gopinathan
  */
 public class ModuleLaunchRequest {
+
+	/**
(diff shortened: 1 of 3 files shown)


ericbottard:
> This is useful when there are multiple module launch requests for the same module

How can this happen, I am confused. Race condition? Because for a regular stream, the requests are already qualified by `label` + `group`, which ought to be different between 2 deploys

ericbottard:
I see now this is related to the JMX clash. So I guess my question is, in a nutshell, "why introduce a new concept? why not use `group+label`, either at the caller or the ML site?"

ericbottard:
Doh! We don't have access to `group+label` at this layer. Sorry for the noise. Then I guess I'd agree with Marius that we should simply pass the dedicated jmx stuff from the s-c-d layer, maybe crafting a default value if not present

markfisher:
+1 for passing the unique ID along. The current use of the full coordinate string is not ideal.

ilayaperumalg:
Closing this PR as it is assumed to be the responsibility of the user to specify the appropriate jmx domain name for the module when launching.

PR #158, opened 2015-10-20
ChannelsEndpoint should return a Map<> (not a domain object)

dsyer:
This is a general rule for endpoints so they work OOTB with JMX.


markfisher:
LGTM; merged as https://github.com/spring-cloud/spring-cloud-stream/commit/5eaba5811c417795254c51a86aca732b5dfb1e66

PR #161, opened 2015-10-21
no longer assuming "exec" classifier for modules

markfisher:


mbogoevici:
We need to update the existing module launcher examples (can add an extra commit on merge).

markfisher:
yes, good point - sorry!

mbogoevici:
Merged, added https://github.com/spring-cloud/spring-cloud-stream/commit/4315b1d9d2f88b2ada9fd3bb4270ed301a012815 to update `README.MD`

PR #162, opened 2015-10-23
Use ChannelFactory

ilayaperumalg:
- ChannelFactory to create message channels and configure message converters


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -102,130 +72,12 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
...
-
-	private String getNamespacePrefixedChannelName(String name) {
-		return this.channelNamespace + "." + name;
-	}
-
-	private void bridgeSubscribableToPollableChannel(SubscribableChannel sharedChannel,
-			MessageChannel inputChannel) {
-		sharedChannel.subscribe(new MessageChannelBinderSupport.DirectHandler(
-				inputChannel));
+		this.channelFactory = beanFactory.getBean(ChannelFactory.class);
```
markfisher: why are we doing a bean lookup here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -252,7 +104,7 @@ public synchronized Object invoke(MethodInvocation invocation) throws Throwable
 	@Override
 	public synchronized Object getObject() throws Exception {
 		if (this.proxy == null) {
-			createChannels(this.type);
+			this.channelFactory.createChannels(this.type);
```
markfisher: this feels like a brittle contract... having to call `createChannels` prior to calling `getInputs` or `getOutputs`

markfisher: this feels like a brittle contract... having to call `createChannels` prior to calling `getInputs` or `getOutputs`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -335,29 +185,13 @@ public void unbindOutputs(ChannelBindingService channelBindingService) {
...
-
-		public MessageChannel getMessageChannel() {
-			return this.messageChannel;
-		}
-
-		public boolean isBindable() {
-			return this.bindable;
-		}
+	@Override
+	public Set<String> getInputs() {
```
markfisher: should this method be called `getInputNames()` instead? (same question for `getOutputs`)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
ilayaperumalg: No, this isn't needed. I was trying something else and forgot to remove this and use `Autowired`. will fix it.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
ilayaperumalg: yeah, I will change.

ilayaperumalg: Do you think we can rename this to `getInputBindingNames()` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
ilayaperumalg: I believe `createChannels` needs to be called before `getInputs` and `getOutputs` as the createChannels method would have the `inputs` and `outputs` set with appropriate channelBindingNames/MessageChannel. What am I missing?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
mbogoevici: Well, maybe I'm missing something, but the channel factory doesn't seem to be injecting anywhere either.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
mbogoevici: The class is already called  `Bindable` ...

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
mbogoevici: This can be done at the end of `afterPropertiesSet()`, I believe - and it'll take care of the issue. What we want is for the channels to be already created when they're injected into their target beans.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactory.java
```diff
@@ -0,0 +1,275 @@
...
+	private ConfigurableListableBeanFactory beanFactory;
+
+	@Autowired(required = false)
+	private SharedChannelRegistry sharedChannelRegistry;
+
+	private CompositeMessageConverterFactory messageConverterFactory;
+
+	private final ChannelBindingServiceProperties channelBindingServiceProperties;
+
+	private Map<String, ChannelHolder> inputs = new HashMap<>();
```
mbogoevici: Let's put this back inside `BindableFactoryProxy` - this is merging all inputs and outputs for all `Bindable` instances across the context. This class shouldn't have state at all, just a strategy for instantiating channels.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -274,10 +126,9 @@ public void bindInputs(ChannelBindingService channelBindingService) {
 		if (log.isDebugEnabled()) {
 			log.debug(String.format("Binding inputs for %s:%s", this.channelNamespace, this.type));
 		}
-		for (Map.Entry<String, ChannelHolder> channelHolderEntry : this.inputs.entrySet()) {
+		for (Map.Entry<String, ChannelHolder> channelHolderEntry : this.channelFactory.getInputs().entrySet()) {
```
mbogoevici: I don't think we need to move the channels managed by a `Bindable` move into `ChannelFactory`, but if we do, then binding should happen elsewhere. This effectively results in binding/unbinding channels multiple times if we have something like `@EnableBindings(Source.class,Sink.class)`.
My suggestion is to keep the channels of a given proxy instance inside the proxy (as it is more intuitive for something like `source.output()` ) and basically revert the move.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
ilayaperumalg: I meant, I will add injection when removing the bean factory lookup.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
ilayaperumalg: Yeah, that was my mistake to move the channels into `ChannelFactory`. Pushed the fix.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -250,9 +95,14 @@ public synchronized Object invoke(MethodInvocation invocation) throws Throwable
 	}
 
 	@Override
+	public void afterPropertiesSet() throws Exception {
+		Assert.notNull(this.channelFactory, "ChannelFactory must not be null.");
```
markfisher: this would have already failed at `@Autowired` given that `required=true` by default

ilayaperumalg: The `channelFactory` bean always exists as this gets created as part of `ChannelBindingServiceConfiguration`.

mbogoevici:
I still don't think that `ChannelFactory` should hold any created channels, but simply have a `createMessageChannel()` method that encapsulates the work done in `createChannels()`, per channel - doing the channel instantiation and convertor setup based on `ChannelBindingProperties`

ilayaperumalg:
@mbogoevici Based on my last push on this PR, the `ChannelFactory` doesn't hold any created channels.

markfisher:
but rather than modifying passed in Maps for the channels, I believe what @mbogoevici is suggesting is that the `ChannelFactory` only creates a single channel on-demand... handling multiple channels would be a responsibility of the caller. And +1 from me on that idea.

ilayaperumalg:
Pushed the changes that has `ChannelFactory` which just creates message channels. There are couple of caveats with this approach though:
1) The create channel methods need to be separate for input/output especially for the case `Processor` especially.
2) The `createChannel` method needs to return `ChannelHolder` so that channel `name` and `bindable` information is passed along when the channel is created.

markfisher:
I think the `ChannelFactory` still has too much responsibility. In my opinion, it shouldn't be handling the introspection of annotations on a class. That should happen on the caller's side, and this should simply provide a method, such as `createChannel(name, type)`, where type is assignable to `MessageChannel`. That would eliminate the need for `ChannelHolder` (which I think is a code smell itself). Not to mention, the current implementation would break for any interface that has more than one input or output channel.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactory.java
```diff
@@ -0,0 +1,266 @@
...
+	@Value("${" + CHANNEL_NAMESPACE_PROPERTY_NAME + ":}")
+	private String channelNamespace;
+
+	@Value("${" + POLLABLE_BRIDGE_INTERVAL_PROPERTY_NAME + ":1000}")
+	private int pollableBridgeDefaultFrequency;
+
+	private ConfigurableListableBeanFactory beanFactory;
+
+	@Autowired(required = false)
+	private SharedChannelRegistry sharedChannelRegistry;
```
mbogoevici: IMO this could go back to BindableChannelProxy, who would consult it and delegate to ChannelFactory if and only if it needs to create a new channel.

mbogoevici:
I agree with @markfisher here. The goal of this whole exercise was to extract the initialization of converters. I still think that `ChannelHolder` is good to have (but as it was - a data structure of `BindableChannelProxy`, which held information about the channels managed by the proxy - i.e. whether they are bound or not and so on) - so perhaps putting it back as an inner class of `BindableProxyChannel`. `ChannelFactory` should be a very simple stateless component, agree with `createChannel(name,type)`. Also, I am not overly fond of its methods being package protected. It may not be a general purpose component, but it makes its contract hard to read. Perhaps we should have an interface (helps with testing, too).

ilayaperumalg:
I agree. Pushed the changes based on the suggestions.

markfisher:
Good point. We should define a single (public) method `ChannelFactory` interface. That will help emphasize the single well-defined responsibility of this component, based on a contract. I could imagine in the future that our logic for resolving channel metadata could require a *separate* abstraction (where we currently assume the annotation-driven approach), but for now that can remain within the `BindableProxyFactory` itself. I'm okay with `ChannelHolder` as a convenience within the `BindableProxyFactory` as well, but would prefer that remain an inner class if possible.

ilayaperumalg:
Pushed after a bit of refactoring. I guess the pendulum has now swung to the other direction ;) The proxy class has too much things; but we have the `ChannelFactory` does only the channel creation. I also moved the message converter configuration separate so that it can be re-used for the `shared` channel configuration at the proxy class.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactory.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Defines methods to create/configure the {@link org.springframework.messaging.MessageChannel} defined
+ * in {@link org.springframework.cloud.stream.annotation.EnableBinding}.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public interface ChannelFactory {
+
+	String SPRING_CLOUD_STREAM_INTERNAL_PREFIX = "spring.cloud.stream.internal";
```
mbogoevici: Let's place the property names elsewhere (maybe in the implementation). `ChannelFactory` should only define the contract in `createChannel`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactoryImpl.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.messaging.PollableChannel;
+
+/**
+ * Class that {@link BindableProxyFactory} uses to create message channels.
+ *
+ * @author Marius Bogoevici
+ * @author David Syer
+ * @author Ilayaperumal Gopinathan
+ */
+public class ChannelFactoryImpl implements ChannelFactory {
```
mbogoevici: We could call this `DefaultChannelFactory` perhaps?

ilayaperumalg: yeah, I like it :-)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/Bindable.java
```diff
@@ -50,11 +50,11 @@
 	/**
 	 * Enumerates all the input binding names.
 	 */
-	Set<String> getInputs();
+	Set<String> getInputNames();
```
mbogoevici: After a quick chat w/ @markfisher - whose initial comment triggered the renaming, we agreed on keeping the original names for consistency with the rest of the methods (will fix on merge).

mbogoevici:
Merged w/changes discussed above: https://github.com/spring-cloud/spring-cloud-stream/commit/ae770cec6997c1b3914e73d26b540c83e171396e

Issue #166, opened 2015-10-28
NPE in the RedisQueueOutboundChannelAdapter caused by the misconfiguration for the RedisMessageChannelBinder


artembilan:
```
2015-10-28 11:45:43,630 ERROR inbound.retry.0-redis:queue-inbound-channel-adapter1 o.s.c.s.b.r.RedisMessageChannelBinder$1:286 - Failed to deliver message
org.springframework.messaging.MessageHandlingException: error occurred in message handler [org.springframework.integration.redis.outbound.RedisQueueOutboundChannelAdapter@5f977c77]; nested exception is java.lang.IllegalArgumentException: EvaluationContext must not be null
	at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:139) ~[spring-integration-core-4.2.1.RELEASE.jar:na]
	at org.springframework.cloud.stream.binder.redis.RedisMessageChannelBinder$1$2.recover(RedisMessageChannelBinder.java:277) ~[classes/:na]
	at org.springframework.cloud.stream.binder.redis.RedisMessageChannelBinder$1$2.recover(RedisMessageChannelBinder.java:267) ~[classes/:na]
	at org.springframework.retry.support.RetryTemplate.handleRetryExhausted(RetryTemplate.java:458) ~[spring-retry-1.1.2.RELEASE.jar:na]
	at org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:320) ~[spring-retry-1.1.2.RELEASE.jar:na]
	at org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:168) ~[spring-retry-1.1.2.RELEASE.jar:na]
	at org.springframework.cloud.stream.binder.redis.RedisMessageChannelBinder$1.doSend(RedisMessageChannelBinder.java:260) ~[classes/:na]
	at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:442) [spring-integration-core-4.2.1.RELEASE.jar:na]
	at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:392) [spring-integration-core-4.2.1.RELEASE.jar:na]
	at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:115) [spring-messaging-4.2.3.BUILD-20151027.143026-12.jar:4.2.3.BUILD-SNAPSHOT]
[...]

PR #167, opened 2015-10-28
GH-166: Fix `ChannelBinder` config inconsistency

artembilan:
Fixes GH-166 (https://github.com/spring-cloud/spring-cloud-stream/issues/166)

Since `onInit()` in the `MessageChannelBinderSupport` was called before the `evaluationContext` population,
the `RedisMessageChannelBinder#errorAdapter` caused `NPE` for its `evaluationContext`.

* Change the initialization order
* Use `ExpressionUtils` to populate `evaluationContext` - creates `StandardEvaluationContext` if there is no one in the `applicationContext`
* Add more initializations for the `RedisMessageChannelBinder#errorAdapter`
* Remove explicit `evaluationContext` population from the `AbstractTestBinder` to rely on the fix for this ticket
* Some other simple polishing

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
@@ -147,8 +147,7 @@ public class RedisMessageChannelBinder extends MessageChannelBinderSupport imple
 			String... headersToMap) {
 		Assert.notNull(connectionFactory, "connectionFactory must not be null");
 		this.connectionFactory = connectionFactory;
-		this.errorAdapter = new RedisQueueOutboundChannelAdapter(
-				parser.parseExpression("headers['" + ERROR_HEADER + "']"), connectionFactory);
+
 		if (headersToMap != null && headersToMap.length > 0) {
 			String[] combinedHeadersToMap =
 					Arrays.copyOfRange(BinderHeaders.STANDARD_HEADERS, 0, BinderHeaders.STANDARD_HEADERS.length
@@ -160,10 +159,14 @@ public class RedisMessageChannelBinder extends MessageChannelBinderSupport imple
 		else {
 			this.headersToMap = BinderHeaders.STANDARD_HEADERS;
 		}
+
+		this.errorAdapter = new RedisQueueOutboundChannelAdapter(
+				parser.parseExpression("headers['" + ERROR_HEADER + "']"), connectionFactory);
 	}
 
 	@Override
-	protected void onInit() {
+	public void afterPropertiesSet() throws Exception {
+		super.afterPropertiesSet();
 		this.errorAdapter.setIntegrationEvaluationContext(this.evaluationContext);
 		this.errorAdapter.setBeanFactory(getBeanFactory());
 		this.errorAdapter.afterPropertiesSet();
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-spi/src/main/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupport.java
```diff
@@ -602,18 +603,17 @@ protected final MessageValues deserializePayloadIfNecessary(Message<?> message)
 	}
 
 	protected final MessageValues deserializePayloadIfNecessary(MessageValues message) {
-		MessageValues messageToSend = message;
 		Object originalPayload = message.getPayload();
-		MimeType contentType = this.contentTypeResolver.resolve(messageToSend);
+		MimeType contentType = this.contentTypeResolver.resolve(message);
 		Object payload = deserializePayload(originalPayload, contentType);
```
garyrussell: Why this change? It is considered bad practice (and Sonar complains about it) to modify a parameter value.

artembilan: `MessageValues messageToSend = message;` - are two vars for the same object, so any modifications have effect

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
```diff
@@ -165,6 +165,8 @@ public RedisMessageChannelBinder(RedisConnectionFactory connectionFactory,
 	@Override
 	protected void onInit() {
 		this.errorAdapter.setIntegrationEvaluationContext(this.evaluationContext);
+		this.errorAdapter.setBeanFactory(getBeanFactory());
+		this.errorAdapter.afterPropertiesSet();
```
mbogoevici: Looking at how the `errorAdapter` is setup, I think we'd best move its entire initialization in `onInit()` - makes it an easier read than the current split.

spring-cloud-stream-binders/spring-cloud-stream-binder-spi/src/main/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupport.java
```diff
@@ -377,10 +378,10 @@ public void setDefaultDurableSubscription(boolean defaultDurableSubscription) {
 	@Override
 	public void afterPropertiesSet() throws Exception {
 		Assert.notNull(this.applicationContext, "The 'applicationContext' property cannot be null");
-		onInit();
 		if (this.evaluationContext == null) {
-			this.evaluationContext = IntegrationContextUtils.getEvaluationContext(getBeanFactory());
+			this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
 		}
+		onInit();
```
mbogoevici: With this move, I think we could remove `onInit()` entirely and just rely on `super.afterPropertiesSet()` calls in subclasses.

artembilan: Yes, makes sense, too.
Having all this `MessageChannelBinderSupport` hierarchy as an internal stuff, there is no so much worry about end-users who won't call `super.afterPropertiesSet()` from his impl.

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
artembilan: I don't mind. Will do together with the `afterPropertiesSet()` refactoring.

artembilan:
Pushed, but I left `errorAdapter` instantiation in the ctor to keep it as `final` for better performance.

mbogoevici:
OK, we'll just need to remember not to touch any of this. 

LGTM, will merge.

mbogoevici:
Rebased, squashed, merged https://github.com/spring-cloud/spring-cloud-stream/commit/2e7f9cc729d37651985eff5485bb684f34c1e2a0

PR #169, opened 2015-10-28
XD-3651 Fix handling of numeric, boolean and null values

trisberg:
- this applies to the Json-Tuple conversions

- we need to be able to support docs like this: {"id": 19, "name": "Anna", "age": null} without loosing the null or the numeric type of the id

### DIFF
--- a/spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/JsonNodeToTupleConverter.java
+++ b/spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/JsonNodeToTupleConverter.java
@@ -45,6 +45,15 @@ public class JsonNodeToTupleConverter implements Converter<JsonNode, Tuple> {
 				else if (node.isArray()) {
 					builder.addEntry(name, nodeToList(node));
 				}
+				else if (node.isNull()) {
+					builder.addEntry(name, null);
+				}
+				else if (node.isBoolean()) {
+					builder.addEntry(name, node.booleanValue());
+				}
+				else if (node.isNumber()) {
+					builder.addEntry(name, node.numberValue());
+				}
 				else {
 					builder.addEntry(name, node.asText());
 				}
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/JsonNodeToTupleConverter.java
```diff
@@ -66,6 +75,15 @@ else if (node.isArray()) {
 			else if (item.isArray()) {
 				list.add(nodeToList(item));
 			}
+			else if (node.isNull()) {
```
mbogoevici: Do these checks belong here? AFAICT it's already been established that the node is an array.

mbogoevici: Actually, should be `item.isNull()` and so on. Will fix on merge.

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/JsonStringToTupleConverter.java
```diff
@@ -73,6 +82,15 @@ else if (node.isArray()) {
 			else if (item.isArray()) {
 				list.add(nodeToList(item));
 			}
+			else if (node.isNull()) {
```
mbogoevici: Same comment as https://github.com/spring-cloud/spring-cloud-stream/pull/169/files#r43447550

trisberg: yeah, for the nodeToList tests they should be item.isXxx()

mbogoevici:
Rebased, merged with changes mentioned above: https://github.com/spring-cloud/spring-cloud-stream/commit/1b10c8801aa96bda1b5903c8a2b204de9539a7a1

PR #189, opened 2015-11-02
XD-3512: Initial GemFire binder

### LINKED ISSUE #191: Implement Gemfire message-channel binder
As a s-c-s user, I'd like to have *Gemfire* message-channel binder, so I can use *Gemfire* as the messaging middleware for low latency use-cases. 

**Acceptance Criteria:**
* *Gemfire* as a binder exist in s-c-s repo under *spring-cloud-stream-binder-gemfire* package
* Including *spring-cloud-stream-binder-gemfire* as binder successfully allows using *Gemfire* as messaging middleware

Resolves #189

### LINKED ISSUE #198: Use GemFire AsyncEventQueue for binder
Experiment with GemFire's `AsyncEventQueue` to pass messages sent to a region to a channel.

pperalta:
Initial implementation of GemFire binder.
See https://docs.google.com/document/d/18gjwnh50CADCVyfnEh2h5cbgrIZm6n4UYQlaHGQW_QU/edit
for more details.

Next steps include:
 * Partitioning support
 * Configuration support (including cache topologies)
 * Automated test suite

This resolves #191.
This closes #198.


sabbyanandan:
This resolves #191.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+	<description>Gemfire binder implementation</description>
+
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-stream-binders-parent</artifactId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
ericbottard: I believe this is inherited

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+			<artifactId>spring-cloud-stream-binder-spi</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-codec</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-autoconfigure</artifactId>
+			<optional>true</optional>
```
ericbottard: Please explain the rationale for all the `optional` here except `spring-boot-configuration-processor`

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/GemfireMessageChannelBinder.java
```diff
@@ -0,0 +1,338 @@
...
+		public void run() {
+			// only messages that are present in this JVM will be processed
+			Region<MessageKey, Message<?>> localMessageRegion =
+					PartitionRegionHelper.getLocalData(this.messageRegion);
+
+			while (isRunning()) {
+				if (localMessageRegion.isEmpty()) {
+					try {
+						synchronized (this.messageListener) {
+							this.messageListener.wait(100);
```
ericbottard: I assume this `100` would be made configurable?

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/GemfireMessageChannelBinder.java
```diff
@@ -0,0 +1,338 @@
...
+					logger.debug("Fetched {} messages", keys.size());
+
+					if (this.inOrder) {
+						Collections.sort(keys);
+					}
+					if (keys.size() > this.batchSize) {
+						keys = keys.subList(0, this.batchSize);
+					}
+
+					List<MessageKey> errorKeys = null;
```
ericbottard: Why not use `clear()` instead if `== null` tests all over the place?

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * Object returned for {@link #getRoutingObject}.
+	 */
+	private int routingHash;
+
+	/**
+	 * Constructor for serialization.
+	 */
+	public MessageKey() {
```
ericbottard: Can the visibility of this be reduced?

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+	 * Construct a {@code MessageKey}.
+	 *
+	 * @param sequenceId unique sequence id used for ordering messages
+	 * @param producerTimestamp time stamp of when the producer process started
+	 * @param producerPid PID of producer process
+	 */
+	public MessageKey(long sequenceId, long producerTimestamp, int producerPid) {
+		this.sequenceId = sequenceId;
+		this.producerId = producerId(producerTimestamp, producerPid);
+		this.routingHash = this.hashCode();
```
ericbottard: I believe this class will never be subclassed, but if it were, this call could be problematic

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+	 * process, rightmost 48 bits are reserved for the timestamp of when the
+	 * producer process was started.
+	 *
+	 * @param producerTimestamp startup timestamp for producer
+	 * @param producerPid PID for producer
+	 * @return producer id
+	 */
+	private long producerId(long producerTimestamp, int producerPid) {
+		long longPid = producerPid;
+		longPid = longPid << 48;
```
ericbottard: this line can be folded in the line above. The optimizer will happily do it for you though

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+	@Override
+	public int compareTo(MessageKey that) {
+		int c = Long.compare(this.getProducerTimestamp(), that.getProducerTimestamp());
+		return (c == 0 ? Long.compare(this.sequenceId, that.sequenceId) : c);
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) {
```
ericbottard: `o == null` is not necessary given line above

ericbottard: scratch that

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+		MessageKey that = (MessageKey) o;
+		return this.sequenceId == that.sequenceId
+				&& this.producerId == that.producerId;
+	}
+
+	@Override
+	public int hashCode() {
+		int result = (int) (sequenceId ^ (sequenceId >>> 32));
+		result = 31 * result + (int) (producerId ^ (producerId >>> 32));
+		result = 31 * result + routingHash;
```
ericbottard: routingHash does not appear in `hashCode()`, on purpose?

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+	@Override
+	public String toString() {
+		return "MessageKey{" +
+				"id=" + sequenceId +
+				", producerId=" + producerId +
+				'}';
+	}
+
+	@Override
+	public Object getRoutingObject(EntryOperation<MessageKey, Message<?>> entryOperation) {
```
ericbottard: Not sure how often this is called, but may be better to store `routingHash` as an `Integer` to avoid repeated boxing

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+	public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
+		sequenceId = in.readLong();
+		producerId = in.readLong();
+		routingHash = in.readInt();
+	}
+
+	/**
+	 * Simple program to print the size of a serialized key.
+	 */
+	public static void main(String[] args) throws IOException {
```
ericbottard: Should this be removed?

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/config/GemfireExecutorConfigurationProperties.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.springframework.cloud.stream.binder.gemfire.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Patrick Peralta
+ */
+@ConfigurationProperties(prefix = "spring.cloud.stream.binder.gemfire.executor")
```
ericbottard: Maybe this should go as a pojo in `GemfireMessageChannelBinderConfiguration`?

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/config/GemfireMessageChannelBinderConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * @author Patrick Peralta
+ */
+@Configuration
+@ConfigurationProperties(prefix = "spring.cloud.stream.binder.gemfire")
+@EnableConfigurationProperties(GemfireExecutorConfigurationProperties.class)
+
```
ericbottard: newline

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,2 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration:\
+org.springframework.cloud.stream.binder.gemfire.config.GemfireBinderAutoConfiguration
```
ericbottard: newline

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/test/java/org/springframework/cloud/stream/binder/gemfire/GemfireBinderTests.java
```diff
@@ -0,0 +1,266 @@
...
+	 * @throws Exception
+	 */
+	@Test
+	public void testMessageSendReceive() throws Exception {
+		LocatorLauncher locatorLauncher = null;
+		JavaApplication consumer = null;
+		JavaApplication producer = null;
+
+		try {
+			locatorLauncher = new LocatorLauncher.Builder()
```
ericbottard: Should that setup be extracted?
What about code that was added when we did the GemFire sink? Shouldn't we consolidate all this?
see https://github.com/spring-cloud/spring-cloud-stream-modules/pull/22#issuecomment-133854769

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/GemfireMessageChannelBinder.java
markfisher: I had the same question initially, but there's really no reason to make this configurable. The wait would be interrupted by a notify (this is not a sleep), so there's no user-facing benefit of adjusting this time... it's simply here to moderate the tight spin.

ericbottard: Ah, very good point

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
pperalta: Nope, it's a mistake. Good catch!

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
pperalta: That's an interesting idea. I'm not sure how often it will be invoked either. My thought is to consume as little memory as possible, but repeated boxing would also be inefficient. My guess is that it won't be called very often; it would just be used to determine the hash for selecting a bucket. I'll look into it.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
pperalta: Yes.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/test/java/org/springframework/cloud/stream/binder/gemfire/GemfireBinderTests.java
pperalta: Yup, we can definitely extract this test code at some point. I purposefully didn't do it as part of this PR. As I add more tests for different parts of the functionality, it will be easier to determine the correct abstractions for GF tests.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
```diff
@@ -0,0 +1,229 @@
...
+		return (int) longPid;
+	}
+
+	/**
+	 * Return the timestamp of when the producer of this message key started.
+	 *
+	 * @return producer startup timestamp
+	 */
+	public long getProducerTimestamp() {
+		return producerId << 16 >> 16;
```
garyrussell: Keep the `producerTimestamp` as a field to avoid the need to strip out the pid on every `compareTo()`.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/GemfireMessageChannelBinder.java
```diff
@@ -0,0 +1,338 @@
...
+						shutdown();
+					}
+				}
+				else {
+					List<MessageKey> keys = new ArrayList<>(localMessageRegion.keySet());
+					logger.debug("Fetched {} messages", keys.size());
+
+					if (this.inOrder) {
+						Collections.sort(keys);
+					}
```
garyrussell: This is not strictly "inOrder" - only in order within each producer.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/test/java/org/springframework/cloud/stream/binder/gemfire/GemfireBinderTests.java
```diff
@@ -0,0 +1,266 @@
...
+import org.springframework.messaging.SubscribableChannel;
+import org.springframework.messaging.support.ExecutorSubscribableChannel;
+import org.springframework.messaging.support.GenericMessage;
+
+/**
+ * Tests for {@link GemfireMessageChannelBinder}.
+ *
+ * @author Patrick Peralta
+ */
+public class GemfireBinderTests {
```
garyrussell: Should extend `PartitionCapableBinderTests` and `@Overrride @Ignore` tests for features not currently supported.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/pom.xml
pperalta: I don't know - but it's the same for all of the binders:

```
$ grep -A 1 -r "<artifactId>spring-boot-autoconfigure" *
spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/pom.xml:			<artifactId>spring-boot-autoconfigure</artifactId>
spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/pom.xml-			<optional>true</optional>
--
spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml:			<artifactId>spring-boot-autoconfigure</artifactId>
spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml-			<optional>true</optional>
--
spring-cloud-stream-binders/spring-cloud-stream-binder-local/pom.xml:			<artifactId>spring-boot-autoconfigure</artifactId>
spring-cloud-stream-binders/spring-cloud-stream-binder-local/pom.xml-			<optional>true</optional>
--
spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/pom.xml:			<artifactId>spring-boot-autoconfigure</artifactId>
spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/pom.xml-			<optional>true</optional>
--
spring-cloud-stream-binders/spring-cloud-stream-binder-redis/pom.xml:			<artifactId>spring-boot-autoconfigure</artifactId>
spring-cloud-stream-binders/spring-cloud-stream-binder-redis/pom.xml-			<optional>true</optional>
--
spring-cloud-stream-codec/pom.xml:			<artifactId>spring-boot-autoconfigure</artifactId>
spring-cloud-stream-codec/pom.xml-			<optional>true</optional>
```

I can issue a separate PR for all of these if everyone agrees.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/GemfireMessageChannelBinder.java
pperalta: Good point - I can just allocate the `ArrayList` once and clear it out every time.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
pperalta: I think so; I'll try it out.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
pperalta: I should mark this class `final`.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
pperalta: Yes, I can add it as a transient field.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/GemfireMessageChannelBinder.java
pperalta: Is there any other kind of ordering other than what a single thread in a producer creates? Otherwise we'd be talking about global ordering, which means a single point of bottleneck for new messages.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/test/java/org/springframework/cloud/stream/binder/gemfire/GemfireBinderTests.java
pperalta: OK, I'll look into it!

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
pperalta: Making this protected results in:

```
Caused by: java.io.InvalidClassException: org.springframework.cloud.stream.binder.gemfire.MessageKey; no valid constructor
	at  in java.io.ObjectStreamClass$ExceptionInfo.newInvalidClassException(ObjectStreamClass.java:150)
	at  in java.io.ObjectStreamClass.checkDeserialize(ObjectStreamClass.java:768)
	at  in java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1775)
	at  in java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1351)
```

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/test/java/org/springframework/cloud/stream/binder/gemfire/GemfireBinderTests.java
pperalta: Once I add real partitioning support (right now it's only supported as part of `MessageKey` but is not yet configurable) I'll introduce `PartitionCapableBinderTests`.

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/MessageKey.java
ericbottard: yep, that's what I feared as soon as I understood we were talking native java serialization. Too bad ;)

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/test/java/org/springframework/cloud/stream/binder/gemfire/GemfireBinderTests.java
ericbottard: You could try to extend from its superclass still, already adds interesting tests "for free" to validate your impl

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/src/main/java/org/springframework/cloud/stream/binder/gemfire/GemfireMessageChannelBinder.java
garyrussell: I am just saying that with other buses; if there are multiple producer instances, the consumer(s) will get the messages in approximate time order (regardless of which producer published).

With this bus, the ordering will be time within producer; I guess it's only really a problem if the producer is running for an extended time with no consumer(s).

spring-cloud-stream-binders/spring-cloud-stream-binder-gemfire/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-configuration-processor</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-binder-spi</artifactId>
```
mbogoevici: This needs to change to `spring-cloud-stream` after the move of the SPI in the core project.

 https://github.com/spring-cloud/spring-cloud-stream/pull/216

pperalta:
Closing PR pending upcoming work on consumer group support.

PR #194, opened 2015-11-03
XD-3519: Rabbit Taps GH-193

### LINKED ISSUE #193: Add TAP support for Rabbit binder
As a user, I'd like to *tap* the primary pipeline, so I can fork the same data and do some ad-hoc analysis without impacting the original stream.

For background and design notes, please refer to [XD-3519](https://jira.spring.io/browse/XD-3519)

garyrussell:
Resolves #193 

- instead of publishing to the default exchange, producers now publish to a topic exchange, to which the main queue is bound
- taps on a stream simply bind a queue to the same exchange with a name including the group for the tapping stream and a
    routing key pattern `#`
 - this means that multiple instances of the same tap stream will compete but multiple tap streams will get a copy
- for partitioned data, a single topic exchange is used, which each partition queue bound with its name as the routing key
 - this means taps on partitioned data see the consolidated messages

Add group to ChannelBindingServiceProperties

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitBindingCleaner.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitBindingCleaner.java
@@ -25,13 +25,13 @@ import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.springframework.cloud.stream.binder.BindingCleaner;
-import org.springframework.web.client.RestTemplate;
-import org.springframework.web.util.UriComponentsBuilder;
 import org.springframework.cloud.stream.binder.BinderUtils;
+import org.springframework.cloud.stream.binder.BindingCleaner;
 import org.springframework.cloud.stream.binder.MessageChannelBinderSupport;
 import org.springframework.cloud.stream.binder.RabbitAdminException;
 import org.springframework.cloud.stream.binder.RabbitManagementUtils;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
 
 
 /**
@@ -44,7 +44,7 @@ public class RabbitBindingCleaner implements BindingCleaner {
 
 	private final static Logger logger = LoggerFactory.getLogger(RabbitBindingCleaner.class);
 
-	public static final String BINDER_PREFIX = "binder.rabbit.";
+	public static final String BINDER_PREFIX = "binder.";
 
 	@Override
 	public Map<String, List<String>> clean(String entity, boolean isJob) {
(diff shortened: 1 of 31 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -645,12 +668,13 @@ public void bindPubSubProducer(String name, MessageChannel moduleOutputChannel,
 			Properties properties) {
 		validateProducerProperties(name, properties, SUPPORTED_PUBSUB_PRODUCER_PROPERTIES);
 		RabbitPropertiesAccessor accessor = new RabbitPropertiesAccessor(properties);
-		String exchangeName = applyPrefix(accessor.getPrefix(this.defaultPrefix), applyPubSub(name));
+		String exchangeName = applyPrefix(accessor.getPrefix(this.defaultPrefix), name);
 		declareExchangeIfNotPresent(new FanoutExchange(exchangeName));
```
TabberGit: Wouldn't you want this to be a TopicExchange as well, instead of a FanoutExchange?

garyrussell: Yes, of course, good catch.

garyrussell: Pushed.

markfisher:
What would be the impact if we replace rather than deprecate the existing `bindPubSubConsumer(..)` method?

garyrussell:
For some reason, I thought the Binder API was being used directly in s-c-df so I was trying to avoid issues there; but now I see it's not so I agree, it can be removed.

garyrussell:
Rebased; pushed.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitBindingCleaner.java
```diff
@@ -192,9 +179,19 @@ private void checkNoConsumers(String queueName, Map<String, Object> queue) {
 		return removedExchanges;
 	}
 
-	private interface ExchangeCandidateCallback {
-
-		boolean isCandidate(String exchangeName);
+	public boolean hasNoForeignBindings(List<Map<String, Object>> bindings, String exchangeNamePrefix) {
```
markfisher: does this need to be public?

garyrussell: No; should be private.

spring-cloud-stream/src/test/resources/org/springframework/cloud/stream/binder/sink-binding-pubsub-test.properties
```diff
@@ -0,0 +1,2 @@
+spring.cloud.stream.bindings.input=topic:testpubsub
+spring.cloud.stream.group=tgroup
```
markfisher: should the group property be at the per-binding level?

mbogoevici: I believe the question here is whether the same component can be part of two separate consumer groups, with respect to its different inputs. I'd be tempted to answer 'no' ...

markfisher: Well... consider a very general wiretap case. Perhaps a component is calculating some simple metrics on traffic across multiple streams, hence multiple "groups". This would be a macro version of what some users accomplish with global channel-interceptors in spring-integration.

garyrussell:
Pushed

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/a375e6126009b111a3992ae47489b5630f4e3942

spencergibb:
@garyrussell @markfisher does this get automatically pushed to snapshots?

garyrussell:
Yes, but only if the build passes :frowning: 

https://build.spring.io/chain/result/viewChainResult.action?planKey=SCS-SMJLE&buildNumber=212

spencergibb:
@garyrussell thanks, I went to see if it fixed my issue and it didn't, so that is why :-)

garyrussell:
@mbogoevici any ideas...

````
java.lang.RuntimeException: org.springframework.beans.factory.BeanInitializationException: Property 'cleanup.policy' must be set to 'compact' on offset topic
	at org.springframework.integration.kafka.listener.KafkaTopicOffsetManager.validateOffsetTopic(KafkaTopicOffsetManager.java:306)
````

mbogoevici:
Yeah,

Not sure how the test environment got broken like this. Need to check the PR.

mbogoevici:
OK, this is fixed now in master. The artifacts have bee successfully deployed as per https://build.spring.io/browse/SCS-SMJLE-214/

PR #197, opened 2015-11-09
Use the SI EvaluationContext by default for Expression properties

ericbottard:


markfisher:
I believe rather than using the `BeanFactory` for lookup, the "integrationEvaluationContext" should be injectable via `@Autowired`. cc @artembilan @garyrussell

ericbottard:
It's always a decision between the lookup approach (urghh) and relying on a name (which could change)

But I'm definitely eager to know what we consider to be the best approach

markfisher:
I don't think that bean name will change, but perhaps a compromise is to use `IntegrationContextUtils.INTEGRATION_EVALUATION_CONTEXT_BEAN_NAME` within the `@Qualifier`. That cannot change for backwards compatibility reasons.

ericbottard:
Changed, added test

markfisher:
LGTM; merging

markfisher:
merged as https://github.com/spring-cloud/spring-cloud-stream/commit/80b1d28be1c8b9a23099b145fe2dcf472bfa9697 (added a newline at the end of the test class)

PR #199, opened 2015-11-12
Add Generic Map converter

ericbottard:


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpelExpressionConverterConfiguration.java
```diff
@@ -70,4 +70,4 @@ public Expression convert(String source) {
 			}
 		}
 	}
-}
+}
```
markfisher: should keep this newline

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/config/GenericMapConverterTests.java
```diff
@@ -0,0 +1,84 @@
...
+
+		TypeDescriptor targetType = TypeDescriptor.map(Map.class, TypeDescriptor.valueOf(String.class), TypeDescriptor.valueOf(String[].class));
+		@SuppressWarnings("unchecked")
+		Map<String, String[]> result = (Map<String, String[]>) conversionService.convert("foo = bar\\,quizz", targetType);
+
+		assertThat(result.values().iterator().next(), Matchers.arrayContaining("bar", "quizz"));
+
+	}
+
+}
```
markfisher: add newline

artembilan:
That's an interesting exercise, but I wonder if we should consider an existing syntax:
http://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#expressions-inline-maps
http://www.groovy-lang.org/groovy-dev-kit.html#Collections-Maps.
And just delegate to the SpEL to parse the `Map` for us!

markfisher:
I think I agree with Artem here. Since SpEL supports Map syntax, we should rely on that.

ericbottard:
Note that this is not tied to SpEL (even though I used it for `Map<String, Expression>`).

This could be applied to `Map<Foo, Bar>` as long as `Foo` and `Bar` are conversion targets.

But you make a point for the `pmml-processor` indeed

artembilan:
@ericbottard , @sabbyanandan , I'm assignee to this PR.
What should I do? 
Just review, comment and merge?
Thanks.

ericbottard:
I guess so :-)

markfisher:
If this is not needed for the PMML processor at this time, can we close it and revisit later, possibly at a lower level than `spring-cloud-stream`?

ericbottard:
It *is* needed for the pmml module

markfisher:
Then I misunderstood your comment: "But you make a point for the pmml-processor indeed"

What does that mean?

ericbottard:
Ah, my mistake. My old memory forgot about that. In its current form, _this_ this needed for the pmml processor. But, as pointed above, we could refactor to use literal map syntax there. Will attempt that.
Closing for now then

artembilan:
We can move this to the `pmml` module.
Or even just do the manual `to map conversion` exactly there and don't impact all other stuff.
I'm really a bit nervous to make this public unconditional feature...

ericbottard:
Note that it was opt-in.
But I'll try with literal maps, and if this doesn't work, move the GenericMapConverter to the pmml-processor

artembilan:
Note how  `@ConfigurationPropertiesBinding` `converters` are applied in the `ConfigurationPropertiesBindingPostProcessor`.
So, they don't impact the global `conversionService`.

ericbottard:
They do, if the conversion service is named with a special name

PR #201, opened 2015-11-13
Add support for arrays in tuple conversion and general tuple tests

markpollack:


spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/JsonStringToTupleConverter.java
```diff
@@ -62,7 +63,7 @@ else if (node.isNumber()) {
 					builder.addEntry(name, node.numberValue());
 				}
 				else {
-					builder.addEntry(name, node.asText());
```
ilayaperumalg: Can we add the same approach to `node.isObject()` as well? I think eventually the logic will get invoked as we call `convert(node.toString)` for `node.isObject` check.

ilayaperumalg:
LGTM, rebased and merged as 3241e46.

PR #202, opened 2015-11-18
Introduce synthetic configuration properties via @ImplicitExpression

### LINKED ISSUE #177: Bind message properties to modules
As a s-c-s developer, I'd like to support XD-like features where modules bind to incoming messages via expressions or other mechanism, so I can bind message properties to every microservice modules.

ericbottard:
This fixes spring-cloud/spring-cloud-stream#177

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.springframework.cloud</groupId>
 	<artifactId>spring-cloud-stream-parent</artifactId>
@@ -9,7 +10,7 @@
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
 		<version>1.1.0.BUILD-SNAPSHOT</version>
-		<relativePath />
+		<relativePath/>
 	</parent>
 	<scm>
 		<url>https://github.com/spring-cloud/spring-cloud-stream</url>
@@ -33,6 +34,7 @@
 		<module>spring-cloud-stream-test-support</module>
 		<module>spring-cloud-stream-test-support-internal</module>
 		<module>spring-cloud-stream-configuration-metadata</module>
+		<module>spring-cloud-stream-implicit-expression-processor</module>
 		<module>docs</module>
 	</modules>
 	<dependencyManagement>
@@ -44,11 +46,11 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
-		    <dependency>
-			  <groupId>org.springframework.cloud</groupId>
-			  <artifactId>spring-cloud-stream</artifactId>
-			  <version>1.0.0.BUILD-SNAPSHOT</version>
-		    </dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream</artifactId>
+				<version>1.0.0.BUILD-SNAPSHOT</version>
+			</dependency>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-stream-codec</artifactId>
@@ -58,6 +60,11 @@
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-stream-common</artifactId>
 				<version>${project.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream-implicit-expression-processor</artifactId>
+				<version>1.0.0.BUILD-SNAPSHOT</version>
 			</dependency>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
@@ -142,7 +149,7 @@
 			<dependency>
 				<groupId>org.objenesis</groupId>
 				<artifactId>objenesis</artifactId>
-                <version>2.1</version>
+				<version>2.1</version>
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
--- a/spring-cloud-stream-implicit-expression-processor/src/main/java/org/springframework/cloud/stream/annotationprocessor/ImplicitExpression.java
+++ b/spring-cloud-stream-implicit-expression-processor/src/main/java/org/springframework/cloud/stream/annotationprocessor/ImplicitExpression.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.annotationprocessor;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Mark a field in an {@literal @ConfigurationProperties} annotated class, so that an alternative synthetic property
+ * be taken into account to populate the value of the annotated field. This is typically used as a way to allow
+ * literal String values to be used for configuration properties that expect a SpEL expression.
+ *
+ * <p>
+ *     Here is an example. Assume you want a module to have a 'name' property, but also allow that named to be
+ *     dervied from a SpEL expression (resolved against the incoming message). You could add fields for both
+ *     {@literal name} and {@literal nameExpression} in your {@literal @ConfigurationProperties} class, and have an
+ *     {@literal if} statement to decide which one to use, <i>etc.</i>
+ * </p>
+ * <p>
+ *     As an alternative, do the following:
+ *<pre>
+ *{@code
+ * &#47;** This is the documentation for the nameExpression property. *&#47;
+ *   &#64;ImplicitExpression("This is the doco for the name property")
+ *   private Expression nameExpression;
+ *}</pre>
+ * </p>
+ *
+ * <p>
+ *     This will document a synthetic {@literal name} property, that can be properly resolved
+ *     at runtime thanks to the ImplicitExpressionConfiguration. Then, passing <i>e.g.</i>
+ *     {@literal --name=foo} would behave as if one had passed {@literal --nameExpression='foo'}.
+ * </p>
+ *
+ * @author Eric Bottard
+ * @author Florent Biville
+ */
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.FIELD)
+@Documented
+public @interface ImplicitExpression {
+
+	/**
+	 * The type of the alternate property. Influences how literal values for SpEL expressions are
+	 * created.
+	 */
+	Class<?> type() default String.class;
+
+	/**
+	 * A simple one-line documentation for the alternate property.
+	 */
+	String value();
+
+}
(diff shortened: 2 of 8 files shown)


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/config/ImplicitExpressionConfigurationTests.java
```diff
@@ -0,0 +1,129 @@
...
+			System.out.println("Setting1 " + otherExpression);
+			this.otherExpression = otherExpression;
+		}
+
+		public Expression getNameExpression() {
+			return nameExpression;
+		}
+
+		public void setNameExpression(Expression nameExpression) {
+			System.out.println("Setting2 " + nameExpression);
```
ericbottard: yes, left over from some fun debugging sessions..

pom.xml
```diff
@@ -203,4 +209,4 @@
 			</pluginRepositories>
 		</profile>
 	</profiles>
-</project>
+</project>
```
artembilan: New line in the end of each file.
There is an option in IDEA on the matter.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+			}
+			configurableEnvironment.getPropertySources().addLast(fakeValues);
+		}
+	}
+
+	/**
+	 * Return a literal SpEL expression with the given value, using proper literal notation according to the wanted
+	 * type.
+	 */
+	private String literalExpression(Object value, String type) {
```
artembilan: How about to take a look into the `org.springframework.integration.expression.ValueExpression` ?

ericbottard: We're working with Strings here. This is invoked in the context of a PropertySource

pom.xml
```diff
@@ -33,6 +33,7 @@
 		<module>spring-cloud-stream-test-support</module>
 		<module>spring-cloud-stream-test-support-internal</module>
 		<module>spring-cloud-stream-configuration-metadata</module>
+		<module>spring-cloud-stream-implicit-expression-processor</module>
```
mbogoevici: Does this really require a new module? Generally speaking, while it is helpful at writing modules, it is a configuration enhancement, not specific to Spring Cloud Stream proper.

ericbottard: Are you saying this should go in e.g. core? it needs to be visible at compile time (it's an annotation processor) but doesn't need to be embedded (but wouldn't hurt if it was, it's 2 classes) in the built artifact.

ericbottard:
Rebased and merged

spring-cloud-stream-implicit-expression-processor/src/main/java/org/springframework/cloud/stream/annotationprocessor/ImplicitExpression.java
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * The type of the alternate property. Influences how literal values for SpEL expressions are
+	 * created.
+	 */
+	Class<?> type() default String.class;
+
+	/**
+	 * A simple one-line documentation for the alternate property.
+	 */
+	String value();
```
markfisher: if `value()` always returns `String` what is the reason for providing the `type` attribute?

spring-cloud-stream/pom.xml
```diff
@@ -31,8 +31,23 @@
 			<artifactId>spring-messaging</artifactId>
 		</dependency>
 		<dependency>
+<<<<<<< 4e5277c0d0d46c2b9644b68a8e6188a12df5d78a
```
markfisher: looks like you've included a file with unresolved conflicts

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+		Class<?> clazz = ClassUtils.resolveClassName(type, Thread.currentThread().getContextClassLoader());
+		if (CharSequence.class.isAssignableFrom(clazz)) {
+			return String.format("'%s'", value);
+		}
+		else if (Number.class.isAssignableFrom(clazz)) {
```
markfisher: the result of this condition is the same as the catch-all `else`, so it can be removed, right?

markfisher:
Ultimately, any time a literal value is provided, we should end up with a `LiteralExpression` instance. Does that happen within the SpEL parser for any value that is completely enclosed within single quotes?

spring-cloud-stream-implicit-expression-processor/src/main/java/org/springframework/cloud/stream/annotationprocessor/ImplicitExpression.java
ericbottard: `value()` is the documentation for the synthetic property. `type` tells the framework the type of the property, which influences how SpEL literals should be crafted (e.g. a String needs to be enclosed in quotes, a number does not)

spring-cloud-stream/pom.xml
ericbottard: oops. Will fix

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
ericbottard: yes. Wanted to make it obvious as a special supported case though. I'm fine with it being removed (or default throwing an exception for now, not sure which is best)

ericbottard:
> Ultimately, any time a literal value is provided, we should end up with a LiteralExpression instance. Does that happen within the SpEL parser for any value that is completely enclosed within single quotes?

Currently doesn't, but ends up with a `SpelExpression` whose AST is a trivial single node. Could microbenchmark if this is what worries you

markfisher:
Given that many cases where this feature will be used the `Expression` would be evaluated in the critical path (for each message received), I do think it's worth benchmarking. It seems like a change we could push down to the parser though, where it would be an even more generalized optimization.

ericbottard:
Addressed review comments. Will have a look at performance

ericbottard:
For the record (note that I first got my hands on jmh 20mins ago, so no expert here...)
```

public class MyBenchmark {

    @State(Scope.Thread)
    public static class MyState {

        public Expression literal;

        public Expression parsed;

        @Setup(Level.Trial)
        public void doSetup() {
            literal = new LiteralExpression("foobar");
            parsed = new SpelExpressionParser().parseExpression("'foobar'");
        }
    }

    @Benchmark @BenchmarkMode(Mode.Throughput) @OutputTimeUnit(TimeUnit.MILLISECONDS)
    public void testLiteral(MyState state, Blackhole blackhole) {
        blackhole.consume(state.literal.getValue());

    }

    @Benchmark @BenchmarkMode(Mode.Throughput) @OutputTimeUnit(TimeUnit.MILLISECONDS)
    public void testParsed(MyState state, Blackhole blackhole) {
        blackhole.consume(state.parsed.getValue());

    }

}
```

gave
```


# Run complete. Total time: 00:01:20

Benchmark                 Mode  Cnt       Score      Error   Units
MyBenchmark.testLiteral  thrpt   20  394157.159 ± 2027.223  ops/ms
MyBenchmark.testParsed   thrpt   20  119627.862 ± 2086.432  ops/ms
```

ericbottard:
Same thing with `-jvmArgs -Dspring.expression.compiler.mode=MIXED`

```
Benchmark                 Mode  Cnt       Score      Error   Units
MyBenchmark.testLiteral  thrpt   20  393834.956 ± 3367.027  ops/ms
MyBenchmark.testParsed   thrpt   20  211024.409 ±  799.405  ops/ms
```

markfisher:
If I understand those results, it does appear worthwhile to aim for `LiteralExpression` creation in the case where a user provides the literal option (not the `-expression` property).

ericbottard:
Yes, 2x différence. But this is not handled at this PR level . rather
parser itself.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+		allProperties = builder.build().getAllProperties();
+	}
+
+	@Override
+	public void setEnvironment(final Environment environment) {
+		if (environment instanceof ConfigurableEnvironment) {
+			ConfigurableEnvironment configurableEnvironment = (ConfigurableEnvironment) environment;
+
+			Map<String, Object> map = new HashMap<>();
+			MapPropertySource fakeValues = new MapPropertySource("Implicit expressions", map);
```
markfisher: what is the intended meaning of "fakeValues"?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
markfisher: what do we expect other than `String` or `Number`?

markfisher:
>  this is not handled at this PR level . rather parser itself.

Until it's addressed in the parser we could do a regex comparison (e.g. `Pattern.compile("'[^']*'")`) in our `SpelConverter` and return `LiteralExpression` instead of parsing for those cases.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
ericbottard: They're fake in the sense that the user did not actually set them. ie when one sets `foo`, this will register `foo-expression`, even though the user did not specify it explicitly.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
ericbottard: That's the question really. I think we'll see in the wild, but thinking about the modules we've already authored, I don't many more. Maybe arrays of the above.

ericbottard:
Yep, was writing on my phone so did not want to write a novel, but the `SpelConverter` could be another thing we could leverage. Actually changing the `SpelExpressionParser` should not be that difficult, but I need to ping @aclement about the case of templates in Strings.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ImplicitExpressionConfiguration.java
markfisher: > the user did not specify it explicitly

...which is why the word "implicit" is used everywhere else. Why add another term ("fake")?

ericbottard: will change

markfisher: alternatively, could just inline the creation:

```
configurableEnvironment.getPropertySources().addLast(new MapPropertySource("Implicit expressions", map));
```

markfisher:
Unfortunately, the more I contemplate this feature, the less I want it to be in the `spring-cloud-stream` project. In general, I'm wary of any generic functionality (not specific to binders, etc) that adds to the otherwise "pure boot" developer experience. We don't want to blur that line. That said, if @philwebb and @dsyer see value in this feature for boot itself, we should discuss that.

cc @sabbyanandan @markpollack

spring-cloud-stream-implicit-expression-processor/src/main/java/org/springframework/cloud/stream/annotationprocessor/ImplicitExpression.java
```diff
@@ -0,0 +1,71 @@
...
+ *     {@literal --name=foo} would behave as if one had passed {@literal --nameExpression='foo'}.
+ * </p>
+ *
+ * @author Eric Bottard
+ * @author Florent Biville
+ */
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.FIELD)
+@Documented
+public @interface ImplicitExpression {
```
markfisher: This name doesn't quite capture the role. The annotated field of type `Expression` *can* still be configured explicitly. The actual feature is that it enables a shortcut option for the case where you want to provide a literal value for a field that is actually an `Expression` instance. Somehow the name should match that role. This is not a great option either, but to provide an example: `@ExposeLiteralValueProperty`

philwebb:
I'm generally against the idea of mixing in SpEL with `@ConfigurationProperties` but feel free to raise an issue if you feel it needs more discussion.

markfisher:
Just to clarify, this is not really about "mixing" SpEL with `@ConfigurationProperties`. It is for handling properties that happen to be `Expression` instances. For example, an `Expression` may be used by a Message handler to evaluate each Message at runtime. If that handler is writing to a file, it might evaluate SpEL against a message to generate the filename. No expression evaluation is happening within `@ConfigurationProperties` itself in that case. The feature that is addressed by this PR is simply to allow a literal value for the filename (when it does not need to be dynamically based on message content) without forcing the user to still deal with the baggage of expression syntax.

markpollack:
Closing and will have further discussion to revisit with Boot team.

PR #207, opened 2015-11-20
Fix deprecated use of boot's showBanner

ilayaperumalg:
- Use setBannerMode instead


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateBuilder.java
```diff
@@ -1,4 +1,5 @@
 /*
+/*
```
markfisher: extra comment start

markfisher: will fix on merge

markfisher:
LGTM; merged with one minor fix as https://github.com/spring-cloud/spring-cloud-stream/commit/73ba7b059dedb779bbfebfb01f28f3811880303c

PR #215, opened 2015-11-30
GH-211: PubSub Group Default to Random UUID

### LINKED ISSUE #211: Why does a consumer have to specify a group name in order to get pub-sub?
I would have thought it would be the default if the destination is a "topic" (I think it used to be). Can anyone explain why this extra step is necessary?

garyrussell:
Resolves #211

Add a default UUID to the `BindingProperties.group` so a pubsub binding is
unique unless the group is specifically set.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
@@ -32,6 +32,7 @@ import org.springframework.util.StringUtils;
  * @author Dave Syer
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
+ * @author Gary Russell
  */
 public class ChannelBindingService {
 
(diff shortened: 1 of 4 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -78,10 +76,22 @@ private String removePrefix(String bindingTarget) {
...
+			this.binder.unbindConsumers(inputChannelName);
+		}
 	}
 
 	public void unbindProducers(String outputChannelName) {
 		this.binder.unbindProducers(outputChannelName);
 	}
+
+	private String consumerGroup(String inputChannelName) {
+		BindingProperties bindingProperties = this.channelBindingServiceProperties.getBindings()
```
mbogoevici: This will work for now, since it is currently impossible to specify a pubsub scenario without using binding properties. When we remove the distinction between queue/topic, this will have to be updated, since there may be cases when no binding properties are present.

mbogoevici:
LGTM; will merge

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/6f1f246dfba422d728f86644aab4be6a97c5ad73

Issue #213, opened 2015-11-30
Merge `spring-cloud-spring-binder-spi` with `spring-cloud-stream`


mbogoevici:
There is very little reason for the separation, seeing as `spring-cloud-stream` cannot really function without the functionality provided by the SPI. The division currently prevents the generic binder functionality to have access to SCS configuration easily, as well as creates the need for intermediate classes such as `ChannelBindingService`. 

The main goal of this exercise would be to simplify the project structure of Spring Cloud Stream, as well as its internal design.

PR #216, opened 2015-11-30
GH-213 Merge Binder SPI into Core

mbogoevici:
- Moved contents of `spring-cloud-stream-binder-spi` into `spring-cloud-stream`
- Removed `spring-cloud-stream-binder-local` and used local binder implementation for tests exclusively

This resolves spring-cloud/spring-cloud-stream#213


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -54,6 +54,7 @@
  * @author Gary Russell
  */
 
+@Ignore
```
markfisher: is this addition intentional?

mbogoevici: Yes, the Kafka tests have been failing for a while after the Rabbit PubSub changes - need to address that separately.

mbogoevici: The failures have been sporadic, though (and unrelated to the PR).

spring-cloud-stream/pom.xml
```diff
@@ -36,29 +36,30 @@
...
 		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-test</artifactId>
-			<scope>test</scope>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-databind</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.httpcomponents</groupId>
+			<artifactId>httpclient</artifactId>
```
markfisher: why is this dependency needed in the core now?

mbogoevici: It was in the SPI. We can remove it, I believe - it's an offshoot of the RabbitAdmin being present in the Binder SPI at some point.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
markfisher: okay, thanks

spring-cloud-stream/pom.xml
markfisher: I will try to remove it. If the build passes I will merge without it.

mbogoevici: Ok, cool - thanks.

markfisher: The build passes with it removed. Actually, the same goes for the jackson dependency. So, I am removing both on merge.

Issue #196, opened 2015-11-06
Port XD History Tracking


garyrussell:
History tracking binder/deployment option.

See XD-2606 and XD-3693.

garyrussell:
closed via #218

PR #218, opened 2015-12-03
GH-196 Add message history tracking

ilayaperumalg:
- Add `MessageChannelConfigurer` that any class that configures the message channel would implement
   - Have `message-converter` and `message-history-tracking` configurer implement this interface
  - Autowire all available configurers and use them to configure the message channel
  - For the message history tracking
    - add `channel-binding` properties as part of message header along with `timestamp`

This resolves spring-cloud/spring-cloud-stream#196

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.config;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.test.SpringApplicationConfiguration;
+import org.springframework.cloud.stream.annotation.Bindings;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.binder.redis.config.RedisMessageChannelBinderConfiguration;
+import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.cloud.stream.messaging.Source;
+import org.springframework.cloud.stream.test.junit.redis.RedisTestSupport;
+import org.springframework.cloud.stream.tuple.Tuple;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.MessagingException;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.springframework.util.Assert;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration({MessageChannelConfigurerTests.TestSource.class,
+		MessageChannelConfigurerTests.TestSink.class})
+public class MessageChannelConfigurerTests {
+
+	@Rule
+	public RedisTestSupport redisTestSupport = new RedisTestSupport();
+
+	@Autowired @Bindings(TestSink.class)
+	private Sink testSink;
+
+	@Autowired
+	@Bindings(TestSource.class)
+	private Source testSource;
+
+	@Test
+	public void testContentTypeConfigurer() throws Exception {
+		final CountDownLatch latch = new CountDownLatch(1);
+		testSink.input().subscribe(new MessageHandler() {
+			@Override
+			public void handleMessage(Message<?> message) throws MessagingException {
+				Assert.isTrue(message.getPayload() instanceof Tuple);
+				Assert.isTrue(((Tuple)message.getPayload()).getFieldNames().get(0).equals("message"));
+				Assert.isTrue(((Tuple) message.getPayload()).getValue(0).equals("Hi"));
+				latch.countDown();
+			}
+		});
+		testSource.output().send(MessageBuilder.withPayload("{\"message\":\"Hi\"}").build());
+		latch.await(10, TimeUnit.SECONDS);
+	}
+
+	@Test
+	public void testHistoryTrackerConfigurer() throws Exception {
+		final CountDownLatch latch = new CountDownLatch(1);
+		testSink.input().subscribe(new MessageHandler() {
+			@Override
+			public void handleMessage(Message<?> message) throws MessagingException {
+				Assert.isTrue(message.getHeaders().containsKey("SPRING_CLOUD_STREAM_HISTORY"),
+						"Message header should have tracking history info");
+				BindingProperties inputBindingProps = (BindingProperties) ((Map)((List) message.getHeaders()
+						.get("SPRING_CLOUD_STREAM_HISTORY")).get(0)).get("input");
+				Assert.isTrue(inputBindingProps.getDestination().equals("configure"));
+				Assert.isTrue(inputBindingProps.getTrackHistory());
+				BindingProperties outputBindingProps = (BindingProperties) ((Map)((List) message.getHeaders()
+						.get("SPRING_CLOUD_STREAM_HISTORY")).get(0)).get("output");;
+				Assert.isTrue(outputBindingProps.getDestination().equals("configure"));
+				Assert.isTrue(!outputBindingProps.getTrackHistory());
+				latch.countDown();
+			}
+		});
+		testSource.output().send(MessageBuilder.withPayload("{\"test\":\"value\"}").build());
+		latch.await(10, TimeUnit.SECONDS);
+	}
+
+	@EnableBinding(Source.class)
+	@EnableAutoConfiguration
+	@Import(RedisMessageChannelBinderConfiguration.class)
+	@PropertySource("classpath:/org/springframework/cloud/stream/config/source-channel-configurers.properties")
+	public static class TestSource {
+
+	}
+
+	@EnableBinding(Sink.class)
+	@EnableAutoConfiguration
+	@Import(RedisMessageChannelBinderConfiguration.class)
+	@PropertySource("classpath:/org/springframework/cloud/stream/config/sink-channel-configurers.properties")
+	public static class TestSink {
+
+	}
+}
+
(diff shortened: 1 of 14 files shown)


ilayaperumalg:
This resolves #196

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -322,6 +323,12 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
 		this.beanFactory = (ConfigurableListableBeanFactory) beanFactory;
 	}
 
+	private void configureMessageChannels(MessageChannel messageChannel, String channelName) {
```
markfisher: this method name should not be plural since it's configuring a single channel

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -322,6 +323,12 @@ public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
...
+		for (MessageChannelConfigurer messageChannelConfigurer: messageChannelConfigurers) {
```
markfisher: technically, this should have a null-check for the list, but a bigger issue is that we really should not be relying on field-injection in this class at all

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageHistoryTrackerConfigurer.java
```diff
@@ -0,0 +1,83 @@
...
+
+	public MessageHistoryTrackerConfigurer(ChannelBindingServiceProperties channelBindingServiceProperties) {
+		this.channelBindingServiceProperties = channelBindingServiceProperties;
+	}
+
+	@Override
+	public void configureMessageChannel(MessageChannel messageChannel, String channelName) {
+		// todo: add history properties
+		BindingProperties bindingProperties = channelBindingServiceProperties.getBindings().get(channelName);
+		if (bindingProperties != null && bindingProperties.getTrackHistory() != null &&
```
markfisher: if you invert you won't need the 2nd null check, e.g.:

```
if (Boolean.TRUE.equals(bindingProperties.getTrackHistory())) {...
```

markfisher:
We may want to create a single `CompositeMessageChannelConfigurer` as the one that gets injected. That addresses 2 concerns I have with this PR: 1) no control of the order of configurers which determines the order of any `ChannelInterceptor`s being added as well and 2) this list of configurers is used in more than one place (`ChannelFactory` and `BindableProxyFactory`).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageHistoryTrackerConfigurer.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.support.ChannelInterceptorAdapter;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+public class MessageHistoryTrackerConfigurer implements MessageChannelConfigurer {
+
+	public static final String HISTORY_TRACKING_HEADER = "DATAFLOW_HISTORY";
```
garyrussell: This isn't really limited to dataflow alone; any s-c-s user might want to (and can) enable history. I don't have a good alternative (yet).

garyrussell:
IN XD, we track output channels for source, processor; input channel for sink.

I guess that will now be the responsibility of s-c-df - is there a story over there? I don't see one.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageHistoryTrackerConfigurer.java
markfisher: Good point. We definitely do not want `"DATAFLOW"` as a value from a `spring-cloud-stream` component like this.

ilayaperumalg: Yeah, I agree. might be `spring-cloud-stream` as a header name? I have just created this one in dataflow: https://github.com/spring-cloud/spring-cloud-dataflow/issues/209

ilayaperumalg:
Addressed the review comments by introducing `CompositeMessageChannelConfigurer`

markfisher:
this PR should include tests for the configurers

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -80,7 +80,7 @@
 	private ChannelFactory channelFactory;
 
 	@Autowired
```
ilayaperumalg: Once https://github.com/spring-cloud/spring-cloud-stream/pull/219 is merged, the changes in this class can go away.

ilayaperumalg:
Rebased and updated the PR against `ChannelFactory` changes.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
```diff
@@ -0,0 +1,119 @@
...
+			@Override
+			public void handleMessage(Message<?> message) throws MessagingException {
+				Assert.isTrue(message.getPayload() instanceof Tuple);
+				Assert.isTrue(((Tuple)message.getPayload()).getFieldNames().get(0).equals("message"));
+				Assert.isTrue(((Tuple) message.getPayload()).getValue(0).equals("Hi"));
+				latch.countDown();
+			}
+		});
+		testSource.output().send(MessageBuilder.withPayload("{\"message\":\"Hi\"}").build());
+		latch.await(10, TimeUnit.SECONDS);
```
garyrussell: Needs `assertTrue(...)`.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
```diff
@@ -0,0 +1,119 @@
...
+				Assert.isTrue(inputBindingProps.getTrackHistory());
+				BindingProperties outputBindingProps = (BindingProperties) ((Map)((List) message.getHeaders()
+						.get("SPRING_CLOUD_STREAM_HISTORY")).get(0)).get("output");;
+				Assert.isTrue(outputBindingProps.getDestination().equals("configure"));
+				Assert.isTrue(!outputBindingProps.getTrackHistory());
+				latch.countDown();
+			}
+		});
+		testSource.output().send(MessageBuilder.withPayload("{\"test\":\"value\"}").build());
+		latch.await(10, TimeUnit.SECONDS);
```
garyrussell: ditto - I'll fix these on merge if I don't find other issues

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
```diff
@@ -0,0 +1,119 @@
...
+				Assert.isTrue(message.getHeaders().containsKey("SPRING_CLOUD_STREAM_HISTORY"),
+						"Message header should have tracking history info");
+				BindingProperties inputBindingProps = (BindingProperties) ((Map)((List) message.getHeaders()
+						.get("SPRING_CLOUD_STREAM_HISTORY")).get(0)).get("input");
+				Assert.isTrue(inputBindingProps.getDestination().equals("configure"));
+				Assert.isTrue(inputBindingProps.getTrackHistory());
+				BindingProperties outputBindingProps = (BindingProperties) ((Map)((List) message.getHeaders()
+						.get("SPRING_CLOUD_STREAM_HISTORY")).get(0)).get("output");;
+				Assert.isTrue(outputBindingProps.getDestination().equals("configure"));
+				Assert.isTrue(!outputBindingProps.getTrackHistory());
```
garyrussell: Is there any particular reason you are using Spring's `Assert` rather than JUnit?

ilayaperumalg: There is no specific reason on that.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.java
```diff
@@ -74,13 +79,29 @@ public ChannelBindingService bindingService(
 	}
 
 	@Bean
-	public MessageConverterConfigurer messageConverterConfigurer(ChannelBindingServiceProperties channelBindingServiceProperties) {
+	MessageConverterConfigurer messageConverterConfigurer
```
garyrussell: Why no longer `public` ?

ilayaperumalg: I thought I fixed this. missed to make it public :-)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -200,4 +200,24 @@ public String getBinder(String channelName) {
...
+		}
+		for (String name : getProducerProperties().stringPropertyNames()) {
+			properties.put(name, getProducerProperties());
+		}
+		for (Map.Entry<String, BindingProperties> entry : getBindings().entrySet()) {
+			properties.put(entry.getKey(), entry.getValue());
+		}
+		return properties;
+	}
+
```
garyrussell: I might have missed part of the conversation, but is this really what we want to track? 
Do we really want to track all the configuration/binding properties?
In XD we had other useful stuff like ip address, pid etc.
Can we capture the boot application/module name somehow?
I am concerned about name collision for a processor module - where we're adding both consumer and producer properties.
Finally, I think we need `toString()` on `BindingProperties` (if we include them) - the history is pretty useless in logs:

    SPRING_CLOUD_STREAM_HISTORY=[{input=org.springframework.cloud.stream.config.BindingProperties@19af3d5b, 
    output=org.springframework.cloud.stream.config.BindingProperties@3fe52e89, instanceIndex=0, instanceCount=1}]}]

garyrussell: At some point (perhaps in a future milestone) we should consider some mechanism where the user (perhaps via config server) can configure exactly what is captured in the history entries.

ilayaperumalg: Good point. Created: https://github.com/spring-cloud/spring-cloud-stream/issues/252

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
garyrussell: When I run these tests in STS, this message goes to the other test....

       : Channel 'bootstrap:-1.input' has 2 subscriber(s).

garyrussell: Need to unbind the handler after each test.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
ilayaperumalg: Since we only use the producer/consumer property **names**, the possibility of collision is nil (as we don't have any overlapping names). To distinguish these properties, I explicitly added `producer.` and `consumer.` prefix added to the `toString` value. I guess this is ok given we only update the message header. I also updated the test case to reflect this case.

garyrussell:
Looks good - I'm still not sure we want to capture all the properties, but I guess we can revisit that with #252

garyrussell:
Test failed after I rebased; investigating.

garyrussell:
Seems to be a timing problem - Dispatcher has no subscribers for an input channel someplace.

garyrussell:
Rebased, squashed, merge with minor test polish.

PR #219, opened 2015-12-04
Add createSharedChannel method in ChannelFactory

ilayaperumalg:
- Remove configuring shared message channel

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -28,7 +28,6 @@ import org.springframework.cloud.stream.messaging.Source;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.integration.channel.DirectChannel;
-import org.springframework.messaging.MessageChannel;
 
 /**
  * Class that is responsible for embedding modules using shared channel registry.
@@ -90,7 +89,7 @@ public class AggregateApplication {
 				.web(false)
 				.headless(true)
 				.properties("spring.jmx.default-domain="
-								+ AggregatorParentConfiguration.class.getName());
+						+ AggregatorParentConfiguration.class.getName());
 		return aggregatorParentConfiguration.run(args);
 	}
 
@@ -120,7 +119,7 @@ public class AggregateApplication {
 
 	private static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] modules,
 			ChannelFactory channelFactory) {
-		MessageChannel sharedChannel = null;
+		DirectChannel sharedChannel = null;
 		for (int i = 0; i < modules.length; i++) {
 			Class<?> module = modules[i];
 			String moduleClassName = module.getName();
@@ -131,8 +130,8 @@ public class AggregateApplication {
 			try {
 				sharedChannel = channelFactory.createSharedChannel(DirectChannel.class);
 			}
-			catch(Exception e) {
-
+			catch (Exception e) {
+				throw new RuntimeException("Exception while creating shared channel " + e);
 			}
 			if (i < modules.length - 1) {
 				sharedChannelRegistry.register(getNamespace(moduleClassName, i)
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -125,12 +125,7 @@ public class AggregateApplication {
 				sharedChannelRegistry.register(getNamespace(moduleClassName, i)
 						+ "." + INPUT_CHANNEL_NAME, sharedChannel);
 			}
-			try {
-				sharedChannel = new DirectChannel();
-			}
-			catch (Exception e) {
-				throw new RuntimeException("Exception while creating shared channel", e);
-			}
+			sharedChannel = new DirectChannel();
 			if (i < modules.length - 1) {
 				sharedChannelRegistry.register(getNamespace(moduleClassName, i)
 						+ "." + OUTPUT_CHANNEL_NAME, sharedChannel);
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -27,7 +27,7 @@ import org.springframework.cloud.stream.messaging.Sink;
 import org.springframework.cloud.stream.messaging.Source;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
-import org.springframework.integration.channel.DirectChannel;
+import org.springframework.messaging.SubscribableChannel;
 
 /**
  * Class that is responsible for embedding modules using shared channel registry.
@@ -119,7 +119,7 @@ public class AggregateApplication {
 
 	private static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] modules,
 			ChannelFactory channelFactory) {
-		DirectChannel sharedChannel = null;
+		SubscribableChannel sharedChannel = null;
 		for (int i = 0; i < modules.length; i++) {
 			Class<?> module = modules[i];
 			String moduleClassName = module.getName();
@@ -128,7 +128,7 @@ public class AggregateApplication {
 						+ "." + INPUT_CHANNEL_NAME, sharedChannel);
 			}
 			try {
-				sharedChannel = channelFactory.createSharedChannel(DirectChannel.class);
+				sharedChannel = channelFactory.createSubscribableSharedChannel();
 			}
 			catch (Exception e) {
 				throw new RuntimeException("Exception while creating shared channel", e);
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -21,6 +21,7 @@ import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.binding.ChannelFactory;
 import org.springframework.cloud.stream.messaging.Processor;
 import org.springframework.cloud.stream.messaging.Sink;
 import org.springframework.cloud.stream.messaging.Source;
@@ -29,7 +30,10 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.integration.channel.DirectChannel;
 
 /**
+ * Class that is responsible for embedding modules using shared channel registry.
+ *
  * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
  */
 public class AggregateApplication {
 
@@ -72,7 +76,8 @@ public class AggregateApplication {
 	public static void runEmbedded(ConfigurableApplicationContext parentContext,
 			Class<?>[] modules, String[][] args) {
 		SharedChannelRegistry bean = parentContext.getBean(SharedChannelRegistry.class);
-		prepareSharedChannelRegistry(bean, modules);
+		ChannelFactory channelFactory = parentContext.getBean(ChannelFactory.class);
+		prepareSharedChannelRegistry(bean, modules, channelFactory);
 		// create child contexts first
 		createChildContexts(parentContext, modules, args);
 	}
@@ -112,7 +117,8 @@ public class AggregateApplication {
 				.parent(applicationContext);
 	}
 
-	private static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] modules) {
+	private static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] modules,
+			ChannelFactory channelFactory) {
 		DirectChannel sharedChannel = null;
 		for (int i = 0; i < modules.length; i++) {
 			Class<?> module = modules[i];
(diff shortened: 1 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactory.java
```diff
@@ -25,5 +25,7 @@
  */
 public interface ChannelFactory {
```
markfisher: the methods here could be parameterized, e.g.:

```
<T extends MessageChannel> T createBindableChannel(String name, Class<T> channelType) throws Exception;

<T extends MessageChannel> T createSharedChannel(Class<T> channelType) throws Exception;
```

That would make it unnecessary to cast as in https://github.com/spring-cloud/spring-cloud-stream/pull/219/files#diff-d2113fc67655244ed5f8c586d6302862R130

ilayaperumalg: While we can use `MessageChannel` as the target type instead of DirectChannel in the case above, I went ahead and parameterize the `ChannelFactory` interface.

markfisher:
To clarify, my suggestion for parameterization was for method-level, not class-level:
https://github.com/spring-cloud/spring-cloud-stream/pull/219#discussion_r46981590

ilayaperumalg:
yeah, sorry. I misunderstood what you suggested earlier. fixed it with the latest push.

ilayaperumalg:
yeah, +1 and pushed the change.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
```diff
@@ -121,7 +127,12 @@ private static void prepareSharedChannelRegistry(SharedChannelRegistry sharedCha
 				sharedChannelRegistry.register(getNamespace(moduleClassName, i)
 						+ "." + INPUT_CHANNEL_NAME, sharedChannel);
 			}
-			sharedChannel = new DirectChannel();
+			try {
+				sharedChannel = channelFactory.createSharedChannel(DirectChannel.class);
+			}
+			catch (Exception e) {
+				throw new RuntimeException("Exception while creating shared channel " + e);
```
markfisher: should include `e` as the root cause (not just the `String`)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -130,10 +127,12 @@ public void doWith(Method method) throws IllegalArgumentException {
 						MessageChannel sharedChannel = locateSharedChannel(name);
 						if (sharedChannel == null) {
 							inputHolders.put(name, new ChannelHolder(
-									channelFactory.createChannel(name, method.getReturnType()), true));
+									channelFactory.createBindableChannel(name,
+											(Class<? extends MessageChannel>)method.getReturnType()), true));
```
markfisher: space after cast parens

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -130,10 +127,12 @@ public void doWith(Method method) throws IllegalArgumentException {
...
 						}
 						else {
-							configureSharedMessageChannel(name, method.getReturnType(), sharedChannel);
+							configureSharedMessageChannel(name,
+									(Class<? extends MessageChannel>) method.getReturnType(), sharedChannel);
```
markfisher: maybe move the `method.getReturnType()` call up so it doesn't need to be duplicated

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -152,10 +151,11 @@ public void doWith(Method method) throws IllegalArgumentException {
 						MessageChannel sharedChannel = locateSharedChannel(name);
 						if (sharedChannel == null) {
 							outputHolders.put(name, new ChannelHolder(
-									channelFactory.createChannel(name, method.getReturnType()), true));
+									channelFactory.createBindableChannel(name, (Class<? extends MessageChannel>) method.getReturnType()), true));
 						}
 						else {
-							configureSharedMessageChannel(name, method.getReturnType(), sharedChannel);
+							configureSharedMessageChannel(name, (Class<? extends MessageChannel>)method.getReturnType(),
```
markfisher: same thing here... could move the `method.getReturnType()` up so it's not duplicated, but if left here add a space after the cast parens

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DefaultChannelFactory.java
```diff
@@ -34,17 +34,23 @@
...
 
-	private MessageChannel createMessageChannel(Class<?> messageChannelType) {
-		return isPollable(messageChannelType) ? new QueueChannel() : new DirectChannel();
+	@Override
+	public <T extends MessageChannel> T createSharedChannel(Class<T> channelType) throws Exception {
+		return createMessageChannel(channelType);
+	}
+
+	private <T extends MessageChannel> T createMessageChannel(Class<T> messageChannelType) {
+		return isPollable(messageChannelType) ? (T) new QueueChannel() : (T) new DirectChannel();
```
markfisher: this makes me think we should only expect `SubscribableChannel` or `PollableChannel` after all... or do we actually need separate methods for pollable and subscribable? cc @mbogoevici

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/aggregation/ModuleAggregationTest.java
```diff
@@ -42,6 +44,8 @@ public void testModuleAggregation() {
 		ConfigurableApplicationContext aggregatedApplicationContext = AggregateApplication.run(TestSource.class,
 				TestProcessor.class);
 		SharedChannelRegistry sharedChannelRegistry = aggregatedApplicationContext.getBean(SharedChannelRegistry.class);
+		ChannelFactory channelFactory= aggregatedApplicationContext.getBean(ChannelFactory.class);
```
markfisher: space before the equals sign

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DefaultChannelFactory.java
ilayaperumalg: Given the `ChannelFactory` and its implementations are only for `internal` purpose and it is going to be either `SubscribableChannel` or `PollableChannel` for binding, we can simplify to have separate methods. But l like the current parameterized approach too which can make the `ChannelFactory` more lenient.

markfisher: Agreed this makes it more lenient. What I don't like is that in this implementation we are casting to a super type of what's passed in. And, at least in the only current implementation, it's essentially a binary decision. It just feels like an unclear contract in its current form. Interested in your thoughts @mbogoevici

ilayaperumalg: After a discussion with @mbogoevici, pushed a change that simplifies the `ChannelFactory` contract by having separate methods.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -176,22 +179,19 @@ private String getNamespacePrefixedChannelName(String name) {
 		return this.channelNamespace + "." + name;
 	}
 
-	private void configureSharedMessageChannel(String name, Class<?> channelType, MessageChannel sharedChannel)
+	private void configureSharedMessageChannel(String name, Class<? extends MessageChannel> channelType, MessageChannel sharedChannel)
```
markfisher: now that no "configuration" happens here, perhaps we should consider pushing the `isAssignableFrom` check up, so that this method is only called when a bridge is necessary... which would also simplify it e.g. `bridgeSharedMessageChannel`. Also, the `name` param is no longer used.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -176,22 +179,19 @@ private String getNamespacePrefixedChannelName(String name) {
...
-		if (channelType.isAssignableFrom(sharedChannel.getClass())) {
-			messageConverterConfigurer.configureMessageConverters(sharedChannel, name);
-		}
-		else {
+		if (!channelType.isAssignableFrom(sharedChannel.getClass())) {
 			// handle the special case where the shared channel is of a different nature
 			// (i.e. pollable vs subscribable) than the target channel
-			final MessageChannel inputChannel = this.channelFactory.createChannel(name, channelType);
 			if (isPollable(sharedChannel.getClass())) {
+				final MessageChannel inputChannel = this.channelFactory.createPollableSharedChannel();
```
markfisher: I believe this should be calling `createSubscribableSharedChannel`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -176,22 +179,19 @@ private String getNamespacePrefixedChannelName(String name) {
...
 				bridgePollableToSubscribableChannel(sharedChannel, inputChannel);
 			}
 			else {
+				final MessageChannel inputChannel = this.channelFactory.createSubscribableSharedChannel();
```
markfisher: I believe this should be calling `createPollableSharedChannel`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DefaultChannelFactory.java
```diff
@@ -30,21 +29,33 @@
...
+	public SubscribableChannel createSubscribableBindableChannel(String name) throws Exception {
+		SubscribableChannel subscribableChannel = new DirectChannel();
+		messageConverterConfigurer.configureMessageConverters(subscribableChannel, name);
+		return subscribableChannel;
 	}
 
-	private boolean isPollable(Class<?> channelType) {
-		return PollableChannel.class.equals(channelType);
+	@Override
+	public SubscribableChannel createSubscribableSharedChannel() throws Exception {
```
mbogoevici: Maybe we can simplify this a little bit by elliminating the 'sharedChannel' methods and having just `createPollableChannel` and `createSubscribableChannel`. At this particular point, we don't configure them, and we can isolate bridging as a concern between the Aggregator and the BindableProxyFactory. Maybe we can rename the interface as `BindableChannelFactory` so it's clear it's used only for that. I believe we need to revisit configuration on aggregation anyway.

markfisher: Just to clarify you are suggesting that we only keep the methods that *do* apply configuration but removing "bindable" from those method names, and then completely removing the 2 methods with "shared" in the name. The responsibility for shared channel creation (which is just a `new` call) would go back to the `BindableProxyFactory` (to be renamed), and the bridging logic would remain in that class also. Is that correct?

mbogoevici: Indeed. Didn't feel like what was happening there warranted the inclusion in the `ChannelFactory` contract.  I was also suggesting renaming the interface, so that the role it plays it's clearer. In a way this *is* a generic factory, but it's only supposed to be used as part of channel binding.

markfisher: For now, I think those changes would make sense.

ilayaperumalg: ok, I have pushed the changes to reflect this.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactory.java
```diff
@@ -25,5 +26,24 @@
...
-	MessageChannel createChannel(String name, Class<?> channelType) throws Exception;
+	/**
+	 * Create a {@link SubscribableChannel} that will be bound to the message channel
+	 * {@link org.springframework.cloud.stream.binder.Binder}.
+	 *
+	 * @param name name of the message channel
+	 * @return Subscribable message channel
+	 * @throws Exception
+	 */
+	SubscribableChannel createSubscribableChannel(String name) throws Exception;
```
markfisher: why do these 2 methods have `throws Exception`? I don't think it's necessary (and if if were we would want to reconsider throwing *checked* Exceptions)

ilayaperumalg: +1. Sure, I will remove those.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactory.java
```diff
@@ -25,5 +26,24 @@
  */
 public interface ChannelFactory {
 
-	MessageChannel createChannel(String name, Class<?> channelType) throws Exception;
+	/**
+	 * Create a {@link SubscribableChannel} that will be bound to the message channel
```
markfisher: I would change "bound to" to either "bound by" or "bound via"... "the message channel binder" (the binder itself is not a destination for binding, but the agent that binds to some broker/middleware)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelFactory.java
markfisher: that will also allow removal of the `catch` block in `BindableProxyFactory`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -123,50 +122,49 @@ public void afterPropertiesSet() throws Exception {
...
-				catch (Exception e) {
-					throw new RuntimeException(e);
-				}
 			}
 
 		});
 	}
 
+	private MessageChannel createBindableChannel(String name, Class<? extends MessageChannel> channelType) {
+			return isPollable(channelType) ? this.channelFactory.createPollableChannel(name) :
```
markfisher: this indent needs to be adjusted now that you removed the try/catch

ilayaperumalg: Sorry. I always thought my IDE would fix it when I save but apparent it didn't do so :(. Pushed the fix.

markfisher:
rebased, squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/3afa3d8bd39988ac23af91963a2ca0c7962bae73 along with https://github.com/spring-cloud/spring-cloud-stream/commit/9dfa9c92186deab9861b6069c3caa264e41cc2cc

PR #220, opened 2015-12-04
Add waffle badges

sabbyanandan:


markpollack:
Added the build status image that comes from Bamboo.  Experimented a bit with formatting, hard to get any text to align before the images.  Merged as ac395f84e52856f72b5fbeece4db7f1918e37027

markpollack:
Also, put the changes in the top level README.adoc file since we are not generating it from the doc tree.

Issue #173, opened 2015-11-01
Create BinderFactory abstraction


sabbyanandan:
As a developer, I want to have a *BinderFactory* abstraction, so that I can support multiple binder types in the future.

**Acceptance Criteria:**
* Default binder support would be *redis* 
* Upon deployment with default binder type uses locally running *redis* instance for message transmission 
* Interface can be unit tested with different *mock* implementations

Issue #172, opened 2015-11-01
Add support for multiple binders per binder type


sabbyanandan:
As a developer, I want to be able to connect to multiple external systems for the same binding type, so that I can read data from a system and write it to another.

**Acceptance Criteria:**
* I can successfully consume from and publish to different instances of the same binder (ex: consume from *rabbit1* but publish to *rabbit2*) 
* Unit tests included to simulate publish and subscribe to an from different instances of the same binder type

Issue #140, opened 2015-09-21
Support for multiple connections to the same binder implementation


spencergibb:
In Spring Cloud Services (managed spring cloud netflix on PCF), SCS manages a rabbit instance for hystrix data. The user can have a user managed rabbitmq.  The app would need to support both rabbit connection factories. I've just ported they hystrix support from something amqp specific to stream (https://github.com/spring-cloud/spring-cloud-netflix/pull/548).

Issue #138, opened 2015-09-21
Support for multiple binder implementations in the same app


dsyer:
As a user, I'd like to consume messages from one middleware type and produce them in another (for instance).

dsyer:
Which issue is it a duplicate of @sabbyanandan ?

mbogoevici:
@dsyer 

I believe @sabbyanandan was referring to https://jira.spring.io/browse/XD-3563 https://jira.spring.io/browse/XD-3564 and https://jira.spring.io/browse/XD-3565

mbogoevici:
Also https://github.com/spring-cloud/spring-cloud-stream/issues/140

mbogoevici:
But I think that is a mistake, as #140 says something different.

PR #221, opened 2015-12-07
Initial implementation of multibinder support

mbogoevici:
This resolves spring-cloud/spring-cloud-stream#138,
This resolves spring-cloud/spring-cloud-stream#172,
This resolves spring-cloud/spring-cloud-stream#173,
This resolves spring-cloud/spring-cloud-stream#140 

- Addresses #138,#172,#173 with support for multiple binders in an application
- Binders can be of multiple types (e.g. Rabbit or Kafka), as well as multiple binders of the same type connecting to different system (e.g. multiple Rabbit binders connecting to different Rabbit clusters)
- Binders are now created in a separate context, allowing for multiple configurations
- Adds the `spring.cloud.spring.binders` namespace that allows for creating configuration properties for multiple binders

- Add support for default binder, modify Kafka tests to use multiple brokers when running embedded


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -0,0 +1,161 @@
...
+		}
+		if (!binderInstanceCache.containsKey(configurationName)) {
+			BinderConfiguration binderConfiguration = binderConfigurations.get(configurationName);
+			if (binderConfiguration == null) {
+				throw new IllegalStateException("Unknown binder configuration: " + configurationName);
+			}
+			Properties binderProperties = binderConfiguration.getProperties();
+			// Convert all properties to arguments, so that they receive maximum precedence
+			ArrayList<String> args = new ArrayList<>();
+			for (Map.Entry<Object, Object> property : binderProperties.entrySet()) {
```
ilayaperumalg: It might be worth to do `null` check on the binderProperties as one can rely on default properties for the `binder` configuration.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -0,0 +1,161 @@
...
+		public ConfigurableApplicationContext getBinderContext() {
+			return binderContext;
+		}
+	}
+
+	/**
+	 * Configuration class that enables autoconfiguration for the binders
+	 */
+	// TODO: Reconsider the use of autoconfiguration as part of binder configuration refactoring
+	@EnableAutoConfiguration
```
ilayaperumalg: Shouldn't the `enable` auto configuration be part of the actual binder `configuration` classes defined from the binder implementors?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+@Configuration
+public class BinderFactoryConfiguration {
+
+	@Bean
+	@ConditionalOnMissingBean(BinderFactory.class)
+	public BinderFactory binderFactory(BinderTypeRegistry binderTypeRegistry,
+									   ChannelBindingServiceProperties channelBindingServiceProperties) {
+
+		Map<String, BinderConfiguration> binderConfigurations = new HashMap<>();
+		if (!CollectionUtils.isEmpty(channelBindingServiceProperties.getBinders())) {
```
ilayaperumalg: Can we add `default` binder configurations that use all the built-in default values for the binders registered from the `BinderTypeRegistry` ? That way, one doesn't explicitly need to create a `binder configuration` even for just to use a  binder without any overriding binder properties.

spring-cloud-stream-tuple/pom.xml
```diff
@@ -47,5 +47,9 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
ilayaperumalg: What is reason behind adding this dependency?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
mbogoevici: Good catch, thanks!

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
mbogoevici: I specifically preferred not touching any of the binders in a first pass - it's easier to deal with this if there are fewer parts in motion IMO.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
mbogoevici: Actually, this is what happens a few lines below, here: https://github.com/spring-cloud/spring-cloud-stream/pull/221/files#diff-d51dcf232884fe56129839cd2934e6f4R77 . What this does is to flip between defaults (when no binders are defined) and specfic configurations (when binders configurations are present). I think that avoiding a mix of explicit configurations and defaults is desirable to avoid confusion.

mbogoevici: See the two newly added `multibinder` and `multibinder-different-systems` for comparision.

spring-cloud-stream-samples/multibinder-differentsystems/pom.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-stream-sample-multibinder-differentsystems</artifactId>
+	<packaging>jar</packaging>
+	<name>spring-cloud-stream-sample-multibinder-different-systems</name>
+	<description>Demo project for extended module using aggregate builder</description>
```
markfisher: looks like a copied description

spring-cloud-stream-samples/multibinder-differentsystems/src/main/java/multibinder/BridgeTransformer.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.integration.annotation.ServiceActivator;
+
+/**
+ * @author Marius Bogoevici
+ */
+@EnableBinding(Processor.class)
+public class BridgeTransformer {
+
+	@ServiceActivator(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT)
+	public Object transform(Object payload) {
```
markfisher: I believe this would be optimized (cc @garyrussell)  if it uses `Message` as the param and return type.

markfisher: actually given that this is only within the context of a sample, that is probably not important

spring-cloud-stream-samples/multibinder/pom.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-stream-sample-multibinder</artifactId>
+	<packaging>jar</packaging>
+	<name>spring-cloud-stream-sample-multibinder</name>
+	<description>Demo project for extended module using aggregate builder</description>
```
markfisher: same copied description here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Configuration for a binder instance, associating a {@link BinderType} with its configuration {@link Properties}.
+ * An application may contain multiple {@link BinderConfiguration}s per {@link BinderType}, when connecting to multiple
+ * systems of the same type.
+ *
+ * @author Marius Bogoevici
+ */
+public class BinderConfiguration {
+
+	private BinderType binderType;
```
markfisher: this can be final

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderType.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package org.springframework.cloud.stream.binder;
+
+import java.util.Arrays;
+
+/**
+ * References one or more {@link org.springframework.context.annotation.Configuration}-annotated classes which
+ * provide a context definition which contains exactly one {@link Binder}, typically for a given type of system (e.g.
+ * Rabbit, Kafka, Redis, etc.). An application may contain multiple instances of a given {@link BinderType}s,
```
markfisher: `BinderType` should not be plural in this sentence

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderTypeRegistry.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package org.springframework.cloud.stream.binder;
+
+import java.util.Map;
+
+/**
+ * A registry of {@link BinderType}s, indexed by name. A {@link BinderTypeRegistry} bean is created automatically
+ * based on information found in the {@literal META-INF/spring.binders} files provided by binder implementors.
+ * This can be overriden by registering a {@link BinderTypeRegistry} bean in the context.
```
markfisher: typo: 2 d's in 'overridden'

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
markfisher: ...or should `BinderConfiguration` itself initialize the `Properties` in its constructor (empty if none provided)? That way it would never be `null`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -76,16 +85,20 @@ private String removePrefix(String bindingTarget) {
...
 		}
 		else {
-			this.binder.unbindConsumers(inputChannelName);
+			binder.unbindConsumers(inputChannelName);
 		}
 	}
 
 	public void unbindProducers(String outputChannelName) {
-		this.binder.unbindProducers(outputChannelName);
+		String transport = this.channelBindingServiceProperties.getBinder(outputChannelName);
```
markfisher: we could factor out a private `getBinderForChannel()` method given how many times these 2 lines are duplicated:

```
String transport = this.channelBindingServiceProperties.getBinder(channelName);
Binder<MessageChannel> binder = binderFactory.getBinder(transport);
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAutoConfiguration.java
```diff
@@ -16,19 +16,42 @@
 
 package org.springframework.cloud.stream.config;
 
+import java.io.IOException;
```
markfisher: what's going on with all of these added imports?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/BinderAwareChannelResolverTests.java
```diff
@@ -55,6 +57,7 @@
  * @author Mark Fisher
  * @author Gary Russell
  */
+@Ignore
```
markfisher: do we have a story to address this so it can be un-`@Ignore`d again?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderProperties.java
```diff
@@ -0,0 +1,47 @@
...
+package org.springframework.cloud.stream.config;
+
+import java.util.Properties;
+
+/**
+ * Contains the properties of a binder.
+ *
+ * @author Marius Bogoevici
+ */
+public class BinderProperties {
```
markfisher: we might want to reconsider the name... for one thing it conflicts with an existing class of the same name (as evidenced by the need for a fully qualified name in this PR), but also it seems odd to have a `BinderProperties` class that is itself primarily used via its `getProperties()` method.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -36,36 +37,44 @@
...
 
 	public void bindConsumer(MessageChannel inputChannel, String inputChannelName) {
 		String channelBindingTarget = this.channelBindingServiceProperties.getBindingDestination(inputChannelName);
+		String transport = this.channelBindingServiceProperties.getBinder(inputChannelName);
+		Binder<MessageChannel> binder = binderFactory.getBinder(transport);
 		if (BinderUtils.isChannelPubSub(channelBindingTarget)) {
-			this.binder.bindPubSubConsumer(removePrefix(channelBindingTarget),
+			BindingProperties bindingProperties = this.channelBindingServiceProperties.getBindings()
+					.get(inputChannelName);
+			String group = bindingProperties == null ? null : bindingProperties.getGroup();
```
markfisher: looks like the `group` String is not being used

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -0,0 +1,131 @@
...
+		}
+		return new DefaultBinderTypeRegistry(binderTypes);
+	}
+
+
+	static Collection<BinderType> parseBinderConfigurations(ClassLoader classLoader, Resource resource)
+			throws IOException, ClassNotFoundException {
+		Properties properties = PropertiesLoaderUtils.loadProperties(resource);
+		Collection<BinderType> parsedBinderConfigurations = new ArrayList<>();
+		for (Map.Entry<?, ?> entry : properties.entrySet()) {
```
markfisher: why not iterate property names here and use getProperty to avoid the String-casts?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderProperties.java
markfisher: The *other* `BinderProperties` is just an interface full of constants. We should either consider relocating those to an existing class or renaming that class more explicitly, such as `BindingPropertyKeys` or `BindingPropertyNames`.

markfisher: For *this* class perhaps something like `BinderInstanceMetadata` would be more appropriate given that it includes a getter for type and a getter for properties.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -174,4 +193,11 @@ public Properties getProducerProperties(String outputChannelName) {
 		}
 	}
 
+	public String getBinder(String channelName) {
```
markfisher: This class has evolved to include several of these "lookup" methods, rather than just playing the role of `@ConfigurationProperties`. Should we consider factoring some of that out?... perhaps as part of the larger changes such as removing `ChannelBindingService` altogether. Obviously, that would be beyond the scope of this PR.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/BinderFactoryConfigurationTests.java
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * @author Marius Bogoevici
+ */
+public class BinderFactoryConfigurationTests {
+
+	@Test
+	public void loadBinderTypeRegistry() throws Exception {
+		try {
+			ConfigurableApplicationContext context = createBinderTestContext(
```
markfisher: should add a `fail()` after this line to ensure the `Exception` is thrown?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
garyrussell: Right - that code was moved to `consumerGroup(inputChannelName)` (below) with the fix for #211

garyrussell: I would delete this variable - `consumerGroup()` is also used during the unbind.

markfisher: +1

spring-cloud-stream-samples/multibinder-differentsystems/pom.xml
mbogoevici: Yep. Will fix.

spring-cloud-stream-tuple/pom.xml
mbogoevici: This shouldn't be here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: Yeah, it's a rebase artifact.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: +1

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
mbogoevici: We can do that - it's a little bit less efficient, but it doesn't matter, if you feel it reads better.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAutoConfiguration.java
mbogoevici: Need to be removed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
mbogoevici: I don't particularly like the methods, but they're an expedient way of doing a null-safe property lookup. Maybe we can invest some effort in that direction ...

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/BinderAwareChannelResolverTests.java
mbogoevici: Good catch. This needs to be fixed before we merge the PR.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
mbogoevici: I like that option better ...

spring-cloud-stream-samples/multibinder-differentsystems/src/main/java/multibinder/BridgeTransformer.java
mbogoevici: Yeah, we can improve this in the interest making a better sample, but it's just a test, really.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderProperties.java
mbogoevici: I'm all for changing the other `BinderProperties`, but don't want to add more unrelated changes to the PR. So let's change the name of this for now. I'd like to retain the suffix `Properties` for the properties class, since it's more or less what we do for every other typesafe property holder. Would `BinderInstanceProperties` a more suitable compromise? I see the issue about `binderProperties.getProperties()` but I don't know how to call the properties holder better than `properties` , so maybe `BinderInstanceConfiguration` or something.

markfisher: I considered "configuration" but then it's generally expected to be annotated with `@Configuration` which is why I suggested "metadata", but maybe there's a better option that avoids both "properties" and "configuration"

mbogoevici: As discussed with @markfisher renaming the field to `environment` is a more accurate solution.

mbogoevici:
Pushes a series of changes, addressing the comments

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/96f96f7621d6d4ccd24b37036a3707a0fe41d669

PR #228, opened 2015-12-10
Remove autoconfiguration for binders

mbogoevici:
Remove reliance on autoconfiguration for binders, preventing them to act on the main application's  dependencies.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
@@ -155,8 +155,6 @@ public class DefaultBinderFactory<T> implements BinderFactory<T>, DisposableBean
 	/**
 	 * Configuration class that enables autoconfiguration for the binders
 	 */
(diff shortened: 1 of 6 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -155,8 +155,6 @@ public ConfigurableApplicationContext getBinderContext() {
 	/**
 	 * Configuration class that enables autoconfiguration for the binders
 	 */
-	// TODO: Reconsider the use of autoconfiguration as part of binder configuration refactoring
-	@EnableAutoConfiguration
 	public static class SeedConfiguration {
```
markfisher: do we still need this class at all?

mbogoevici: No.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/bf178360eebaf6eb00ff48bbbfc5c1a182b6f22b

PR #230, opened 2015-12-11
Binders with default settings inherit from module's configuration

mbogoevici:
Resolves #229
Optimizes connection factory creation for binders that can use Boot's autoconfiguration

PR #232, opened 2015-12-15
Inherit Environment from main context when creating a Binder

dsyer:
The DefaultBinderFactory makes use of SpringApplicationBuilder to
build a context containing a binder. It has access to the Environment
of the surrounding context, but doesn't use it, and doesn't use it
(before this change). Users will be surprised if (for instance) they
change the default rabbitmq hostname and the change doesn't propagate
to the binder.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
@@ -26,7 +26,9 @@ import org.springframework.boot.Banner.Mode;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.EnvironmentAware;
+import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.Environment;
+import org.springframework.core.env.StandardEnvironment;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
It looks like a similar issue is resolved [here] (https://github.com/spring-cloud/spring-cloud-stream/pull/230/files#diff-bf51553f8e5cf51a0af7c8fb68d5fb95R120)

mbogoevici:
I prefer the solution in https://github.com/spring-cloud/spring-cloud-stream/pull/230, as it also ensures that autoconfig beans (and for example connection factory customization) beans are inherited by the binder contexts. For binder contexts that specify a completely new environment that can be disabled or made optional.

dsyer:
Maybe a combination of the two is needed. I'm not really clear on when you would not want to be a child context yet.

mbogoevici:
If you are specifying two Rabbit binders with different Rabbit hosts, you will need to create separate connection factories, and that's easier (and clearer) done when the binder context is not a child of the 'default' main context. That shouldn't prevent some environment propagation, though, so in that sense, adding this to the original fix makes sense - that's what I meant by "For binder contexts that specify a completely new environment that can be disabled or made optional." . But the original issue is still better fixed if the binder context inherits the connection factory created by Boot autoconfig, I would believe, rather than creating a new one that reuses the environment settings of the main context.

dsyer:
I think we are agreeing we need both (parent context is the default, but when not a parent context the environment should be propagated).

mbogoevici:
Yes, indeed. If it's okay with you @dsyer, I'd like to add this to complement the other PR.

dsyer:
Sounds like a plan

mbogoevici:
Closed in favour of merging this into https://github.com/spring-cloud/spring-cloud-stream/pull/230

PR #258, opened 2015-12-21
Remove tap sample

ilayaperumalg:


spring-cloud-stream-samples/pom.xml
```diff
@@ -4,7 +4,6 @@
 	
 	<artifactId>spring-cloud-stream-samples</artifactId>
 	<packaging>pom</packaging>
-	<url>http://projects.spring.io/spring-xd/</url>
```
mbogoevici: We can use the correct URL `http://cloud.spring.io/spring-cloud-stream/` instead.

ilayaperumalg: yeah, sure

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/62865679a73bf55d29e9b14de9d06226e0f1f41e

PR #259, opened 2015-12-21
scsm-88 Support Dynamic Binding Channel Limits

garyrussell:
Required by scsm-88.

If `outputChannelNames` is not zero length, only dynamically bind if the channel is is in the list.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
@@ -93,15 +93,16 @@ public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestina
 								" [<transport>:]<name>");
 					}
 				}
-				String[] outboundChannelNames = null;
+				String[] dynamicDestinations = null;
 				Properties producerProperties = null;
 				if (this.channelBindingServiceProperties != null) {
-					outboundChannelNames = this.channelBindingServiceProperties.getOutboundChannelNames();
+					dynamicDestinations = this.channelBindingServiceProperties.getDynamicDestinations();
 					// TODO: need the props to return some defaults if not found
 					producerProperties = this.channelBindingServiceProperties.getProducerProperties(name);
 				}
-				boolean dynamicAllowed = outboundChannelNames == null || outboundChannelNames.length == 0
-						|| Arrays.asList(outboundChannelNames).contains(name);
+				boolean dynamicAllowed = dynamicDestinations == null
+						|| dynamicDestinations.length == 0
+						|| Arrays.asList(dynamicDestinations).contains(name);
 				if (dynamicAllowed) {
 					Binder<MessageChannel> binder = binderFactory.getBinder(transport);
 					binder.bindProducer(name, channel, producerProperties);
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
@@ -16,17 +16,19 @@
 
 package org.springframework.cloud.stream.binding;
 
+import java.util.Arrays;
 import java.util.Properties;
 
 import org.springframework.beans.factory.BeanFactory;
-import org.springframework.beans.factory.config.ConfigurableBeanFactory;
-import org.springframework.beans.factory.support.DefaultListableBeanFactory;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.cloud.stream.binder.Binder;
 import org.springframework.cloud.stream.binder.BinderFactory;
+import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
 import org.springframework.integration.channel.DirectChannel;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.core.BeanFactoryMessageChannelDestinationResolver;
 import org.springframework.messaging.core.DestinationResolutionException;
+import org.springframework.util.Assert;
 
 /**
  * A {@link org.springframework.messaging.core.DestinationResolver} implementation that
@@ -40,36 +42,41 @@ public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestina
 
 	private final BinderFactory<MessageChannel> binderFactory;
 
-	private final Properties producerProperties;
+	private final ChannelBindingServiceProperties channelBindingServiceProperties;
 
-	private DefaultListableBeanFactory beanFactory;
+	private ConfigurableListableBeanFactory beanFactory;
 
-	public BinderAwareChannelResolver(BinderFactory<MessageChannel> binderFactory, Properties producerProperties) {
+	public BinderAwareChannelResolver(BinderFactory<MessageChannel> binderFactory,
+			ChannelBindingServiceProperties channelBindingServiceProperties) {
+		Assert.notNull(binderFactory, "'binderFactory' cannot be null");
 		this.binderFactory = binderFactory;
-		this.producerProperties = producerProperties;
+		this.channelBindingServiceProperties = channelBindingServiceProperties;
 	}
 
 	@Override
 	public void setBeanFactory(BeanFactory beanFactory) {
 		super.setBeanFactory(beanFactory);
-		if (beanFactory instanceof ConfigurableBeanFactory) {
-			this.beanFactory = (DefaultListableBeanFactory) beanFactory;
+		if (beanFactory instanceof ConfigurableListableBeanFactory) {
+			this.beanFactory = (ConfigurableListableBeanFactory) beanFactory;
 		}
 	}
 
 	@Override
 	public MessageChannel resolveDestination(String name) {
 		MessageChannel channel = null;
+		DestinationResolutionException destinationResolutionException;
 		try {
 			return super.resolveDestination(name);
 		}
 		catch (DestinationResolutionException e) {
+			destinationResolutionException = e;
 		}
 		synchronized (this) {
 			try {
 				return super.resolveDestination(name);
 			}
 			catch (DestinationResolutionException e) {
+				destinationResolutionException = e;
 			}
 			if (this.beanFactory != null && this.binderFactory != null) {
 				channel = new DirectChannel();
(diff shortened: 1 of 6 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -49,7 +49,7 @@ public MessageChannel resolveDestination(String name) {
 		}
 		catch (DestinationResolutionException e) {
 		}
-		if (name.contains(":")) {
+		if (name.contains(":")) { // TODO : optional - only needed if transport provided
 			if (binderFactory != null) {
 				String[] tokens = name.split(":", 2);
```
mbogoevici: Must change the maximum to 3. Will fix on merge.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -66,14 +66,11 @@ else if (tokens.length == 3) {
 							" [<transport>:]<type>:<name>");
 				}
 				Binder<MessageChannel> binder = binderFactory.getBinder(transport);
-				if ("queue".equals(type)) {
+				if ("queue".equals(type)) { // TODO: remove
 					channel = binder.bindDynamicProducer(name, this.producerProperties);
 				}
-				else if ("topic".equals(type)) {
```
mbogoevici: Ditching the 'topic' prefix makes it a little ambiguous, as '[transport]:[name]' and 'queue:[name]' will take the branch where the `type` becomes `[transport]`. IMO should add another check that the first token is not `queue` before deciding which scheme are we going to use on parsing. Of course, that will cause issues with folks naming a transport `queue`, but that's fine, I would suppose. And `queue` is rather not long lived anyway.

mbogoevici:
LGTM - with the two notes above, which can be both fixed on merge. 
Could also use a unit test for multibinder.

garyrussell:
Thanks, @mbogoevici , I am going to rework this as a PoC based on the slack discussion I had with @markfisher this morning (white list of allowable dynamic channels). I will add test case(s).

artembilan:
Looks like you are talking here about https://jira.spring.io/browse/INT-1926

garyrussell:
Yes; we should probably go ahead and implement that, in any case, but here it's a bit more complicated than that.

Consider: `expression="headers['foo']"` to select the channel name.

In this case, we can't do it in the router; the router has no information about the binding properties; we need to make available the allowable channel names to the binder aware channel resolver so it will only dynamically bind if the channel is in the white list (or there are no restrictions).

In other words, if the binding properties contains an outputChannel list, only they are allowed; if no list is provided the resolver needs to bind the channels dynamically.

i.e. the decision has to be made in the resolver, not in the router.

Let me hack together a straw-man and we can discuss further.

garyrussell:
Pushed

garyrussell:
Updated for the newly refactored `Binder`; ready for review, but please don't merge until I rework the router PR in s-c-s-m.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -86,8 +93,22 @@ else if (tokens.length != 1) {
 								" [<transport>:]<name>");
 					}
 				}
-				Binder<MessageChannel> binder = binderFactory.getBinder(transport);
-				binder.bindProducer(name, channel, this.producerProperties);
+				String[] outboundChannelNames = null;
+				Properties producerProperties = null;
+				if (this.channelBindingServiceProperties != null) {
+					outboundChannelNames = this.channelBindingServiceProperties.getOutboundChannelNames();
+					// TODO: need the props to return some defaults if not found
```
mbogoevici: As mentioned on Slack, I strongly believe we should restore global producer and consumer properties in `ChannelBindingServiceProperties`

garyrussell: +1 :+1:

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -50,6 +50,8 @@
 
 	private String defaultBinder;
 
+	private String[] outboundChannelNames = new String[0];
```
mbogoevici: This property might be slightly confusing, since it does not include the outputs of `Source`, `Processor` or anything else specified via `@EnableBinding`, which definitely *are* output channels (granted, this feature is intended to be used mainly by Sinks). This is very specific to dynamic binding so I think we should find a way to express that.

markfisher: Also we're not strictly talking about "channels" here, but "destinations". Perhaps the property could be renamed something like `routableDestinations`?

garyrussell: How about `allowedDynamicDestinations` ?

mbogoevici: +1 on @garyrussell 's suggestion

markfisher: sounds good - I like the inclusion of "dynamic"

garyrussell:
Pushed testing support; I went with `dynamicDestinations` for now; we can talk more.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -86,8 +93,23 @@ else if (tokens.length != 1) {
...
-				Binder<MessageChannel> binder = binderFactory.getBinder(transport);
-				binder.bindProducer(name, channel, this.producerProperties);
+				String[] dynamicDestinations = null;
+				Properties producerProperties = null;
+				if (this.channelBindingServiceProperties != null) {
+					dynamicDestinations = this.channelBindingServiceProperties.getDynamicDestinations();
+					// TODO: need the props to return some defaults if not found
+					producerProperties = this.channelBindingServiceProperties.getProducerProperties(name);
+				}
+				boolean dynamicAllowed = dynamicDestinations == null
```
mbogoevici: Possibly simplify as o.s.u.ObjectUtils.isEmpty()

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binding/ChannelBindingServiceTests.java
```diff
@@ -94,10 +110,51 @@ public void testExplicitGroup() throws Exception {
...
 		Binding<MessageChannel> mockBinding = Binding.forConsumer("foo", "fooGroup", Mockito.mock(AbstractEndpoint.class),
 				inputChannel, null);
-		Mockito.when(binder.bindConsumer("foo", "fooGroup", inputChannel, new Properties()))
+		when(binder.bindConsumer("foo", "fooGroup", inputChannel, new Properties()))
 				.thenReturn(mockBinding);
 		Binding<MessageChannel> binding = service.bindConsumer(inputChannel, name);
-		Assert.assertThat(binding, sameInstance(mockBinding));
+		assertThat(binding, sameInstance(mockBinding));
+
+		final AtomicReference<MessageChannel> dynamic = new AtomicReference<>();
```
mbogoevici: I'd make the second part a separate test, just to avoid a possible confusion. Also because it reuses the mockBinding.

mbogoevici:
These are my only observations, LGTM in general. Can fix on merge if agreed.

garyrussell:
OK; thanks; sounds good

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -40,36 +42,41 @@
...
 		MessageChannel channel = null;
+		DestinationResolutionException destinationResolutionException;
 		try {
 			return super.resolveDestination(name);
 		}
 		catch (DestinationResolutionException e) {
+			destinationResolutionException = e;
 		}
 		synchronized (this) {
 			try {
```
mbogoevici: Seems like the superclass call is duplicate, once inside the synchronized block, and once outside. Is this a double-check locking?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -86,8 +93,23 @@ else if (tokens.length != 1) {
...
+				Properties producerProperties = null;
+				if (this.channelBindingServiceProperties != null) {
+					dynamicDestinations = this.channelBindingServiceProperties.getDynamicDestinations();
+					// TODO: need the props to return some defaults if not found
+					producerProperties = this.channelBindingServiceProperties.getProducerProperties(name);
+				}
+				boolean dynamicAllowed = dynamicDestinations == null
+						|| dynamicDestinations.length == 0
+						|| Arrays.asList(dynamicDestinations).contains(name);
+				if (dynamicAllowed) {
```
mbogoevici: Also, should we do this check before registering the channel? It seems like there would be little reason to do the channel registration unless we can actually bind to it.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -86,8 +93,23 @@ else if (tokens.length != 1) {
...
+					Binder<MessageChannel> binder = binderFactory.getBinder(transport);
+					binder.bindProducer(name, channel, producerProperties);
```
mbogoevici: While reviewing this, I realized that we never unbind these channels. While the endpoints are in principle stopped eventually, I still think it'd be preferable if the bindings would be part of the managed input/output binding lifecycle.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
mbogoevici: Actually it is. IntelliJ threw me off I guess.

mbogoevici:
Minor polishing here: https://github.com/mbogoevici/spring-cloud-stream/commit/fbb51771741a4af40f935e3ebdbaea795350a3c0 

(Does not address unbinding of dynamic channels, that can be done separately)

garyrussell:
Polishing looks good; thanks; yes the double-check locking is to avoid synchronizing once the channel has been registered with the context.

I thought about unbinding - we might be able to add something in the router when it's stopped.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/16e7893e497ddf2ab03c229545e4326230c73dc9, additional minor polishing https://github.com/spring-cloud/spring-cloud-stream/commit/2f330c7bc477c65bd2dc571a5507164eca3d3bc3

PR #261, opened 2015-12-22
Handle binder config properties in BindingProperties

### LINKED ISSUE #256: Move producer/consumer properties as per-binding properties
Currently, the producer/consumer properties are set at `ChannelBindingService` level (as `ChannelBindingServiceProperties`). Since, they are per-binding level properties, we need to move the producer/consumer properties as `BindingProperties`.

ilayaperumalg:
- Per-binding properties defined in `BindingProperties` need to hold the binder specific configurations needed when binding producer/consumers.
Currently, these are added in `ChannelBindingServiceProperties` as producer/consumer properties. Since, these configuration properties are for
per-binding based on the input/output (consumer/producer) channel being bound it would be better to move these properties into `BindingProperties`.
- Add `config` key in `BindingProperties` which can hold all the properties that can be passed as `per-binding` properties (including the producer/consumer properties)
- Remove `partition`, `trackHistory` related properties with the assumption they are handled in BindingProperties' config

- This fixes #256


ilayaperumalg:
Closing this for now to provide strongly typed `BindingProperties` rather than using `Map` for all the per-binding properties.

PR #262, opened 2015-12-23
Handle binder config properties in BindingProperties

### LINKED ISSUE #256: Move producer/consumer properties as per-binding properties

ilayaperumalg:
- Per-binding properties defined in `BindingProperties` need to hold the binder specific configurations needed when binding producer/consumers.
Currently, these are added in `ChannelBindingServiceProperties` as producer/consumer properties. Since, these configuration properties are for
per-binding based on the input/output (consumer/producer) channel being bound it would be better to move these properties into `BindingProperties`.
 - Add some of the supported producer/consumer properties into BindingProperties

- This fixes #256


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/CommonBinderProperties.java
```diff
@@ -22,7 +22,7 @@
  *
  * @author Gary Russell
  */
-public interface BinderProperties {
+public interface CommonBinderProperties {
```
markfisher: I think it would be better to rename this something that clarifies it is only a holder for property names (it's not actually properties in the sense of key/value pairs). Perhaps `BinderPropertyKeys`?

artembilan: From other side any variable in the `interface` is exactly `public static final`.
I mean that we don't need those modification there.

I don't remember the discussion, but decided to stay away from interfaces for the constants containers in the SI. e.g. `FileHeaders`, `AmqpHeaders` etc.

I would agree with @markfisher regarding `BinderPropertyKeys` name...

ilayaperumalg: +1 for both the comments. I will make this interface to `abstract` class (as used in SI) and rename the class name.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderPropertyKeysAccessor.java
```diff
@@ -29,13 +29,13 @@
  *
  * @author Gary Russell
  */
-public abstract class AbstractBinderPropertiesAccessor implements BinderProperties {
+public abstract class AbstractBinderPropertyKeysAccessor extends BinderPropertyKeys {
```
markfisher: This should *not* be renamed since it is the accessor for actual property *value*. It's not an *implementation* of the `BinderPropertyKeys` interface. In fact, I don't think we should have the *extends* keyword here, but rather should just have a dependency on `BinderPropertyKeys` where those constants are needed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderPropertiesAccessor.java
```diff
@@ -29,7 +29,7 @@
  *
  * @author Gary Russell
  */
-public abstract class AbstractBinderPropertiesAccessor implements BinderProperties {
+public abstract class AbstractBinderPropertiesAccessor {
```
markfisher: I believe this should be renamed `AbstractBindingPropertiesAccessor` now, but it's a bit of a mix between *binder* an *binding* properties. Perhaps we should be splitting those, and maybe even splitting producer vs. consumer binding property accessors? WDYT @garyrussell ?

ilayaperumalg: Yeah, I believe it is hard to distinguish most of these properties whether they are `binding` or `binder` properties and they mostly belong to the same. I like the idea of separating them based on `producer/consumer` properties. Also, it would be better we name these as `producer/consumer` **binding** properties rather than **binder** properties since `binder` properties can only be [this] (https://github.com/ilayaperumalg/spring-cloud-stream/blob/per-binding-properties-fix/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderProperties.java)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
```diff
@@ -172,9 +273,48 @@ public String toString() {
...
+		if (this.concurrency != null) {
+			sb.append("concurrency=" + this.concurrency);
+			sb.append(COMMA);
 		}
 		sb.deleteCharAt(sb.lastIndexOf(COMMA));
 		return "BindingProperties{" + sb.toString() + "}";
 	}
 
-}
+}
```
markfisher: add newline

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderPropertiesAccessor.java
markfisher: For now, I think we should rename to `AbstractBindingPropertiesAccessor`. We can consider splitting into multiple classes later.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/fa89e5bcb6bfc808015dce4c884a1ff06e990e5b

Issue #266, opened 2016-01-04
Support pub/sub only semantics for Redis binder


sabbyanandan:
As a developer, I'd like to support only `pub/sub` semantics for `redis` binder.

*Acceptance:*
`queue` semantics are not supported
`pub/sub` specs are supported and documented with samples

mbogoevici:
Closed via https://github.com/spring-cloud/spring-cloud-stream/commit/c3758b9dc0d0ccebbf162a7f75affb84a1b08ff0

Issue #263, opened 2015-12-23
Explore options to support only pub/sub for binders


sabbyanandan:
As a developer, I'd like to investigate how to only support `pub/sub` (vs. `queue`) semantics at the binder level. With this spike, we would also want to address the following.

- would the `pub/sub` semantics apply to producer, consumer, or both sides?
- what happens to `redis` as a binder if we do not support `queue:` semantics?
- perhaps drop the notion of `request` and `reply` at the binder level

markfisher:
The (review only) PR #276 shows the results of this spike. In the next sprint, that will be ready for merging, and it addresses all current `Binder` implementations.

Issue #243, opened 2015-12-18
Support PubSub in Kafka binder 


ericbottard:
Un-@Ignore KafkaBinderTests.testSendAndReceivePubSub()

sabbyanandan:
This depends on spring-cloud/spring-cloud-stream#268

PR #276, opened 2016-01-14
[Sprint 66] Binder simplification

markfisher:
The best place to start reviewing is the `Binder` interface itself. Several methods have been removed (and as a follow-up, we will likely reduce the `unbind` methods to 2 as well - one for producer and one for consumer). Most notably, there are no longer separate "pubSub" methods. Instead, we now use consumer *group* for every consumer bind (and there is a "default" group if `null` or an empty String is provided). Within a group, messages will be load-balanced if more than one consumer instance is bound for that group. On the producer side, it's essentially always pub/sub, and if there are multiple groups represented, each group will get each Message. This means that our notion of *tap* is now no different than any additional consumer group, and if a "tap" contains more than one module, load-balancing will occur across those module instances (this is the original fix that @garyrussell introduced for Rabbit pub/sub behavior that led to the idea of this convergence in the first place).

For Rabbit, this model maps directly to a producer-side `TopicExchange` with the consumer-side having a `Queue` bound per group (routing-keys are used when partitioning is enabled).

For Kafka, the consumer group model aligns directly with Kafka's own consumer model, and the producer always sends to a "topic".

For Redis, the model is not as closely aligned (and we may decide not to support Redis in the future, but at the very least will not be recommending it for production). The consumer groups are being tracked in Redis ZSET collections where the name of each ZSET is `groups.<destination>`, and the members are group names (with the "score" representing the count of bound consumer instances). The producer then consults this ZSET to find each explicit queue (Redis LIST) and sends to each. Obviously there are many opportunities for optimizing... could be a background thread to manage the set contents as a cache, or even a pub/sub topic in Redis for set update notifications, or simply parameters to configure how often the ZSET is checked (a refresh interval based on time and/or the number of messages having been sent since the last check). Before going down any of those paths, we need to decide if we are even going to continue with the Redis implementation. If we decide to keep it but clarify that it is for dev-time convenience only, then perhaps we do not need to pursue any optimizations. In any case, that's out of scope for this PR.

This resolves spring-cloud/spring-cloud-stream#263
This resolves spring-cloud/spring-cloud-stream#243
This resolves spring-cloud/spring-cloud-stream#266
This resolves spring-cloud/spring-cloud-stream#267
This resolves spring-cloud/spring-cloud-stream#268


mbogoevici:
Regarding Redis - not sure of the benefit in continuing to try to support it. The only reasonable argument is that it was supported before.
I would argue that setting up RabbitMQ locally is in the end as easy as using Redis (sans the ability of MONITORing, but I'm pretty sure we can solve this with a better tracing/debugging strategy).  
I'd be much more keen to explore other lightweight (possibly in-memory) binders such as Chronicle Queue, Aeron, etc. instead.

markfisher:
@mbogoevici Agreed 100% - would be great to have one of those binders that would be even simpler in terms of setup

garyrussell:
The question, then, is whether we should expend any significant time reviewing the Redis Binder?

My vote is to trust @markfisher :smile:

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
```
mbogoevici: 2016

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -49,38 +65,32 @@ public MessageChannel resolveDestination(String name) {
...
-					type = tokens[0];
-				}
-				else if (tokens.length == 3) {
-					transport = tokens[0];
-					type = tokens[1];
-				}
-				else {
-					throw new IllegalArgumentException("Unrecognized channel naming scheme: " + name + " , should be" +
-							" [<transport>:]<type>:<name>");
+				if (name.contains(":")) {
```
mbogoevici: Maybe we can use a different character here and get done with the colon? Indirectly, this is a concern for Kafka, where the colon is a legal character.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitBinderTests.java
```diff
@@ -154,7 +150,7 @@ public void testConsumerProperties() throws Exception {
 		assertEquals(1000L, TestUtils.getPropertyValue(retry, "retryOperations.backOffPolicy.initialInterval"));
 		assertEquals(10000L, TestUtils.getPropertyValue(retry, "retryOperations.backOffPolicy.maxInterval"));
 		assertEquals(2.0, TestUtils.getPropertyValue(retry, "retryOperations.backOffPolicy.multiplier"));
-		binder.unbindConsumers("props.0");
+		binder.unbindConsumers("props.0", "default");
```
mbogoevici: There is something weird about binding with the `null` consumer group but having to pass back `"default"` on unbinding.

markfisher: Good point. We should probably have the same approach with a final superclass method that always ensures "default" is a replacement for either `null` or empty String on unbind as well.

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
```diff
@@ -300,55 +243,39 @@ public Boolean recover(RetryContext context) throws Exception {
...
-						parser.parseExpression(buildPartitionRoutingExpression(queueName)), this.connectionFactory);
-			}
-			queue.setIntegrationEvaluationContext(this.evaluationContext);
-			queue.setBeanFactory(this.getBeanFactory());
-			queue.afterPropertiesSet();
-			doRegisterProducer(name, moduleOutputChannel, queue, accessor);
-		}
+	public void unbindConsumers(String name, String group) {
+		super.unbindConsumers(name, group);
+		this.redisOperations.boundSetOps(CONSUMER_GROUPS_KEY_PREFIX + name).remove(group);
```
mbogoevici: This will work only there will be only one application instance per consumer group (i.e. no competing consumers or partitioning). If there are multiple clients with the same 'group', the first to unbind will deregister the others. I think we'll need to do something sophisticated (e.g. reference counting or something like that).

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
```diff
@@ -359,82 +286,54 @@ private void doRegisterProducer(final String name, MessageChannel moduleOutputCh
...
-			delegate.handleMessage(MessageBuilder.withPayload(messageToSend).copyHeaders(transformed).build());
+			
+			refreshChannelAdapters();
+			for (RedisQueueOutboundChannelAdapter adapter : adapters.values()) {
+				adapter.handleMessage((MessageBuilder.withPayload(messageToSend).copyHeaders(transformed).build()));
+			}
 		}
 
+		private void refreshChannelAdapters() {
+			Set<String> groups = redisOperations.boundSetOps(CONSUMER_GROUPS_KEY_PREFIX + bindingName).members();
```
mbogoevici: If groups (and their subscriptions) are not durable - based on https://github.com/spring-cloud/spring-cloud-stream/pull/276/files#diff-abf0f621ebbf51a773c2d6ab5129ac24R248, then maybe it makes sense to shut down adapters for groups that aren't listening anymore ?

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
markfisher: I think using a ZSET might work - with the count as the value. Still a "set" so that's good.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -277,9 +208,8 @@ public void testSendAndReceiveNoOriginalContentType() throws Exception {
 
 	}
 
-	@Override
 	@Test
-	public void testSendAndReceivePubSub() throws Exception {
+	public void testSendAndReceiveWithExplicitConsumerGroup() {
```
mbogoevici: This will need to be pulled up and a customization created for RawKafkaBinder.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -288,15 +218,16 @@ public void testSendAndReceivePubSub() throws Exception {
...
 		QueueChannel module3InputChannel = new QueueChannel();
 		binder.bindProducer("baz.0", moduleOutputChannel, null);
-		binder.bindConsumer("baz.0", moduleInputChannel, null);
+		binder.bindConsumer("baz.0", "test", moduleInputChannel, null);
 		moduleOutputChannel.addInterceptor(new WireTap(tapChannel));
 		// A new module is using the tap as an input channel
 		String fooTapName = "baz.0";
-		binder.bindPubSubProducer(fooTapName, tapChannel, null);
-		binder.bindPubSubConsumer(fooTapName, module2InputChannel, null, null);
+		binder.bindProducer(fooTapName, tapChannel, null);
```
mbogoevici: This will need to be rewritten, as it seems to duplicate messages.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -288,15 +218,16 @@ public void testSendAndReceivePubSub() throws Exception {
 		QueueChannel module2InputChannel = new QueueChannel();
 		QueueChannel module3InputChannel = new QueueChannel();
 		binder.bindProducer("baz.0", moduleOutputChannel, null);
-		binder.bindConsumer("baz.0", moduleInputChannel, null);
+		binder.bindConsumer("baz.0", "test", moduleInputChannel, null);
 		moduleOutputChannel.addInterceptor(new WireTap(tapChannel));
```
mbogoevici: This needs to be removed, we don't do wiretapping for taps anymore. Also, this causes duplicate messages to be resent to the main topic.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -435,51 +387,29 @@ public void afterPropertiesSet() throws Exception {
...
+		declareExchange(exchangeName, exchange);
+
+		String queueName = applyPrefix(prefix, baseQueueName);
 		int partitionIndex = accessor.getPartitionIndex();
-		String dlqNameRoot = name;
+		String dlqNameRoot = baseQueueName;
 		if (partitionIndex >= 0) {
 			String partitionSuffix = "-" + partitionIndex;
 			queueName += partitionSuffix;
 			dlqNameRoot += partitionSuffix;
```
garyrussell: FYI, I [just fixed a bug in XD](https://github.com/spring-projects/spring-xd/pull/1876) - over there, we only have one DLQ, not a DLQ for each partition.

Having a DLQ for each partition might be a bit of overkill, given that we can get the originating partition from a header, especially with a large number of partitions.

Not necessarily something for this PR, but something to think about, perhaps.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binder.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
```
ilayaperumalg: 2016

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
markfisher: Sure; that makes sense and is something I thought about while in this code (could imagine use-cases for draining the DLQ where it's much easier to address as a single queue per group). Would you like to create a commit on top of this one, or should we address it after merging?

garyrussell: I'll add a commit - and I'll port the test case from XD.

markfisher: Perfect. Thanks!

garyrussell: https://github.com/garyrussell/spring-cloud-stream/commit/4e8d56bff9b828198d8921c1035141d82a28050a

I didn't push directly to your branch in case you have some local WIP.

garyrussell: Don't merge - it's wrong :frowning:

garyrussell: I added a second commit fixing DLQ binding on the producer side. My branch now has 2 commits above yours.

@markfisher as discussed on skype, we still have some holes in unbind/cleanup.

markfisher: Thanks for the changes, I squashed them into a single commit and updated this PR to include that commit.

spencergibb:
@markfisher @mbogoevici I wonder if this would help with the consul impl I [started](https://github.com/spring-cloud/spring-cloud-consul/issues/86) (and admittedly haven't tried super hard to finish). I already have an `InboundChannelAdapter` and an `OutboundEndpoint`.

mbogoevici:
@spencergibb I think that the simplifications definitely would help.

mbogoevici:
Notes: this requires a Spring Cloud Data Flow update, for populating the group info.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupport.java
```diff
@@ -82,13 +80,17 @@
...
-	protected static final String P2P_NAMED_CHANNEL_TYPE_PREFIX = "queue:";
-
-	protected static final String PUBSUB_NAMED_CHANNEL_TYPE_PREFIX = "topic:";
+	protected static final String PARTITION_HEADER = "partition";
 
-	protected static final String JOB_CHANNEL_TYPE_PREFIX = "job:";
+	/**
+	 * Default group name (used if <code>null</code> or empty String is provided).
+	 */
+	protected static final String DEFAULT_CONSUMER_GROUP = "default";
```
mbogoevici: This is not really effective in the context of SCS, where the default is to set a UUID when no group property is provided for a binding. That has been done to fix https://github.com/spring-cloud/spring-cloud-stream/issues/211, and I think we should preserve it - applications wishing to act as competing consumers should rather explicitly register as part of a group than the other way around (apps acting as competing consumers unless they are registered as part of distinct consumer groups). I think we should disable the current defaulting mechanism, and put in place an alternative as part of upcoming refactoring.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -587,54 +521,36 @@ private MessageRecoverer determineRecoverer(String name, RabbitPropertiesAccesso
...
-			declareQueue(name, queue);
-			autoBindDLQ(name, properties);
-			endpoint.setRoutingKey(queueName);
-			org.springframework.amqp.core.Binding binding = BindingBuilder.bind(queue).to(exchange).with(queueName);
-			declareBinding(queueName, binding);
+			Queue queue = new Queue(baseQueueName, true, false, false, queueArgs(properties, baseQueueName));
+			declareQueue(baseQueueName, queue);
+			autoBindDLQ(baseQueueName, baseQueueName, properties);
+			endpoint.setRoutingKey(name);
+			org.springframework.amqp.core.Binding binding = BindingBuilder.bind(queue).to(exchange).with(name);
```
mbogoevici: There is a slight difference between the way the default queue is bound (with the routing key matching the `name`) and the way explicitly named consumer groups are bound (hash). While this may not have any practical consequences (the outbound producers will always use the proper routing key), I think it would be better if we were consistent, and avoid some subtle bugs down the road.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -652,18 +568,14 @@ private void configureOutboundHandler(AmqpOutboundEndpoint handler, RabbitProper
 	}
 
 	@Override
-	public void bindPubSubProducer(String name, MessageChannel moduleOutputChannel,
-			Properties properties) {
-		validateProducerProperties(name, properties, SUPPORTED_PUBSUB_PRODUCER_PROPERTIES);
+	public void bindProducer(String name, MessageChannel outputChannel, Properties properties) {
+		validateProducerProperties(name, properties, SUPPORTED_PRODUCER_PROPERTIES);
 		RabbitPropertiesAccessor accessor = new RabbitPropertiesAccessor(properties);
 		String exchangeName = applyPrefix(accessor.getPrefix(this.defaultPrefix), name);
```
ilayaperumalg: It looks like the method invocation `buildOutboundEndpoint` also declares the `exchange` with the same name and the steps at lines 574:576 can be removed?

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
markfisher: Good point and +1 for consistency. I'm actually wondering if we need to use the name in the routingKey at all. We could simply use the partition ID as the routingKey (or 0 if no partitioning), then for binding keys use the partition index on the consumer side else # for non-partitioned consumption. WDYT @garyrussell ?

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
markfisher: Yes, also we can remove the `doRegisterProducer()` method that accepts a `replyTo` since that is no longer used (request/reply behavior has been removed from the core `Binder` contract).

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
garyrussell: My reason for routing with the full queue name is to make DLQ re-queueing easier.

We route to the DLX using the full queue name (we set it explicitly on the queue but that's not really necessary because rabbit will route to the DLX using the original routing key if none is specified on the queue).

If we route to the primary queue using just the partition, we'd still need to route to the DLX using the full queue name.

If I want to requeue a dead lettered message, I can simply route to the proper exchange using the message's `receivedRoutingKey`.

If we change the primary routing key we'd need to manipulate the received routing key when re-queuing - or use a separate DLX for each binding. (Currently all DLQs are bound to a single DLX).

That said, the `x-death` header has all the information we need to re-route (exchange, routing key), which we'd need to examine to get the exchange anyhow.

But yes, we could bind the default queue with a hash.

markfisher: Thanks for the detailed explanation. So, I guess we can just make the minimal change to ensure both cases use a consistent bindingKey (can be hash).

mbogoevici: Thanks @garyrussell. Can make the change during merge.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
mbogoevici: Ok, will remove on merge.

mbogoevici:
Added further binder simplifications from @markfisher and removed consumer group defaulting here: https://github.com/mbogoevici/spring-cloud-stream/tree/binder-simplification-with-unbind

mbogoevici:
Addressed branch comments here: https://github.com/mbogoevici/spring-cloud-stream/commit/813fdadec16352913b32e68631446305d1bf0e52

mbogoevici:
Rebased, squashed, closed via https://github.com/spring-cloud/spring-cloud-stream/commit/c3758b9dc0d0ccebbf162a7f75affb84a1b08ff0, additional enhancements in https://github.com/spring-cloud/spring-cloud-stream/commit/192aa79baa881e01f197ebf1a58651de4b4a0106 and https://github.com/spring-cloud/spring-cloud-stream/commit/e2485e3290b599582dbb11bddfc128a921ee6bef

PR #281, opened 2016-01-19
Remove Spring XD tuple dependency

### LINKED ISSUE #280: Remove Spring XD tuple dependency
Currently, spring-cloud-stream-binder implementations use Spring XD tuple as a dependency (mostly in tests). We could use the tuple artifact from spring-cloud-stream itself.

ilayaperumalg:
- Use spring-cloud-stream-tuple instead

This resolves #280


markfisher:
LGTM; merged

PR #283, opened 2016-01-20
Support proxy settings for Aether Module resolver

### LINKED ISSUE #272: AetherModuleResolver may not be picking up proxy settings
As a user, I'm deploying a `stream` behind the firewall with an artifact repository that acts as a proxy; however, I'm unable to successfully deploy the stream, as the `AetherModuleResolver` resolver may not be picking up the proxy settings to resolve and download the module jar's required by the stream.

**Acceptance:**
* Module jar's are successfully resolved by `AetherModuleResolver`
* Streams successfully deploy behind the firewall with an artifact repository that acts as a proxy

ilayaperumalg:
- Add `AetherProxyProperties` as a ConfigurationProperties for the proxy settings needed for the ProxySelector in Aether system
 - Update proxy settings in two places:
   1) Where the remote repositories are setup
   2) WHere the repository system session is created to resolve the artifacts

This resolves #272

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
@@ -110,14 +110,14 @@ public class AetherModuleResolver implements ModuleResolver {
 			this.authentication = new Authentication() {
 				@Override
 				public void fill(AuthenticationContext context, String key, Map<String, String> data) {
-					context.put(context.USERNAME, proxyProperties.getUsername());
-					context.put(context.PASSWORD, proxyProperties.getPassword());
+					context.put(context.USERNAME, proxyProperties.getAuth().getUsername());
+					context.put(context.PASSWORD, proxyProperties.getAuth().getPassword());
 				}
 
 				@Override
 				public void digest(AuthenticationDigest digest) {
-					digest.update(AuthenticationContext.USERNAME, proxyProperties.getUsername(),
-							AuthenticationContext.PASSWORD, proxyProperties.getPassword());
+					digest.update(AuthenticationContext.USERNAME, proxyProperties.getAuth().getUsername(),
+							AuthenticationContext.PASSWORD, proxyProperties.getAuth().getPassword());
 				}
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
I couldn't get to test this change behind proxy/firewall settings yet. I tried to setup local proxy but not progressing much. 
I would like to submit this PR so that someone could test this out against proxy/firewall settings.

The proxy properties introduced in this PR have `maven.proxy` as the prefix.

spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/resolver/AetherModuleResolver.java
```diff
@@ -133,6 +182,13 @@ private DefaultRepositorySystemSession newRepositorySystemSession(RepositorySyst
 		LocalRepository localRepo = new LocalRepository(localRepoPath);
 		session.setLocalRepositoryManager(system.newLocalRepositoryManager(session, localRepo));
 		session.setOffline(this.offline);
+		if (isProxyEnabled()) {
+			DefaultProxySelector proxySelector = new DefaultProxySelector();
+			Proxy proxy = new Proxy(proxyProperties.getProtocol(), proxyProperties.getHost(),
+					Integer.parseInt(proxyProperties.getPort()), authentication);
```
ericbottard: Why not make the port property an int in the first place?

ericbottard:
Haven't tested either, but I have one question: why is authentication tied to proxy? Seems like we could benefit from authentication on the remote repo directly, no?

ilayaperumalg:
Good point. We can have authentication for remoteRepositories as well as the proxy (that also includes authentication). Only caveat with the remoteRepositories though is that, we need to have credentials mapped for each remote repository values provided. may be we can address that in a separate PR. I will add proxy configuration for the remoteRepositories and address the direct authentication for remote repo in a separate PR.

ilayaperumalg:
Pushed the changes that sets the authentication **only** for proxy settings.

mbogoevici:
Rebased, squashed, merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/b98d91da394d4d98dbf08e3cb1bd30455825c76d

PR #287, opened 2016-01-21
Module launcher JMX domain name

### LINKED ISSUE #286: Module launcher to use unique JMX domain name
The JMX MBeans exposed from the module launcher application needs to have unique JMX domain name so that they don't collide with any other default MBean names from a module's application context.

ilayaperumalg:
- Module launcher needs to use its own jmx domain name so that it doesn't collide with any other default domains regisetered.

This resolves #286


mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/a2486db9050a3ef21d276842e199f0f1365ff9ac

Issue #290, opened 2016-01-26
Make Tuple mutable + PropertyAccessor.write()


ericbottard:
This is to support e.g. adding new fields to the tuple in the case of the pmml processor (adding the results fields)

PR #291, opened 2016-01-27
Make Tuple mutable.

ericbottard:
This fixes spring-cloud/spring-cloud-stream#290
This resolves spring-cloud/spring-cloud-stream-modules#166

### DIFF
--- a/spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultMutableTuple.java
+++ b/spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultMutableTuple.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.tuple;
+
+import java.util.List;
+
+import org.springframework.core.convert.support.ConfigurableConversionService;
+
+/**
+ * Extension of {@link DefaultTuple} that also allows tuple {@link MutableTuple mutation}.
+ *
+ * @author Eric Bottard
+ */
+public class DefaultMutableTuple extends DefaultTuple implements MutableTuple {
+
+	public DefaultMutableTuple(List<String> names, List<Object> values, ConfigurableConversionService configurableConversionService) {
+		super(names, values, configurableConversionService);
+	}
+
+	@Override
+	public synchronized void setValue(int index, Object value) {
+		if (index < 0 || index >= size()) {
+			throw new IndexOutOfBoundsException();
+		}
+		values.set(index, value);
+	}
+
+	@Override
+	public synchronized void setValue(String name, Object value) {
+		int index = indexOf(name);
+		if (index != -1) {
+			setValue(index, value);
+		} else {
+			names.add(name);
+			values.add(value);
+		}
+	}
+
+
+}
--- a/spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultMutableTuple.java
+++ b/spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultMutableTuple.java
@@ -23,16 +23,18 @@ import org.springframework.core.convert.support.ConfigurableConversionService;
 /**
  * Extension of {@link DefaultTuple} that also allows tuple {@link MutableTuple mutation}.
  *
+ * <p>Note that this implementation is not threadsafe.</p>
+ *
  * @author Eric Bottard
  */
 public class DefaultMutableTuple extends DefaultTuple implements MutableTuple {
 
-	public DefaultMutableTuple(List<String> names, List<Object> values, ConfigurableConversionService configurableConversionService) {
+	/*package*/ DefaultMutableTuple(List<String> names, List<Object> values, ConfigurableConversionService configurableConversionService) {
 		super(names, values, configurableConversionService);
 	}
 
 	@Override
-	public synchronized void setValue(int index, Object value) {
+	public void setValue(int index, Object value) {
 		if (index < 0 || index >= size()) {
 			throw new IndexOutOfBoundsException();
 		}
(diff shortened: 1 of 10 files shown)


markpollack:
How about we keep an immutable option and in addition provide a new mutable one.

ericbottard:
Yep, thought about it, as an option on the builder. We'd want the default to be mutable though (or create a new mime type)

markpollack:
We should err on the side of keeping backwards comparability with XD to minimize change when users convert to the new spring cloud stream model.

ericbottard:
So are you advocating for a new mime-type for example? And when users want a mutable tuple, they'd explicitly ask for it?

mbogoevici:
How exactly can this affect backwards compatibility though - the old contract remains unchanged.

mminella:
+1 for allowing for an immutable option in some form.

ericbottard:
I added a way to create immutable Tuples on the builder

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultMutableTuple.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Eric Bottard
+ */
+public class DefaultMutableTuple extends DefaultTuple implements MutableTuple {
+
+	public DefaultMutableTuple(List<String> names, List<Object> values, ConfigurableConversionService configurableConversionService) {
+		super(names, values, configurableConversionService);
+	}
+
+	@Override
+	public synchronized void setValue(int index, Object value) {
```
pperalta: Unless reads are also `synchronized`, I'm not sure that making these methods `synchronized` is worthwhile. IMO, either the entire class needs to be made thread safe, or the JavaDoc should explicitly state that it isn't thread safe.

ericbottard: very good point

markpollack:
You are right Marius, it would not impact existing code, but it would break a rather important contract going forward.  The analogy in SI would be MutableMessage and Message.  I had though that perhaps one day this would also be a nice replacement to FieldSet in batch - since it was created before the conversion service.  The current set of tests - and in some cases behavior - is mostly derived from the FieldSet tests in batch.

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultMutableTuple.java
markpollack: I'd go for avoiding putting synchronized on it and stating it in the docs.  There could always be another sync'd variant created by the builder.

ericbottard:
Removed synchronized, Forced use of TupleBuilder (see also https://github.com/spring-cloud/spring-cloud-stream-modules/pull/168) and documented threadsafety

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultTuple.java
```diff
@@ -36,22 +36,23 @@
...
+	protected List<Object> values;
 
-	private transient ConfigurableConversionService configurableConversionService;
+	protected transient ConfigurableConversionService configurableConversionService;
 
-	private transient Converter<Tuple, String> tupleToStringConverter = new TupleToJsonStringConverter();
+	protected transient Converter<Tuple, String> tupleToStringConverter = new TupleToJsonStringConverter();
 
-	public DefaultTuple(List<String> names, List<Object> values, ConfigurableConversionService 
+	/*package*/ DefaultTuple(List<String> names, List<Object> values, ConfigurableConversionService
```
markpollack: Why change to package level protection?  Was thinking the hard enforcement of not having a ctor vs. only builder was a bit too restrictive.

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/JsonStringToTupleConverter.java
```diff
@@ -38,6 +38,7 @@ public JsonStringToTupleConverter() {
 
 	@Override
 	public Tuple convert(String source) {
+		System.out.println("Converting " + source);
```
markpollack: No system.out...

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/JsonStringToTupleConverter.java
```diff
@@ -70,7 +71,9 @@ else if (node.isNumber()) {
 		catch (Exception e) {
 			throw new RuntimeException(e);
 		}
-		return builder.build();
+		Tuple result = builder.build();
```
markpollack: Should just be `return builder.build()`

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/DefaultMutableTuple.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Extension of {@link DefaultTuple} that also allows tuple {@link MutableTuple mutation}.
+ *
+ * <p>Note that this implementation is not threadsafe.</p>
+ *
+ * @author Eric Bottard
+ */
+public class DefaultMutableTuple extends DefaultTuple implements MutableTuple {
+
+	/*package*/ DefaultMutableTuple(List<String> names, List<Object> values, ConfigurableConversionService configurableConversionService) {
```
markpollack: Similar comment as with protected ctor for DefaultTuple

spring-cloud-stream-tuple/src/main/java/org/springframework/cloud/stream/tuple/TupleBuilder.java
```diff
@@ -68,8 +70,22 @@
...
+	public static TupleBuilder immutableTuple() {
+		return new TupleBuilder(false);
+	}
+
+	/**
+	 * Return a new builder that will create {@link MutableTuple}.
+	 */
 	public static TupleBuilder tuple() {
-		return new TupleBuilder();
+		return new TupleBuilder(true);
```
markpollack: The default should be immutable, as it was before, with mutable being the new option.

markpollack:
I made changes and merged.  TubleBuilder.tuple() returns an immutable tuple like before.  I also made the ctors public, mostly concerned about backward compat, but we can discuss more.  Fixed the other minor issues.  

See 7fb834981ab6e987d1d0c784aebfbe642677e9a9

ericbottard:
We absolutely need a new mime type for mutable tuple then. One of the aim of this PR is to use mutable tuples with the pmml module (via `inputType=`)

markpollack:
This opens up another discussion, Mark F. mention some changes to how type conversion is done.  He can explain more.  We haven't had any perf complaints wrt to using immutable objects, so given that I'd suggest that we defer the goal of pmml using mutable tuples until we can plan out this change better.

markfisher:
My comment about type-conversion is that we should move away from representing Java types as custom MIME types altogether, but instead explicitly recognize the distinction between conversion to/from the *wire* vs. conversion before/after message handling (i.e. it should be a 2-step process: could be `text/plain` or `application/octet-stream` on the wire and either of those then convert to `Tuple` or `String` or `Foo`).

PR #294, opened 2016-01-28
Add durableSubscription as a binding property

### LINKED ISSUE #293: Create durableSubscription as a binding property
As a SCS user, I would like to set `durableSubscription` as a binding property so that the consumer property is set with the same when binding.

ilayaperumalg:
- This will be updated as a binder consumer property for the key `durableSubscription`

This resolves #293

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
```diff
@@ -243,6 +245,14 @@ public void setPartitioned(Boolean partitioned) {
 		this.partitioned = partitioned;
 	}
 
+	public Boolean getDurableSubscription() {
```
markfisher: shouldn't this be `isDurableSubscription`?

mbogoevici: I'll leave it as is for now, since we do `get` with other booleans in the same class. Maybe some cleanup after.

mbogoevici:
Rebased, merged w/ https://github.com/spring-cloud/spring-cloud-stream/commit/40709fbf140e0cf79dd3485fbff67e05b186bc15

Issue #278, opened 2016-01-18
Add binder SPI overview section to reference docs


sabbyanandan:
As a developer, I'd like to add overview section for binder SPIs, so it could be as reference for anyone ~~selecting and using the supported binders~~ interested in learning more about the Binders and the underpinning contracts. 

**Acceptance:**
* Overview section is available along with visuals to articulate core details
* All the available binder properties for override are included
* Binder properties for both SCS and SCDF developer experience are defined

ericbottard:
Isn't that already covered here: http://docs.spring.io/spring-cloud-stream/docs/1.0.0.BUILD-SNAPSHOT/reference/html/spring-cloud-stream-overview.html#_binder_selection ?

markfisher:
That existing doc doesn't cover anything about the `Binder` interface itself, nor the semantics for bound producers and consumers.

ericbottard:
Hmm, I think I got mislead by 
> be as reference for anyone selecting and using the supported binders.

Is this supposed to be an "end-user reference" (how to select binders) or a "binder implementor reference" (how to write a new `Binder`)?

markfisher:
I agree that particular bullet point sounds like it describes what is already documented. I don't think we're talking so much about a guide for implementors as an overview of the contract and description of the behavior.

markfisher:
That said, we also should be including the documentation of how that contract is mapped to each of the existing implementations, and that is relevant for users since they need to know what actually occurs within their broker, for monitoring etc.

mbogoevici:
Actually, I think we need to do even more than just the method contracts in terms of signatures. A significant amount of the Binder behaviour is captured in the Properties passed as arguments. We need to capture what changes between different implementations (if it does).

sabbyanandan:
@ericbottard: My bad. I updated the description/acceptance to reflect the missing links.

mbogoevici:
Also should document infrastructural things like health management infrastructure support (see https://github.com/spring-cloud/spring-cloud-stream/issues/233)

mbogoevici:
Fixed via https://github.com/spring-cloud/spring-cloud-stream/issues/516

PR #296, opened 2016-01-29
Add health indicators infrastructure

### LINKED ISSUE #233: Health indicator does not reflect actual connection status
Spring Cloud Stream probably should start adding new health indicators for its middleware connections, otherwise the user will think that the app is UP, when it might be DOWN. None of the configuration in the binder factories shows up in the actuator endpoints, so you can't really tell what's broken when it is.

mbogoevici:
Resolves #233

- Adds a generic 'bindersHealthIndicator' bean controlled by `management.health.binders.enabled` that tallies results from binders
- Binders are expected to expose one or more health indicators based on the status of the middleware connection. If no health indicators are exposed, the status is deemed to be 'UNKNOWN'
- Add support for Redis and Rabbit based on existing health indicators

Note: Kafka will require a separate health indicator - see https://github.com/spring-cloud/spring-cloud-stream/issues/297

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
@@ -26,6 +26,9 @@ import java.util.Map;
 import java.util.Properties;
 
 import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.boot.actuate.autoconfigure.ConditionalOnEnabledHealthIndicator;
+import org.springframework.boot.actuate.health.CompositeHealthIndicator;
+import org.springframework.boot.actuate.health.OrderedHealthAggregator;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.cloud.stream.binder.BinderConfiguration;
 import org.springframework.cloud.stream.binder.BinderFactory;
@@ -112,6 +115,13 @@ public class BinderFactoryConfiguration {
 			throw new BeanCreationException("Cannot create binder factory:", e);
 		}
 		return new DefaultBinderTypeRegistry(binderTypes);
+	}
+
+	@Bean
+	@ConditionalOnEnabledHealthIndicator("binders")
+	@ConditionalOnMissingBean(name = "bindersHealthIndicator")
+	public CompositeHealthIndicator bindersHealthIndicator() {
+		return new CompositeHealthIndicator(new OrderedHealthAggregator());
 	}
 
 
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAutoConfiguration.java
```diff
@@ -40,6 +45,7 @@
 @Configuration
 @ConditionalOnBean(ChannelBindingService.class)
 @EnableConfigurationProperties(DefaultPollerProperties.class)
+@AutoConfigureBefore(EndpointAutoConfiguration.class)
```
ilayaperumalg: Though I agree that it is better to create the `Channels Endpoint` defined in this configuration class before the `EndpointAutoConfiguration`, I don't see anything specific in this `EndpointAutoConfiguration` class that needs to be done before this configuration class. What am I missing?

Also, does this change anyway impact the way health indicators are aligned in this PR?

ilayaperumalg:
Please update the `copyright` year in some of the modified classes.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -114,6 +117,13 @@ public BinderTypeRegistry binderTypeRegistry(ConfigurableApplicationContext conf
 		return new DefaultBinderTypeRegistry(binderTypes);
 	}
 
+	@Bean
+	@ConditionalOnEnabledHealthIndicator("binders")
+	@ConditionalOnMissingBean(name = "bindersHealthIndicator")
```
ilayaperumalg: Is this needed only for the case of `AggregateApplication`? Since we don't `enable` web application for aggregate application, I wonder how health endpoints are used in there.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -114,6 +117,13 @@ public BinderTypeRegistry binderTypeRegistry(ConfigurableApplicationContext conf
 		return new DefaultBinderTypeRegistry(binderTypes);
 	}
 
+	@Bean
+	@ConditionalOnEnabledHealthIndicator("binders")
```
ilayaperumalg: Can we document this health indicator property `management.health.binders.enabled` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAutoConfiguration.java
mbogoevici: In the current state of the PR, this is not necessary, since the health infrastructure is created in a separate configuration - this is a leftover from an earlier approach. However, now that the question has been asked, I'd be much more keen on moving the health endpoint bean here. The rationale of having it close to the other Binder infrastructure beans was to have all the binder-related functionality in one place. However, I'd rather think that the autoconfig should have everything that relates to Boot autoconfig (including health endpoints).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
mbogoevici: No, this just follows the general pattern of other autoconfiguration beans. Also, endpoint usage is not conditional of enabling web endpoints.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
mbogoevici: Added a note to expand our documentation to cover it https://github.com/spring-cloud/spring-cloud-stream/issues/278#issuecomment-177990814.

mbogoevici:
Pushed an update.

ilayaperumalg:
LGTM. Squashed, rebased and merged as 7477fcb

PR #299, opened 2016-02-01
Add `-s .settings.xml` to `maven.config`

### LINKED ISSUE #298: Add .settings.xml to maven.config
All other spring-cloud-stream/spring-cloud-dataflow projects include `-s .settings.xml` in the `maven.config` file. We should add it to spring-cloud-stream as well, to make it easier to just build with `./mvnw clean install`

trisberg:
Resolves #298


markfisher:
Closing for now. We need to agree upon a cross-project plan that is consistent with other `spring-cloud` projects and then make those necessary changes (apparently that has already happened for `spring-cloud-task` via https://github.com/spring-cloud/spring-cloud-task/commit/f7a9f17236e42d51fcfc46d150e8ee078ca383c7)

PR #300, opened 2016-02-01
Add spring-rabbit explicitly to the dependencies BOM

dsyer:
It actually still has to be managed using a property in
spring-cloud-stream-binders-parent, but this change is for users.
Any user that wants an app using stream and rabbit needs to use the
spring-cloud-dependencies BOM. A gradle user has to do nothing else
with this change. A maven user still has to add an explicit
<spring-aqmp.version/> to the application POM.

### DIFF
--- a/spring-cloud-stream-dependencies/pom.xml
+++ b/spring-cloud-stream-dependencies/pom.xml
@@ -6,7 +6,7 @@
 		<artifactId>spring-cloud-dependencies-parent</artifactId>
 		<groupId>org.springframework.cloud</groupId>
 		<version>1.1.0.BUILD-SNAPSHOT</version>
-        <relativePath/>
+		<relativePath/>
 	</parent>
 	<artifactId>spring-cloud-stream-dependencies</artifactId>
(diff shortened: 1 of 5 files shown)


spring-cloud-stream-dependencies/pom.xml
```diff
@@ -16,18 +16,19 @@
 	<properties>
 		<rxjava.version>1.0.14</rxjava.version>
 		<spring-integration.version>4.2.4.RELEASE</spring-integration.version>
+        <spring-amqp.version>1.5.4.BUILD-SNAPSHOT</spring-amqp.version>
```
markfisher: indent

mbogoevici:
Will merge (with indentation fixes). Also, should remove workaround for the NPE bug when declaring queues.

mbogoevici:
Actually, an additional change is required for the starters, I believe. `mvn dependency:tree` still brings up `1.5.3.RELEASE` there.

dsyer:
Additional change made to starters (and rebased on master).

dsyer:
Plus I fixed the indentation to tabs only

PR #303, opened 2016-02-02
ModuleJarLauncher class loader to filter `classes` directory URLs

ilayaperumalg:
- When launching from tests, the module launcher adds `classes` directory in its system classloader URLs.
This results in triggering auto configuration/loading resources from the classes directory and causes issues
mentioned in https://github.com/spring-cloud/spring-cloud-dataflow/issues/362

### DIFF
--- a/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/utils/ClassloaderUtils.java
+++ b/spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/utils/ClassloaderUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,8 @@ package org.springframework.cloud.stream.module.utils;
 
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.List;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-module-launcher/src/main/java/org/springframework/cloud/stream/module/utils/ClassloaderUtils.java
```diff
@@ -50,16 +53,23 @@ public static ClassLoader createModuleClassloader(URL[] urls) {
 			// to compensate for LaunchedURLClassLoader not delegating to parent to retrieve resources
 			@SuppressWarnings("resource")
 			URLClassLoader systemUrlClassLoader = (URLClassLoader) systemClassLoader;
-			URL[] mergedUrls = new URL[urls.length + systemUrlClassLoader.getURLs().length];
+			List<URL> systemURLs = new ArrayList<>();
+			for (URL url: systemUrlClassLoader.getURLs()) {
+				// this will help avoiding any tests using the classes directory in the system class loader.
+				if (!url.getFile().endsWith("classes/")) {
```
mbogoevici: Can we perhaps control this through a configuration parameter/method argument - maybe with an override that preserves the default behaviour? Not worth investing much time in the configuration, but I don't think we should do exclusions based on a hardcoded path.

ilayaperumalg: @mbogoevici While I agree we shouldn't be doing exclusions on `hardcoded` path, I thought this specific change makes sense at least in the context of excluding the path that **ends** with `classes` directory in order to remove **non-jar** resources out of the path here (that happens only in case of test run). This is a non-issue on `runtime` environment where we wouldn't have path that ends with `classes/` and I believe all the URLs would end with `jar` or `jar` inside the `exec` jar. What do you think?

markfisher:
Closing as @ilayaperumalg mentioned on a call that this is not actually necessary.

PR #305, opened 2016-02-03
Allow underscore in kafka topic name

### LINKED ISSUE #217: KafkaMessageChannelBinder escapeTopicName and underscores
We have a set of Kafka topics that reflect table names from an Oracle database. These get auto-generated and we'd prefer not to change it on the Oracle producer side.

When we try something like this:
spring.cloud.stream.bindings.kafkaInputChannel=FOO_BAR

Spring is escaping this name so that you see FOO5FBAR or FOO_5FBAR, because it contains an underscore. See KafkaMessageChannelBinder#escapeTopicName.

Is there a way to disable this encoding, or express the topic name so that we can keep the underscore?

ilayaperumalg:
- Instead of escaping the topic name, perform validation that throws `RuntimeException` when the given kafka topic name doesn't meet the criteria set by Kafka.

This resolves #217

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -376,28 +376,21 @@ public class KafkaMessageChannelBinder extends MessageChannelBinderSupport {
 	}
 
 	/**
-	 * Allowed chars are ASCII alphanumerics, '.', '_' and '-'. '_' is used as escaped char in the form '_xx' where xx
-	 * is the hexadecimal value of the byte(s) needed to represent an illegal char in utf8.
+	 * Allowed chars are ASCII alphanumerics, '.', '_' and '-'.
 	 */
-	/*default*/
-	public static String escapeTopicName(String original) {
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -376,28 +376,21 @@ public void afterPropertiesSet() throws Exception {
...
 			for (byte b : utf8) {
-				if ((b >= 'a') && (b <= 'z') || (b >= 'A') && (b <= 'Z') || (b >= '0') && (b <= '9') || (b == '.')
-						|| (b == '-')) {
-					result.append((char) b);
-				}
-				else {
-					result.append(String.format("_%02X", b));
+				if (!((b >= 'a') && (b <= 'z') || (b >= 'A') && (b <= 'Z') || (b >= '0') && (b <= '9') || (b == '.')
+						|| (b == '-') || (b == '_'))) {
+					throw new RuntimeException("Topic name can only have ASCII alphanumerics, '.', '_' and '-'");
```
markfisher: `IllegalArgumentException` is probably better here

markfisher:
rebased, changed Exception type, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/8f795074e9983f471f92f179189ab31c61a4b3a0

PR #308, opened 2016-02-03
Add support for resetting offsets in the Kafka bus

mbogoevici:
- add two additional binder properties `startOffset` (with supported values `earliest` and `latest`) and `resetOffsets` that allow the binder to reset consumers to the earliest and latest offsets in the bus. 
- ensure that environment is merged correctly
- tests 

TBD: provide configuration options per-binding (separate issue)

In summary, allows users to start a Kafka-bound app with the following options:

````
java -jar log-sink/target/log-sink-1.0.0.BUILD-SNAPSHOT-exec.jar --spring.cloud.stream.bindings.input.destination=mydestination --spring.cloud.stream.bindings.input.group=group1 --server.port=9003 --spring.cloud.stream.binder.kafka.resetOffsets=true --spring.cloud.stream.binder.kafka.startOffset=earliest
````
(`latest` is also supported for `startOffset`)

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -153,6 +153,10 @@ public class KafkaMessageChannelBinder extends MessageChannelBinderSupport {
 
 	private static final boolean DEFAULT_AUTO_COMMIT_ENABLED = true;
 
+	private static final boolean DEFAULT_RESET_ON_START = false;
+
+	private static final StartingOffset DEFAULT_START = StartingOffset.latest;
+
 	private RetryOperations retryOperations;
 
 	/**
@@ -273,6 +277,10 @@ public class KafkaMessageChannelBinder extends MessageChannelBinderSupport {
 	private int offsetUpdateShutdownTimeout = 2000;
 
 	private Mode mode = Mode.embeddedHeaders;
+
+	private boolean resetOnStart = false;
+
+	private StartingOffset startingOffset = StartingOffset.latest;
 
 	public KafkaMessageChannelBinder(ZookeeperConnect zookeeperConnect, String brokers, String zkAddress,
 			 String... headersToMap) {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: spring-cloud-stream-samples/source/pom.xml [hunk 1]
```diff
@@ -24,7 +24,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-stream-binder-redis</artifactId>
+			<artifactId>spring-cloud-stream-binder-kafka</artifactId>
```
markfisher: will revert to 'redis' while merging

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -153,6 +153,10 @@
 
 	private static final boolean DEFAULT_AUTO_COMMIT_ENABLED = true;
 
+	private static final boolean DEFAULT_RESET_ON_START = false;
```
markfisher: these 2 constants are not used, will fix on merge

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/8271478e2775a1b1532dc7479a96e3bb50baec1b along with https://github.com/spring-cloud/spring-cloud-stream/commit/252aee40795ea939232ff42c57b23f02bb622318

garyrussell:
FYI "earliest" seems to be a misnomer - at least with 0.9.0.1 it means firstUnconsumed.

See http://stackoverflow.com/questions/35830327/spring-cloud-stream-kafka-offline-consumer-messages-lost/35831670#35831670

mbogoevici:
@garyrussell I can't speak fully to it, since I haven't tested this with 0.9 yet, but `earliest` and `latest` are the values used by `auto.offset.reset` to indicate the start and end of a topic/partition set, even with 0.9, replacing the previously used `smallest` and `largest` - see http://kafka.apache.org/documentation.html#newconsumerconfigs. 

Of note however, that if a group is specified and offsets for the group are already saved then the `start-offset` value is ignored (we need to document this, just as we need to document all binder options, but in essence it is a hint when a group is specified) and the consumer will resume from the first unconsumed message. A client may specifically request to reset its offset by specifying `reset-offsets=true`.

garyrussell:
>I would expect issues in a 0.8/0.9 mix

From the kafka doc:

>To ensure a smooth upgrade path for users, we still maintain the old 0.8 consumer clients that continue to work on an 0.9 Kafka cluster. 

My observations were empirical - I didn't dig into the code.

>more details on how did you test?

A simple boot app...

````
@EnableBinding(Sink.class)
public class Test {

	@ServiceActivator(inputChannel = Sink.INPUT)
	@Bean
	public MessageHandler handler() {
		return new MessageHandler() {

			@Override
			public void handleMessage(Message<?> message) throws MessagingException {
				System.out.println(new String((byte[]) message.getPayload()) + ": " + message.getHeaders());
			}

		};
	}

}
````

with

````
spring.cloud.stream.bindings.input.destination=foo
spring.cloud.stream.bindings.input.group=foogroup
spring.cloud.stream.binder.kafka.mode=raw
#spring.cloud.stream.binder.kafka.start-offset=earliest
````

Post some messages with the console producer.

Started the app - no output; post some more; output received.

Stop the app; post some more messages.

Uncomment earliest

start the app;

I just see the new messages.


However, I am getting different results today; the app starts at the unconsumed point regardless of the start-offset - weird.

garyrussell:
Working as you described in your SO answer; must've been pilot error yesterday - teach me to work on Sundays.

mbogoevici:
I believe it has a lot to do with lack of documentation :)

Issue #314, opened 2016-02-05
Add installation of kafka for travis build


ericbottard:
See https://github.com/spring-cloud/spring-cloud-dataflow/issues/333, but for s-c-s

PR #315, opened 2016-02-05
Use kafka on the travis infrastructure

ericbottard:
Fixes spring-cloud/spring-cloud-stream#314

* Adds setup commands for kafka on travis
* re-enables tests on PR branch builds
* as a consequence of that, allows a bit more time (currently x3 and x10, x3 & x3 did not work) for tests that interact with the binder on travis, as it seems to be way slower than what we're usually used to

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
@@ -33,6 +33,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
 
+import kafka.api.OffsetRequest;
 import org.junit.ClassRule;
 import org.junit.Test;
 
@@ -53,8 +54,6 @@ import org.springframework.integration.kafka.support.ZookeeperConnect;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.support.GenericMessage;
-
-import kafka.api.OffsetRequest;
 
 
 /**
@@ -186,7 +185,7 @@ public class KafkaBinderTests extends PartitionCapableBinderTests {
 		// Let the consumer actually bind to the producer before sending a msg
 		binderBindUnbindLatency();
 		moduleOutputChannel.send(message);
-		Message<?> inbound = moduleInputChannel.receive(2000);
+		Message<?> inbound = moduleInputChannel.receive((int)(2000 * timeoutMultiplier));
 		assertNotNull(inbound);
 		assertArrayEquals(ratherBigPayload, (byte[]) inbound.getPayload());
 		Collection<Partition> partitions = binder.getCoreBinder().getConnectionFactory().getPartitions(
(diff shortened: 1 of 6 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -134,6 +135,7 @@ public Object receive(boolean expectNull) throws Exception {
 
 	@Test(expected = IllegalArgumentException.class)
 	public void testValidateKafkaTopicName() {
+		Assert.fail("Should fail now");
```
mbogoevici: This is a debug statement I suppose - since it will always fail the test?

ericbottard: Yes, this is to make sure Tests are triggered

Made me discover they were not o.O

Did not expect anyone to jump on this PR so quick. Still in flux, but obviously I need it to be in PR form

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -186,7 +185,7 @@ public void testCustomPartitionCountOverridesDefaultIfLarger() throws Exception
 		// Let the consumer actually bind to the producer before sending a msg
 		binderBindUnbindLatency();
 		moduleOutputChannel.send(message);
-		Message<?> inbound = moduleInputChannel.receive(2000);
+		Message<?> inbound = moduleInputChannel.receive((int)(2000 * timeoutMultiplier));
```
mbogoevici: While this improves the configurability of the test (which is good) - it adds something that will need to be copied in any newly added test (not that the hardcoded values were all that good to begin with). The first impulse is to perhaps isolate the constant and the multiplier in a utility method available to testers - e.g. receive(PollableChannel).

ericbottard: The constant is not ... constant.
Even though in a given test it is often (arbitrarily) the same value (2000 in the snippet here), it could be different.

So really, should be `receiveWith(channel, baseAmount)`. Do we gain anything?

mbogoevici: Yeah, I know - but I was hoping we could pick up a value. Just looking through all of them, there doesn't seem to be any criterion by which those values are set - in most cases it's 1s and they're just bumped up for certain tests which I suppose were the ones that happened to fail at some point - e.g. Kafka tends to have larger values. But we already introduced the concept of per-binder test multiplier, so we can just start with different default values - wdyt?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -53,6 +54,14 @@
  */
 public class RawModeKafkaBinderTests extends KafkaBinderTests {
 
+	@Before
+	public void init() {
+		String multiplier = System.getenv("KAFKA_TIMEOUT_MULTIPLIER");
```
mbogoevici: Needs to be moved to superclass, which currently runs with defaults.

mbogoevici: Can fix on merge.

mbogoevici:
Rebased, squashed, merged w/minor changes https://github.com/spring-cloud/spring-cloud-stream/commit/412b1dd2945992c2959d6cc870323642527d6918

Issue #151, opened 2015-10-13
Acknowledgement for Sending Message


VikasJindal84:
Hi,

Currently there is no way to find out if Kafka message is successfully delivered . I tried to use 
MessageChannel.send , it always returns true ,even if zookeeper and kakfa server is down.

So, my requirement is to have a way to get a confirmation about delivery of message to kafka server.

mbogoevici:
Hi Vikas,

Thanks for reporting this. Noted. This needs to be an enhancement to Spring Integration Kafka, as well as properly handling this inside Spring Cloud Stream.

Cheers,
Marius

artembilan:
Looks like this should be fixed together with the https://github.com/spring-projects/spring-integration-kafka/issues/62

mbogoevici:
It would depend on it, but more to it than just that.

VikasJindal84:
Hi,

Now spring-integration-kafka has support for ProducerListener : https://github.com/spring-projects/spring-integration-kafka/commit/325d0b294c825252f5313bd9696229d198b0d38e

Any plan for providing support for ProducerListener in spring-cloud-stream-binder-kafka ?

mbogoevici:
This is currently addressed by https://github.com/spring-cloud/spring-cloud-stream/issues/316

sabbyanandan:
This is already addressed by spring-cloud/spring-cloud-stream#253, so closing this issue.

PR #316, opened 2016-02-05
Upgrade Spring Integration Kafka to 1.3 in Kafka binder

### LINKED ISSUE #253: Upgrade Spring Integration Kafka to 1.3

sobychacko:
resolves #253 

Upgrade scala/SIK versions in mvn pom

Migrage to KafkaNativeOffsetManager from KafkaTopicOffsetManager

Introduce LoggingProducerListener for errors - See GH #151

Remove properties specific to the Topic offset manager

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -55,15 +55,8 @@ import org.springframework.integration.kafka.core.DefaultConnectionFactory;
 import org.springframework.integration.kafka.core.Partition;
 import org.springframework.integration.kafka.core.ZookeeperConfiguration;
 import org.springframework.integration.kafka.inbound.KafkaMessageDrivenChannelAdapter;
-import org.springframework.integration.kafka.listener.Acknowledgment;
-import org.springframework.integration.kafka.listener.KafkaMessageListenerContainer;
-import org.springframework.integration.kafka.listener.KafkaTopicOffsetManager;
-import org.springframework.integration.kafka.listener.OffsetManager;
-import org.springframework.integration.kafka.support.KafkaHeaders;
-import org.springframework.integration.kafka.support.ProducerConfiguration;
-import org.springframework.integration.kafka.support.ProducerFactoryBean;
-import org.springframework.integration.kafka.support.ProducerMetadata;
-import org.springframework.integration.kafka.support.ZookeeperConnect;
+import org.springframework.integration.kafka.listener.*;
+import org.springframework.integration.kafka.support.*;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
@@ -124,6 +117,7 @@ import scala.collection.Seq;
  * @author David Turanski
  * @author Gary Russell
  * @author Mark Fisher
+ * @author Soby Chacko
  */
 public class KafkaMessageChannelBinder extends MessageChannelBinderSupport {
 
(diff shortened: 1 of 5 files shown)


artembilan:
For the `KafkaNativeOffsetManager` upgrade consider to pay attention to the issue: http://stackoverflow.com/questions/34538314/spring-integration-kafka-slow-message-processing-when-using-kafkanativeoffsetman.

Sorry haven't found the mind to figure out the problem there...

mbogoevici:
I haven't seen that issue before, we should definitely investigate it ...

sobychacko:
I will take a look at that issue.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -55,15 +55,8 @@
...
-import org.springframework.integration.kafka.listener.KafkaMessageListenerContainer;
-import org.springframework.integration.kafka.listener.KafkaTopicOffsetManager;
-import org.springframework.integration.kafka.listener.OffsetManager;
-import org.springframework.integration.kafka.support.KafkaHeaders;
-import org.springframework.integration.kafka.support.ProducerConfiguration;
-import org.springframework.integration.kafka.support.ProducerFactoryBean;
-import org.springframework.integration.kafka.support.ProducerMetadata;
-import org.springframework.integration.kafka.support.ZookeeperConnect;
+import org.springframework.integration.kafka.listener.*;
+import org.springframework.integration.kafka.support.*;
```
artembilan: No asterisk imports at all.
You can configure your IDE on the matter.
Not sure if Spring Cloud team has some configs, but you can consider to use the SF's one: https://github.com/spring-projects/spring-framework/tree/master/src/eclipse

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -715,18 +674,12 @@ private KafkaMessageListenerContainer createMessageListenerContainer(KafkaProper
 
 	private OffsetManager createOffsetManager(String group, long referencePoint) {
 		try {
-			KafkaTopicOffsetManager kafkaOffsetManager =
-					new KafkaTopicOffsetManager(zookeeperConnect, offsetStoreTopic, Collections.<Partition,
+
+			KafkaNativeOffsetManager kafkaOffsetManager =
+					new KafkaNativeOffsetManager(connectionFactory, zookeeperConnect,Collections.<Partition,
```
artembilan: A whitespace after `,` and I don't like that generics rip :smile:

artembilan:
I don't see any test changes, but I guess this high change should just demonstrate non-breakage in the test stuff :smile: .

That's all from my side.

Have a good weekend!

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
sobychacko: sorry, that was an IDE configuration. Will fix it. Thnx!!

mbogoevici:
One reason why the KafkaNativeOffsetManager may be slower than the KafkaTopicOffsetManager is the synchronous vs/asynchronous nature of the updates (synchronous in the former case, asynchronous in the latter case). I believe we can wrap the KafkaNativeOffsetManager to operate asynchronously by default, as long as we ensure a proper flush on shutdown. In any case, the WindowingOffsetManager wrapper will significantly reduce the number of calls to the actual offset manager per total number of messages, to the point that the differences become insignificant.

mbogoevici:
Coming back to the issue raised by @artembilan, the `KafkaNativeOffsetManager` does indeed impact performance significantly, due to the relatively expensive blocking calls made by `SimpleConsumer.commitOffset()`. This can be observed by running https://github.com/mbogoevici/spring-integration-kafka/blob/perftest/src/test/java/org/springframework/integration/kafka/performance/OffsetManagerPerformanceTests.java

At the same time though, introducing windowing, as the binder does, eliminates the performance penalty if the update window is kept reasonably large (e.g. every second or so), as shown below (100000 updates with a 1s update window - default is 10s).

````
KafkaTopicOffsetManager with windowing completed in 0.052s
KafkaNativeOffsetManager with windowing completed in 0.03s
KafkaNativeOffsetManager completed in 9.404s
KafkaTopicOffsetManager completed in 0.417s
````
I would argue that the use of the native offset manager is less liable to produce a performance impact in the binder than thought and it provides a better upgrade path (since the committed offsets will be present when upgrading to 0.9).

mbogoevici:
As a corollary, we should think of porting the windowing functionality to Spring Integration Kafka, ideally not RxJava based (so that we can avoid that dependency).

artembilan:
@mbogoevici , thank you very much for the investigation!

So, you're good here with the `KafkaNativeOffsetManager` if you are really wrapped to the `WindowingOffsetManager`.

I'd be glad to contribute to the SIK with the `WindowingOffsetManager` utilization and migration to the Reactor-2.5 for it, bypassing RxJava.
Plus that your test-case look good to be included to the Framework as well.

mbogoevici:
Rebased, merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/7138f125eecf8a36848f5c411c383190f63a4f4d + minor adjustments in https://github.com/spring-cloud/spring-cloud-stream/commit/c4a64b37fba8a42c0d4cdab65f20e6d3d0695944

PR #318, opened 2016-02-08
Rabbit binder environment should be named rabbitmq

donovanmuller:
In the context of trying to configure Rabbit, the docs mention that the Rabbit config properties be added under `rabbit`:

```
      ...
      binders:
        rabbit1:
          type: rabbit
          environment:
            spring:
              rabbit: <-- here
                host: <host1>
      ...
```

however, the properties should be configured under `rabbitmq` as reference in the [Spring Boot Appendix A](https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html)

### DIFF
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -284,13 +284,13 @@ spring:
           type: rabbit
           environment:
             spring:
-              rabbit:
+              rabbitmq:
                 host: <host1>
         rabbit2:
           type: rabbit
           environment:
             spring:
-              rabbit:
+              rabbitmq:
                 host: <host2>
 ----
 


ilayaperumalg:
@donovanmuller thanks for catching this and the PR. 
In fact, Spring Cloud Stream uses its [own] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/SpringRabbitMQProperties.java#L25) RabbitMQProperties which also has the same properties that boot uses with prefix `spring.rabbitmq`.

Could you confirm if you have signed the [contributor agreement] (https://support.springsource.com/spring_committer_signup)? Would be happy to merge once that is done.

ilayaperumalg:
@donovanmuller thanks for catching this and the PR. 
In fact, Spring Cloud Stream uses its [own] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/SpringRabbitMQProperties.java#L25) RabbitMQProperties which also has the same properties that boot uses with prefix `spring.rabbitmq`.

Could you confirm if you have signed the [contributor agreement] (https://support.springsource.com/spring_committer_signup)? Would be happy to merge once that is done.

donovanmuller:
@ilayaperumalg Good to know :+1: 
Have signed (161920160208070017)

ilayaperumalg:
LGTM, merged. @donovanmuller thanks again!

PR #328, opened 2016-02-09
Refactor Kafka binder configuration

### LINKED ISSUE #332: Kafka Binder configuration/properties
Currently, KafkaBinderConfiguration and properties are combined in the same configuration class which makes it weird to include the configuration class to refer to the properties.

Also, the defaultProperties bean injected in KafkaServiceAutoConfiguration class can be simplified to include locations attribute on the ConfigurationProperties annotation.

ilayaperumalg:
- Move KafkaBinderConfigurationProperties out of the configuration class
 - Remove defaultProperties bean creation and use `locations` in `ConfigurationProperties` to load the default properties

This resolves #332


mbogoevici:
As discussed with @ilayaperumalg, we should hold onto this until https://github.com/spring-cloud/spring-cloud-stream/pull/316 is merged.

ilayaperumalg:
Rebased after the #316 merge.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfigurationProperties.java
```diff
@@ -17,110 +17,136 @@
...
-	
-	private boolean autoCommitEnabled;
-	
-	private int fetchSize;
-	
-	private int minPartitionCount;
-	
-	private int queueSize;
+@ConfigurationProperties(prefix = "spring.cloud.stream.binder.kafka",
+		locations = "classpath:/META-INF/spring-cloud-stream/kafka-binder.properties")
```
mbogoevici: We can remove `kafka-binder-lattice.properties` too I guess.

ilayaperumalg: That was already removed via [this] (https://github.com/spring-cloud/spring-cloud-stream/commit/7ffe9070e3957f5097e8bb76289275a456c61dd8).

ilayaperumalg:
I believe `PropertyPlaceholderAutoConfiguration` isn't needed as it was used to resolve the properties loaded from the file explicitly. Now that we use `@ConfigurationProperties`'s locations and that removes the need of the `PropertyPlaceholderAutoConfiguration`. Do you see any issue without this?

mbogoevici:
Added a few changes here: https://github.com/mbogoevici/spring-cloud-stream/commit/d2bd28dd34defb41889cfc52249ec8542d00a694. @ilayaperumalg let me know what you think

ilayaperumalg:
@mbogoevici your changes look good. I created [this] (https://github.com/spring-projects/spring-boot/issues/5135) spring-boot issue that sets the lowest precedence level for the property source loaded from `@ConfigurationProperties`' locations which is the reason for requiring `PropertyPlaceholderConfiguration` in our case explicitly.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/47d1d69402c1a6564adcfbc1fb34cd8a0a2a09a2

PR #331, opened 2016-02-10
Add Kafka binder health indicator

### LINKED ISSUE #297: Create health indicator for Kafka binder
Since there's no native health indicator for Kafka, we shall create one that reflects the status of the current connections, i.e. is healthy if the all the leaders for all the partitions connected to by the binder are up. Optionally this can be done with a composite health indicator.

ilayaperumalg:
- Health indicator fetches broker addresses using ZK configuration and matches it with the leaders of the paritions being used in the binder

This resolves #297


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -161,31 +161,7 @@
...
-		public Object deserialize(byte[] bytes) throws ZkMarshallingError {
-			try {
-				return new String(bytes, "UTF-8");
-			}
-			catch (UnsupportedEncodingException e) {
-				throw new ZkMarshallingError(e);
-			}
-		}
-	};
+	private Map<String, Collection<Partition>> topicsInUse = new HashMap<>();
```
ilayaperumalg: Given binding lifecycle is same for all the bindings in the context, I believe it is ok not removing the `topics` from this map when the `unbind` operation occurs.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -0,0 +1,74 @@
...
+	public Health health() {
+		ZkClient zkClient = new ZkClient(binder.getZkAddress(), 10000, 10000, ZKStringSerializer$.MODULE$);
+		Seq<Broker> allBrokersInCluster = ZkUtils$.MODULE$.getAllBrokersInCluster(zkClient);
+		Collection<Broker> brokersInCluster = JavaConversions.asJavaCollection(allBrokersInCluster);
+		Set<String> brokersInClusterSet = new HashSet<>();
+		for (Broker broker: brokersInCluster) {
+			brokersInClusterSet.add(broker.connectionString());
+		}
+		List<String> downMessages = new ArrayList<>();
+		for (Map.Entry<String, Collection<Partition>> entry : binder.getTopicsInUse().entrySet()) {
```
ilayaperumalg: Explicit verification using the `broker addresses` returned by the binder's `topics` is needed to avoid `hostname` issues with what user specified as `broker` addresses and what ZK client returned as broker addresses in cluster.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -0,0 +1,74 @@
...
+			for (Partition partition : entry.getValue()) {
+					BrokerAddress address = binder.getConnectionFactory().getLeader(partition);
+					if (!brokersInClusterSet.contains(address.toString())) {
+						downMessages.add(address.toString());
```
mbogoevici: We should only generate a message per broker - i.e. use a Set for storing error messages.

ilayaperumalg: Good point. fixed and pushed the change.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaMessageChannelBinderConfiguration.java
```diff
@@ -113,6 +115,11 @@ KafkaMessageChannelBinder kafkaMessageChannelBinder() {
 		return kafkaMessageChannelBinder;
 	}
 
+	@Bean
```
mbogoevici: Should move to the main configuration file after the refactoring.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -0,0 +1,72 @@
...
+
+	private final KafkaMessageChannelBinder binder;
+
+	public KafkaBinderHealthIndicator(KafkaMessageChannelBinder binder) {
+		this.binder = binder;
+	}
+
+	@Override
+	public Health health() {
+		ZkClient zkClient = new ZkClient(binder.getZkAddress(), 10000, 10000, ZKStringSerializer$.MODULE$);
```
mbogoevici: Should make ZK connectivity parameters configurable (optional).

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -0,0 +1,72 @@
...
+		Seq<Broker> allBrokersInCluster = ZkUtils$.MODULE$.getAllBrokersInCluster(zkClient);
```
mbogoevici: We should catch the exception thrown if ZK itself is down and return that as an error message. (ZK down definitely means that Kafka is down, too)

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
ilayaperumalg: I believe the ZK address always needs to be obtained from what is set inside the `binder` and hence it is not an option at the health indicator level?

ilayaperumalg:
Updated the PR to handle exception when connecting to ZK.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
mbogoevici: Sorry for the confusion, I meant the timeouts.

ilayaperumalg: No problem :-). Pushed the changes to make ZK session/connection timeout properties configurable at the binder so that the same can be used for health indicator. Also, rebased the PR.

ilayaperumalg:
Updated the PR to rename the ZK properties and also moved those properties to `KafkaBinderConfigurationProperties` instead of the default properties.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -452,6 +460,8 @@ public void setStartOffset(StartOffset startOffset) {
 
 		Collection<Partition> partitions = ensureTopicCreated(name, numPartitions, defaultReplicationFactor);
 
+		topicsInUse.put(name, partitions);
```
mbogoevici: needs to be done for consumer, too

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderHealthIndicator.java
```diff
@@ -0,0 +1,88 @@
...
+	public KafkaBinderHealthIndicator(KafkaMessageChannelBinder binder) {
+		this.binder = binder;
+	}
+
+	@Override
+	public Health health() {
+		ZkClient zkClient = new ZkClient(binder.getZkAddress(), binder.getZkSessionTimeout(),
+				binder.getZkConnectionTimeout(), ZKStringSerializer$.MODULE$);
+		Set<String> brokersInClusterSet = new HashSet<>();
+		try {
```
mbogoevici: try needs to wrap the client creation as well.

mbogoevici:
Squashed, rebased, merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/0fb431db3ae326cb6d518c080ceac6297cc8d4c4 with additional changes in https://github.com/spring-cloud/spring-cloud-stream/commit/b1b0e960d78d77c830fb02acbdbe88d476b3a25f

PR #333, opened 2016-02-10
Make history tracker properties configurable

### LINKED ISSUE #252: Make message track history properties configurable
As a user, I would like to configure what properties are updated in message track history header.

ilayaperumalg:
- Add property `spring.cloud.stream.bindings.<channelName>.trackHistoryProperties` with comma separated string values that specify the exact
property names in `ChannelBindingServiceProperties` to be added in the message tracker header
 - Add tests

This resolves #252

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageHistoryTrackerConfigurer.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageHistoryTrackerConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,11 @@ package org.springframework.cloud.stream.binding;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
 
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.stream.config.BindingProperties;
 import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
 import org.springframework.integration.channel.ChannelInterceptorAware;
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
```diff
@@ -53,6 +53,8 @@
 
 	private Boolean trackHistory;
 
+	private String trackHistoryProperties = "all";
```
mbogoevici: Maybe we should call this 'trackedProperties' only?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageHistoryTrackerConfigurer.java
```diff
@@ -40,17 +42,31 @@
...
+			MessageBuilderFactory messageBuilderFactory) {
 		this.channelBindingServiceProperties = channelBindingServiceProperties;
+		this.messageBuilderFactory = messageBuilderFactory;
 	}
 
 	@Override
 	public void configureMessageChannel(MessageChannel messageChannel, String channelName) {
 		BindingProperties bindingProperties = channelBindingServiceProperties.getBindings().get(channelName);
 		if (bindingProperties != null && Boolean.TRUE.equals(bindingProperties.isTrackHistory())) {
+			final Set<String> trackHistoryProperties = StringUtils.commaDelimitedListToSet(bindingProperties.getTrackHistoryProperties());
```
mbogoevici: We should probably check for an empty array of tracked properties too. What should be the behaviour in that case?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
ilayaperumalg: +1

mbogoevici: Ok, will do the changes on merge.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageHistoryTrackerConfigurer.java
ilayaperumalg: In this case, the default track properties (thread, timestamp) will only be available. I will add a test case for this.

ilayaperumalg:
Rebased and updated the PR after addressing the comments.

mbogoevici:
Rebases, squashed, merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/81e6769b744be2c6554a8537a4c2badf85e1e05e

mbogoevici:
Will close as is. Documentation to be added separately.

PR #341, opened 2016-02-11
Fix autowiring CompositeHealthIndicator

### LINKED ISSUE #339: Health indicator autowiring breaks existing apps
This line in `DefaultBinderFactory` breaks existing apps (e.g. anyone using `@EnableDiscoveryClient`):

```
	@Autowired(required = false)
	public void setBindersHealthIndicator(CompositeHealthIndicator bindersHealthIndicator) {
		this.bindersHealthIndicator = bindersHealthIndicator;
	}
```

it is making a false assumption that the only bean of type `CompositeHealthIndicator` in the context is the one that belongs to the binders. I suggest maybe using a subclass instead, or a qualifier, to be more precise about the type.

ilayaperumalg:
- The setter method `setBindersHealthIndicator(CompositeHealthIndicator bindersHealthIndicator)` is sufficient enough to autowire the candidate bean with the name `bindersHealthIndicator` based on the hint at the method argument.
   This gets interesting when the health indicator is disabled via `management.health.binders.enabled` and thereby the `bindersHealthIndicator` bean is not instantiated. In this case, if there are any matching beans of the type
`CompositeHealthIndicator` then that competes for the autowiring in the setter.

 - Add `Qualifier` with the name `bindersHealthIndicator` which makes the autowiring happen only with the bean named `bindersHealthIndicator` and thereby no `type` matching
 - Add test that demonstrates this uses case

This resolves #339

### DIFF
--- a/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/HealthIndicatorsConfigurationTests.java
+++ b/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/HealthIndicatorsConfigurationTests.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.stream.binder;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/HealthIndicatorsConfigurationTests.java
```diff
@@ -73,6 +82,29 @@ public void healthIndicatorsCheck() throws Exception {
...
+		Binder binder2 = context.getBean(BinderFactory.class).getBinder("binder2");
+		assertThat(binder2, instanceOf(StubBinder2.class));
+		boolean exceptionThrown = false;
+		try {
+			context.getBean("bindersHealthIndicator", CompositeHealthIndicator.class);
+		}
+		catch (NoSuchBeanDefinitionException e) {
+			exceptionThrown = true;
+		}
+		assertTrue(exceptionThrown);
```
mbogoevici: We should probably add a message explaining the nature of the failure, or perhaps even swallow the exception but add a fail() in the try clause.

mbogoevici:
Merged w/minor updates https://github.com/spring-cloud/spring-cloud-stream/commit/528b90b6b5bbaf4d160eec3958908ea6d1a62789

PR #342, opened 2016-02-11
Binder API simplification

### LINKED ISSUE #330: Extract Binding interface
Refactor the Binder API, removing the `unbind` method and moving it to the Binding. Also, make `Binding` an interface with just the `unbind` methods and have the invokers of `Binder.unbind()` call it directly instead.

mbogoevici:
Resolves #330

- removes `unbind` from the binder and moves it to the `Binding` instance itself;
- `Binding` is now an interface with a default implementation provided by SCS
- Removed old and unused code


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -192,7 +193,8 @@ private boolean isPollable(Class<?> channelType) {
 	}
 
 	private void bridgeSubscribableToPollableChannel(SubscribableChannel sharedChannel, MessageChannel inputChannel) {
-		sharedChannel.subscribe(new MessageChannelBinderSupport.DirectHandler(inputChannel));
+		BridgeHandler bridgeHandler = new BridgeHandler();
```
ilayaperumalg: bridgeHandler is not used here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DirectHandler.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.MessagingException;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class DirectHandler implements MessageHandler {
+
+	private final MessageChannel outputChannel;
```
ilayaperumalg: The name `outputChannel` here could be misleading. Can we just have it as a `messageChannel` or something?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private final Type type;
+
+	private final DefaultBindingPropertiesAccessor properties;
+
+	private final AbstractBinder<T> abstractBinder;
+
+	private DefaultBinding(String name, String group, T target, AbstractEndpoint endpoint, Type type,
+						   DefaultBindingPropertiesAccessor properties, AbstractBinder<T> abstractBinder) {
+		Assert.notNull(target, "target must not be null");
```
ilayaperumalg: Would be good to add `notNull` check on `abstractBinder` as well.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -31,141 +27,33 @@
...
+	/**
+	 * Returns the binding type.
+	 * @return the binding type
+	 */
+	Type getType();
+
+	/**
+	 * Unbinds the target component represented by this instance and stop any active components. Implementations must
+	 * be idempotent. After this method is invoked, the target is not expected to receive any message, this instance
+	 * should be discarded, and a new Binding should be created instead,
```
ilayaperumalg: typo at the end: comma

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -373,60 +353,25 @@ public void afterPropertiesSet() throws Exception {
...
 	}
 
-	protected void stopBindings() {
-		for (Lifecycle bean : this.bindings) {
+	@Override
+	public void destroy() throws Exception {
+		for (Binding<?> binding : this.bindings) {
 			try {
-				bean.stop();
+				binding.unbind();
```
ilayaperumalg: The `stop` lifecycle operation on the Input/OutputBindingLifecycle beans would also invoke the `unbind()` operation on the bindings. Given the `binder` also needs to be notified on `unbind` to perform `afterUnbind` operations, may be the `unbind` on Input/OutputBindingLifecycle beans can be removed?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
mbogoevici: Oops. Will fix.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DirectHandler.java
mbogoevici: I don't see how it can be. It's the output channel of the handler. Should probably add some JavaDoc (this was moved verbatim from the `MessageChannelBinderSupport`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
mbogoevici: I don't think that would be good. For one thing, the Input/OutputBindingLifecycle allows binding/unbinding whenever the application is started or stopped (which in principle can happen multiple times during its lifecycle). This is redundant and needs to be removed on the long run, however it currently ensures that dynamic bindings are unbound as well (we need to fix that).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DirectHandler.java
markfisher: This is functionally equivalent to `BridgeHandler`, but more efficient.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -0,0 +1,112 @@
...
+	public String getGroup() {
+		return group;
+	}
+
+	@Override
+	public T getTarget() {
+		return target;
+	}
+
+	public AbstractEndpoint getEndpoint() {
```
markfisher: I feel like we should not expose this. Other than where `Binder` impls call `getEndpoint().start()`, it appears to be only used in tests (in some cases unnecessarily). I think it would be cleaner to have the equivalent of `start()` directly on the `Binding` interface so that the endpoint can remain hidden (might be easier when we adapt `Binding` to other scenarios in the future). Of course, if we DO add that method to `Binding`, it might make sense to rename `unbind()` so they are symmetrical.

mbogoevici: This is part of the implementation though. I would prefer not to have start on the Binding itself. I don't think that the client should be able to start and stop bindings at their discretion.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -0,0 +1,112 @@
...
+
+	@Override
+	public Type getType() {
+		return type;
+	}
+
+	@Override
+	public void unbind() {
+		endpoint.stop();
+		abstractBinder.notifyUnbind(this);
```
markfisher: This might seem like overkill, but I am not too crazy about having the direct dependency on `Binder` here (even though we're dealing with both interfaces, it is a cyclical relationship). Perhaps we should introduce a `BindingListener` that `Binder` itself would extend - or perhaps just have `AbstractBinder` implement it? If you agree with my comment above about encapsulating the `start()`, we could maybe change to start/stop methods (where `unbind()` becomes `stop()`) and have both `onStart()` and `onStop()` for the listener. I don't think it's necessary to implement `Lifecycle` even if we do have those methods, considering we are dealing with an object that will typically be created via a constructor call (i.e. *not* a Spring-managed bean).

mbogoevici: Yeah, that made me think a little bit. It's an implementation detail though. See my comment above about calling `start`. A BindingListener, which I've considered raises the issue of where does the registration take place - would it be possible for binders to listen to unbind events raised by 'foreign' bindings ? We can do away with exposing the endpoint, since the binder will call start() on it anyway. I'm almost keen on having `DefaultBinding` a superclass that Binder implementors can extend to provide additional behaviour on unbind - e.g. region/binding cleanup for Rabbit/Gemfire.

mbogoevici: That would allow us to remove any kind of state in the binder, too.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
mbogoevici: Yeah, looking again - we can kill this without even exposing any method - the binder starts the endpoint anyway. The endpoint remains internal and we can access it in the tests via a `DirectFieldAccessor` - that should be good enough.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -31,141 +27,33 @@
...
+	/**
+	 * The name of the destination bound by the current binding.
+	 */
+	String getName();
+
+	/**
+	 * Returns the target component wrapped by this instance.
+	 * @return the target component
+	 */
+	T getTarget();
```
markfisher: `getTarget()` is currently unused - can we drop it?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
markfisher: +1

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -31,141 +27,33 @@
...
-			return false;
-		if (type != other.type)
-			return false;
-		return true;
-	}
-
+	/**
+	 * The name of the destination bound by the current binding.
+	 */
+	String getName();
```
markfisher: none of these methods are currently being used via the *interface*: `getName()`, `getTarget()`, `getType()`

that said, there are a couple cases within `Binder` implementations where `getName()` and `getType()` are invoked on the `DefaultBinding` class itself

It would be preferable if those references could be relying on the interface only, so I think we should keep `getName()` and `getType()` here, but we should try to decouple those implementations from the concrete class if possible

BUT I do think we can remove `getTarget()`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DirectHandler.java
mbogoevici: I guess the point is that the same `outputChannel` property is present on both and we can keep the name?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
mbogoevici: If we remove Target, we might as well remove Name. I'm keen on only keeping unbind() as the interface method.

markfisher: If there's a way to avoid using `getName()` within the Binder impls that currently invoke it, then +1 for only having `unbind()` on the interface. I do think we need to avoid any reliance on the concrete `DefaultBinding` in those Binder impls as well though.

mbogoevici: "I do think we need to avoid any reliance on the concrete DefaultBinding in those Binder impls as well though." I think it's the other way around, since the Binder impls create the bindings and will know exactly what kind of Binding they are working with. My idea here is to break the circular dependency by allowing Binders to subclass DefaultBinding and let it do its own cleanup code. Which basically means that the Binder will never see the Binding after creation anymore.  And viceversa.

mbogoevici: Surely, the binding will need to get access to something that allows them to do the cleanup - e.g. the administration utils for Rabbit, and something else in the case of Gemfire, but I guess it's okay to share the reference.

markfisher: "the Binder impls create the binding" is key... and so it does make sense to call any methods on the concrete type at the same point where it has been created. To clarify, what I want to avoid is any downcasting anywhere an interface-based reference is held. If that does not apply to `getName()` then we're fine.

mbogoevici: Yeah, we're on the same page I think. 'getName' is only used during 'unbind'.

mbogoevici:
Pushed rebased version with a number of changes: removal of all methods from `Binding` except for `unbind`, removed `Binding.Type`, removed circular references between binders and bindings, removed internal list of bindings (bindings are unbound on context destroy anyway).

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -635,9 +640,9 @@ private RabbitTemplate determineRabbitTemplate(RabbitPropertiesAccessor properti
 		consumer.setBeanFactory(getBeanFactory());
 		consumer.setBeanName("outbound." + name);
 		consumer.afterPropertiesSet();
-		Binding<MessageChannel> producerBinding = Binding.forProducer(name, moduleOutputChannel, consumer, properties);
-		addBinding(producerBinding);
-		producerBinding.start();
+		DefaultBinding<MessageChannel> producerBinding = new DefaultBinding<>(name, null, moduleOutputChannel, consumer, properties);
```
mbogoevici: Some cleanup to use the interfaces may be necessary.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+
+	@Override
+	public final void unbind() {
+		endpoint.stop();
+		afterUnbind();
+	}
+
+	protected void afterUnbind() {
```
mbogoevici: or maybe `doAfterUnbind`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DirectHandler.java
markfisher: I think the name is fine.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
markfisher: I think `afterUnbind` is sufficient actually.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -179,10 +169,9 @@ public void testPartitionedModuleSpEL() throws Exception {
 						((byte[]) receive2.getPayload())[0]),
 				containsInAnyOrder((byte)0, (byte)1, (byte)2));
 
-		binder.unbind(input0Binding);
-		binder.unbind(input1Binding);
-		binder.unbind(input2Binding);
-		binder.unbind(outputBinding);
+		input0Binding.unbind();
+		input1Binding.unbind();
```
markfisher: missing `input2Binding.unbind()`

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -302,11 +294,11 @@ public void testSendAndReceiveMutipleTopics() throws Exception {
 		assertMessageReceive(moduleInputChannel, "foo-x-payload");
 		assertMessageReceive(moduleInputChannel, "foo-y-payload");
 
-		binder.unbind(producerBinding1);
-		binder.unbind(consumerBinding1);
+		producerBinding1.unbind();
+		producerBinding2.unbind();
```
markfisher: 2nd should be 'consumerBinding1'

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -302,11 +294,11 @@ public void testSendAndReceiveMutipleTopics() throws Exception {
...
 
-		binder.unbind(producerBinding2);
-		binder.unbind(consumerBinding2);
+		producerBinding1.unbind();
+		producerBinding2.unbind();
```
markfisher: these should be: `producerBinding2` and `consumerBinding2`

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitBinderTests.java
```diff
@@ -122,8 +122,8 @@ public void handleMessage(Message<?> message) throws MessagingException {
 		});
 		moduleOutputChannel.send(message);
 		assertTrue(latch.await(10, TimeUnit.SECONDS));
-		binder.unbind(consumerBinding);
-		binder.unbind(producerBinding);
+		consumerBinding.unbind();
+		consumerBinding.unbind();
```
markfisher: 2nd one should be `producerBinding`

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/test/java/org/springframework/cloud/stream/binder/redis/RedisBinderTests.java
```diff
@@ -91,15 +92,12 @@ public void testConsumerProperties() throws Exception {
 		Properties properties = new Properties();
 		properties.put("maxAttempts", "1"); // disable retry
 		Binding<MessageChannel> binding = binder.bindConsumer("props.0", "test", new DirectChannel(), properties);
-		@SuppressWarnings("unchecked")
-		List<Binding<MessageChannel>> bindings = TestUtils.getPropertyValue(binder, "binder.bindings", List.class);
-		assertEquals(1, bindings.size());
-		assertEquals(binding, bindings.get(0));
-		AbstractEndpoint endpoint = binding.getEndpoint();
+		assertEquals(binding, binding);
```
markfisher: this looks pointless

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/test/java/org/springframework/cloud/stream/binder/redis/RedisBinderTests.java
```diff
@@ -91,15 +92,12 @@ public void testConsumerProperties() throws Exception {
...
+		AbstractEndpoint endpoint = extractEndpoint(binding);
 		assertThat(endpoint, instanceOf(RedisQueueMessageDrivenEndpoint.class));
 		assertSame(DirectChannel.class, TestUtils.getPropertyValue(endpoint, "outputChannel").getClass());
-		binder.unbind(binding);
-		assertEquals(0, bindings.size());
+		binding.unbind();
+		assertFalse(extractEndpoint(binding).isRunning());
```
markfisher: already have `endpoint` assigned above

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/test/java/org/springframework/cloud/stream/binder/redis/RedisBinderTests.java
```diff
@@ -109,13 +107,11 @@ public void testConsumerProperties() throws Exception {
...
-		assertEquals(1, bindings.size());
-		assertEquals(binding, bindings.get(0));
-		endpoint = binding.getEndpoint();
+		endpoint = extractEndpoint(binding);
 		verifyConsumer(endpoint);
 
-		binder.unbind(binding);
-		assertEquals(0, bindings.size());
+		binding.unbind();
+		assertFalse(extractEndpoint(binding).isRunning());
```
markfisher: already have `endpoint` assigned above

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/test/java/org/springframework/cloud/stream/binder/redis/RedisBinderTests.java
```diff
@@ -145,16 +137,16 @@ public void testProducerProperties() throws Exception {
...
 		assertEquals(
 				"'props.0.test-' + headers['partition']",
 				TestUtils.getPropertyValue(adapter, "queueNameExpression", Expression.class).getExpressionString());
 
-		binder.unbind(producerBinding);
-		binder.unbind(consumerBinding);
-		assertEquals(0, bindings.size());
+		producerBinding.unbind();
+		consumerBinding.unbind();
+		assertFalse(extractEndpoint(producerBinding).isRunning());
```
markfisher: already have `endpoint` assigned above (but should rename to `producerEndpoint` for clarity)

spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
```diff
@@ -135,11 +134,11 @@ public void testSendAndReceiveMutipleTopics() throws Exception {
...
 		assertMessageReceive(moduleInputChannel, "foo-y-payload");
 
-		binder.unbind(producerBinding1);
-		binder.unbind(consumerBinding1);
+		producerBinding1.unbind();
+		consumerBinding1.unbind();
 
-		binder.unbind(producerBinding2);
-		binder.unbind(consumerBinding2);
+		consumerBinding1.unbind();
```
markfisher: this should be `producerBinding2`

spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/PartitionCapableBinderTests.java
```diff
@@ -227,10 +225,11 @@ public boolean matches(Object item) {
 					containsOur3Messages);
 
 		}
-		binder.unbind(input0Binding);
-		binder.unbind(input1Binding);
-		binder.unbind(input2Binding);
-		binder.unbind(outputBinding);
+		input0Binding.unbind();
+		input1Binding.unbind();
+		input2Binding.unbind();
```
markfisher: unbinding `input2Binding` twice here

spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/PartitionCapableBinderTests.java
```diff
@@ -333,4 +329,9 @@ protected String getPubSubEndpointRouting(AbstractEndpoint endpoint) {
 
 	protected abstract String getClassUnderTestName();
 
+	protected AbstractEndpoint extractEndpoint(Binding<MessageChannel> binding) {
+		Assert.isInstanceOf(DefaultBinding.class, binding, "Binding not of expected type, found " + binding.getClass().getName() + " instead");
```
markfisher: this might be too brittle in the future; probably fine for now though

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -362,7 +335,7 @@ public void afterPropertiesSet() throws Exception {
 	}
 
 	@Override
-	public final Binding<MessageChannel> bindConsumer(String name, String group, MessageChannel inputChannel, Properties properties) {
+	public final Binding<T> bindConsumer(String name, String group, T inputChannel, Properties properties) {
```
markfisher: we should not be calling the `T` parameter `inputChannel` now that it's generic

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -31,141 +27,11 @@
...
-		} else if (!target.equals(other.target))
-			return false;
-		if (type != other.type)
-			return false;
-		return true;
-	}
-
+public interface Binding<T> {
+	/**
+	 * Unbinds the target component represented by this instance and stop any active components. Implementations must
```
markfisher: s/stop/stops/

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -192,7 +192,7 @@ private boolean isPollable(Class<?> channelType) {
 	}
 
 	private void bridgeSubscribableToPollableChannel(SubscribableChannel sharedChannel, MessageChannel inputChannel) {
```
markfisher: maybe these param names should be: `subscribableChannel` and `pollableChannel` - also could we use the `PollableChannel` type for the latter?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -192,7 +192,7 @@ private boolean isPollable(Class<?> channelType) {
...
-		sharedChannel.subscribe(new MessageChannelBinderSupport.DirectHandler(inputChannel));
+		sharedChannel.subscribe(new DirectHandler(inputChannel));
 	}
 
 	private void bridgePollableToSubscribableChannel(MessageChannel pollableChannel,
```
markfisher: can these parameters be typed? (`PollableChannel` and `SubscribableChannel`, respectively)

spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/PartitionCapableBinderTests.java
mbogoevici: We can remove it for leniency. Maybe add a check for the endpoint field instead. Just to give a more meaningful error.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binding/ChannelBindingServiceTests.java
```diff
@@ -189,9 +188,10 @@ public void checkDynamicBinding () {
 		Binder<MessageChannel> binder = binderFactory.getBinder("mock");
 
 		MessageChannel inputChannel = new DirectChannel();
-		Binding<MessageChannel> mockBinding = Binding.forConsumer("bar", null, Mockito.mock(AbstractEndpoint.class),
-				inputChannel, null);
-
+		@SuppressWarnings("unchecked")
+		Binding<MessageChannel> mockBinding = Mockito.mock(Binding.class);
+		ChannelBindingService service = new ChannelBindingService(properties, binderFactory);
```
markfisher: it doesn't look like `service` is ever used

mbogoevici:
Addressed PR, rebased.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/6c880726b8099a92a0f6d4a12c84d9b322375591

PR #346, opened 2016-02-13
Allow binding to multiple input destinations in the binder #334

### LINKED ISSUE #334: Allow binding to multiple input destinations in the binder
As a user, I want to be able to connect an input channel to multiple destinations, so that I can merge the incoming data in a single flow.

sobychacko:
This resolves #334

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
@@ -16,7 +16,10 @@
 
 package org.springframework.cloud.stream.binding;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.logging.Log;
@@ -28,6 +31,8 @@ import org.springframework.cloud.stream.binder.Binding;
 import org.springframework.cloud.stream.config.BindingProperties;
 import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
 import org.springframework.messaging.MessageChannel;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.StringUtils;
 
 /**
  * Handles the operations related to channel binding including binding of input/output channels by delegating
@@ -49,7 +54,7 @@ public class ChannelBindingService {
 
 	private final Map<String, Binding<MessageChannel>> producerBindings = new HashMap<>();
 
-	private final Map<String, Binding<MessageChannel>> consumerBindings = new HashMap<>();
+	private final Map<String, List<Binding<MessageChannel>>> consumerBindings = new HashMap<>();
 
 	public ChannelBindingService(ChannelBindingServiceProperties channelBindingServiceProperties,
 								 BinderFactory<MessageChannel> binderFactory) {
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -57,13 +62,18 @@ public ChannelBindingService(ChannelBindingServiceProperties channelBindingServi
...
 		String channelBindingTarget = this.channelBindingServiceProperties.getBindingDestination(inputChannelName);
-		Binder<MessageChannel> binder = getBinderForChannel(inputChannelName);
-		Binding<MessageChannel> binding = binder.bindConsumer(channelBindingTarget, consumerGroup(inputChannelName), inputChannel,
-				this.channelBindingServiceProperties.getConsumerProperties(inputChannelName));
-		this.consumerBindings.put(inputChannelName, binding);
-		return binding;
+		String[] channelBindingTargets = StringUtils.commaDelimitedListToStringArray(channelBindingTarget);
+		List<Binding<MessageChannel>> bindings = new ArrayList<>();
+		for (String target : channelBindingTargets) {
+			Binder<MessageChannel> binder = getBinderForChannel(inputChannelName);
```
ilayaperumalg: ` getBinderForChannel(inputChannelName)` can be done before the iteration.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -57,13 +62,18 @@ public ChannelBindingService(ChannelBindingServiceProperties channelBindingServi
...
+			Binding<MessageChannel> binding = binder.bindConsumer(target, consumerGroup(inputChannelName), inputChannel,
```
ilayaperumalg: Same here. Since the `consumerGroup` and the `consumerProperties` are same for all the `destinations` for the given `inputChannelName`, these values can be obtained before iterating over the destinations to bind.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -57,13 +62,18 @@ public ChannelBindingService(ChannelBindingServiceProperties channelBindingServi
 		this.binderFactory = binderFactory;
 	}
 
-	public Binding<MessageChannel> bindConsumer(MessageChannel inputChannel, String inputChannelName) {
+	public Collection<Binding<MessageChannel>> bindConsumer(MessageChannel inputChannel, String inputChannelName) {
 		String channelBindingTarget = this.channelBindingServiceProperties.getBindingDestination(inputChannelName);
```
ilayaperumalg: Do we require to change this property/method name to destination**s**/getBindingDestination**s**? But, the `destination` doesn't differentiate whether it is for `@Output` or `@Input` type and I believe we won't support multiple destination names for `@Output` type. What do you think?

sobychacko: @ilayaperumalg I think since it is a method to bind a consumer, it is implicitly understood that we are binding on multiple destinations to an input channel. In that sense, `destination` here means `@Input`. We don't support multiple destination names for producers (`@output`)although for fanout type of use cases, it might be a conceivable idea.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
sobychacko: Updated the PR. Thanks!!

ilayaperumalg:
LGTM. Squashed, rebased and merged as 76c64e3763053e04eee8e376534cc132563c47ac

PR #347, opened 2016-02-15
Fix RxJava processor lifecycle

### LINKED ISSUE #351: RxJava processor lifecycle issue
RxJava processor's subject message handler should starter `after` the output channel binding and `before` the input channel binding.

mbogoevici:
-  Make SubjectMessageHandler a SmartLifecycle and assign it to phase 0, allowing it to be started after the outputs are bound and before the inputs are bound;
- Setup Rx in the start() method;
- Set the ServiceActivator associated with the handler to phase 0 as well;

This resolves #351

### DIFF
--- a/spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/SubjectMessageHandler.java
+++ b/spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/SubjectMessageHandler.java
@@ -19,7 +19,7 @@ package org.springframework.cloud.stream.annotation.rxjava;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import org.springframework.beans.factory.DisposableBean;
+import org.springframework.context.SmartLifecycle;
 import org.springframework.integration.handler.AbstractMessageProducingHandler;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.support.MessageBuilder;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/SubjectMessageHandler.java
```diff
@@ -62,46 +62,80 @@
...
+		return running;
+	}
+
+	@Override
+	public boolean isAutoStartup() {
+		return false;
+	}
+
+	@Override
+	public void stop(Runnable callback) {
```
mbogoevici: should check for running here too

spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/SubjectMessageHandler.java
```diff
@@ -110,7 +144,13 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
 	}
 
 	@Override
-	public void destroy() throws Exception {
-		subscription.unsubscribe();
+	public synchronized void stop() {
```
mbogoevici: should delegate to stop(Runnable)

ilayaperumalg:
Since we want the `phase` to be **0** for the `MessageHandler`, I think we can just have `SubjectMessagerHandler` implement the `Lifecycle` bean as all the `Lifecycle` beans are started with the phase **0**.

mbogoevici:
That would be ok, but I'd rather think we should implement SmartLifecycle for auto start control as well.

mbogoevici:
I any case, it is the phase of the ServiceActivator that also overrides the phase of the MessageHandler bean.

ilayaperumalg:
> it is the phase of the ServiceActivator that also overrides the phase of the MessageHandler bean.

yeah, the `ServiceActivator` definitely needs to have `0` here otherwise it assigns lowest phase value for the event driven consumer endpoint there by default. 

Since the messageHandler is always started (with phase 0 value) between the `OutboundLifecycle` (low phase value <  0) and `InboundLifecycle`  (high phase value > 0) beans, I am wondering how would the `autoStartup` control on the SubjectMessageHandler's SmartLifecycle help us here.

Irrespective of the comments, the changes look good as it fixes the underlying issue on starting the `message handler` after the `output` binding and before the `input` binding happens.

spring-cloud-stream-rxjava/src/main/java/org/springframework/cloud/stream/annotation/rxjava/SubjectMessageHandler.java
```diff
@@ -62,46 +62,80 @@
...
+
+				@Override
+				public void call(Throwable throwable) {
+					logger.error(throwable.getMessage(), throwable);
+				}
+			}, new Action0() {
+
+				@Override
+				public void call() {
+					logger.error("Subscription close for [" + subscription + "]");
```
ilayaperumalg: Not from the changes on this PR but this logger could be at `info` level as this method is invoked `onCompleted` and we may not want to log it as `error`.

mbogoevici: Agreed, it looks ugly.

ilayaperumalg:
LGTM, rebased and merged as 3874620d2a98857f14423bc94924232ebefa1695 along with some minor polishing 1216b4b797e083a43430f22946919b33c6f828e0

PR #354, opened 2016-02-16
Maven configuration simplification

### LINKED ISSUE #353: Maven configuration simplification
As a developer, I would like to simplify the maven configuration files in `spring-cloud-stream` and its sub-projects.

Currently, the project version needs to be specified in many places in POM files though those projects can simply use the `project.version` from the parent projects. This is difficult to manage in future.

Also, there are some duplicate properties which can be removed.

ilayaperumalg:
- Use `project.version` wherever possible
 - Move kafka related dependencies from parent `dependencyManagement` to `kafka` binder project
 - Remove duplicate properties
 - Fix typo

This resolves #353

### DIFF
--- a/spring-cloud-stream-binders/pom.xml
+++ b/spring-cloud-stream-binders/pom.xml
@@ -4,7 +4,7 @@
 	
 	<artifactId>spring-cloud-stream-binders-parent</artifactId>
 	<packaging>pom</packaging>
-	<url>http://projects.spring.io/spring-xd/</url>
+	<url>https://github.com/spring-cloud/spring-cloud-stream</url>
 	<organization>
 		<name>Pivotal Software, Inc.</name>
 		<url>http://www.spring.io</url>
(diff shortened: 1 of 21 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/pom.xml
```diff
@@ -10,11 +10,12 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-stream-binders-parent</artifactId>
-		<version>1.0.0.BUILD-SNAPSHOT</version>
+		<version>${project.version}</version>
 	</parent>
 
 	<properties>
 		<kafka.version>0.8.2.2</kafka.version>
```
mbogoevici: I believe these version properties need to move to the parent, along with the <dependencyManagement> statements, leaving only <dependencies> in modules.

spring-cloud-stream-binders/pom.xml
```diff
@@ -25,48 +22,6 @@
 		<module>spring-cloud-stream-binder-kafka</module>
 	</modules>
 	
-	<dependencyManagement>
```
mbogoevici: As per my other comment, let's consolidate this in a single <depenndencyManagement> section.

ilayaperumalg:
@sobychacko The problem with that approach is that when we install the individual POMs of each child projects, the `project.version` doesn't get set. May be there is an option to set when doing install?

sobychacko:
LGTM - rebased, squashed and merged with 4dfd217db6e7580ae67b5889a1c119d116720926

Issue #317, opened 2016-02-06
Rabbit Binder Improvements


garyrussell:
- Consider an option to __not__ bind the `default` queue on the producer side (just declare the topic exchange) for cases outside of SCDF. Perhaps it should even be true by default and set to false by SCDF.

- Consider an option to provide a name other than `default` for the default queue (when the above option is set to declare/bind the queue).

See [Stack Overflow Question here](http://stackoverflow.com/questions/35215810/spring-could-stream-with-rabbit-binder-source-sink-queue-names-dont-match/35232255#35232255).

markfisher:
I think we should have a binder property (common for all binders), such as "defaultConsumerGroup", and it could (probably should) be empty by default. IF provided, it would trigger eager creation of the queue with its name (or whatever that particular `Binder` impl needs to create to support a consumer group). SCDF could always set that property's value to "default".

garyrussell:
Sounds good to me bind this default consumer group when binding the producer. It's up to the binder as to exactly what that means.

mbogoevici:
I don't think it makes a lot of sense outside of rabbit though - so it's a split vote between Rabbit and Kafka, which makes me reluctant to generalize at first. Maybe Gemfire could be a tie-breaker here.

I could see the default being not to do anything as @garyrussell suggest  - consistent with the notion that default consumers get a new queue in the first place, so won't see any of the 'durable' default stuff.

markfisher:
Actually @pperalta can confirm/correct/clarify, but I believe GemFire will not fit the default consumer group model well at all, since the actual partitioned regions that hold the message content are only generated when the consumer process starts.

garyrussell:
In any case, we don't want scdf to have to know what binder(s) he's talking to so a generic property that may, or may not, be meaningful to a particular binder implementation seems appropriate.

mbogoevici:
If we make this a general purpose property, then I believe it should be something more generic, that would give binders some freedom for implementation, while providing strong guidance on the expected outcome. We are doing the `defaultGroup` for Rabbit in order to ensure that consumers in that group created after the producer will receive all the messages that the producing component has emitted. It seems like not all binders will have this capability (as per @markfisher 's comment), so we should make it optional.

TabberGit:
[...]

Issue #307, opened 2016-02-03
Add support for passing generic binder configuration properties per binding


mbogoevici:
Currently, `BindingProperties` allows only for specific, general purpose binding properties to be set, but it may be desirable to configure individual settings per-binding so adding a way to map them (perhaps via a generic `properties` map) would allow more granular configuration. In the process, we may want to consider how prefixes for binders currently work with that, since the resulting property names might be long and confusing (e.g. `spring.cloud.stream.bindings.output.properties`)

mbogoevici:
Closed w/https://github.com/spring-cloud/spring-cloud-stream/commit/9b0c4bd627f7d122be60d11caf993b18799b3ae6

PR #357, opened 2016-02-17
Ignore binder property keys when not supported

### LINKED ISSUE #355: Ignore the unsupported producer/consumer properties for the binder
As a user, I would like to have the `ChannelBindingService` filters the unsupported producer/consumer properties that are passed along when binding the producer/consumer channel.

For instance, the `durableSubscription` consumer property is only supported for the `Rabbit` binder and when this property is set and the `Kafka` binder is used, then it needs to be filtered out.

ilayaperumalg:
- When the binder implementation doesn't support the producer/consumer properties that are passed along by the `ChannelBindingService` when binding,
those properties can be ignored(instead of failing the validation) during the validation of producer/consumer properties.

This resolves #355

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
@@ -136,4 +136,11 @@ public abstract class BinderPropertyKeys {
 	 */
 	public static final String MIN_PARTITION_COUNT = "minPartitionCount";
 
+	/**
+	 * The property keys which can be ignored by the binder when they are not supported by the corresponding
+	 * implementation.
+	 */
+	public static final String[] IGNORABLE_KEYS = { DURABLE, MIN_PARTITION_COUNT, BATCHING_ENABLED, BATCH_SIZE,
+			BATCH_BUFFER_LIMIT, BATCH_TIMEOUT };
+
 }
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupport.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupport.java
@@ -668,8 +668,9 @@ public abstract class MessageChannelBinderSupport
 	private void validateProperties(String name, Properties properties, Set<Object> supported, String type) {
 		StringBuilder builder = new StringBuilder();
 		int errors = 0;
+		List<String> ignorableKeys = Arrays.asList(BinderPropertyKeys.IGNORABLE_KEYS);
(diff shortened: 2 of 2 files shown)


garyrussell:
I don't agree with the premise of this change; just ignoring some specific instruction from an upstream component just seems wrong to me, especially for user-specified properties (such as batching etc).

At a minimum it should be logged as a warning.

I agree there is a case for ignoring properties that are set by our software implicitly (such as whatever we come up with for #317 to control binding the default queue) but I don't like ignoring user-specified properties.

While `durableSubscription` might only apply to rabbit (i.e. subscriptions are always durable with kafka), it seems to me that a user setting `durableSubscription=false` when using kafka should get an error (or WARN log).

ilayaperumalg:
@garyrussell I agree and +1 for logging as a `warning`

mbogoevici:
Warning is fair, but the current strategy of failing the deployment on unenforceable properties seems to harsh as well.

ilayaperumalg:
Updated the PR to include the `warning` logging when the property is ignored.

mbogoevici:
I think this raises the issue of how the binders and their implementations will evolve over time. It feels like introducing new options in the binder superclass is liable to break all implementations easily, so validation is rather brittle IMO if we keep crashing the binder for any property that is not understood (I also believe that any 'must-have' type of behaviour should be captured in the Binder interface rather than the property list). I think this change is necessary, but at the same time, in the particular case of durability, we may want to remove it as a general purpose property.

ilayaperumalg:
Based on the discussion related to the concerns raised, I think having to refer the properties `per` binder would help avoiding these concerns. Do you think my earlier version of the PR would make sense here: https://github.com/ilayaperumalg/spring-cloud-stream/commit/2ff5831088e9bb325918b5f793d5bb60f87965dd

mbogoevici:
I don't filtering them upstream does anything substantially better than issuing warnings downstream. In both cases the properties are ignored, with the disadvantage that `ChannelBindingService` becomes dependent on `MessageChannelBinderSupport` and violates encapsulation some. If warnings are acceptable for leniency, we can live with that.

mbogoevici:
One idea to explore here is to extract the Binding settings as two POJO(s): for Producer and Consumer, thus making it clearer what properties should be supported as part of a Binder's general contract. The Properties object is just un-wieldy and doesn't fit well with the Boot model.

mbogoevici:
We would still have to pass some binder-specific properties in an extensible way (i.e. properties), but at the very least those could be validated by each binder individually.

ilayaperumalg:
We can revisit/discuss this PR once #307 is addressed. Once that is addressed, we will move most of  the uncommon `binder` specific properties away from `BindingProperties`.

mbogoevici:
Closing this favour of handling unrecognized properties as an enhancement of https://github.com/spring-cloud/spring-cloud-stream/pull/409

PR #358, opened 2016-02-18
Support sync producer option in Kafka binder

### LINKED ISSUE #343: Add an option for synchronous  producers with Kafka
Expose the synchronous producer option in Spring Integration Kafka as a binder property.

ilayaperumalg:
- Add a binding property `spring.cloud.stream.bindings.<outputChannelName>.syncProducer`
   When set to `true`, the Kafka producer metadata is set to use `sync` producer
 - Update DefaultPropertiesAccessor for Kafka to use the `syncProducer` property
 - Add test

This resolves #343

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
@@ -136,9 +136,4 @@ public abstract class BinderPropertyKeys {
 	 */
 	public static final String MIN_PARTITION_COUNT = "minPartitionCount";
 
-	/**
-	 * Flag to indicate if the Kafka producer needs to be synchronous/asynchronous.
-	 */
-	public static final String SYNC_PRODUCER = "syncProducer";
-
 }
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
@@ -132,8 +132,13 @@ public abstract class BinderPropertyKeys {
 	public static final String DURABLE = "durableSubscription";
 
 	/**
-	 * Minimum partition count, if the transport supports partitioning natively (e.g. Kafka)
+	 * Minimum partition count, if the transport supports partitioning natively (e.g. Kafka).
 	 */
 	public static final String MIN_PARTITION_COUNT = "minPartitionCount";
+
+	/**
+	 * Flag to indicate if the Kafka producer needs to be synchronous/asynchronous.
+	 */
+	public static final String SYNC_PRODUCER = "syncProducer";
 
 }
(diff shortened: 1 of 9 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -122,7 +122,7 @@ public Properties getConsumerProperties(String inputChannelName) {
 				channelConsumerProperties.setProperty(BinderPropertyKeys.CONCURRENCY,
 						Integer.toString(bindingProperties.getConcurrency()));
 			}
-			if (bindingProperties.isDurableSubscription() != null) {
+			if (Boolean.TRUE.equals(bindingProperties.isDurableSubscription())) {
```
mbogoevici: Actually this prevents passing `durable=false` to any binder, including Rabbit.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
```diff
@@ -132,8 +132,13 @@
...
 
 	/**
-	 * Minimum partition count, if the transport supports partitioning natively (e.g. Kafka)
+	 * Minimum partition count, if the transport supports partitioning natively (e.g. Kafka).
 	 */
 	public static final String MIN_PARTITION_COUNT = "minPartitionCount";
 
+	/**
+	 * Flag to indicate if the Kafka producer needs to be synchronous/asynchronous.
+	 */
```
mbogoevici: Should we perhaps make it a Kafka property only and set it on the binder directly (with a default option) ? Doesn't feel like a generic property. (minPartitionCount is an unfortunate anomaly, as for durable, we might want to consider making removing it at the top level too).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
ilayaperumalg: Since the default `durableSubscription` value is always `false` at the binder level, I thought one doesn't need to pass the value `durable=false` explicitly. no?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -139,6 +139,14 @@ public Properties getConsumerProperties(String inputChannelName) {
 	 */
 	public Properties getProducerProperties(String outputChannelName) {
 		Properties channelProducerProperties = new Properties();
+		BindingProperties bindingProperties = this.bindings.get(outputChannelName);
+		if (bindingProperties != null) {
+			// Setting Kafka producer properties
+			if (Boolean.TRUE.equals(bindingProperties.isSyncProducer())) {
```
mbogoevici: Same here. The binder should be able to decide how they implement this.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
mbogoevici: Well, the discussion on durability is a longer one, but as a norm, I think it adds some confusion if we filter out properties based on their value, before letting the binder to act on it.

ilayaperumalg:
Updated the PR to move the `syncProducer` property into Kafka `binder` configuration properties.

mbogoevici:
Rebased, squashed merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/047a3f8d4fcc08f876394fc01fbf3207ba8a3b35

PR #359, opened 2016-02-18
Binder interface javadoc

ilayaperumalg:
- `module` to `app`
- Add more detail to the javadoc

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binder.java
@@ -19,7 +19,7 @@ package org.springframework.cloud.stream.binder;
 import java.util.Properties;
 
 /**
- * A strategy interface used to bind a module interface to a logical name. The name is intended to identify a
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binder.java
```diff
@@ -32,21 +32,21 @@
 public interface Binder<T> {
 
 	/**
-	 * Bind a message consumer on a channel
+	 * Bind the target component as a message consumer to the logical entity identified by the name.
```
mbogoevici: I think we should specify exactly what the logical entity is. Sounds a bit obscure.

ilayaperumalg: I thought about it when adding this and convinced myself with the explanation at the class level what the logical entity could be. do you think that isn't enough?

ericbottard: Sorry guys, maybe I was too quick on this. If you come up with a better phrasing, just confirm it here, and I can add an additional commit

PR #360, opened 2016-02-19
Durability configuration and default groups

### LINKED ISSUE #317: Rabbit Binder Improvements

mbogoevici:
Resolves #317

Remove the `durable` binder configuration property
Make subscriber groups durable by default
Introduce `requiredGroups` property
Kafka groups (non-anonymous) now start by default at EARLIEST, which is more appropriate for new stream consumers

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -191,6 +191,7 @@ public class KafkaMessageChannelBinder extends AbstractBinder<MessageChannel> {
 
 	private static final Set<Object> KAFKA_PRODUCER_PROPERTIES = new SetBuilder()
 			.add(BinderPropertyKeys.MIN_PARTITION_COUNT)
+			.add(BinderPropertyKeys.REQUIRED_GROUPS)
 			.build();
 
 	/**
@@ -249,7 +250,7 @@ public class KafkaMessageChannelBinder extends AbstractBinder<MessageChannel> {
 
 	private boolean resetOffsets = DEFAULT_RESET_OFFSETS;
 
-	private StartOffset startOffset = DEFAULT_START_OFFSET;
+	private StartOffset startOffset = null;
 
 	private int zkSessionTimeout = DEFAULT_ZK_SESSION_TIMEOUT;
 
@@ -440,9 +441,12 @@ public class KafkaMessageChannelBinder extends AbstractBinder<MessageChannel> {
 		// Consumers reset offsets at the latest time by default, which allows them to receive only
 		// messages sent after they've been bound. That behavior can be changed with the
 		// "resetOffsets" and "startOffset" properties.
-		String consumerGroup = group == null ? "anonymous." + UUID.randomUUID().toString() : group;
+		boolean anonymous = !StringUtils.hasText(group);
+		String consumerGroup = anonymous ? "anonymous." + UUID.randomUUID().toString() : group;
+		// The reference point, if not set explicitly is the latest time for anonymous subscriptions and the
+		// earliest time for group subscriptions. This allows subscriptions to receive er
 		long referencePoint = this.startOffset != null ?
-				startOffset.getReferencePoint() : OffsetRequest.LatestTime();
+				startOffset.getReferencePoint() : (anonymous ? OffsetRequest.LatestTime() : OffsetRequest.EarliestTime());
 		return createKafkaConsumer(name, inputChannel, properties, consumerGroup, referencePoint);
 	}
 
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -444,7 +444,8 @@ public class KafkaMessageChannelBinder extends AbstractBinder<MessageChannel> {
 		boolean anonymous = !StringUtils.hasText(group);
 		String consumerGroup = anonymous ? "anonymous." + UUID.randomUUID().toString() : group;
 		// The reference point, if not set explicitly is the latest time for anonymous subscriptions and the
-		// earliest time for group subscriptions. This allows subscriptions to receive er
+		// earliest time for group subscriptions. This allows the latter to receive messages published before the group
+		// has been created.
 		long referencePoint = this.startOffset != null ?
 				startOffset.getReferencePoint() : (anonymous ? OffsetRequest.LatestTime() : OffsetRequest.EarliestTime());
 		return createKafkaConsumer(name, inputChannel, properties, consumerGroup, referencePoint);
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
@@ -134,7 +134,6 @@ public class RabbitMessageChannelBinder extends AbstractBinder<MessageChannel> {
 	private static final String DEAD_LETTER_EXCHANGE = "DLX";
 
 	private static final Set<Object> RABBIT_CONSUMER_PROPERTIES = new HashSet<Object>(Arrays.asList(new String[] {
-
 		BinderPropertyKeys.MAX_CONCURRENCY,
 		RabbitPropertiesAccessor.ACK_MODE,
 		RabbitPropertiesAccessor.PREFETCH,
@@ -144,7 +143,8 @@ public class RabbitMessageChannelBinder extends AbstractBinder<MessageChannel> {
 		RabbitPropertiesAccessor.TRANSACTED,
 		RabbitPropertiesAccessor.TX_SIZE,
 		RabbitPropertiesAccessor.AUTO_BIND_DLQ,
-		RabbitPropertiesAccessor.REPUBLISH_TO_DLQ
+		RabbitPropertiesAccessor.REPUBLISH_TO_DLQ,
+		RabbitPropertiesAccessor.DURABLE
 	}));
 
 	/**
@@ -161,7 +161,6 @@ public class RabbitMessageChannelBinder extends AbstractBinder<MessageChannel> {
 	 */
 	private static final Set<Object> SUPPORTED_CONSUMER_PROPERTIES = new SetBuilder()
 			.addAll(SUPPORTED_BASIC_CONSUMER_PROPERTIES)
-			.add(BinderPropertyKeys.DURABLE)
 			.add(BinderPropertyKeys.CONCURRENCY)
 			.add(BinderPropertyKeys.PARTITION_INDEX)
 			.build();
@@ -175,6 +174,7 @@ public class RabbitMessageChannelBinder extends AbstractBinder<MessageChannel> {
 			.add(RabbitPropertiesAccessor.PREFIX)
 			.add(RabbitPropertiesAccessor.REQUEST_HEADER_PATTERNS)
 			.add(BinderPropertyKeys.COMPRESS)
+			.add(BinderPropertyKeys.REQUIRED_GROUPS)
 			.build();
 
 	/**
@@ -231,6 +231,8 @@ public class RabbitMessageChannelBinder extends AbstractBinder<MessageChannel> {
 	private volatile int defaultPrefetchCount = DEFAULT_PREFETCH_COUNT;
 
 	private volatile int defaultTxSize = DEFAULT_TX_SIZE;
+
+	protected volatile boolean defaultDurableSubscription = true;
 
 	private volatile String defaultPrefix = DEFAULT_RABBIT_PREFIX;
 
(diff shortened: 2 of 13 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -431,9 +432,12 @@ public void setStartOffset(StartOffset startOffset) {
 		// Consumers reset offsets at the latest time by default, which allows them to receive only
 		// messages sent after they've been bound. That behavior can be changed with the
 		// "resetOffsets" and "startOffset" properties.
-		String consumerGroup = group == null ? "anonymous." + UUID.randomUUID().toString() : group;
+		boolean anonymous = !StringUtils.hasText(group);
+		String consumerGroup = anonymous ? "anonymous." + UUID.randomUUID().toString() : group;
+		// The reference point, if not set explicitly is the latest time for anonymous subscriptions and the
+		// earliest time for group subscriptions. This allows subscriptions to receive er
```
garyrussell: er?

mbogoevici: Er ...

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -325,6 +327,15 @@ public void setDefaultTxSize(int defaultTxSize) {
 		this.defaultTxSize = defaultTxSize;
 	}
 
+	/**
+	 * Set whether subscriptions to taps/topics are durable.
```
mbogoevici: Should remove reference to taps.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -232,6 +232,8 @@ public MessageProperties toMessageProperties(AMQP.BasicProperties source, Envelo
 
 	private volatile int defaultTxSize = DEFAULT_TX_SIZE;
 
+	protected volatile boolean defaultDurableSubscription = true;
+
```
garyrussell: I don't think this is the behavior the spring-cloud-bus guys want - I believe they want an anonymous auto-delete queue by default. The javadocs below say the default is `false`.

Default false is more aligned with default JMS topic behavior too.

ilayaperumalg:
@mbogoevici Could you create a corresponding a change in dataflow that removes the `durableSubscription` option?

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
mbogoevici: This affects only groups/taps. Anonymous queues would remain as they were. But we can discuss if it's too confusing. We're covered via requiredGroups anyway.

mbogoevici:
Sure, was planning to do that. But meanwhile the setting will be simply ignored by Boot.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
mbogoevici: This would make things more in line with Kafka, where we'd need to start consumer groups at `earliest` by default, since the producer cannot safely reset the offset of a consumer.

mbogoevici: Actually, there is another reason for doing it this way. If we make this non-durable, tap subscribers in SCDF will not be able to create durable subscriptions easily (i.e. without passing rabbit-specific properties), since there is no global property that governs that. Given that an SCS user can opt for either, I think that the defaults should cover the most likely to be used use cases.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBindingPropertiesAccessor.java
```diff
@@ -337,14 +337,17 @@ public boolean isCompress(boolean defaultValue) {
 	}
 
 	/**
-	 * If true, subscriptions to taps/topics will be durable.
-	 * @param defaultValue the default value.
-	 * @return the property or default value.
+	 * A list of groups for which the binder will make ensure message delivery, even if their consumers bind
```
ilayaperumalg: typo: make ensure

spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/PartitionCapableBinderTests.java
```diff
@@ -106,6 +106,66 @@ public void testAnonymousGroup() throws Exception {
...
+
+	@Test
+	public void testTwoRequiredGroups() throws Exception {
+		Binder<MessageChannel> binder = getBinder();
+		DirectChannel output = new DirectChannel();
+		Properties properties = new Properties();
+
+		String testDestination = "testDestination" + UUID.randomUUID().toString().replace("-", "");
+
+		// one group
```
ilayaperumalg: two groups

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderPropertyKeys.java
```diff
@@ -127,13 +127,14 @@
...
 	/**
-	 * Durable pub/sub consumer.
+	 * Minimum partition count, if the transport supports partitioning natively (e.g. Kafka)
 	 */
-	public static final String DURABLE = "durableSubscription";
+	public static final String MIN_PARTITION_COUNT = "minPartitionCount";
 
 	/**
-	 * Minimum partition count, if the transport supports partitioning natively (e.g. Kafka)
+	 * Required groups. The binder will ensure that consumers from these groups binding after
```
ilayaperumalg: Unless I miss something, can we reframe this javadoc?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -210,6 +210,7 @@ public Object deserialize(byte[] bytes) throws ZkMarshallingError {
 
 	private static final Set<Object> KAFKA_PRODUCER_PROPERTIES = new SetBuilder()
 			.add(BinderPropertyKeys.MIN_PARTITION_COUNT)
+			.add(BinderPropertyKeys.REQUIRED_GROUPS)
```
ilayaperumalg: I am trying to understand the role of `requiredGroups` in Kafka binder as it is considered as a `supported` property. Could you clarify?

mbogoevici: It's a supported property, in the sense that Kafka can satisfy the contract out of the box without additional changes. Being a global property, we must pass it here, or else the Kafka binder will crash. We will have to pass the `requiredGroups` there, to ensure that Rabbit and Redis work as expected - since the ordering of start after deployment cannot be guaranteed.

mbogoevici: I'm not sure about how other binders can satisfy this, but if they cannot, they wouldn't be usable with SCDF.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
garyrussell: >This affects only groups/taps.

I don't follow - it affects anyone who binds a consumer - we're switching the defaults on them again.

I am not convinced this should be pushed down to the `Binder` implementation(s) - some binders support it (RabbitMQ, JMS [future]) some don't.

It seems to me it's a common property that should be ignored by those who don't support it rather than making it a custom property of each binder that does support it.

garyrussell: Sorry - I see it now 

	boolean durable = !anonymousConsumer && accessor.isDurable(this.defaultDurableSubscription);

But we need to make it very clear in the docs.

The problem is if someone is using a `UUID.randomUUID()` for a group (after the last change we imposed on them :smile: )

But spring-cloud-bus has backed out that change so we're ok there.

mbogoevici: The goal is to find the right defaults before the RC. To me, it seems natural that anonymous consumers will be non-durable by default (eg. bus consumers are unaffected), whereas named/grouped will be durable by default (i.e. will reconnect where they left), and Rabbit users have the option of turning that off (not every other binder does). I could be wrong though, and I'm happy to revert that.

But I am trying to understand your stance on demoting durability properties. The notion of ignoring global options is not consistent with https://github.com/spring-cloud/spring-cloud-stream/pull/357#issuecomment-185267078. Are you suggesting that we switch the validation process to property values, instead? I'm fine with that, too, as long as we don't keep going back and forth on the validation process.

mbogoevici: I think https://github.com/spring-cloud/spring-cloud-stream/pull/360/files#r53472658 answers it :)

garyrussell: I don't **think** I'm being inconsistent - in #357 I'm saying we should log or reject invalid global properties; in this case (if durableSubscription is global), the kafka binder should reject an invalid setting for it. The solution here (for #357) is to demote it instead.

Pushing properties like this down to the implementations removes any possibility of doing such logging and the properties are simply ignored.

I guess I am ok (but not thrilled) with very specific properties being ignored by other binders, but there is definitely a grey area with "somewhat" common properties. One concern is how do we ensure consistency across binders (e.g. someone creates a JMS binder and names the property something else, e.g. `durable`).

mbogoevici: Ok. I see what you are saying. The question for me is whether `durableSubscription` should be global or not. I believe that we should prevent proliferating global properties that apply only for certain types of middleware, or rather set up a contract for default behaviour (as in this case) and force binders to deal with it or provide options to override it if necessary. The tie-breaker, in my opinion, is whether the default setting makes sense from a logical point of view wrt the SCSt model (e.g. making groups durable by default for Rabbit ensures that it is consistent with Kafka, where a group subscription has the same effect). But I agree what we shouldn't ignore what we would expect to be global settings either (e.g. no binder can ignore the `groups` functionality for instance). My take was that if someone used a JMS binder, they would (and should) have the binder specific properties documented (which we don't, but should before 1.0.0.GA), so that they understand what are the options that they have for their own binder.

mbogoevici: Ok, here's what I think we need to do, based on the feedback: reinstate `DURABLE` as a global property, make changes to the current validation process to do validation on values as well (not only on key presence), so binders can declare options they only support. Also, I think we need a mechanism at the global level that allows to detect what properties have been accessed by a binder and which ones weren't, and act accordingly - e.g. if a binder ignores a mandatory property in its implementation.

garyrussell: I am fine with the status quo of this PR - but perhaps we can introduce some notion of 'optional' properties that binders **may** support (instead of making it a rabbit-specific property) - it might be as simple as an interface with constants.

mbogoevici: OK, let's revisit that as a separate issue. I'd be keen on exploring more options on making binder implementations as lenient as possible, without introducing massive inconsistencies.

mbogoevici:
Pushed rebased version + addressing PR comments.

garyrussell:
Rebased, squashed, merged as 2fa9cda89c162bdaa4cb439c9c57e5f1f0066dea

PR #363, opened 2016-02-22
Add AggregateApplicationBuilder

### LINKED ISSUE #362: Add support for command line args in AggregateBuilder
Currenlty `AggregateApplication` has convenience methods accepting `String[]` arrays as module arguments. The API is a bit clumsy, and the builder is nicer, so it would be good to support it there, e.g. instead of

```
AggregateApplication.run(
	new Class<?>[] { TwitterStreamApplication.class, StreamTransformer.class },
	args, new String[][] { {"--language=en"}, {} });
```

we could write

```
new AggregateApplicationBuilder()
  .from(TwitterStreamApplication.class).args(new String[] {"--language=en"})
  .to(StreamTransformer.class)
  .run(args);
```

ilayaperumalg:
- Uses builder pattern to create aggregate application that supports specifying
  - args
  - profiles
  - config name for each child app and parent aggregate app separately.

An example config would look like this:

```
@SpringBootApplication
public class DoubleApplication {

		new AggregateApplicationBuilder().
				from(SourceApplication.class).args("--fixedDelay=5000")
				.to(SinkApplication.class).args("--debug=true").run("--spring.application.name=aggregate-test");
	}
```

This resolves #362

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -17,20 +17,16 @@
 package org.springframework.cloud.stream.aggregate;
 
 import org.springframework.boot.Banner.Mode;
-import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.builder.SpringApplicationBuilder;
-import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.messaging.Processor;
 import org.springframework.cloud.stream.messaging.Sink;
 import org.springframework.cloud.stream.messaging.Source;
 import org.springframework.context.ConfigurableApplicationContext;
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
The existing `AggregateBuilder` (and its associate classes such as `AggregateConfigurer` and `AggregateBuilderConfiguration`) probably need different names as they relate to a different type of `aggregation` where there the apps are connected via **bindable** channels rather than `AggregateApplication'` **non-bindable** shared channels.

ilayaperumalg:
Associated sample is updated [here] (https://github.com/spring-cloud/spring-cloud-stream-samples/pull/2)

ilayaperumalg:
May be we could use our previous term `DirectBinding` for builder that uses `AggregateApplication` via `shared` channels (e.g: DirectBindingBuilder) and keep the `AggregateBuilder` for composing applications with `bindable` channels.

mbogoevici:
I believe we should just keep one aggregate builder for in-JVM aggregation (i.e. harmonize the API).

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/7d7fcc9184b95455090f0b0accf53db3354a323c

PR #366, opened 2016-02-22
Partition handling refactoring

pperalta:
Moved partition calculation from `AbstractBinder` to its own class `PartitionHandler`.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
@@ -1,0 +1,232 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.binder;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.expression.EvaluationContext;
+import org.springframework.expression.Expression;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * Utility class to determine if a binding is configured for partitioning
+ * (based on the binder properties provided in the constructor) and
+ * what partition a message should be delivered to.
+ *
+ * @author Patrick Peralta
+ * @author David Turanski
+ * @author Gary Russell
+ * @author Ilayaperumal Gopinathan
+ * @author Mark Fisher
+ * @author Marius Bogoevici
+ */
+public class PartitionHandler {
+	private static final Logger logger = LoggerFactory.getLogger(PartitionHandler.class);
+
+	private final ConfigurableListableBeanFactory beanFactory;
+
+	private final EvaluationContext evaluationContext;
+
+	private final PartitionSelectorStrategy partitionSelector;
+
+	private final PartitioningMetadata partitioningMetadata;
+
+
+	/**
+	 * Construct a {@code PartitionHandler}.
+	 *
+	 * @param beanFactory bean factory for binder
+	 * @param evaluationContext evaluation context for binder
+	 * @param partitionSelector configured partition selector; may be {@code null}
+	 * @param properties binder properties
+	 * @param partitionCount number of partitions configured for binder
+	 */
+	public PartitionHandler(ConfigurableListableBeanFactory beanFactory,
+			EvaluationContext evaluationContext,
+			PartitionSelectorStrategy partitionSelector,
+			DefaultBindingPropertiesAccessor properties, int partitionCount) {
+		this.beanFactory = beanFactory;
+		this.evaluationContext = evaluationContext;
+		this.partitionSelector = partitionSelector == null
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
```diff
@@ -0,0 +1,232 @@
...
+		}
+		try {
+			Object extractor = clazz.newInstance();
+			Assert.isInstanceOf(PartitionKeyExtractorStrategy.class, extractor);
+			this.beanFactory.registerSingleton(partitionSelectorClassName, extractor);
+			this.beanFactory.initializeBean(extractor, partitionSelectorClassName);
+			return ((PartitionSelectorStrategy) extractor).selectPartition(key, partitionCount);
+		}
+		catch (Exception e) {
+			logger.error("Failed to instantiate partition selector", e);
```
ericbottard: Why log AND rethrow ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
```diff
@@ -0,0 +1,232 @@
...
+		if (this.beanFactory.containsBean(partitionSelectorClassName)) {
+			return this.beanFactory.getBean(partitionSelectorClassName, PartitionSelectorStrategy.class)
+					.selectPartition(key, partitionCount);
+		}
+		Class<?> clazz;
+		try {
+			clazz = ClassUtils.forName(partitionSelectorClassName, this.beanFactory.getBeanClassLoader());
+		}
+		catch (Exception e) {
+			logger.error("Failed to load partition selector", e);
```
ericbottard: Why log AND rethrow

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
```diff
@@ -0,0 +1,232 @@
...
+		}
+		partition = partition % partitioningMetadata.partitionCount;
+		if (partition < 0) { // protection in case a user selector returns a negative.
+			partition = Math.abs(partition);
+		}
+		return partition;
+	}
+
+	private Object invokeExtractor(String partitionKeyExtractorClassName, Message<?> message) {
+		if (this.beanFactory.containsBean(partitionKeyExtractorClassName)) {
```
ericbottard: Do we need some form of synchronization here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
```diff
@@ -0,0 +1,232 @@
...
+			return ((PartitionKeyExtractorStrategy) extractor).extractKey(message);
+		}
+		catch (Exception e) {
+			logger.error("Failed to instantiate key extractor", e);
+			throw new BinderException("Failed to instantiate key extractor: " + partitionKeyExtractorClassName, e);
+		}
+	}
+
+	private int invokePartitionSelector(String partitionSelectorClassName, Object key, int partitionCount) {
+		if (this.beanFactory.containsBean(partitionSelectorClassName)) {
```
ericbottard: idem synchronization

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
pperalta: +1. This was a copy/paste from `AbstractBinder` that I should have caught.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
pperalta: +1

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
pperalta: I think that we do.

Furthermore, I'm not sure why we register the extractor with the `beanFactory` - could we just simply hold a reference to it and use that instead of going back to the `beanFactory` every time?

ericbottard: > I'm not sure why we register the extractor with the beanFactory

I was tempted to state that as well, but figured you were just porting previous code.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * @author Patrick Peralta
+ * @author David Turanski
+ * @author Gary Russell
+ * @author Ilayaperumal Gopinathan
+ * @author Mark Fisher
+ * @author Marius Bogoevici
+ */
+public class PartitionHandler {
+	private static final Logger logger = LoggerFactory.getLogger(PartitionHandler.class);
```
markfisher: `logger` is not being used

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
```diff
@@ -0,0 +1,238 @@
...
+	 * @param evaluationContext evaluation context for binder
+	 * @param partitionSelector configured partition selector; may be {@code null}
+	 * @param properties binder properties
+	 * @param partitionCount number of partitions configured for binder
+	 */
+	public PartitionHandler(ConfigurableListableBeanFactory beanFactory,
+			EvaluationContext evaluationContext,
+			PartitionSelectorStrategy partitionSelector,
+			DefaultBindingPropertiesAccessor properties, int partitionCount) {
+		this.beanFactory = beanFactory;
```
markfisher: should add an `Assert.notNull(..)` since the `BeanFactory` is required

markfisher:
rebased and squashed as https://github.com/spring-cloud/spring-cloud-stream/commit/d34aafb87beeaec2e276ea3709dba825ce261eeb

also fixed an issue with the `BeanFactory` being `null` for tests

and followed with this commit: https://github.com/spring-cloud/spring-cloud-stream/commit/1dc1303b548b0b43dfdfe908cf0e8b933db3c2c1

PR #371, opened 2016-02-24
Binding error channel

### LINKED ISSUE #329: Add support for error channel binding

ilayaperumalg:
- bind Spring Integration `errorChannel` as a message producer when there is at least one outbound endpoint exists in the SCSt context.
  - By default the `errorChannel` is bound with the destination name `${spring.config.name}.error` and this can be overridden via `spring.cloud.stream.errorDestination`
     - This destination name can be used to subscribe to the error channel output messages.
  - Add test

This resolves #329


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -62,7 +66,7 @@
  * @see EnableBinding
  */
 public class BindableProxyFactory implements MethodInterceptor, FactoryBean<Object>, Bindable, BeanFactoryAware,
```
mbogoevici: The BindableProxyFactory is created for each bound interface. I don't think that the error binding should happen within this class.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -260,12 +269,23 @@ public void bindOutputs(ChannelBindingService channelBindingService) {
...
 			}
 		}
+		//Make sure to bind error channel as a message producer only if there is at least a single output bindable exists.
+		//i.e. binding error channel of the  outbound endpoints.
+		if (this.outputBindableExists) {
+			bindErrorChannel(channelBindingService);
+		}
+	}
+
+	private void bindErrorChannel(ChannelBindingService channelBindingService) {
```
mbogoevici: Maybe we can move this into a standalone Bindable implementation, `ConditionalOnProperty("spring.cloud.stream.bindings.error.destination"))`

ilayaperumalg:
Updated the PR with the approach to bind error channel via implicit `Bindable` for the error channel.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableErrorChannel.java
```diff
@@ -0,0 +1,45 @@
...
+package org.springframework.cloud.stream.binding;
+
+import org.springframework.integration.channel.PublishSubscribeChannel;
+
+/**
+ * A {@link Bindable} component that represents an error channel.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class BindableErrorChannel extends NoopBindable {
```
mbogoevici: Should return the error channel via `getOutputs`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.java
```diff
@@ -69,6 +73,12 @@
 @EnableConfigurationProperties(ChannelBindingServiceProperties.class)
 public class ChannelBindingServiceConfiguration {
 
+	private static final String ERROR_CHANNEL_NAME = "error";
+
+	@Autowired
+	@Qualifier(IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME)
```
mbogoevici: can we just provide this as an argument to the bean creation method?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/NoopBindable.java
```diff
@@ -0,0 +1,53 @@
...
+package org.springframework.cloud.stream.binding;
+
+import java.util.Set;
+
+/**
+ * A No-op Bindable implementation.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class NoopBindable implements Bindable {
```
mbogoevici: We should call this BindableAdapter (following the naming patterns of default implementations in Spring projects).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/NoopBindable.java
```diff
@@ -0,0 +1,53 @@
...
+	public void unbindInputs(ChannelBindingService adapter) {
+	}
+
+	@Override
+	public void unbindOutputs(ChannelBindingService adapter) {
+	}
+
+	@Override
+	public Set<String> getInputs() {
+		return null;
```
mbogoevici: Should return an empty collection rather.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/NoopBindable.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	@Override
+	public Set<String> getOutputs() {
+		return null;
```
mbogoevici: Should return an empty collection.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/NoopBindable.java
ilayaperumalg: ok

ilayaperumalg:
Updated the PR based on comments and rebased as well.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
```diff
@@ -31,7 +31,7 @@
 import org.springframework.integration.config.EnableIntegration;
 
 /**
- * Enables the binding of inputs and outputs to a broker, according to the list
+ * Enables the binding of {@link Input} and {@link Output} message channels to a broker, according to the list
```
markfisher: I think we intentionally removed the mention of "channels" here earlier.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -260,7 +260,7 @@ public void bindOutputs(ChannelBindingService channelBindingService) {
 			ChannelHolder channelHolder = channelHolderEntry.getValue();
 			String outputChannelName = channelHolderEntry.getKey();
 			if (channelHolderEntry.getValue().isBindable()) {
-				if (log.isDebugEnabled()) {
+			if (log.isDebugEnabled()) {
```
markfisher: fix this indent

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -39,6 +39,7 @@
 @JsonInclude(Include.NON_DEFAULT)
 public class ChannelBindingServiceProperties {
 
+
```
markfisher: no need for this extra line here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+	public void unbindInputs(ChannelBindingService adapter) {
+	}
+
+	@Override
+	public void unbindOutputs(ChannelBindingService adapter) {
+	}
+
+	@Override
+	public Set<String> getInputs() {
+		return new HashSet<String>();
```
markfisher: I believe the return values from `getInputs` and `getOutputs` should be empty *unmodifiable* collections.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ErrorChannelTests.java
```diff
@@ -0,0 +1,140 @@
...
+
+	@EnableBinding(ErrorSink1.class)
+	@EnableAutoConfiguration
+	@Import(RedisMessageChannelBinderConfiguration.class)
+	@PropertySource("classpath:/org/springframework/cloud/stream/config/errorchannel/errorsink1-channel.properties")
+	public static class TestErrorSink1 {
+
+	}
+
+	public interface ErrorSink1 {
```
markfisher: I don't see any reason for adding a new sink interface here, the OOTB `Sink` would be fine... just binding to the same name that the outbound error binding specifies if I understand correctly.

ilayaperumalg: I intentionally added these couple of interfaces to have it look more `generic`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
ilayaperumalg: ok, that was my first impression too. I think `@Input` and `@Output` annotations are always used to denote the message channels in our implementation. Hence I added the same here as well.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ErrorChannelTests.java
markfisher: But what's the point? As far as I understand, the test is just ensuring that the outbound error channel binding works. In my opinion, adding this just complicates things unnecessarily.

ilayaperumalg: ok, I can revert this to use the default `Sink` interface.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
markfisher: it's fine to refer to those annotations but should leave out "message channels" probably - wdyt @mbogoevici ?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ErrorChannelTests.java
ilayaperumalg: Now that I recall, there is one caveat here. Since we wanted to test this for multiple destinations (error channel being pub/sub), we have to create one more interface with the `@Input` annotation but with a different name than **input**. I think that was the original reason behind adding multiple explicit interfaces here. May be I will keep the default `Sink` and create just one more?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
mbogoevici: We do refer to those annotations in the `values()` method, but I suppose we can use the same here. "Enables the binding of {@link Input} and {@link Output} - annotated components to a broker".

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ErrorChannelTests.java
markfisher: Why? We test pub/sub behavior for outbound bindings elsewhere. The error binding is just another outbound binding. So, a "pub sub" consumer test does not add anything new.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
mbogoevici: (we use "components" in "value()" and if we decide on something else here, it should be consistent)

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ErrorChannelTests.java
ilayaperumalg: yeah, ok :-)

mbogoevici: Yeah, I think we can safely rely on the fact that if multi-destination binding works here it should work elsewhere.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ErrorChannelTests.java
```diff
@@ -0,0 +1,140 @@
...
+		String INPUT = "errorInput1";
+
+		@Input(ErrorSink1.INPUT)
+		SubscribableChannel errorInput();
+
+	}
+
+	@EnableBinding(ErrorSink2.class)
+	@EnableAutoConfiguration
+	@Import(RedisMessageChannelBinderConfiguration.class)
```
mbogoevici: Also, let's not rely on specific binder implementations. We can use a mock binder here and just test that the channel has been bound.

mbogoevici: Based on the assumption that once a channel has been bound it will send messages.

ilayaperumalg: This test is specifically added in `integration` tests to test the end-to-end scenario. May be I can add one in unit test that tests just the binding?

mbogoevici: I see it's an integration test though - can we use the test binder instead?

mbogoevici: Ok, let's add a unit test in the core, and we can revise what kind of binder gets used in ITs later. I'm keen on using the test binder on the long run.

ilayaperumalg: I think it is better to test against the real binder implementation when running on `integration` tests.

mbogoevici: I disagree for the following reasons: these tests should always run, regardless whether the target middleware is started or not, and we want to encourage users to use the `TestBinder` for their own testing, as it is supposed the encapsulate the general contract for the binder. If we want binder-specific tests then we should put them in the `AbstractBinderTest` hierarchy and test them across the board. However, this is beyond the scope of this PR.

markfisher: I agree with @mbogoevici. They are "integration" tests as long as any binder is involved. The `TestBinder` *must* be sufficient for such tests, so that any binder would be guaranteed to work - as long as each binder implementation satisfies the `AbstractBinderTest`-based "compatibility test" suite (which should probably become more formalized with time).

Furthermore, even if do continue to host binder implementations, such as Kafka and Rabbit, within this repository, we know that other binder implementations will exist outside of the scope of this repository. In theory, (this will take some time to achieve but should a goal IMO), we should be able to move all but the `TestBinder` out of this repository and not have any dependencies on other binder(s).

ilayaperumalg: ok, sure. Created https://github.com/spring-cloud/spring-cloud-stream/issues/380 to track this.

ilayaperumalg:
Updated the PR to address the latest review comments.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+	public void unbindInputs(ChannelBindingService adapter) {
+	}
+
+	@Override
+	public void unbindOutputs(ChannelBindingService adapter) {
+	}
+
+	@Override
+	public Set<String> getInputs() {
+		return Collections.unmodifiableSet(Collections.EMPTY_SET);
```
mbogoevici: EMPTY_SET is already immutable.

ilayaperumalg: oh, yeah. fixed it.

mbogoevici:
Rebased, squashed, merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/cae56385e0356011e64396df2245b8eeb963d5fa and additional changes in https://github.com/spring-cloud/spring-cloud-stream/commit/5e2e4ecd6796f965ae6405970e5ed2bb289cbcb3

PR #372, opened 2016-02-24
Make AggregateApplicationBuilder more fluent

dsyer:
So you can add a parent for instance, like this:

```
new AggregateApplicationBuilder().from(TwitterStreamApplication.class)
	.args(new String[] { "--language=en" }).to(StreamTransformer.class)
	.parent(
		SpringApplication.run(ModuleApplication.class, args))
	.run();
```

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
@@ -41,6 +41,8 @@ public class AggregateApplicationBuilder {
 
 	private AggregateApplicationBuilder applicationBuilder = this;
 
+	ConfigurableApplicationContext parentContext;
+
 	public SourceConfigurer from(Class<?> app) {
 		SourceConfigurer sourceConfigurer = new SourceConfigurer(app);
 		this.sourceConfigurer = sourceConfigurer;
@@ -48,9 +50,12 @@ public class AggregateApplicationBuilder {
 	}
 
 	public void run(String[] parentArgs) {
-		ConfigurableApplicationContext parentContext = AggregateApplication.createParentContext(parentArgs);
-		SharedChannelRegistry sharedChannelRegistry = parentContext.getBean(SharedChannelRegistry.class);
-		List<AppConfigurer> apps = new ArrayList<AppConfigurer>();
+		ConfigurableApplicationContext parentContext = this.parentContext != null
+				? this.parentContext
+				: AggregateApplication.createParentContext(parentArgs);
+		SharedChannelRegistry sharedChannelRegistry = parentContext
+				.getBean(SharedChannelRegistry.class);
+		List<AppConfigurer<?>> apps = new ArrayList<AppConfigurer<?>>();
 		if (this.sourceConfigurer != null) {
 			apps.add(sourceConfigurer);
 		}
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -48,9 +50,12 @@ public SourceConfigurer from(Class<?> app) {
 	}
 
 	public void run(String[] parentArgs) {
-		ConfigurableApplicationContext parentContext = AggregateApplication.createParentContext(parentArgs);
-		SharedChannelRegistry sharedChannelRegistry = parentContext.getBean(SharedChannelRegistry.class);
-		List<AppConfigurer> apps = new ArrayList<AppConfigurer>();
+		ConfigurableApplicationContext parentContext = this.parentContext != null
+				? this.parentContext
+				: AggregateApplication.createParentContext(parentArgs);
+		SharedChannelRegistry sharedChannelRegistry = parentContext
```
ilayaperumalg: I think we handle the scenario where the `parentContext` (when set explicitly) doesn't have the `SharedChannelRegistry` bean.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -164,7 +125,7 @@ public ProcessorConfigurer via(Class<?> processor) {
 
 	}
 
-	private abstract class AppConfigurer {
+	public abstract class AppConfigurer<T extends AppConfigurer<T>> {
```
ilayaperumalg: Is this meant to be just `AppConfigurer<T>` unless I miss something very basic :-)

artembilan: With just the `T` you won't be able to chain methods like in the `embed()` method.
Just try it locally :smile:

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
mbogoevici: If there's not `SharedChannelRegistry`, then channels will not be shared, I believe. It shouldn't be fundamentally hard to do a `registerSingleton` in the parent though.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -174,20 +135,40 @@ public ProcessorConfigurer via(Class<?> processor) {
...
+		}
+
+		public T profiles(String... profiles) {
+			this.profiles = profiles;
+			return getConfigurer();
+		}
+
+		public T parent(ConfigurableApplicationContext parentContext) {
+			AggregateApplicationBuilder.this.parentContext = parentContext;
+			return getConfigurer();
```
mbogoevici: We can register a `SharedChannelRegistry` here if not already defined. Or we can generalize the problem and support shared channels as via `@EnableBinding` and use a `@ConditionalOnMissingBean` with parent lookup for preventing them to be created in the children.

dsyer: That's not necessarily a feature of this PR, though. I'm happy if you throw the PR away, as long as I get something equivalent.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -174,20 +135,40 @@ public ProcessorConfigurer via(Class<?> processor) {
...
+			this.args = args;
+			return getConfigurer();
+		}
+
+		public T profiles(String... profiles) {
+			this.profiles = profiles;
+			return getConfigurer();
+		}
+
+		public T parent(ConfigurableApplicationContext parentContext) {
```
mbogoevici: I'd be tempted to move this to the `AggregateApplicationBuilder` itself since it cannot be meaningfully invoked on each child module.

mbogoevici: I'd be tempted to move this to the `AggregateApplicationBuilder` itself since it cannot be meaningfully invoked on each child module.

mbogoevici:
@dsyer Here's my proposed change to AbstractApplicationBuilder https://github.com/mbogoevici/spring-cloud-stream/commit/995e268da51ce89fc57e695707bb2c9fe846ebe7

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/22f5b7a15e47f53ae59a2218a59da99ad16d4ed3 with additional changes here:https://github.com/spring-cloud/spring-cloud-stream/commit/995e268da51ce89fc57e695707bb2c9fe846ebe7

PR #378, opened 2016-02-25
Solve problem proxy without authentication

Mwea:


ilayaperumalg:
Hi @Mwea,
 
 Thanks for the PR. From the change list, it is hard to find the `diff` because of formatting changes in the entire file. Can you please update the PR without those formatting changes?

Also, please confirm if you have signed the contributor agreement: https://support.springsource.com/spring_committer_signup

Thanks again!

artembilan:
@ilayaperumalg  (and others),

to make your life easier in the GitHub for such a case you can review the changes with the `?w=1` HTTP request parameter in the `/files` tab. E.g. for this PR it would be like:
https://github.com/spring-cloud/spring-cloud-stream/pull/378/files?w=1

FYI: https://github.com/blog/967-github-secrets :smile: 

But yes, @Mwea , we prefer `tabs` not `whitespaces` for indents in Java code.

Thank you for the contribution, of course!

ilayaperumalg:
@artembilan wow, thanks for the guidance. it is so cool!

mbogoevici:
We should apply this to the SCDF app launcher in fact, since the module launcher will move out of here.

mbogoevici:
Hi @Mwea,

Thanks for the contribution. We've moved the code from this repository to here: https://github.com/spring-cloud/spring-cloud-dataflow/tree/master/spring-cloud-dataflow-app-launcher - so would you mind re-issuing a PR against that repository, please?

We would also need the contributor agreement signed, and it would be best if you'd use our formatting (tabs not spaces) - Also, the only comment is https://github.com/Mwea/spring-cloud-stream/commit/105ab95c54be4b613eddfc2417f0f001ccc08018#commitcomment-16343795 - we'd need to check if a proxy is configured at all, otherwise this will fail when no proxy is configured.

Thanks!

ilayaperumalg:
Based on the discussion above, the intended fix is applied [here] (https://github.com/spring-cloud/spring-cloud-dataflow/commit/b3e4d2f63b08d5fcfcccab11932a311350561f77)

PR #385, opened 2016-02-29
Allow setting ChannelCacheSize in RabbitMQ binder

### LINKED ISSUE #368: Rabbit Binder ChannelCacheSize Property
As a user, I would like to configure the `channelCacheSize` property on the `CachingConnectionFactory`.

http://stackoverflow.com/questions/35563064/processing-messages-through-namedchannels-with-prefetch-1/35584333#35584333

ilayaperumalg:
- Set ChannelCacheSize property for the binder's CacheConnectionFactory

- Add test

This resolves #368

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitMessageChannelBinderConfiguration.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitMessageChannelBinderConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,14 @@
 package org.springframework.cloud.stream.binder.rabbit.config;
 
 /**
+ * Configuration class for RabbitMQ message channel binder.
+ *
  * @author David Turanski
+ * @author Ilayaperumal Gopinathan
  */
 
 import org.springframework.amqp.core.MessagePostProcessor;
+import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.support.postprocessor.DelegatingDecompressingPostProcessor;
 import org.springframework.amqp.support.postprocessor.GZipPostProcessor;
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitMessageChannelBinderConfiguration.java
```diff
@@ -45,15 +49,19 @@
...
+
 	@Autowired
 	private SpringRabbitMQProperties springRabbitMQProperties;
 
 	@Bean
 	RabbitMessageChannelBinder rabbitMessageChannelBinder() {
+		if (rabbitConnectionFactory instanceof CachingConnectionFactory) {
+			((CachingConnectionFactory) rabbitConnectionFactory).
+					setChannelCacheSize(springRabbitMQProperties.getChannelCacheSize());
+		}
```
garyrussell: This just seems wrong to me (reconfiguring Boot's object) - given that we are already overriding the `RabbitProperties` from boot auto-configure, we should really disable the boot auto-config altogether here and create our own factory.

For example, the connection factory `addresses` is populated from `RabbitProperties` but the binder `addresses` is populated from `SpringRabbitMQProperties`. Yes, the source of the property is the same but it's a bit smelly IMHO.

Either `SpringRabbitMQProperties` should just have those properties over and above those in `RabbitProperties` or we should do all the wiring here.

I think I prefer the latter, at least until (or unless) the `channelCachesize` property is added to boot.

I am just not keen on reconfiguring boot's object here.

ilayaperumalg:
+1 for addressing this in `spring-boot` itself as I felt the same when working on this.

mbogoevici:
Can we just close this until Boot supports it then?

garyrussell:
Boot PR: https://github.com/spring-projects/spring-boot/pull/5303

ilayaperumalg:
Given the boot PR https://github.com/spring-projects/spring-boot/pull/5303 is scheduled for **1.4.0.M2**, are we okay waiting for this fix post GA as I believe we can not depend on milestone version of boot?

Re-opening this PR to clarify the question above.

mbogoevici:
@garyrussell can confirm or deny, but this doesn't seem to be a particular issue that Stream should worry about - we don't do request reply scenarios, and we don't open temporary channels all that often. If anything, we can set up a decent default until Boot supports it.

garyrussell:
@mbogoevici No; it has nothing to do with request/reply; the cache is for channels to talk to RabbitMQ over, not queues (similar to a JMS Session).

The problem is on each send a channel is borrowed from the cache; consumer channels are long-lived but producer channels are very short-lived. The default cache size is only 1 so, under extreme conditions, we end up creating/destroying many channels instead of caching them.

garyrussell:
The BOOT PR completely solves it, so it's not really an issue for Stream per se; but I would like to see us clean up the binder config (duplicate properties) before RC, though.

mbogoevici:
Ok, sorry - I was confused by the initial issue originating in an XD job launch scenario. I would close this as is though and keep just https://github.com/spring-cloud/spring-cloud-stream/pull/387 for the binder config.

garyrussell:
Do you mean to augment that PR to remove the duplicate properties?

mbogoevici:
Yes, or create a new one for removing the duplicates specifically. This issue can be closed IMO.

PR #387, opened 2016-03-01
Remove ConnectionFactorySettings configuration class

### LINKED ISSUE #386: Remove ConnectionFactorySettings configuration class
Currently, the rabbit `ConnectionFactorySettings` **configuration** class is loaded as `bean` which looks incorrect [here] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitMessageChannelBinderConfiguration.java#L98). 

Also, the properties (useSSL, sslPropertiesLocation) that are set in `ConnectionFactorySettings` are appropriately set via `RabbitMessageChannelBinder` bean creation.

Hence, it is better to remove this configuration class.

ilayaperumalg:
This resolves #386


ilayaperumalg:
Irrespective of the changes that will be added into `spring-boot`, I believe the `ConnectionFactorySettings` class in SCSt is currently not in use/not configured appropriately (currently it is invoked with `@Bean` annotation).

Given we always configure the `useSSL` and `sslPropertiesLocation` in the connectionFactory at the `RabbitMessageChannelBinderConfiguration`, I think this PR still makes sense (to remove this configuration class completely).

garyrussell:
Agreed.

garyrussell:
I thought we agreed (in #385) that we'd use this PR to resolve the duplicate properties in the RabbitBinder - i.e. also get rid of `SpringRabbitMQProperties`.

ilayaperumalg:
@garyrussell ok, sorry. I misunderstood. I will update the PR with those changes.

ilayaperumalg:
@garyrussell It looks like the properties such as `adminAddresses`, `nodes` and `sslPropertiesLocation` are still missing from boot. Do you think they need to be addressed in `spring-boot` as RabbitProperties?

garyrussell:
We need to configure `adminAddresses` and `nodes` but we can do away with `sslPropertiesLocation` since the properties therein are now top-level props.

My biggest concern here is the duplication.

Thanks

ilayaperumalg:
I will update this PR once the PR https://github.com/spring-cloud/spring-cloud-stream/pull/409 is merged. Hope that makes sense.

mbogoevici:
Sounds good!

ilayaperumalg:
This PR changes are applied in this branch: https://github.com/ilayaperumalg/spring-cloud-stream/tree/PR-387 after rebasing the #409

ilayaperumalg:
With the changes from #409, I believe the RabbitBinder configuration properties doesn't have duplicates now. I think `sslPropertiesLocation` still needs to be supported. @garyrussell Is this property also needs to be available in spring-boot ?

garyrussell:
I thought I already left a comment to this - I don't think we need sslPropertiesLocation - boot made the properties from that file "top level" properties.

mbogoevici:
I would suppose we can also remove `useSSL` as redundant then? Seing as `spring.rabbitmq.ssl.enabled` is a Boot property as well.

mbogoevici:
Rebased and merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/6a125b601b8172c7e9da8ad063f881e36a02b2fc

PR #389, opened 2016-03-01
Unbind dynamic producers

### LINKED ISSUE #370: Ensure that dynamic producer bindings are unbound properly

ilayaperumalg:
- Add `SmartLifecycle` to BinderAwareChannelResolver
  - Add resolved destinations' bindings to the list of `Bindings` that need to be onbound
  - Add lowest phase for its lifecycle so that bindings are unbound at the latest

This resolves #370

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
@@ -42,8 +42,6 @@ import org.springframework.util.ObjectUtils;
  */
 public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestinationResolver {
 
-	private final ChannelBindingService channelBindingService;
-
 	private final BinderFactory<MessageChannel> binderFactory;
 
 	private final ChannelBindingServiceProperties channelBindingServiceProperties;
@@ -52,11 +50,11 @@ public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestina
 
 	private ConfigurableListableBeanFactory beanFactory;
 
-	public BinderAwareChannelResolver(ChannelBindingService channelBindingService, DynamicBindable dynamicBindable) {
-		Assert.notNull(channelBindingService.getBinderFactory(), "'binderFactory' cannot be null");
-		this.channelBindingService = channelBindingService;
-		this.binderFactory = channelBindingService.getBinderFactory();
-		this.channelBindingServiceProperties = channelBindingService.getChannelBindingServiceProperties();
+	public BinderAwareChannelResolver(BinderFactory binderFactory,
+			ChannelBindingServiceProperties channelBindingServiceProperties, DynamicBindable dynamicBindable) {
+		Assert.notNull(binderFactory, "'binderFactory' cannot be null");
+		this.binderFactory = binderFactory;
+		this.channelBindingServiceProperties = channelBindingServiceProperties;
 		this.dynamicBindable = dynamicBindable;
 	}
 
@@ -69,11 +67,11 @@ public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestina
 	}
 
 	@Override
-	public MessageChannel resolveDestination(String name) {
+	public MessageChannel resolveDestination(String destinationName) {
 		MessageChannel channel = null;
 		DestinationResolutionException destinationResolutionException;
 		try {
-			return super.resolveDestination(name);
+			return super.resolveDestination(destinationName);
 		}
 		catch (DestinationResolutionException e) {
 			destinationResolutionException = e;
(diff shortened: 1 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -35,10 +40,12 @@
  * resolves the channel from the bean factory and, if not present, creates a new channel
  * and adds it to the factory after binding it to the binder. The binder is optionally
  * determined with a prefix preceding a colon.
+ *
  * @author Mark Fisher
  * @author Gary Russell
+ * @author Ilayaperumal Gopinathan
  */
-public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestinationResolver {
+public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestinationResolver implements SmartLifecycle {
```
mbogoevici: I think this would be much simpler if made bindable and obeyed the `unbindOutputs` contract (the other methods would be no-op since there are no inputs nor predefined outputs)

ilayaperumalg: It looks like we may not make `BinderAwareChannelResolver` as Bindable since it would create unresolvable circular reference.

For instance, while `BinderAwareChannelResolver` is required as a bean for destination resolver the `ChannelBindingAutoConfiguration` requires all the `Bindable`s in it.

mbogoevici: Maybe I am missing something, but `BinderAwareChannelResolver` is created  by `ChannelBindingServiceConfiguration`, and only depends on the `BinderFactory` and the `ChannelBindingServiceProperties` whereas  `Bindable`s are referenced in `ChannelBindingAutoConfiguration`, which happens later. We already have a `Bindable` created in `ChannelBindingServiceConfiguration`. Not sure where the circular dependency is coming from.

mbogoevici: As a side note, we need to find better names for the configuration classes.

ilayaperumalg: Here is the circular reference:

Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'binderAwareChannelResolver' defined in class path resource [org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.class]: Unsatisfied dependency expressed through constructor argument with index 0 of type [org.springframework.cloud.stream.binder.BinderFactory]: : Error creating bean with name 'binderFactory': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanCreationException: Could not autowire method: public void org.springframework.cloud.stream.binder.DefaultBinderFactory.setBindersHealthIndicator(org.springframework.boot.actuate.health.CompositeHealthIndicator); nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.springframework.cloud.stream.config.ChannelBindingAutoConfiguration': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanCreationException: Could not autowire field: private java.util.List org.springframework.cloud.stream.config.ChannelBindingAutoConfiguration.adapters; nested exception is org.springframework.beans.factory.BeanCurrentlyInCreationException: Error creating bean with name 'binderAwareChannelResolver': Requested bean is currently in creation: Is there an unresolvable circular reference?; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'binderFactory': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanCreationException: Could not autowire method: public void org.springframework.cloud.stream.binder.DefaultBinderFactory.setBindersHealthIndicator(org.springframework.boot.actuate.health.CompositeHealthIndicator); nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.springframework.cloud.stream.config.ChannelBindingAutoConfiguration': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanCreationException: Could not autowire field: private java.util.List org.springframework.cloud.stream.config.ChannelBindingAutoConfiguration.adapters; nested exception is org.springframework.beans.factory.BeanCurrentlyInCreationException: Error creating bean with name 'binderAwareChannelResolver': Requested bean is currently in creation: Is there an unresolvable circular reference?

mbogoevici: I see. I didn't count on the HealthIndicators being produced by an instance method of the same `ChannelBindingAutoConfiguration`. I suppose having a separate bean for holding the dynamic bindings (a collection of them) would break the circular dependency if we injected it into the `BinderAwareChannelResolver` instead?

ilayaperumalg:
Updated the PR that uses `BindableAdapter` to handle the unbinding.

The changes include:
    
- Add `DyanamicBindable` which is a BindableAdapter that stores the dynamic destination names and handle their unbinding
- Inject `ChannelBindingService` into BinderAwareChannelResolver so that the dynamic binding gets registered with `producerBindings` and can subsequently be used when unbinding
- Update tests

ilayaperumalg:
@mbogoevici based on our discussion, updated the PR that uses `Binding.unbind` instead of involving `ChannelBindingService`.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/66a78bf68530ad2a26f92f7f45332ca463547f51

PR #394, opened 2016-03-03
Set string value for originalContentType in message header

### LINKED ISSUE #383: Explicit content type in bindings configuration broken for redis
If you use `spring.cloud.stream.bindings.output.content-type: application/json` (for instance) with a Redis transport, the message header gets converted to JSON in `EmbeddedHeadersMessageConverter`, and the `MimeType` is converted to a JSON object (not just the String representation). When it lands in the consumer there is an exception because it tries to convert back from JSON with no type information:

```
Caused by: java.lang.IllegalArgumentException: Unknown type for contentType header value: class java.util.LinkedHashMap
	at org.springframework.messaging.converter.DefaultContentTypeResolver.resolve(DefaultContentTypeResolver.java:69) ~[spring-messaging-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.messaging.converter.AbstractMessageConverter.getMimeType(AbstractMessageConverter.java:237) ~[spring-messaging-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.messaging.converter.AbstractMessageConverter.supportsMimeType(AbstractMessageConverter.java:224) ~[spring-messaging-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.messaging.converter.AbstractMessageConverter.canConvertFrom(AbstractMessageConverter.java:179) ~[spring-messaging-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.cloud.stream.converter.AbstractFromMessageConverter.canConvertFrom(AbstractFromMessageConverter.java:146) ~[spring-cloud-stream-1.0.0.BUILD-SNAPSHOT.jar:1.0.0.BUILD-SNAPSHOT]
	at org.springframework.messaging.converter.AbstractMessageConverter.fromMessage(AbstractMessageConverter.java:172) ~[spring-messaging-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.messaging.converter.AbstractMessageConverter.fromMessage(AbstractMessageConverter.java:167) ~[spring-messaging-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.messaging.converter.CompositeMessageConverter.fromMessage(CompositeMessageConverter.java:55) ~[spring-messaging-4.2.5.RELEASE.jar:4.2.5.RELEASE]
	at org.springframework.integration.channel.AbstractMessageChannel.convertPayloadIfNecessary(AbstractMessageChannel.java:482) [spring-integration-core-4.2.5.RELEASE.jar:na]
	at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:426) [spring-integration-core-4.2.5.RELEASE.jar:na]
	... 30 common frames omitted
```

It works with Rabbit, I guess because the transport has explicit support for headers, so there is no coercion to JSON and back again.

ilayaperumalg:
- Since the `originalContentType` header is preserved across the wire while the serialization/de-serialization processes take place using the deduced `contentType`,
it is convenient to set the `String` value of the `originalContentType` instead of setting the actual `MimeType` object in the header.
 - Once the `de-serialization` has taken place, the deduced `contentType` is always replaced with the `originalContentType` as the `contentType` in the message header and this preserved contentType can be resolved to its corresponding MimeType using `ContentTypeResolver` anytime. Hence, using the `String` object for `originalContentType` doesn't do any harm.

 - Fix the issue while setting the deduced contentType based on the `payload` when the `originalContentType` is `application/json`
   - Currently, the contentType is set to `text/plain` which seems incorrect
 - Also, add the `application/json` contentType to return `String` object during de-serialization process.

This resolves #383

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.stream.binder;
 
-import static org.springframework.util.MimeTypeUtils.APPLICATION_JSON;
 import static org.springframework.util.MimeTypeUtils.APPLICATION_OCTET_STREAM;
 import static org.springframework.util.MimeTypeUtils.TEXT_PLAIN;
 
@@ -364,7 +363,7 @@ public abstract class AbstractBinder<T> implements ApplicationContextAware, Init
 		Object originalContentType = message.getHeaders().get(MessageHeaders.CONTENT_TYPE);
 
 		//Pass content type as String since some transport adapters will exclude CONTENT_TYPE Header otherwise
-		Object contentType = JavaClassMimeTypeConversion.mimeTypeFromObject(originalPayload, originalContentType).toString();
+		Object contentType = JavaClassMimeTypeConversion.mimeTypeFromObject(originalPayload).toString();
 		Object payload = serializePayloadIfNecessary(originalPayload);
 		MessageValues messageValues = new MessageValues(message);
 		messageValues.setPayload(payload);
@@ -426,7 +425,7 @@ public abstract class AbstractBinder<T> implements ApplicationContextAware, Init
 	}
 
 	private Object deserializePayload(byte[] bytes, MimeType contentType) {
-		if (TEXT_PLAIN.equals(contentType) || APPLICATION_JSON.equals(contentType)) {
+		if (TEXT_PLAIN.equals(contentType)) {
 			try {
 				return new String(bytes, "UTF-8");
 			}
@@ -546,16 +545,12 @@ public abstract class AbstractBinder<T> implements ApplicationContextAware, Init
 
 		private static ConcurrentMap<String, MimeType> mimeTypesCache = new ConcurrentHashMap<>();
 
-		static MimeType mimeTypeFromObject(Object payload, Object originalContentType) {
+		static MimeType mimeTypeFromObject(Object payload) {
 			Assert.notNull(payload, "payload object cannot be null.");
 			if (payload instanceof byte[]) {
 				return MimeTypeUtils.APPLICATION_OCTET_STREAM;
 			}
 			if (payload instanceof String) {
-				if (originalContentType != null &&
-						originalContentType.toString().equals(MimeTypeUtils.APPLICATION_JSON_VALUE)) {
-					return MimeTypeUtils.APPLICATION_JSON;
-				}
 				return MimeTypeUtils.TEXT_PLAIN;
 			}
 			String className = payload.getClass().getName();
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -16,10 +16,9 @@
 
 package org.springframework.cloud.stream.binder;
 
+import static org.springframework.util.MimeTypeUtils.APPLICATION_JSON;
 import static org.springframework.util.MimeTypeUtils.APPLICATION_OCTET_STREAM;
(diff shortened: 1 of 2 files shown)


dsyer:
LGTM. Tested with a simple source/sink combo and a POJO payload.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -538,31 +538,31 @@ protected RetryTemplate buildRetryTemplateIfRetryEnabled(DefaultBindingPropertie
...
-
-		public static final MimeType TEXT_PLAIN_MIME_TYPE = MimeType.valueOf(TEXT_PLAIN_VALUE);
-
 		private static ConcurrentMap<String, MimeType> mimeTypesCache = new ConcurrentHashMap<>();
 
-		static MimeType mimeTypeFromObject(Object obj) {
-			Assert.notNull(obj, "object cannot be null.");
-			if (obj instanceof byte[]) {
-				return APPLICATION_OCTET_STREAM_MIME_TYPE;
+		static MimeType mimeTypeFromObject(Object payload, Object originalContentType) {
```
mbogoevici: I'm wondering if we can avoid passing an argument here, and move the JSON test outside of it. It seems to imply that the originalContentType plays a bigger role than just preserving the APPLICATION_JSON content type if already set.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -538,31 +538,31 @@ protected RetryTemplate buildRetryTemplateIfRetryEnabled(DefaultBindingPropertie
...
+			Assert.notNull(payload, "payload object cannot be null.");
+			if (payload instanceof byte[]) {
+				return MimeTypeUtils.APPLICATION_OCTET_STREAM;
 			}
-			if (obj instanceof String) {
-				return TEXT_PLAIN_MIME_TYPE;
+			if (payload instanceof String) {
+				if (originalContentType != null &&
+						originalContentType.toString().equals(MimeTypeUtils.APPLICATION_JSON_VALUE)) {
```
mbogoevici: Trying to understand why this is necessary. Since the original content type will replace the CONTENT_TYPE header in the deserialized message, the only function of the `text/plain` content type header is to signal that the payload must be deserialized as a String, before the header is replaced with the original `application/json` value.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -426,7 +426,7 @@ private Object deserializePayload(Object payload, MimeType contentType) {
 	}
 
 	private Object deserializePayload(byte[] bytes, MimeType contentType) {
-		if (TEXT_PLAIN.equals(contentType)) {
+		if (TEXT_PLAIN.equals(contentType) || APPLICATION_JSON.equals(contentType)) {
```
mbogoevici: This is only necessary because of https://github.com/spring-cloud/spring-cloud-stream/pull/394/files#r54971080, which may not be necessary itself. I believe we can just rely on `text/plain`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
ilayaperumalg: Yes, we can avoid this as the payload type is `String` anyway. My intention was to be consistent with the `contentType` we infer from the payload.

ilayaperumalg:
Updated the PR to use `text/plain` contentType(the intermediate contentType that is used for serialization/de-serialization) when the payload is of type `String`.

mbogoevici:
Squashed, rebased, merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/7021d2195d8431d3f89318aac8065cd4e934d02b

PR #395, opened 2016-03-03
Remove XD references in SCSt code base

ilayaperumalg:


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -92,10 +93,8 @@
 import scala.collection.Seq;
 
 /**
- * A binder that uses Kafka as the underlying middleware. The general implementation mapping between XD concepts
- * and Kafka concepts is as follows:
- * A binder that uses Kafka as the underlying middleware.
- * The general implementation mapping between XD concepts and Kafka concepts is as follows:
+ * A {@link Binder} that uses Kafka as the underlying middleware.
```
ilayaperumalg: I guess we can probably remove this javadoc completely as we may not want `dataflow` references in SCSt too.

markpollack:
Removed a few more refs and merged as 6acb825ad58107c42d5cc3ea7bc84035ef750315

PR #398, opened 2016-03-04
Set message header using BindingProperties' contentType

### LINKED ISSUE #397: At the producer, set message contentType header to channel's content-type property value (if available)
At the message producer side, if the message doesn't have `contentType` header, set the `content-type` header to the channel's content-type (if set via binding property).

ilayaperumalg:
- Add a `MessageChannelConfigurer` that uses a channel interceptor to check if the message contentType header is missing and if there is a binding property for `contentType` header is provided, then set that value as the message contentType header

- This resolves #397

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ContentTypeConfigurerTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/ContentTypeConfigurerTests.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
Updated the PR based on the suggestion from @mbogoevici that uses `ChannelInterceptor` to handle the message contentType header from binding properties of the channel.

ilayaperumalg:
This PR changes are rebased on top of the PR #409 here: https://github.com/ilayaperumalg/spring-cloud-stream/tree/PR-398

ilayaperumalg:
Rebased the PR after updating the change to use `MessageConverterConfigurer` to configure the `contentType` instead of using a separate Configurer.

PR #402, opened 2016-03-07
Support text/* contentType sent by non-SCSt applications

### LINKED ISSUE #403: Support text/* content type for non-SCSt applications
When deserializing the payload at the consumer endpoint, the non-byte stream payload type requires to use `String` object when the underlying message content-type is of any `text` type contentType (text/plain, text/xml and text/html)

ilayaperumalg:
- When deserializing the payload at the consumer endpoint, the non-byte stream payload type requires to use `String` object when the underlying message content-type is of any `text` type contentType (text/plain, text/xml and text/html).
 - This fix is only needed to support any non-SCSt applications that will have the 'text/*` contentType of the deserializing message
 - Add test

This resolves #403

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/test/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupportTests.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/test/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupportTests.java
@@ -112,6 +112,19 @@ public class MessageChannelBinderSupportTests {
 	}
 
 	@Test
+	public void testStringXML() throws IOException {
+		Message<?> message = MessageBuilder
+				.withPayload("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\"?><test></test>")
+				.setHeader(MessageHeaders.CONTENT_TYPE, MimeTypeUtils.TEXT_XML)
+				.build();
+		Message<?> converted = binder.serializePayloadIfNecessary(message).toMessage();
+		assertEquals(MimeTypeUtils.TEXT_PLAIN, contentTypeResolver.resolve(converted.getHeaders()));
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/test/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupportTests.java
```diff
@@ -112,6 +112,19 @@ public void testString() throws IOException {
 	}
 
 	@Test
+	public void testStringXML() throws IOException {
+		Message<?> message = MessageBuilder
+				.withPayload("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\"?><test></test>")
+				.setHeader(MessageHeaders.CONTENT_TYPE, MimeTypeUtils.TEXT_XML)
+				.build();
+		Message<?> converted = binder.serializePayloadIfNecessary(message).toMessage();
+		assertEquals(MimeTypeUtils.TEXT_PLAIN, contentTypeResolver.resolve(converted.getHeaders()));
```
ericbottard: I'm not sure I understand that step

ilayaperumalg: Here the contentType `text/plain` is inferred from the payload type which is `String`.

ericbottard: This definitely goes beyond this simple PR, but this really seems weird to me.
IMHO, when "serialization" happens, there should either
* be no more concept of contentType, as everything "is just bytes"
* or if there is the concept of contentType, the original contentType should apply, especially when it has been explicitly set (as opposed to inferred).

I'd be curious to hear what @dsyer @markfisher and @mbogoevici think about this, but this should not block this PR I guess

PR #405, opened 2016-03-07
Preserve contentType when de-serializing messages from non-SCSt app

### LINKED ISSUE #404: Preserve contentType at the consumer when receiving messages from non-SCSt applications
As a user, I would like to preserve my message `contentType` header info when receiving the messages from non-SCSt applications.

The deserialisation process at the binder's `ReceivingHandler` expects the `originalContentType` header when reseting the `contentType` of the message.

The `originalContentType` is set at the `serialisation` process only by the SCSt applications. When receiving messages from non-SCSt applications, this `originalContentType` header may not be available. Hence the message will miss the `contentType` (set to null) after deserialisation.

ilayaperumalg:
- When deserialising the message, replace `contentType` with the `originalContentType` only if the `originalContentType` isn't null which means the original content type was set by the producer before serialising.
 - Add and fix tests

This resolves #404

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
@@ -41,6 +41,7 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.PollableChannel;
+import org.springframework.util.MimeTypeUtils;
 
 /**
  * @author Gary Russell
@@ -169,7 +170,7 @@ public abstract class AbstractBinderTests {
 		assertNotNull(inbound);
 		assertEquals("foo", inbound.getPayload());
 		assertNull(inbound.getHeaders().get(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE));
-		assertNull(inbound.getHeaders().get(MessageHeaders.CONTENT_TYPE));
+		assertEquals(MimeTypeUtils.TEXT_PLAIN_VALUE, inbound.getHeaders().get(MessageHeaders.CONTENT_TYPE));
 		producerBinding.unbind();
 		consumerBinding.unbind();
 	}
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
```diff
@@ -169,7 +170,7 @@ public void testSendAndReceiveNoOriginalContentType() throws Exception {
 		assertNotNull(inbound);
 		assertEquals("foo", inbound.getPayload());
 		assertNull(inbound.getHeaders().get(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE));
-		assertNull(inbound.getHeaders().get(MessageHeaders.CONTENT_TYPE));
+		assertEquals(MimeTypeUtils.TEXT_PLAIN_VALUE, inbound.getHeaders().get(MessageHeaders.CONTENT_TYPE));
```
ilayaperumalg: This test case exposes a `caveat` where the producer message doesn't actually set the `contentType` which means the `originalContentType` was never set by the producer SCSt app. Even in that case, we would set the `contentType` as the deserialised message `contentType` which I believe doesn't do any harm.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -405,10 +405,13 @@ protected final MessageValues deserializePayloadIfNecessary(MessageValues messag
 		Object payload = deserializePayload(originalPayload, contentType);
 		if (payload != null) {
 			messageValues.setPayload(payload);
-
 			Object originalContentType = messageValues.get(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE);
-			messageValues.put(MessageHeaders.CONTENT_TYPE, originalContentType);
-			messageValues.remove(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE);
+			// Reset content-type only if the original content type is not null (when receiving messages from
+			// non-SCSt applications.
```
ericbottard: missing closing ), will add

spring-cloud-stream-binders/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
ericbottard: How do we know which content-type to pick then? Why `text/plain`?

ilayaperumalg: Here the `text/plain` is the contentType deduced based on the payload [this] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java#L367) way.

ericbottard:
Rebased and merged as 6bbf688

PR #407, opened 2016-03-07
Always show destination name via ChannelsEndpoint

### LINKED ISSUE #406: ChannelsEndpoint's binding properties to show default destination name
When there are no explicit `binding` properties provided for the channels, the default properties aren't shown up via ChannelsEndpoint.

It would be better to show at least the destination name with its default value.

ilayaperumalg:
- Retrieve the destination name from `ChannelBindingServiceProperties` for the given message channel name

This resolves #406 #426

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DynamicDestinationsBindable.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DynamicDestinationsBindable.java
@@ -32,7 +32,7 @@ import org.springframework.cloud.stream.binder.Binding;
 public final class DynamicDestinationsBindable extends BindableAdapter {
 
 	/**
-	 * Map containing dynamic destination names and their bindings.
+	 * Map containing channel names and their bindings.
 	 */
 	private Map<String, Binding> outputBindings = new HashMap<>();
 
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
@@ -21,9 +21,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.TreeMap;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonInclude.Include;
-
+import org.springframework.beans.BeanUtils;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Value;
@@ -36,7 +34,9 @@ import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.integration.support.utils.IntegrationUtils;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 
 /**
  * @author Dave Syer
@@ -189,7 +189,6 @@ public class ChannelBindingServiceProperties implements ApplicationContextAware,
 		return consumerProperties;
 	}
 
-
 	public ProducerProperties getProducerProperties(String outputChannelName) {
 		Assert.notNull(outputChannelName, "The output channel name cannot be null");
 		ProducerProperties producerProperties = getBindingProperties(outputChannelName).getProducer();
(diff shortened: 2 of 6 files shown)


ilayaperumalg:
We aren't updating the existing configuration object. Instead intention here is to update the newly created BindingProperties object for the channel that is deemed as `bindable` but the `destinationName` isn't already set (the default case).

The main issue this PR addresses is to always show the `destination-name` of the channel whether set explicitly or the default one.

ilayaperumalg:
Also, noticed that the `dynamicBindable` destination names need to be explicitly registered and exposed for the ChannelEndpoint invocations. Updated the PR for that change.

mbogoevici:
The reason why I said we update the existing object is that https://github.com/ilayaperumalg/spring-cloud-stream/commit/7846711ffe2d522df3b958e4d004e299ed26abdb#diff-21b3ca6ac22a24d6058e707e7a3b5fb1R72 returns the value stored in the map if it exists and then https://github.com/ilayaperumalg/spring-cloud-stream/commit/7846711ffe2d522df3b958e4d004e299ed26abdb#diff-21b3ca6ac22a24d6058e707e7a3b5fb1R74 sets the property. Am I missing something?

ilayaperumalg:
yeah, even in this case I thought we should be ok to update the binding properties that uses the `default` destination name set and we are setting it here `just` to show it for the Channels endpoint. not ok?

mbogoevici:
The main problem is that the value is set directly on the instance - and it remains set after we display it. It would be fine if we created and updated a copy, but as a general rule we shouldn't modify the binding values at runtime (even if we set the correct default value).

ilayaperumalg:
The PR changes are rebased on top of #409 here: https://github.com/ilayaperumalg/spring-cloud-stream/tree/PR-407

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -240,8 +240,12 @@ public String getBinder(String channelName) {
 	}
 
 	public BindingProperties getBindingProperties(String channelName) {
-		BindingProperties bindingProperties = bindings.containsKey(channelName) ?
-				bindings.get(channelName) : new BindingProperties();
+		if (this.bindings.containsKey(channelName)) {
+			return this.bindings.get(channelName);
+		}
+		// Update the bindings map with the new Binding properties.
+		BindingProperties bindingProperties = new BindingProperties();
```
mbogoevici: I don't think we should ever update the property map for bindings nor should we update the binding instance. What can we do instead is to return a copy that is populated with the default value (i.e. the value that is supposed to override the unset value).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -254,6 +258,7 @@ public String getBindingDestination(String channelName) {
 		if (bindingProperties != null && StringUtils.hasText(bindingProperties.getDestination())) {
 			return bindingProperties.getDestination();
 		}
+		bindingProperties.setDestination(channelName);
```
mbogoevici: Also see https://github.com/spring-cloud/spring-cloud-stream/pull/407/files#r57359433

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
ilayaperumalg: Given, we **only** create the BindingProperties when there is no explicit binding properties defined in the environment the `channelName` (since BindingProperties themselves boot ConfigurationProperties), this seems to be the only way to populate the default values for the BindingProperties (especially when the value is set to something like channelName). I made a change to update the BindingProperties with the destination name when the object is created here. This will at least ensure that all `ChannelNames` will have their `BindingProperties` in the map. 

What issue do you see if we do this way?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DynamicDestinationsBindable.java
```diff
@@ -32,7 +32,7 @@
 public final class DynamicDestinationsBindable extends BindableAdapter {
 
 	/**
-	 * Map containing dynamic destination names and their bindings.
+	 * Map containing channel names and their bindings.
```
mbogoevici: We can still indicate that the channels are bound dynamically, for reference.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -200,8 +199,13 @@ public ProducerProperties getProducerProperties(String outputChannelName) {
 	}
 
 	public BindingProperties getBindingProperties(String channelName) {
-		BindingProperties bindingProperties = bindings.containsKey(channelName) ?
-				bindings.get(channelName) : new BindingProperties();
+		BindingProperties bindingProperties = new BindingProperties();
+		if (this.bindings.containsKey(channelName)) {
+			BeanUtils.copyProperties(this.bindings.get(channelName), bindingProperties);
+		}
+		if (bindingProperties.getDestination() == null) {
```
mbogoevici: Minor: can use StringUtils.hasText() for a more comprehensive check.

mbogoevici:
Closed via https://github.com/spring-cloud/spring-cloud-stream/commit/f390feb42b25818f74825cf8260e9592a5db3a2d. Minor additional updates: https://github.com/spring-cloud/spring-cloud-stream/commit/0626cc8c07caeb16539774aa6fc56f6793d129e9 .

Issue #369, opened 2016-02-23
Allow defining binder-specific consumer and producer properties for bindings


mbogoevici:
This depends on spring-cloud/spring-cloud-stream#307

mbogoevici:
Closed https://github.com/spring-cloud/spring-cloud-stream/commit/9b0c4bd627f7d122be60d11caf993b18799b3ae6

PR #409, opened 2016-03-10
Typesafe consumer and producer configurations

mbogoevici:
- Change the Binder interface to support ConsumerProperties/ProducerProperties beans and subclasses
- Binders can subclass the property beans to add new supported properties that will be automatically populated
- Spring Cloud Stream will  infer the target type and populate the beans from the environment based on a `spring.cloud.stream.bindings..<bindingName>` prefix
- Remove binder defaults and retain only general binder configurations

TODO: a) decide on instanceIndex/partitionIndex alignment (we do not need both)
b) support `defaultProducer`/`defaultConsumer` properties
c) add leniency control on binding (fail/ignore for unknown properties)
d) add a `requiredProperties` configuration for consumer/producer properties to finely tune the mandatory properties expected to be supported by a bound application

Resolved spring-cloud/spring-cloud-stream#412
Resolved spring-cloud/spring-cloud-stream#413
Resolved spring-cloud/spring-cloud-stream#369
Resolved spring-cloud/spring-cloud-stream#307


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/DefaultSettingsTests.java
```diff
@@ -0,0 +1,127 @@
...
+	public void testDefaultSettings() {
+		ConfigurableApplicationContext applicationContext = SpringApplication.run(TestFooChannels.class,
+				"--spring.cloud.stream.bindings.foo.destination=fooDest",
+				"--spring.cloud.stream.bindings.bar.destination=barDest",
+				"--spring.cloud.stream.bindings.baz.destination=bazDest",
+				"--spring.cloud.stream.bindings.qux.destination=quxDest",
+				"--spring.cloud.stream.bindings.foo.group=fooBarGroup",
+				"--spring.cloud.stream.bindings.bar.group=fooBarGroup",
+				"--spring.cloud.stream.consumerDefaults.concurrency=2",
+				"--spring.cloud.stream.producerDefaults.requiredGroups=quxbazReq1,quxbazReq2");
```
ilayaperumalg: I think these tests require `--server.port=0` arg to run on random ports.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/DefaultSettingsTests.java
```diff
@@ -0,0 +1,127 @@
...
+				"--spring.cloud.stream.bindings.foo.destination=fooDest",
+				"--spring.cloud.stream.bindings.bar.destination=barDest",
+				"--spring.cloud.stream.bindings.baz.destination=bazDest",
+				"--spring.cloud.stream.bindings.qux.destination=quxDest",
+				"--spring.cloud.stream.bindings.foo.group=fooBarGroup",
+				"--spring.cloud.stream.bindings.bar.group=fooBarGroup",
+				"--spring.cloud.stream.consumerDefaults.concurrency=2",
+				"--spring.cloud.stream.bindings.bar.concurrency=4",
+				"--spring.cloud.stream.producerDefaults.requiredGroups=quxbazReq1,quxbazReq2",
+				"--spring.cloud.stream.bindings.qux.requiredGroups=quxbazReq3,quxbazReq4");
```
ilayaperumalg: `--server.port=0`

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -418,7 +304,7 @@ public void setZkConnectionTimeout(int zkConnectionTimeout) {
 	}
 
 	@Override
-	protected Binding<MessageChannel> doBindConsumer(String name, String group, MessageChannel inputChannel, Properties properties) {
+	protected Binding<MessageChannel> doBindConsumer(String name, String group, MessageChannel inputChannel, KafkaConsumerProperties properties) {
```
ilayaperumalg: It might be better to make the `KafkaConsumerProperties` **final** ?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -429,57 +315,55 @@ public void setZkConnectionTimeout(int zkConnectionTimeout) {
...
-				startOffset.getReferencePoint() : (anonymous ? OffsetRequest.LatestTime() : OffsetRequest.EarliestTime());
+		long referencePoint = properties.getStartOffset() != null ?
+				properties.getStartOffset().getReferencePoint() : (anonymous ? OffsetRequest.LatestTime() : OffsetRequest.EarliestTime());
 		return createKafkaConsumer(name, inputChannel, properties, consumerGroup, referencePoint);
 	}
 
+
 	@Override
-	public Binding<MessageChannel> bindProducer(final String name, MessageChannel moduleOutputChannel, Properties properties) {
+	public Binding<MessageChannel> doBindProducer(final String name, MessageChannel moduleOutputChannel, KafkaProducerProperties properties) {
```
ilayaperumalg: Same here: Making `KafkaProducerProperties` **final**?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -429,57 +315,55 @@ public void setZkConnectionTimeout(int zkConnectionTimeout) {
...
 		if (logger.isInfoEnabled()) {
 			logger.info("Using kafka topic for outbound: " + name);
 		}
 
 		validateTopicName(name);
 
-		int numPartitions = producerPropertiesAccessor.getNumberOfKafkaPartitionsForProducer();
+		int partitionCount = properties.getPartitionCount();
+		if (partitionCount == 0) {
+			throw new IllegalArgumentException("Module count cannot be zero");
```
ilayaperumalg: `Module` -> `App`

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -542,17 +426,16 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
...
-
-		int maxConcurrency = accessor.getConcurrency(defaultConcurrency);
-
 		validateTopicName(name);
-
-		int numPartitions = accessor.getNumberOfKafkaPartitionsForConsumer();
+		int minKafkaPartitions = properties.getMinPartitionCount();
+		int moduleCount = properties.getCount();
+		if (moduleCount == 0) {
+			throw new IllegalArgumentException("Module count cannot be zero");
```
ilayaperumalg: May be `App instances count`?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -632,25 +507,25 @@ protected void doStop() {
...
+		DefaultBinding<MessageChannel> consumerBinding = new DefaultBinding<>(name, group, moduleInputChannel, edc);
 		edc.start();
 		return consumerBinding;
 	}
 
-	public KafkaMessageListenerContainer createMessageListenerContainer(Properties properties, String group,
-			int maxConcurrency, String topic, long referencePoint) {
-		return createMessageListenerContainer(new KafkaPropertiesAccessor(properties), group, maxConcurrency, topic,
+	public KafkaMessageListenerContainer createMessageListenerContainer(KafkaConsumerProperties properties, String group,
+																		int maxConcurrency, String topic, long referencePoint) {
```
ilayaperumalg: weird formatting.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -664,15 +539,15 @@ private KafkaMessageListenerContainer createMessageListenerContainer(KafkaProper
...
+		messageListenerContainer.setConcurrency(Math.min(consumerProperties.getConcurrency(), listenedPartitions.size()));
 		OffsetManager offsetManager = createOffsetManager(group, referencePoint);
-		if (resetOffsets) {
+		if (consumerProperties.isResetOffsets()) {
 			offsetManager.resetOffsets(listenedPartitions);
 		}
 		messageListenerContainer.setOffsetManager(offsetManager);
-		messageListenerContainer.setQueueSize(accessor.getProperty(QUEUE_SIZE, defaultQueueSize));
-		messageListenerContainer.setMaxFetch(accessor.getProperty(FETCH_SIZE, defaultFetchSize));
+		messageListenerContainer.setQueueSize(defaultQueueSize);
```
ilayaperumalg: I wonder if we can remove `default` from the variable names for all the `binder` configuration properties. Having `default` makes it look like constant values in a way. If we remove `default`, the `setter` and `getter` methods on the binder would look better (like setQueueSize, setFetchSize) on the binder.

This also means that `binder` configuration properties don't need to carry `.default` in them (as they are the actual property values and not the default ones)?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -711,60 +586,22 @@ public void doManualAck(LinkedList<MessageHeaders> messageHeadersList) {
...
-
-		public int getMinPartitionCount(int defaultPartitionCount) {
-			return getProperty(BinderPropertyKeys.MIN_PARTITION_COUNT, defaultPartitionCount);
-		}
-
-	}
-
 	private class ReceivingHandler extends AbstractReplyProducingMessageHandler {
 
 		public ReceivingHandler() {
```
ilayaperumalg: I think this default constructor can be removed? and we still need to invoke `setBeanFactory` in the other constructor? not sure what's the significance of setting beanfactory here.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderDefaultProperties.java
```diff
@@ -25,18 +25,12 @@
 @ConfigurationProperties(value = "spring.cloud.stream.binder.kafka.default")
 public class KafkaBinderDefaultProperties {
```
ilayaperumalg: Unless there is a specific reason, I think it would be nice to avoid `Default` and having the name [Kafka/Rabbit etc.,]`BinderProperties`  based on the discussion from [here] (https://github.com/spring-cloud/spring-cloud-stream/pull/409/files#r55822262)

ilayaperumalg: If the reason is not to conflict with the `broker` specific properties such as `KafkaBinderConfigurationProperties`, may be we need to name the broker specific properties as `KafkaBrokerProperties` with the prefix `spring.cloud.stream.broker.kafka`? and this class can be `KafkaBinderProperties` with the prefix `spring.cloud.stream.binder.kafka`?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
```diff
@@ -49,41 +46,38 @@
  * @author Gary Russell
  * @author Mark Fisher
  */
+@Ignore
```
ilayaperumalg: Is there a reason why do we ignore this?

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
```diff
@@ -0,0 +1,152 @@
...
+
+	public boolean isRepublishToDlq() {
+		return republishToDlq;
+	}
+
+	public void setRepublishToDlq(boolean republishToDlq) {
+		this.republishToDlq = republishToDlq;
+	}
+
+	public boolean getRequeueRejected() {
```
ilayaperumalg: Couple of getters for this property name. Could be better if we use `isRequeueRejected` and remove this method.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -641,19 +460,19 @@ private RabbitTemplate determineRabbitTemplate(RabbitPropertiesAccessor properti
 	}
 
 	private Binding<MessageChannel> doRegisterProducer(final String name, MessageChannel moduleOutputChannel,
-			AmqpOutboundEndpoint delegate, RabbitPropertiesAccessor properties) {
+													   AmqpOutboundEndpoint delegate, ProducerProperties properties) {
```
ilayaperumalg: May be better to use `RabbitProducerProperties` instead of just the `ProducerProperties` (thought it looks sufficient).

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -641,19 +460,19 @@ private RabbitTemplate determineRabbitTemplate(RabbitPropertiesAccessor properti
...
 		return this.doRegisterProducer(name, moduleOutputChannel, delegate, null, properties);
 	}
 
 	private Binding<MessageChannel> doRegisterProducer(final String name, MessageChannel moduleOutputChannel,
-			AmqpOutboundEndpoint delegate, String replyTo, RabbitPropertiesAccessor properties) {
+													   AmqpOutboundEndpoint delegate, String replyTo, ProducerProperties properties) {
```
ilayaperumalg: RabbitProducerProperties instead of `ProducerProperties`.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -789,13 +606,13 @@ public void doManualAck(LinkedList<MessageHeaders> messageHeadersList) {
 
 		private final PartitionHandler partitionHandler;
 
-		private SendingHandler(MessageHandler delegate, String replyTo, RabbitPropertiesAccessor properties) {
+		private SendingHandler(MessageHandler delegate, String replyTo, ProducerProperties properties) {
```
ilayaperumalg: Here as well: RabbitProducerProperties

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
```diff
@@ -83,7 +83,7 @@ public void handleMessage(Message<?> message) throws MessagingException {
 				String inputBindingProps = headerValue.get("input");
 				assertTrue(inputBindingProps.contains("destination=configure"));
 				assertTrue(inputBindingProps.contains("trackHistory=true"));
-				assertTrue(inputBindingProps.contains("concurrency=1"));
+//				assertTrue(inputBindingProps.contains("concurrency=1"));
```
ilayaperumalg: This can be removed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -593,26 +408,6 @@ static String classNameFromMimeType(MimeType mimeType) {
...
-
-		public Set<Object> build() {
-			return this.set;
-		}
-
-	}
-
 	/**
 	 * Perform manual acknowledgement based on the metadata stored in the binder.
 	 */
```
ilayaperumalg: I couldn't find any usage for this class. If so, can this be removed?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -212,110 +121,10 @@ public void setCodec(Codec codec) {
...
 
-	protected IdGenerator getIdGenerator() {
-		return this.idGenerator;
-	}
-
 	public void setIntegrationEvaluationContext(EvaluationContext evaluationContext) {
 		this.evaluationContext = evaluationContext;
 	}
 
-	/**
```
ilayaperumalg: The unused methods related to the validation of properties (validateConsumerProperties, validateProducerProperties and validateProperties) can be removed as well.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
```diff
@@ -0,0 +1,126 @@
...
+
+	private String partitionKeyExtractorClass = null;
+
+	private String partitionSelectorClass = null;
+
+	private String partitionSelectorExpression = null;
+
+	private boolean batchingEnabled = false;
+
+	private int batchSize;
```
ilayaperumalg: I wonder if we need default values for `batchSize`, `batchBufferLimit` and `batchTimeout`. Otherwise, having the default `0` may not makes sense?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -234,15 +166,71 @@ public String getBinder(String channelName) {
...
+		catch (InstantiationException | IllegalAccessException e) {
+			throw new BeanInitializationException(e.getMessage());
+		}
+		// bind defaults first
+		RelaxedDataBinder dataBinder = new RelaxedDataBinder(beanInstance);
+		dataBinder.setIgnoreUnknownFields(false);
+		dataBinder.setDisallowedFields(bindingPropertyFields);
+		dataBinder.bind(new MutablePropertyValues(defaults));
+		// bind configured properties next, if available
+		dataBinder = new RelaxedDataBinder(beanInstance, "spring.cloud.stream.bindings." + channelName);
```
ilayaperumalg: Can we move this prefix `spring.cloud.stream.bindings` to a constant? and probably use a constant for `spring.cloud.stream` prefix in `ChannelBindingServiceProperties` itself?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/DefaultSettingsTests.java
mbogoevici: Actually, I wanted to disable web/jmx altogether since that also helps with startup time.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: you mean the class or the parameter? we could - I'm not a huge fan of the unnecessary verbosity. The first parameter is final because we use it in in inner class, but the other ones aren't used that way.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: Probably will remove the qualifier altogether (`Instance count`)

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/RawModeKafkaBinderTests.java
mbogoevici: No, those tests just needed to be fixed.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
```diff
@@ -0,0 +1,152 @@
...
+import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.cloud.stream.binder.ConsumerProperties;
+import org.springframework.util.Assert;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class RabbitConsumerProperties extends ConsumerProperties {
+
+	private String prefix = "binder.";
```
garyrussell: We discussed maybe making this `""` by default - was a decision made to keep it `binder.` ??

mbogoevici: Not really - I just left things as they originally were. Would be happy to change it if that's what we want.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
markfisher: actually this should say "Partition count..."

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
markfisher: +1 for "Instance count"

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -569,34 +450,29 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
...
-		ReceivingHandler rh = new ReceivingHandler();
+		ReceivingHandler rh = new ReceivingHandler(properties);
 		rh.setOutputChannel(moduleInputChannel);
 
 		final FixedSubscriberChannel bridge = new FixedSubscriberChannel(rh);
 		bridge.setBeanName("bridge." + name);
 
 		final KafkaMessageListenerContainer messageListenerContainer =
-				createMessageListenerContainer(accessor, group, maxConcurrency, listenedPartitions,
+				createMessageListenerContainer(properties, group, properties.getConcurrency(), listenedPartitions,
```
markfisher: probably don't need the separate concurrency param since that is included in the properties

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: I'm wondering if we should remove this test here and put it on the bean - maybe with some JSR-303 validation on bind.  It's sort of awkward that we test it only for Kafka (it can't be zero in general).

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaServiceAutoConfiguration.java
```diff
@@ -77,29 +77,19 @@ KafkaMessageChannelBinder kafkaMessageChannelBinder() {
...
-		kafkaMessageChannelBinder.setDefaultBatchSize(kafkaBinderDefaultProperties.getBatchSize());
-		kafkaMessageChannelBinder.setDefaultBatchTimeout(kafkaBinderDefaultProperties.getBatchTimeout());
-		kafkaMessageChannelBinder.setDefaultCompressionCodec(kafkaBinderDefaultProperties.getCompressionCodec());
-		kafkaMessageChannelBinder.setDefaultConcurrency(kafkaBinderDefaultProperties.getConcurrency());
 		kafkaMessageChannelBinder.setDefaultFetchSize(kafkaBinderDefaultProperties.getFetchSize());
 		kafkaMessageChannelBinder.setDefaultMinPartitionCount(kafkaBinderDefaultProperties.getMinPartitionCount());
 		kafkaMessageChannelBinder.setDefaultQueueSize(kafkaBinderDefaultProperties.getQueueSize());
 		kafkaMessageChannelBinder.setDefaultReplicationFactor(kafkaBinderDefaultProperties.getReplicationFactor());
 		kafkaMessageChannelBinder.setDefaultRequiredAcks(kafkaBinderDefaultProperties.getRequiredAcks());
+		kafkaMessageChannelBinder.setMaxWait(kafkaBinderConfigurationProperties.getMaxWait());
```
markfisher: nitpick but it would be nice if this were located alongside the other `kafkaBinderConfigurationProperties` values rather than here with the "defaults"... although based on some other comments, maybe these 2 could be combined into a single configuration properties class?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderDefaultProperties.java
mbogoevici: Yes, that makes sense.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaServiceAutoConfiguration.java
mbogoevici: Agree on that. We should only have one configuration class.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -217,17 +220,17 @@ public void testCustomPartitionCountDoesNotOverrideModuleCountAndConcurrencyIfSm
...
 		QueueChannel moduleInputChannel = new QueueChannel();
-		Properties producerProps = new Properties();
-		producerProps.put(BinderPropertyKeys.MIN_PARTITION_COUNT, "5");
-		producerProps.put(BinderPropertyKeys.NEXT_MODULE_CONCURRENCY, "6");
-		Properties consumerProps = new Properties();
-		consumerProps.put(BinderPropertyKeys.MIN_PARTITION_COUNT, "5");
-		consumerProps.put(BinderPropertyKeys.CONCURRENCY, "6");
+		KafkaProducerProperties producerProps = new KafkaProducerProperties();
+		producerProps.setPartitionCount(5);
+		producerProps.setPartitionCount(6);
```
markfisher: this overrides the call on the line above... previously these were 2 separate props: `MIN_PARTITION_COUNT` and `NEXT_MODULE_CONCURRENCY`

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -282,16 +284,15 @@ public void testCustomPartitionCountDoesNotOverridePartitioningIfSmaller() throw
...
 		DirectChannel moduleOutputChannel = new DirectChannel();
 		QueueChannel moduleInputChannel = new QueueChannel();
-		Properties producerProperties = new Properties();
-		producerProperties.put(BinderPropertyKeys.MIN_PARTITION_COUNT, "3");
-		producerProperties.put(BinderPropertyKeys.NEXT_MODULE_COUNT, "5");
-		producerProperties.put(BinderPropertyKeys.PARTITION_KEY_EXPRESSION, "payload");
-		Properties consumerProperties = new Properties();
-		consumerProperties.put(BinderPropertyKeys.MIN_PARTITION_COUNT, "3");
+		KafkaProducerProperties producerProperties = new KafkaProducerProperties();
+		producerProperties.setPartitionCount(5);
```
markfisher: are we missing anything here by not having a new property equivalent for this?: `producerProperties.put(BinderPropertyKeys.MIN_PARTITION_COUNT, "3");`

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -314,16 +315,15 @@ public void testCustomPartitionCountOverridesPartitioningIfLarger() throws Excep
...
 		DirectChannel moduleOutputChannel = new DirectChannel();
 		QueueChannel moduleInputChannel = new QueueChannel();
-		Properties producerProperties = new Properties();
-		producerProperties.put(BinderPropertyKeys.MIN_PARTITION_COUNT, "5");
-		producerProperties.put(BinderPropertyKeys.NEXT_MODULE_COUNT, "3");
-		producerProperties.put(BinderPropertyKeys.PARTITION_KEY_EXPRESSION, "payload");
-		Properties consumerProperties = new Properties();
-		consumerProperties.put(BinderPropertyKeys.MIN_PARTITION_COUNT, "5");
+		KafkaProducerProperties producerProperties = new KafkaProducerProperties();
+		producerProperties.setPartitionCount(5);
```
markfisher: similar question here... it seems as though the intent of these 2 tests was to verify overriding of the partition count based on these 2 different properties... but now, if there are no longer 2 different properties, do we even need these tests?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
mbogoevici: No, the old property just didn't make sense anymore. The partition count used to be computed from the next module count and the next module concurrency, so instead of that we just use the partition count. The property was an override, which doesn't make sense. If a producer can set the partition count, they might as well set it right.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
mbogoevici: I will need to check if there is an overlap, but the client can request a minimum number of partitions if it creates the topic.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -552,12 +373,12 @@ private MessageRecoverer determineRecoverer(String name, RabbitPropertiesAccesso
 		}
 	}
 
-	private AmqpOutboundEndpoint buildOutboundEndpoint(final String name, RabbitPropertiesAccessor properties,
-			RabbitTemplate rabbitTemplate) {
-		String prefix = properties.getPrefix(this.defaultPrefix);
+	private AmqpOutboundEndpoint buildOutboundEndpoint(final String name, RabbitProducerProperties properties,
+													   RabbitTemplate rabbitTemplate) {
```
markfisher: extreme indentation

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitProducerProperties.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.springframework.amqp.core.MessageDeliveryMode;
+import org.springframework.cloud.stream.binder.ProducerProperties;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class RabbitProducerProperties extends ProducerProperties {
+
+	private String prefix = "binder.";
```
markfisher: if we do keep a default prefix that is not an empty String, maybe we should share a constant?

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitMessageChannelBinderConfiguration.java
```diff
@@ -45,12 +45,9 @@
...
 	@Autowired
 	private ConnectionFactory rabbitConnectionFactory;
-	
-	@Autowired
-	private RabbitBinderConfigurationProperties rabbitBinderConfigurationProperties;
-	
+
 	@Autowired
-	private SpringRabbitMQProperties springRabbitMQProperties;
+	private RabbitBinderConfigurationProperties springRabbitMQProperties;
```
markfisher: should use the other var name here? (rabbitBinderConfigurationProperties)

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitBinderTests.java
```diff
@@ -152,20 +158,20 @@ public void testConsumerProperties() throws Exception {
...
+		properties.setBackOffInitialInterval(2000);
+		properties.setBackOffMaxInterval(20000);
+		properties.setBackOffMultiplier(5.0);
+		properties.setConcurrency(2);
+		properties.setMaxAttempts(23);
+		properties.setMaxConcurrency(3);
+		properties.setPrefix("foo.");
+		properties.setPrefetch(20);
+		properties.setRequestHeaderPatterns(new String[] {"foo"});
+		properties.setRequeueRejected(false);
```
markfisher: I think this is different - the property it's replacing above was "requeue"

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitBinderTests.java
```diff
@@ -554,21 +562,21 @@ public void handleMessage(Message<?> message) throws MessagingException {
...
-		properties.put("batchBufferLimit", "100000");
-		properties.put("batchTimeout", "30000");
-		properties.put("compress", "true");
-		properties.put("requiredGroups", "default");
+		RabbitTestBinder binder = getBinder();
+		RabbitProducerProperties properties = new RabbitProducerProperties();
+		properties.setDeliveryMode(MessageDeliveryMode.NON_PERSISTENT);
+		properties.setBatchingEnabled(true);
+		properties.setBatchSize(2);
+		properties.setBatchBufferLimit(10000);
```
markfisher: in the corresonding deleted line, this was 100k not 10k

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
```diff
@@ -207,7 +183,7 @@ protected void afterUnbind() {
 	 * @return The channel, or a wrapper.
 	 */
 	private MessageChannel addRetryIfNeeded(final String name, final DirectChannel bridgeToModuleChannel,
-			RedisPropertiesAccessor properties) {
+											ConsumerProperties properties) {
```
markfisher: crazy indent

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
```diff
@@ -256,16 +232,14 @@ public Boolean recover(RetryContext context) throws Exception {
...
-		return doRegisterProducer(name, moduleOutputChannel, accessor);
+		return doRegisterProducer(name, moduleOutputChannel, properties);
 	}
 
-	private RedisQueueOutboundChannelAdapter createProducerEndpoint(String name, RedisPropertiesAccessor accessor) {
-		String partitionKeyExtractorClass = accessor.getPartitionKeyExtractorClass();
-		Expression partitionKeyExpression = accessor.getPartitionKeyExpression();
+	private RedisQueueOutboundChannelAdapter createProducerEndpoint(String name, ProducerProperties properties) {
+		String partitionKeyExtractorClass = properties.getPartitionKeyExtractorClass();
+		String partitionKeyExpression = properties.getPartitionKeyExpression();
```
markfisher: should the `partitionKeyExpression` be an `Expression` type within the properties object? (there is a Converter available for that conversion when properties are bound... we rely on that for some of the out-of-the-box `spring-cloud-stream` apps)

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/test/java/org/springframework/cloud/stream/binder/redis/RedisBinderTests.java
```diff
@@ -128,14 +138,14 @@ public void testProducerProperties() throws Exception {
...
 		assertFalse(producerEndpoint.isRunning());
 
-		Properties properties = new Properties();
-		properties.put("partitionKeyExpression", "'foo'");
-		properties.put("partitionKeyExtractorClass", "foo");
-		properties.put("partitionSelectorExpression", "0");
-		properties.put("partitionSelectorClass", "foo");
-		properties.put(BinderPropertyKeys.NEXT_MODULE_COUNT, "1");
+		ProducerProperties producerProperties = new ProducerProperties();
+		producerProperties.setPartitionKeyExpression("'foo'");
```
markfisher: these properties should really be mutually exclusive (validated within the props?):
-  `partitionKeyExpression` OR `partitionKeyExtractorClass`
-  `partitionSelectorExpression` OR `partitionSelectorClass`

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/RedisMessageChannelBinder.java
mbogoevici: Possibly, then the extractor Class could be a Class too.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitBinderTests.java
mbogoevici: We can restore the value, but I don't think it affects the test.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java
```diff
@@ -57,9 +60,9 @@
 	 * @param partitionCount number of partitions configured for binder
 	 */
 	public PartitionHandler(ConfigurableListableBeanFactory beanFactory,
-			EvaluationContext evaluationContext,
-			PartitionSelectorStrategy partitionSelector,
-			DefaultBindingPropertiesAccessor properties, int partitionCount) {
+							EvaluationContext evaluationContext,
+							PartitionSelectorStrategy partitionSelector,
+							ProducerProperties properties, int partitionCount) {
```
markfisher: isn't `partitionCount` also in the `ProducerProperties`?

spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/test/java/org/springframework/cloud/stream/binder/redis/RedisBinderTests.java
mbogoevici: Yes. In fact I think we should support JSR-303 validation

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -104,8 +101,16 @@ else if (tokens.length != 1) {
...
-					Binder<MessageChannel> binder = binderFactory.getBinder(transport);
-					this.dynamicDestinationsBindable.addOutputBinding(beanName, binder.bindProducer(destinationName, channel, producerProperties));
+					@SuppressWarnings("unchecked")
+					Binder<MessageChannel, ?, ProducerProperties> binder =
+							(Binder<MessageChannel, ?, ProducerProperties>) binderFactory.getBinder(binderName);
+					// TODO: need the props to return some defaults if not found
+					Class<? extends ProducerProperties> producerPropertiesClass =
+							ChannelBindingService.resolveProducerPropertiesType(binder);
+					ProducerProperties producerProperties =
+							this.channelBindingServiceProperties.getProducerProperties(destinationName, producerPropertiesClass);
```
markfisher: the first param of `getProducerProperties(..)` is *channelName*, but here the "destinationName" is passed... is that intentional?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -234,15 +166,71 @@ public String getBinder(String channelName) {
...
+		BindingProperties bindingProperties = bindings.containsKey(channelName) ?
+				bindings.get(channelName) : new BindingProperties();
+		return bindingProperties;
+	}
+
+	public String getGroup(String channelName) {
+		return getBindingProperties(channelName).getGroup();
+	}
+
+	public String getBindingDestination(String channel) {
```
markfisher: nitpick: for consistency, this param should probably be "channelName"

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
mbogoevici: that is a mistake

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/BinderAwareChannelResolverTests.java
```diff
@@ -125,25 +125,24 @@ public void resolveNonRegisteredChannel() {
...
 
 	@Test
-	@SuppressWarnings("rawtypes")
+	@SuppressWarnings({"rawtypes", "unchecked"})
 	public void propertyPassthrough() {
 		ChannelBindingServiceProperties bindingServiceProperties = new ChannelBindingServiceProperties();
 		DynamicDestinationsBindable dynamicDestinationsBindable = new DynamicDestinationsBindable();
-		Map<String, BindingProperties> bindings = new HashMap<String, BindingProperties>();
-		BindingProperties bindingProperties = new BindingProperties();
-		bindingProperties.setContentType("text/plain");
```
markfisher: is the contentType relevant to this test? if so, I think it's been dropped

mbogoevici: No.

mbogoevici: Although we have a separate issue about BindingProperties not being properly applied to channels.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -104,8 +101,16 @@ else if (tokens.length != 1) {
 					channel = new DirectChannel();
 					this.beanFactory.registerSingleton(beanName, channel);
 					channel = (MessageChannel) this.beanFactory.initializeBean(channel, beanName);
-					Binder<MessageChannel> binder = binderFactory.getBinder(transport);
-					this.dynamicDestinationsBindable.addOutputBinding(beanName, binder.bindProducer(destinationName, channel, producerProperties));
+					@SuppressWarnings("unchecked")
+					Binder<MessageChannel, ?, ProducerProperties> binder =
+							(Binder<MessageChannel, ?, ProducerProperties>) binderFactory.getBinder(binderName);
+					// TODO: need the props to return some defaults if not found
```
markfisher: would be a good idea to create a GH issue for this TODO if it won't be address in the scope of this PR

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/ChannelsEndpoint.java
```diff
@@ -56,10 +56,12 @@ public ChannelsEndpoint(List<Bindable> adapters,
 			for (String name : factory.getInputs()) {
 				inputs.put(name, bindings.containsKey(name) ? bindings.get(name)
 						: new BindingProperties());
+				// TODO - add consumer properties
```
markfisher: would be a good idea to create a GH issue for this TODO if it won't be address in the scope of this PR

...same for the one a few lines below

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -542,62 +419,53 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
...
+		ReceivingHandler rh = new ReceivingHandler(properties);
 		rh.setOutputChannel(moduleInputChannel);
 
 		final FixedSubscriberChannel bridge = new FixedSubscriberChannel(rh);
 		bridge.setBeanName("bridge." + name);
 
 		final KafkaMessageListenerContainer messageListenerContainer =
-				createMessageListenerContainer(accessor, group, maxConcurrency, listenedPartitions,
-						referencePoint);
+				createMessageListenerContainer(properties, group, null, listenedPartitions, referencePoint);
```
markfisher: is `null` for the topic parameter intentional here?

markfisher: Actually I see below... that is mutually exclusive with the `listenedPartitions`. It might be worth having 2 separate methods instead. WDYT?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -542,62 +419,53 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
...
-		int maxConcurrency = accessor.getConcurrency(defaultConcurrency);
+	private Binding<MessageChannel> createKafkaConsumer(String name, final MessageChannel moduleInputChannel,
+														KafkaConsumerProperties properties, String group, long referencePoint) {
 
 		validateTopicName(name);
-
-		int numPartitions = accessor.getNumberOfKafkaPartitionsForConsumer();
-		Collection<Partition> allPartitions = ensureTopicCreated(name, numPartitions, defaultReplicationFactor);
+		int minKafkaPartitions = properties.getMinPartitionCount();
+		int instance = properties.getCount();
```
markfisher: I think `instanceCount` or at least `instances` would be more clear. Otherwise it seems like it's tied to a single instance (like `instanceIndex` is).

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaProducerProperties.java
```diff
@@ -0,0 +1,67 @@
...
+
+	public void setBufferSize(int bufferSize) {
+		this.bufferSize = bufferSize;
+	}
+
+	public ProducerMetadata.CompressionType getCompressionCodec() {
+		return compressionCodec;
+	}
+
+	public void setCompressionCodec(ProducerMetadata.CompressionType compressionCodec) {
```
markfisher: even though I do like "codec" better, I wonder if the name of the property should match the enum type: `setCompressionType(..)`

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: I'd prefer to have a single method, as ugly as it is, based on the fact that this will be much simpler when we migrate to 0.9.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: I like `instanceCount` too.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -664,15 +478,13 @@ private RabbitTemplate determineRabbitTemplate(RabbitPropertiesAccessor properti
 	 * queue name because we use default exchange routing by queue name for the original message.
 	 * @param queueName The base name for the queue (including the binder prefix, if any).
 	 * @param routingKey The routing key for the queue.
```
markfisher: need to add `@param` for `autoBindDlq`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
```diff
@@ -0,0 +1,132 @@
...
+package org.springframework.cloud.stream.binder;
+
+import org.springframework.expression.Expression;
+
+/**
+ * Common producer properties.
+ *
+ * @author Marius Bogoevici
+ */
+public class ProducerProperties {
```
markfisher: could be added as a follow-up issue (not in the scope of this PR), but I still think we should enforce mutual exclusivity for the partition properties (keyExpression OR keyExtractorClass; selectorExpression OR selectorClass)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -79,33 +79,38 @@ public MessageChannel resolveDestination(String destinationName) {
...
-							transport = tokens[0];
-							destinationName = tokens[1];
+							binderName = tokens[0];
+							channelName = tokens[1];
 						}
 						else if (tokens.length != 1) {
-							throw new IllegalArgumentException("Unrecognized channel naming scheme: " + destinationName + " , should be" +
-									" [<transport>:]<destinationName>");
+							throw new IllegalArgumentException("Unrecognized channel naming scheme: " + channelName + " , should be" +
+									" [<transport>:]<channelName>");
```
markfisher: should change "transport" to "binder" in the exception message too

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -79,33 +79,38 @@ public MessageChannel resolveDestination(String destinationName) {
...
 						}
 					}
 					channel = new DirectChannel();
 					this.beanFactory.registerSingleton(beanName, channel);
 					channel = (MessageChannel) this.beanFactory.initializeBean(channel, beanName);
-					Binder<MessageChannel> binder = binderFactory.getBinder(transport);
-					this.dynamicDestinationsBindable.addOutputBinding(beanName, binder.bindProducer(destinationName, channel, producerProperties));
+					@SuppressWarnings("unchecked")
+					Binder<MessageChannel, ?, ProducerProperties> binder =
```
markfisher: should this delegate to `ChannelBindingService` instead of having all the logic here, or do we need to keep this separate?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -100,128 +136,36 @@ public void setDynamicDestinations(String[] dynamicDestinations) {
...
-				consumerProperties.setProperty(BinderPropertyKeys.COUNT,
-						Integer.toString(getInstanceCount()));
-				consumerProperties.setProperty(BinderPropertyKeys.PARTITION_INDEX,
-						Integer.toString(getInstanceIndex()));
-			}
-		}
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		DefaultConversionService conversionService = new DefaultConversionService();
+		conversionService.addConverter(new SpelExpressionConverterConfiguration.SpelConverter());
```
markfisher: should this be injected as a bean, and/or should we consider annotating that converter bean in the config itself with `@IntegrationConverter`... could then rely on the `integrationConversionService` and the converter would already be registered

mbogoevici: I'd rather go the `@IntegrationConverter` route if available.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
mbogoevici: Set it to "".

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
mbogoevici: Let's address this after rebasing https://github.com/spring-cloud/spring-cloud-stream/pull/424

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
mbogoevici: Added https://github.com/spring-cloud/spring-cloud-stream/issues/425

mbogoevici:
Pushed

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
```diff
@@ -0,0 +1,132 @@
...
+
+	private Class<?> partitionKeyExtractorClass = null;
+
+	private Class<?> partitionSelectorClass = null;
+
+	private Expression partitionSelectorExpression = null;
+
+	private boolean batchingEnabled = false;
+
+	private int batchSize;
```
ilayaperumalg: Is it ok to have these batch properties with the default `0` values?

mbogoevici: At this point, I'm wondering if we shouldn't make these implementation specific (Even duplicate). They mean different things in different contexts (Redis, Rabbit batch messages, Kafka batches bytes).

mbogoevici: Actually @ilayaperumalg points out that they're used only by Rabbit. There we go.

markfisher: so then they should move to `RabbitProducerProperties` only?

mbogoevici: ya

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
markfisher: okay

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/9b0c4bd627f7d122be60d11caf993b18799b3ae6

PR #431, opened 2016-03-15
Add basic validation for producer/consumer properties

### LINKED ISSUE #425: Validate consumer and producer properties
Ideally, we'd want to be able to use JSR-303 to do it, but at the bare minimum, mutually exclusive values should be validated, same for nulls, etc.

ilayaperumalg:
This resolves #425

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.stream.binder.rabbit;
 
+import javax.validation.constraints.Min;
+
 import org.springframework.amqp.core.AcknowledgeMode;
 import org.springframework.util.Assert;
 
@@ -73,6 +75,7 @@ public class RabbitConsumerProperties {
 		this.acknowledgeMode = acknowledgeMode;
 	}
 
+	@Min(value = 1, message = "Max Concurrency should be greater than or equal to 1")
 	public int getMaxConcurrency() {
 		return maxConcurrency;
 	}
@@ -81,6 +84,7 @@ public class RabbitConsumerProperties {
 		this.maxConcurrency = maxConcurrency;
 	}
 
+	@Min(value = 1, message = "Prefetch should be greater than or equal to 1")
 	public int getPrefetch() {
 		return prefetch;
 	}
@@ -97,6 +101,7 @@ public class RabbitConsumerProperties {
 		this.requestHeaderPatterns = requestHeaderPatterns;
 	}
 
+	@Min(value = 1, message = "Tx Size should be greater than or equal to 1")
 	public int getTxSize() {
 		return txSize;
 	}
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
@@ -75,7 +75,7 @@ public class RabbitConsumerProperties {
 		this.acknowledgeMode = acknowledgeMode;
 	}
 
-	@Min(value = 1, message = "Max Concurrency should be greater than or equal to 1")
+	@Min(value = 1, message = "Max Concurrency should be greater than zero.")
 	public int getMaxConcurrency() {
 		return maxConcurrency;
 	}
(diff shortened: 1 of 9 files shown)


ilayaperumalg:
This PR is submitted with the changes on top of the PR #409 and the original commit for this PR is: 2823abc

ilayaperumalg:
Updated the PR after rebasing off master and change to include `AssertTrue` instead of `AssertFalse` on validation.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
```diff
@@ -89,4 +94,14 @@ public void setRequiredGroups(String... requiredGroups) {
 		this.requiredGroups = requiredGroups;
 	}
 
+	@AssertTrue(message = "ParitionKeyExpression and PartitionKeyExtractorClass properties are mutually exclusive.")
```
markfisher: typo: "Parition..."

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
```diff
@@ -89,4 +94,14 @@ public void setRequiredGroups(String... requiredGroups) {
...
+	public boolean isValidPartitionKeyProperty() {
+		return !((this.partitionKeyExpression != null) && (this.partitionKeyExtractorClass != null));
```
markfisher: simpler:

```
return (this.partitionSelectorClass == null) || (this.partitionSelectorExpression == null);
```

mbogoevici: The condition must be a XOR.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
```diff
@@ -74,6 +76,7 @@ public void setAcknowledgeMode(AcknowledgeMode acknowledgeMode) {
 		this.acknowledgeMode = acknowledgeMode;
 	}
 
+	@Min(value = 1, message = "Max Concurrency should be greater than or equal to 1")
```
markfisher: might be worth defining a constant for: "should be greater than or equal to 1"

in fact, I'm wondering if it would be better to say "must be a positive value" (and for the >=0 case, "must not be negative)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
mbogoevici: Actually, you're right @markfisher. Scrap previous comment.

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitProducerProperties.java
```diff
@@ -108,6 +111,7 @@ public void setBatchSize(int batchSize) {
 		this.batchSize = batchSize;
 	}
 
+	@Min(value = 1, message = "Batch Buffer Limit should be greater than or equal to 1")
```
garyrussell: I am not sure a buffer limit of 1 makes much sense; maybe all these should just be "must be greater than zero".

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitConsumerProperties.java
garyrussell: I think "greater than zero" is better.

ilayaperumalg:
Updated the PR after addressing the review comments.

mbogoevici:
Some tests exercising validation would be useful, too.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -235,6 +236,13 @@ public String getBinder(String channelName) {
 			dataBinder.setIgnoreUnknownFields(ignoreUnknownProperties);
 			dataBinder.setDisallowedFields(bindingPropertyFields);
 			dataBinder.bind(new PropertySourcesPropertyValues(applicationContext.getEnvironment().getPropertySources()));
+			CustomValidatorBean customValidator = new CustomValidatorBean();
```
mbogoevici: This has been rendered obsolete, as we don't do our own binding anymore. However, it would be useful to have an integration tests where invalid values are actually detected.

ilayaperumalg:
Rebased and updated the PR with the changes required. Also added tests to verify the validation.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/173002154737524f9b62ce36b6fe3b9d625624b5

PR #432, opened 2016-03-16
Introduce BindingListener annotation

### LINKED ISSUE #156: If consumer and producer agree on the content type of a message coercion should be automatic
If a consumer and producer *don't* specify a content type (via `spring.cloud.stream.bindings.*.contentType`) then as long as they share a class definition for the payload (a domain object) then the message will flow. That's uncomfortable for 2 reasons: first it forces  them to share the domain class (or have an identical class with the same FQCN), and second it uses a default content type which is not interopable (Java serialization basically, not JSON or protobuf or whatever).

As soon as you specify the content type (e.g. as `application/json` for interop / human readable messages), the system breaks because JSON gets converted to a String and the receiver can't process it as a domain object any more (he has to manually convert the String to a domain object, going through 2 layers of conversion, from String to JSON to Object).

I propose that if the consumer and producer agree on the content type (e.g. `application/json`) then Spring Cloud Stream (or Spring Integration) should be able to handle the conversion and type coercion.

Marius has pointed out that you have a similar situation in REST endpoints where the solution is `HttpMessageConverters` that are mime-type as well as Java type aware. The payload is essentially bytes (or Strings) as far as the "dumb pipes" model is concerned (Spring Cloud Stream), but it carries content type headers, a bit like an HTTP request.

mbogoevici:
Fixes #156

- Supports Spring Messaging infrastructure for argument mapping and type conversion, sharing the same configured converters as the ones configured on input/output channels;

- Also: refactored AbstractFromMessageConverter to return only payload - conversion to a message enriched with the content type of the channel is done by a separate wrapping converter;

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/BindingListenerTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/BindingListenerTests.java
@@ -1,0 +1,313 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.config;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasProperty;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.collection.IsMapContaining.hasEntry;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Test;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.cloud.stream.annotation.BindingListener;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.messaging.Processor;
+import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.cloud.stream.test.binder.MessageCollector;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.handler.annotation.Header;
+import org.springframework.messaging.handler.annotation.Headers;
+import org.springframework.messaging.handler.annotation.Payload;
+import org.springframework.messaging.handler.annotation.SendTo;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class BindingListenerTests {
+
+	@Test
+	public void testContentTypeConversion() throws Exception {
+		ConfigurableApplicationContext context = SpringApplication.run(TestSink.class);
+		@SuppressWarnings("unchecked")
+		TestSink testSink = context.getBean(TestSink.class);
+		Sink sink = context.getBean(Sink.class);
+		String id = UUID.randomUUID().toString();
+		sink.input().send(MessageBuilder.withPayload("{\"bar\":\"barbar" + id + "\"}")
+				.setHeader("contentType","application/json").build());
+		assertTrue(testSink.latch.await(10, TimeUnit.SECONDS));
+		assertThat(testSink.receivedArguments, hasSize(1));
+		assertThat(testSink.receivedArguments.get(0), hasProperty("bar", equalTo("barbar" + id)));
+		context.close();
+	}
+
+	@Test
+	public void testAnnotatedArguments() throws Exception {
+		ConfigurableApplicationContext context = SpringApplication.run(TestPojoWithAnnotatedArguments.class);
+		@SuppressWarnings("unchecked")
+		TestPojoWithAnnotatedArguments testPojoWithAnnotatedArguments = context.getBean(TestPojoWithAnnotatedArguments.class);
+		Sink sink = context.getBean(Sink.class);
+		String id = UUID.randomUUID().toString();
+		sink.input().send(MessageBuilder.withPayload("{\"bar\":\"barbar" + id + "\"}")
+								  .setHeader("contentType", "application/json").setHeader("testHeader", "testValue").build());
+		assertThat(testPojoWithAnnotatedArguments.receivedArguments, hasSize(3));
+		assertThat(testPojoWithAnnotatedArguments.receivedArguments.get(0), instanceOf(FooPojo.class));
+		assertThat(testPojoWithAnnotatedArguments.receivedArguments.get(0), hasProperty("bar", equalTo("barbar" + id)));
+		assertThat(testPojoWithAnnotatedArguments.receivedArguments.get(1), instanceOf(Map.class));
+		assertThat((Map<String, String>) testPojoWithAnnotatedArguments.receivedArguments.get(1), hasEntry(MessageHeaders.CONTENT_TYPE, "application/json"));
+		assertThat((Map<String, String>) testPojoWithAnnotatedArguments.receivedArguments.get(1), hasEntry(equalTo("testHeader"), equalTo("testValue")));
+		assertThat((String) testPojoWithAnnotatedArguments.receivedArguments.get(2), equalTo("application/json"));
+		context.close();
+	}
+
+	@Test
+	public void testReturn() throws Exception {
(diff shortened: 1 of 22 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -305,7 +305,7 @@ protected RetryTemplate buildRetryTemplateIfRetryEnabled(ConsumerProperties prop
 	 * @author David Turanski
 	 * @author Ilayaperumal Gopinathan
 	 */
-	abstract static class JavaClassMimeTypeConversion {
+	public abstract static class JavaClassMimeTypeConversion {
```
mbogoevici: Not sure that we need this.

mbogoevici:
Also, will try to align imports and indentation.

mbogoevici:
Also, requires support for proxied beans.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/BindingListener.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @see {@link MessageMapping}
+ * @see {@link org.springframework.messaging.handler.annotation.SendTo}
+ * @author Marius Bogoevici
+ */
+@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@MessageMapping
+@Documented
+public @interface BindingListener {
```
mbogoevici: Another candidate name could be `@BoundChannelListener` although in theory the binding can be something else than a channel

mbogoevici: @markpollack suggests StreamListener

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/BindingListener.java
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * The name of the binding that the method subscribes to.
```
markfisher: I guess technically it's the name of the "channel" that is itself bound.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/BindingListener.java
```diff
@@ -0,0 +1,61 @@
...
+	 *
+	 */
+	String value() default "";
```
markfisher: could "input" be the default? ... or what do you do with blank?

...and sorry for not looking elsewhere before asking ;)

mbogoevici: I've considered INPUT as a default, ended up forcing them to provide a value. Either way works I guess. INPUT is a bit too magical IMO

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/BindingListener.java
mbogoevici: Yeah, I'm trying to stay away from mentioning channels in the same way as the `Binder` interface does. As in theory this could apply to other bound entities.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+			@Override
+			public void doWith(final Method method) throws IllegalArgumentException, IllegalAccessException {
+				BindingListener bindingListener = AnnotationUtils.findAnnotation(method, BindingListener.class);
+				if (bindingListener != null) {
+					Assert.hasText(bindingListener.value(), "The binding name cannot be null");
+					final InvocableHandlerMethod invocableHandlerMethod = messageHandlerMethodFactory.createInvocableHandlerMethod(bean, method);
+					if (!StringUtils.hasText(bindingListener.value())) {
+						throw new BeanInitializationException("Only one of 'bindings' or 'value' can be specified");
+					}
+					SubscribableChannel channel = applicationContext.getBean(bindingListener.value(),
```
markpollack: There could be more than one listener wired to the same channel, maybe want to warn or have a 'strict' mode to control this case.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/BindingListenerTests.java
```diff
@@ -0,0 +1,313 @@
...
+			receivedMessages.add(fooMessage);
+			BazPojo bazPojo = new BazPojo();
+			bazPojo.setQux(fooMessage.getPayload());
+			return bazPojo;
+		}
+	}
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class TestPojoThrowingException {
```
markpollack: not used.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+import org.springframework.messaging.handler.annotation.SendTo;
+import org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory;
+import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
+import org.springframework.util.Assert;
+import org.springframework.util.ObjectUtils;
+import org.springframework.util.ReflectionUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Marius Bogoevici
```
markpollack: missing JavaDoc

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -27,57 +25,49 @@
...
-import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.converter.SmartMessageConverter;
 import org.springframework.messaging.support.ChannelInterceptorAdapter;
 import org.springframework.util.Assert;
-import org.springframework.util.CollectionUtils;
 import org.springframework.util.MimeType;
 import org.springframework.util.StringUtils;
 
 /**
+<<<<<<< HEAD
```
markpollack: some funky leftover merge artifacts...

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -117,10 +93,10 @@ public void configureMessageChannel(MessageChannel channel, String channelName)
 		final String contentType = bindingProperties.getContentType();
 		if (bindingProperties != null && StringUtils.hasText(contentType)) {
```
markpollack: can't be null, just accessed it on the previous line

markpollack: bindingProps can't be null, just accessed on the previous line.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,157 @@
...
+																						  "can return a value");
+					}
+					BindingMessageHandler handler = new BindingMessageHandler(invocableHandlerMethod);
+					handler.setApplicationContext(applicationContext);
+					handler.setChannelResolver(binderAwareChannelResolver);
+					if (!StringUtils.isEmpty(defaultOutputChannel)) {
+						handler.setOutputChannelName(defaultOutputChannel);
+					}
+					handler.afterPropertiesSet();
+					// // TODO: add support for pollable channels
```
markpollack: Create an issue and link here..

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.java
```diff
@@ -221,4 +236,13 @@ public Object postProcessAfterInitialization(Object bean, String beanName) throw
...
 		}
 	}
+
+	@Bean
+	public static BindingListenerAnnotationBeanPostProcessor bindToAnnotationBeanPostProcessor(@Lazy BinderAwareChannelResolver binderAwareChannelResolver, @Lazy CompositeMessageConverterFactory compositeMessageConverterFactory) {
+		DefaultMessageHandlerMethodFactory messageHandlerMethodFactory = new DefaultMessageHandlerMethodFactory();
+		messageHandlerMethodFactory.setMessageConverter(compositeMessageConverterFactory.getMessageConverterForAllRegistered());
+		messageHandlerMethodFactory.afterPropertiesSet();
+		BindingListenerAnnotationBeanPostProcessor bindingListenerAnnotationBeanPostProcessor = new BindingListenerAnnotationBeanPostProcessor(binderAwareChannelResolver, messageHandlerMethodFactory);
+		return bindingListenerAnnotationBeanPostProcessor;
```
markpollack: just return new instance w/o local var.

mbogoevici:
Pushed.

garyrussell:
I don't know how easy it will be to incorporate this in the general `StreamListener` case, but I just issued a PR for RabbitMQ to add information to the `MessageProperties` to help the JSON message converter - instead of requiring `__TypeId__` header(s), we now add an `inferredArgumentType` property (for the target method) as well as the `targetMethod` and `targetObject` properties - so custom converters will have full context of the target bean/method when performing conversion.

https://github.com/spring-projects/spring-amqp/pull/390

This is primarily to support receiving JSON over AMQP from foreign systems (where no type information headers are available), but might also be applicable here.

markpollack:
squashed and merged as 4d2ef3c4d0c7102a1dfdb8a1e1bb22064adb5f32

PR #440, opened 2016-03-17
GH-439: Add 'transacted' to Rabbit Producer Props

### LINKED ISSUE #401: Remove code in binders related to original use in Spring XD
In https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java#L469  there is code that creates a new classloader that was originally done to avoid errors such as https://jira.spring.io/browse/XD-2509

As there are not multiple classloaders in spring-cloud-stream's runtime, it is quite likely this code should be removed.

### LINKED ISSUE #439: Rabbit Binder Transacted Producer

garyrussell:
Fixes #439
Resolves #401 

I suggest reviewing on GitHub with `?w=1` in the URL to ignore whitespace changes - lots of them a big indented try/finally is removed.

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
@@ -23,9 +23,6 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.Map;
 
-import com.rabbitmq.client.AMQP;
-import com.rabbitmq.client.Channel;
-import com.rabbitmq.client.Envelope;
 import org.aopalliance.aop.Advice;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -89,6 +86,10 @@ import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
+import com.rabbitmq.client.AMQP;
+import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.Envelope;
+
 /**
  * A {@link org.springframework.cloud.stream.binder.Binder} implementation backed by RabbitMQ.
  *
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
```diff
@@ -291,70 +291,63 @@ public void onInit() {
...
-			if (maxAttempts > 1 || properties.isRepublishToDlq()) {
-				RetryOperationsInterceptor retryInterceptor = RetryInterceptorBuilder.stateless()
-						.maxAttempts(maxAttempts)
-						.backOffOptions(properties.getBackOffInitialInterval(),
-								properties.getBackOffMultiplier(),
-								properties.getBackOffMaxInterval())
-						.recoverer(determineRecoverer(name, properties.getPrefix(), properties.isRepublishToDlq()))
-						.build();
-				listenerContainer.setAdviceChain(new Advice[] { retryInterceptor });
+		ClassUtils.overrideThreadContextClassLoader(SimpleMessageListenerContainer.class.getClassLoader());
```
mbogoevici: We can do away with overriding the TCCL here, too, I think, especially as it is not restored at the end of the method. This has little relevance for Boot applications anyway.

mbogoevici:
Everything else LGTM. Will merge.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/fa035296f2338d024fcf5f1178d26e69d3a4c360

spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/RabbitMessageChannelBinder.java
garyrussell: Thanks; good catch!

PR #448, opened 2016-03-24
Elliminate duplicate configuration options for Rabbit MQ

### LINKED ISSUE #447: Use `RabbitProperties` for setting Rabbit MQ properties
Currently the Rabbit MQ binder uses a configuration set of its own for configuring the clustered settings. Remove the duplicate properties and use the already bound `RabbitProperties` class.

mbogoevici:
Fixes #447

Some binder configuration properties duplicate options already found in Spring Boot.

### DIFF
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -584,26 +584,15 @@ implementations.
 
 ==== Rabbit MQ Binder properties
 
-The binder supports the all Spring Boot properties for Rabbit MQ configuration.
+By default, the binder uses the Spring Boot `ConnectionFactory` and therefore it supports all the Spring Boot configuration options for Rabbit MQ.
+For reference, consult the [Spring Boot documentation](http://docs.spring.io/spring-boot/docs/1.3.3.RELEASE/reference/htmlsingle/#common-application-properties). Rabbit MQ configuration options use the `spring.rabbitmq` prefix.
 
 In addition to that, it also supports the following properties:
 
-spring.cloud.stream.rabbit.binder.addresses::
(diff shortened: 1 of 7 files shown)


spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -584,26 +584,15 @@ implementations.
 
 ==== Rabbit MQ Binder properties
 
-The binder supports the all Spring Boot properties for Rabbit MQ configuration.
+By default, the binder uses the Spring Boot `ConnectionFactory` and therefore it supports all the Spring Boot configuration options for Rabbit MQ.
+For reference, consult the [Spring Boot documentation](http://docs.spring.io/spring-boot/docs/1.3.3.RELEASE/reference/htmlsingle/#common-application-properties). Rabbit MQ configuration options use the `spring.rabbitmq` prefix.
```
garyrussell: Better to use the version-less symlink

http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/

will fix on merge

mbogoevici: Agreed, but on the long run I think we need to match built documentation with release versions of dependencies. Documentation for older builds may end up referencing more recent versions of Spring Boot (e.g. this may end up referencing Spring Boot 1.4 when 1.4 is `current`, which may contain in turn properties that aren't really supported here - it's a minimal risk, so for now I think it is fine).

garyrussell:
Rebased, merged as ba371ab67a43367884f4d64537fed90bbc283f2b

garyrussell:
Good point - I already merged with current - for our projects, we maintain, e.g. 4.2.x symlinks.

However, one problem with referencing lots of different versions is it makes SEO less effective.

Issue #428, opened 2016-03-15
AggregateBuilder ignores user-supplied namespace


dsyer:
E.g. with this

```
		new AggregateApplicationBuilder(IngesterApplication.class, args)
				.from(FileSourceApplication.class).to(IngesterProcessor.class)
				.namespace("ingester").run();
```

I would expect be able to extract the input channel from the `IngesterProcessor` as "ingester.input" but it is registered as `com.example.IngesterProcessor_1.input`. This is because the namespace is discarded inside the `AggregateBuilder`:

```
	private static void createChildContexts(ConfigurableApplicationContext parentContext,
			Class<?>[] apps, String args[][]) {
		for (int i = apps.length - 1; i >= 0; i--) {
			String appClassName = apps[i].getName();
			embedApp(parentContext, getNamespace(appClassName, i), apps[i])
					.run(args != null ? args[i] : new String[0]);
		}
	}
```

PR #452, opened 2016-03-25
Enabled namespace propagation in AggregateApplication*

venilnoronha:
I've added conditional logic to `AggregateApplicationBuilder#getNamespace()` so that it returns the namespace if set else generates one as earlier.

Resolves spring-cloud/spring-cloud-stream#428

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -16,6 +16,9 @@
 
 package org.springframework.cloud.stream.aggregate;
 
+import java.util.LinkedHashMap;
+import java.util.Map.Entry;
+
 import org.springframework.boot.Banner.Mode;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.cloud.stream.messaging.Processor;
@@ -73,7 +76,7 @@ public class AggregateApplication {
 	public static void runEmbedded(ConfigurableApplicationContext parentContext,
 			Class<?>[] apps, String[][] args) {
 		SharedChannelRegistry bean = parentContext.getBean(SharedChannelRegistry.class);
-		prepareSharedChannelRegistry(bean, apps, null);
+		prepareSharedChannelRegistry(bean, apps);
 		// create child contexts first
 		createChildContexts(parentContext, apps, args);
 	}
@@ -117,21 +120,32 @@ public class AggregateApplication {
 				.parent(applicationContext);
 	}
 
-	protected static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] apps,
-			String namespace) {
+	protected static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] apps) {
+		LinkedHashMap<Class<?>, String> appsToRegister = new LinkedHashMap<>();
+		for (Class<?> app : apps) {
+			appsToRegister.put(app, null);
+		}
+		prepareSharedChannelRegistry(sharedChannelRegistry, appsToRegister);
+	}
+
+	protected static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry,
+			LinkedHashMap<Class<?>, String> appsWithNamespace) {
+		int i = 0;
 		SubscribableChannel sharedChannel = null;
-		for (int i = 0; i < apps.length; i++) {
-			Class<?> app = apps[i];
+		for (Entry<Class<?>, String> appEntry : appsWithNamespace.entrySet()) {
+			Class<?> app = appEntry.getKey();
+			String namespace = appEntry.getValue();
 			String appClassName = app.getName();
 			if (i > 0) {
 				sharedChannelRegistry.register(getNamespace(namespace, appClassName, i)
 						+ "." + INPUT_CHANNEL_NAME, sharedChannel);
 			}
 			sharedChannel = new DirectChannel();
-			if (i < apps.length - 1) {
+			if (i < appsWithNamespace.size() - 1) {
 				sharedChannelRegistry.register(getNamespace(namespace, appClassName, i)
 						+ "." + OUTPUT_CHANNEL_NAME, sharedChannel);
 			}
+			i ++;
 		}
 	}
 
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -30,6 +30,7 @@ import org.springframework.messaging.SubscribableChannel;
  *
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
+ * @author Venil Noronha
  */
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -116,15 +117,20 @@ public ConfigurableApplicationContext run(String[] parentArgs) {
 			apps.add(sinkConfigurer);
 		}
 		List<Class<?>> appsToEmbed = new ArrayList<>();
+		String namespace = null;
 		for (int i = 0; i < apps.size(); i++) {
-			appsToEmbed.add(apps.get(i).getApp());
+			AppConfigurer<?> appConfigurer = apps.get(i);
+			appsToEmbed.add(appConfigurer.getApp());
+			if (namespace == null) {
+				namespace = appConfigurer.namespace;
```
venilnoronha: A doubt that I had, is it desired behavior for all apps (i.e. FileSourceApplication and IngesterProcessor apps) to use a common namespace (i.e. ingester)? I'm referring to the example in #428. @dsyer could you throw some light here?

venilnoronha:
Kindly review and merge. My CLA number is 149720151120072904. Thanks!

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
mbogoevici: The goal of the namespace is to disambiguate between channels with the same name (e.g. two `output` channels that belong to successive `Processors`), but that belong to different applications, as they are registered in the `SharedChannelRegistry`. Therefore namespaces should be different for each embedded app.

mbogoevici: So instead of using the first encountered namespace throughout, we should decide for each app whether we want to use the default (generated) value or the value set through the builder (which should be app-specific).

venilnoronha: Thank you @mbogoevici. I'll modify my implementation.

mbogoevici: No problem, thank you for picking up the issue for fixing! :)

venilnoronha:
@mbogoevici I've made the changes. Could you please review it?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
```diff
@@ -92,12 +96,15 @@ private static void createChildContexts(ConfigurableApplicationContext parentCon
...
 		for (int i = apps.length - 1; i >= 0; i--) {
 			String appClassName = apps[i].getName();
-			embedApp(parentContext, getNamespace(appClassName, i), apps[i])
+			embedApp(parentContext, getNamespace(null, appClassName, i), apps[i])
 					.run(args != null ? args[i] : new String[0]);
 		}
 	}
 
-	protected static String getNamespace(String appClassName, int index) {
+	protected static String getNamespace(String namespace, String appClassName, int index) {
```
ilayaperumalg: Instead of passing `namespace` as a parameter here, we can use this method only when the appConfigurer doesn't have `namespace` set already.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
```diff
@@ -92,12 +96,15 @@ private static void createChildContexts(ConfigurableApplicationContext parentCon
 			Class<?>[] apps, String args[][]) {
 		for (int i = apps.length - 1; i >= 0; i--) {
 			String appClassName = apps[i].getName();
-			embedApp(parentContext, getNamespace(appClassName, i), apps[i])
+			embedApp(parentContext, getNamespace(null, appClassName, i), apps[i])
```
ilayaperumalg: Based on [this] (https://github.com/spring-cloud/spring-cloud-stream/pull/452/files#r58666648), we can avoid passing null here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -115,16 +117,17 @@ public ConfigurableApplicationContext run(String[] parentArgs) {
 		if (this.sinkConfigurer != null) {
 			apps.add(sinkConfigurer);
 		}
-		List<Class<?>> appsToEmbed = new ArrayList<>();
+		LinkedHashMap<Class<?>, String> appsToEmbed = new LinkedHashMap<>();
 		for (int i = 0; i < apps.size(); i++) {
-			appsToEmbed.add(apps.get(i).getApp());
+			AppConfigurer<?> appConfigurer = apps.get(i);
+			Class<?> appToEmbed = appConfigurer.getApp();
+			appsToEmbed.put(appToEmbed, appConfigurer.namespace);
```
ilayaperumalg: At this step we can make sure the namespace is always set before preparing shared channel registry.

ilayaperumalg:
@venilnoronha LGTM, will make a minor change related to setting namespace before preparing shared channel registry while merging.

ilayaperumalg:
Merged as 99f978a262a500c16e8c12c7d3bc9d775ff431c5 along with a minor polishing 68169378279cb3e368fcb860874115bb19aadab4

ilayaperumalg:
@venilnoronha thank you for the contribution!

venilnoronha:
Thank you @ilayaperumalg!

venilnoronha:
@ilayaperumalg I think you missed out a change while making the `getNamespace` changes.

Following is the code in [AggregateApplication](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java)

```
protected static void prepareSharedChannelRegistry(SharedChannelRegistry sharedChannelRegistry, Class<?>[] apps) {
	LinkedHashMap<Class<?>, String> appsToRegister = new LinkedHashMap<>();
	for (Class<?> app : apps) {
		appsToRegister.put(app, null);
	}
	prepareSharedChannelRegistry(sharedChannelRegistry, appsToRegister);
}
```

I think `appsToRegister.put(app, null)` should rather be `appsToRegister.put(app, getNamespace(app.getName(), i))` after your change.

ilayaperumalg:
@venilnoronha yeah, Sorry for missing that out. will fix it shortly.

Also, I think it would be good to support creating AggregateApplication via the `builder` only instead of directly calling the `run` on AggregateApplication (because of its limiting options to override what one can achieve with the builder).

ilayaperumalg:
@venilnoronha thanks for catching this.
Pushed the change 446633126690ad669087326b783690ca2da8ff96 that sets the namespace for the apps when aggregating using AggregateApplication.run(). 
Also, created https://github.com/spring-cloud/spring-cloud-stream/issues/474 to use builder only for creating aggregate application.

venilnoronha:
@ilayaperumalg thanks!

PR #457, opened 2016-03-29
Topic cleanup in KafkaBinder tests

### LINKED ISSUE #327: Add topic cleanup for the Kafka Binder tests

ilayaperumalg:
This resolves #327

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
@@ -36,8 +36,6 @@ import java.util.Collection;
 import java.util.Properties;
 import java.util.UUID;
 
-import kafka.admin.AdminUtils;
-import kafka.api.TopicMetadata;
 import org.hamcrest.collection.IsCollectionWithSize;
 import org.junit.Before;
 import org.junit.ClassRule;
@@ -70,6 +68,9 @@ import org.springframework.messaging.support.MessageBuilder;
 import org.springframework.retry.backoff.FixedBackOffPolicy;
 import org.springframework.retry.policy.SimpleRetryPolicy;
 import org.springframework.retry.support.RetryTemplate;
+
+import kafka.admin.AdminUtils;
+import kafka.api.TopicMetadata;
 
 
 /**
@@ -140,7 +141,7 @@ public class KafkaBinderTests extends PartitionCapableBinderTests<KafkaTestBinde
 
 		DirectChannel moduleOutputChannel = new DirectChannel();
 		DirectChannel moduleInputChannel = new DirectChannel();
-		QueueChannel  dlqChannel = new QueueChannel();
+		QueueChannel dlqChannel = new QueueChannel();
 		FailingInvocationCountingMessageHandler handler = new FailingInvocationCountingMessageHandler();
 		moduleInputChannel.subscribe(handler);
 		ExtendedProducerProperties<KafkaProducerProperties> producerProperties = createProducerProperties();
@@ -185,7 +186,7 @@ public class KafkaBinderTests extends PartitionCapableBinderTests<KafkaTestBinde
 		final ProducerMetadata.CompressionType[] codecs = new ProducerMetadata.CompressionType[] {
 				ProducerMetadata.CompressionType.none,
 				ProducerMetadata.CompressionType.gzip,
-				ProducerMetadata.CompressionType.snappy };
+				ProducerMetadata.CompressionType.snappy};
 
 		byte[] ratherBigPayload = new byte[2048];
 		Arrays.fill(ratherBigPayload, (byte) 65);
@@ -308,12 +309,7 @@ public class KafkaBinderTests extends PartitionCapableBinderTests<KafkaTestBinde
 	@Test
 	@SuppressWarnings("unchecked")
 	public void testDefaultConsumerStartsAtEarliest() throws Exception {
-		KafkaMessageChannelBinder binder = new KafkaMessageChannelBinder(new ZookeeperConnect(kafkaTestSupport.getZkConnectString()),
-				kafkaTestSupport.getBrokerAddress(), kafkaTestSupport.getZkConnectString());
-		GenericApplicationContext context = new GenericApplicationContext();
-		context.refresh();
-		binder.setApplicationContext(context);
-		binder.afterPropertiesSet();
+		KafkaTestBinder binder = getBinder();
 		DirectChannel output = new DirectChannel();
 		QueueChannel input1 = new QueueChannel();
 
@@ -405,12 +401,7 @@ public class KafkaBinderTests extends PartitionCapableBinderTests<KafkaTestBinde
 	@Test
 	@SuppressWarnings("unchecked")
 	public void testResume() throws Exception {
-		KafkaMessageChannelBinder binder = new KafkaMessageChannelBinder(new ZookeeperConnect(kafkaTestSupport.getZkConnectString()),
-				kafkaTestSupport.getBrokerAddress(), kafkaTestSupport.getZkConnectString());
-		GenericApplicationContext context = new GenericApplicationContext();
-		context.refresh();
-		binder.setApplicationContext(context);
-		binder.afterPropertiesSet();
+		KafkaTestBinder binder = getBinder();
 		DirectChannel output = new DirectChannel();
 		QueueChannel input1 = new QueueChannel();
 
@@ -443,15 +434,8 @@ public class KafkaBinderTests extends PartitionCapableBinderTests<KafkaTestBinde
 
 	@Test
 	public void testSyncProducerMetadata() throws Exception {
-		KafkaMessageChannelBinder binder = new KafkaMessageChannelBinder(new ZookeeperConnect(kafkaTestSupport.getZkConnectString()),
-				kafkaTestSupport.getBrokerAddress(), kafkaTestSupport.getZkConnectString());
-		GenericApplicationContext context = new GenericApplicationContext();
-		context.refresh();
-		binder.setApplicationContext(context);
-		binder.afterPropertiesSet();
-
+		KafkaTestBinder binder = getBinder();
 		DirectChannel output = new DirectChannel();
-
 		String testTopicName = UUID.randomUUID().toString();
 		ExtendedProducerProperties<KafkaProducerProperties> properties = createProducerProperties();
 		properties.getExtension().setSync(true);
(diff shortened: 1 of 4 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -107,6 +116,19 @@ public void init() {
...
+	@After
+	public void cleanup() {
+		try {
+			Seq<String> topics = ZkUtils.getAllTopics(kafkaTestSupport.getZkClient());
+			for (String topic: scala.collection.JavaConversions.seqAsJavaList(topics)) {
+				AdminUtils.deleteTopic(kafkaTestSupport.getZkClient(), topic);
+			}
+		}
+		catch (Exception e) {
+			log.error(e.getMessage());
```
markpollack: I am testing with an external broker, I don't think one needs to clean up the topics if the broker is embedded.  I get the following error.  The loop is exited and only "null" is printed. Turns out one need to have `delete.topic.enable=true` in the broker config file to delete topics.  Also, this is being a bit too aggressive, it will delete all topics in my broker, so the test introduces side effects for an external broker.  

A few suggestions:

* Log the exception message.
* Only perform the cleanup if not in embedded mode
* Check if the server property `delete.topic.enable` is true
* In a @Before method, get a list of the topics, then in the @After method get the list again and compute the diff.  That should limit the side effects.
* Maybe make this a @BeforeClass/@AfterClass to do the checking of all topics only twice
* This cleanup method is overriding the base class call which oddly has in it `System.clearProperty("SCS_KAFKA_TEST_EMBEDDED");`  Not sure what that is about.

```
2016-03-29 16:36:26,591 ERROR main o.s.c.s.b.k.KafkaBinderTests:128 - org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /admin/delete_topics/foo.y
org.I0Itec.zkclient.exception.ZkNodeExistsException: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /admin/delete_topics/foo.y
	at org.I0Itec.zkclient.exception.ZkException.create(ZkException.java:55)
	at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:685)
	at org.I0Itec.zkclient.ZkClient.create(ZkClient.java:304)
	at org.I0Itec.zkclient.ZkClient.createPersistent(ZkClient.java:243)
	at kafka.utils.ZkUtils$.createPersistentPath(ZkUtils.scala:306)
	at kafka.admin.AdminUtils$.deleteTopic(AdminUtils.scala:159)
	at kafka.admin.AdminUtils.deleteTopic(AdminUtils.scala)
	at org.springframework.cloud.stream.binder.kafka.KafkaBinderTests.cleanup(KafkaBinderTests.java:124)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:33)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.springframework.cloud.stream.test.junit.AbstractExternalResourceTestSupport$1.evaluate(AbstractExternalResourceTestSupport.java:70)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:78)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:212)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:68)
Caused by: org.apache.zookeeper.KeeperException$NodeExistsException: KeeperErrorCode = NodeExists for /admin/delete_topics/foo.y
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:119)
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:51)
	at org.apache.zookeeper.ZooKeeper.create(ZooKeeper.java:783)
	at org.I0Itec.zkclient.ZkConnection.create(ZkConnection.java:87)
	at org.I0Itec.zkclient.ZkClient$1.call(ZkClient.java:308)
	at org.I0Itec.zkclient.ZkClient$1.call(ZkClient.java:304)
	at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:675)
	... 29 common frames omitted
```

ilayaperumalg: I believe the scope of this PR is intended to delete the topics on **embedded** kafka server in `KafkaBinderTests` in which the test methods share the **same** embedded kafka server. 
That's the reason I went ahead with delete all topics during the cleanup.
Also, I think the KafkaBinderTests always runs with the embbedded kafka server based on what we set as `ClassRule` there. What am I missing?

markpollack: The tests can run with and external kafka server depending on the setting of some environment variables, check the ctor in the rule.  I don't see the need to clean up anything for embedded since everything is destroyed at the end of testing (server/zk instance are ephemeral)

markpollack: Looks like in the current config of the tests it isn't possible to use the external broker as 'true' is harcoded.  (I changed it by hand in the code for my testing).  So either we should support using an external broker or now.  also, even w/ the embedded brokers, if the tests run in parallel then deleting all of them would run into some race conditions.  Might not be worth the effort at this point... it looks like all the destinations created are unique, so if we just for embedded, get rid of the option to use external and we don't need to cleanup anything.

ilayaperumalg: I believe it is still worth to keep the cleanup code just to avoid any overlap in future (given tests run sequentially). Also, it would be good to keep the `AbstractBinderTests` as is (with the embedded mode) and revisit if we need to configure for external broker support. I agree that the KafkaTestSupport code requires some cleanup related to the external broker support.

mbogoevici: No cleanup is not needed if we run things embedded (that being said, we shouldn't run embedded by default by hardcoding the environment - we can just change the default in the rule). We can use the rule for checking whether we're running embedded or external. I guess for deciding what topics have been created we can do the same thing that Rabbit does, and record them in the test binder.

ilayaperumalg:
Updated the PR to use `KafkaTestBinder` for cleanup and to support running from external broker.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaTestBinder.java
```diff
@@ -18,8 +18,8 @@
 
 import java.util.List;
 
-import com.esotericsoftware.kryo.Kryo;
-import com.esotericsoftware.kryo.Registration;
+import org.slf4j.Logger;
```
mbogoevici: Use `commons-logging` instead.

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -32,10 +32,11 @@
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
 
-import kafka.api.OffsetRequest;
 import org.junit.Before;
 import org.junit.ClassRule;
 import org.junit.Test;
+import org.slf4j.Logger;
```
mbogoevici: Use commons-logging (and throughout the project).

spring-cloud-stream-test-support-internal/src/main/java/org/springframework/cloud/stream/test/junit/kafka/KafkaTestSupport.java
```diff
@@ -70,8 +68,14 @@
 	private static boolean hasFailedAlready = false;
 
 	static {
-		// check if either the environment or Java property is set to use embedded tests
-		defaultEmbedded = "true".equals(System.getenv(SCS_KAFKA_TEST_EMBEDDED)) || "true".equals(System.getProperty(SCS_KAFKA_TEST_EMBEDDED));
+		if (System.getProperty(SCS_EXTERNAL_SERVERS_REQUIRED) != null &&
```
mbogoevici: This feels like a slight abuse of `SCS_EXTERNAL_SERVERS_REQUIRED`, whose role is to forcibly fail tests if the external servers are not present and the rules are not satisfied (thus preventing certain potential failures to get unnoticed). I believe we should reinstate the separate setting for an embedded Kafka server. Thougts?

ilayaperumalg: If someone specifies `SCS_EXTERNAL_SERVERS_REQUIRED` to **true** explicitly, then the resource existence check will make sure to fail with the following error:
```
java.lang.AssertionError: KAFKA IS NOT AVAILABLE
	at org.junit.Assert.fail(Assert.java:88)
```

The above check is only needed to make sure `SCS_EXTERNAL_SERVERS_REQUIRED` is not set to `true` and proceed to use the `embedded` server for tests.

ilayaperumalg:
Pushed the change for `commons-logging` fix. Will create a new PR for rest of the codebase.

ilayaperumalg:
Updated the PR to use `SCS_KAFKA_TEST_EMBEDDED` along with `SCS_EXTERNAL_SERVERS_REQUIRED`

mbogoevici:
Closing this for now as it belongs to the Kafka binder implementation.

PR #415, opened 2016-03-14
Support pollable channel as producer output channel

ilayaperumalg:
- When the message producer has the output channel of type `PollableChannel`, bridge the pollable channel to a subscribable channel and bind the subscribable channel to the broker.
  - This is needed since the binder expects the producer output channel is of type `SubscribableChannel`

 - Assign bean name for the `ConsumerEndpointFactoryBean` that is created when bridging the pollable channel to a subscribable channel
 - Fix isPollable method to use AssignableFrom instead of the `equals` check

This resolves #410

PR #460, opened 2016-03-30
Support @StreamListener in PollableChannels

### LINKED ISSUE #436: Support @StreamListener with PollableChannels

ilayaperumalg:
- If the StreamListener is annotated with PollableChannel, then bridge the pollable channel to subscribable channel
and use the subscribable channel to subscribe to the StreamListenerMessageHandler

This resolves #436

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTests.java
@@ -41,6 +41,7 @@ import org.springframework.beans.factory.BeanCreationException;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.annotation.Input;
 import org.springframework.cloud.stream.annotation.StreamListener;
 import org.springframework.cloud.stream.messaging.Processor;
 import org.springframework.cloud.stream.messaging.Sink;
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
Created this PR on top of the PR https://github.com/spring-cloud/spring-cloud-stream/pull/415 so that we can make use of the `PollableToSubscribableBridge`. The actual change for this PR is from this commit: https://github.com/spring-cloud/spring-cloud-stream/pull/460/commits/985d2f9a298c7d75206c5d7a849a779248f1f3b3

garyrussell:
I need to look at this more closely, but it seems much more complicated than needed.

I think we simply need to wrap the `StreamListenerMessageHandler` in a `PollingConsumer` rather than creating a bridge.

But, I see the bridging mechanism is already in the `BindableProxyFactory`, using a `ConsumerEndpointFactoryBean` - I need to talk to @mbogoevici to understand why that was the approach.

garyrussell:
That said, does this "feature" even make sense? With an unbounded pollable input channel, the binder is likely to outpace any app and memory will fill up. That can be avoided with a bounded channel but, even then, the binder will immediately ack queued messages and they will be lost in the event of a failure.

ilayaperumalg:
> I think we simply need to wrap the StreamListenerMessageHandler in a PollingConsumer rather than creating a bridge

yeah, that's a good idea. I will give that a try.

For the feature question, I think we still want to support `PollableChannel` with the `@StreamListener` when someone has a usecase where the binder is likely to receive messages no faster than the polling frequency.

ilayaperumalg:
Also, the bridging functionality in `BindableProxyFactory` is used **only** for the shared channels in Aggregate applications. 
The feature we introduced as part of the PR #415 uses the bridging mechanism but that is for the output channels only.

ilayaperumalg:
Updated the PR that uses `PollingConsumer` instead of the bridge.

mbogoevici:
@garyrussell 

> That said, does this "feature" even make sense? With an unbounded pollable input channel, the binder is likely to outpace any app and memory will fill up. That can be avoided with a bounded channel but, even then, the binder will immediately ack queued messages and they will be lost in the event of a failure.

Good question. I think that using `PollableChannels` in general has some loss potential (whether inbound and outbound), so perhaps we can document that or just decide whether we want to support that.

mbogoevici:
Closing this as per https://github.com/spring-cloud/spring-cloud-stream/issues/436#issuecomment-205430607.

PR #465, opened 2016-04-01
Allow AggregateApplicationBuilder with no arg constructor

### LINKED ISSUE #464: AggregateApplicationBuilder doesn't set parentContext correctly when using no sources
When invoking the `AggregateApplicationBuilder` without passing any sources for the parent configuration, the parent context doesn't get set which results in missing `SharedChannelRegistry` for the child contexts to find the shared channels when binding.

For example, this doesn't work now: https://github.com/spring-cloud/spring-cloud-stream-samples/blob/master/double/src/main/java/demo/DoubleApplication.java#L29

ilayaperumalg:
- When the AggregateApplicationBuilder doesn't have constructor args set, then invoke parent configuration setup with no args.
  - This will make sure to set the parent context for the child application contexts

This resolves #464


mbogoevici:
Merged via https://github.com/spring-cloud/spring-cloud-stream/commit/12bd4effa1f715d89980cd56da184160b1887f1d

PR #470, opened 2016-04-04
Edit reference manual

fifthposition:
A few notes:

* There were a couple of headers without content (“Type-Converting Message Channels” and “@StreamListener and Conversion”). I’ve commented those out.
* Images were missing `images/` in their file paths. I’m not sure whether they were still being rendered without that—they weren’t on my machine (locally), and they seem not to render currently in [GitHub’s view of the manual file](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc). I added the directory.
* The image [SCSt-with-binder.png](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-docs/src/main/asciidoc/images/SCSt-with-binder.png) is referenced twice, once with the caption “Spring Cloud Stream Application” and once with the caption “Spring Cloud Stream Application topologies”. (I’ve left those both.)
* The heading “Programming Model” is used twice (as I recall, one is a subsection of the other; I’ve left these both as well).

### DIFF
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -3,20 +3,13 @@
 
 [partintro]
 --
-This section goes into more detail about how you can work with Spring Cloud Stream. It covers topics
-such as creating and running stream applications.
+This section goes into more detail about how you can work with Spring Cloud Stream. It covers topics such as creating and running stream applications.
 --
 
 == Introducing Spring Cloud Stream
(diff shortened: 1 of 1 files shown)


mbogoevici:
Hi Ben,

Thanks for the updates. A couple of things:

a) generally it would be preferable if the PR was created from a fork. This prevents proliferating branches in the main repository (which we usually reserve for maintenance)

b) we prefer [one sentence per line](http://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line), which helps tracking changes across versions easier.

Anyway, appreciate the PR!

fifthposition:
Thanks @mbogoevici! (Sorry—should have checked before making My First PR.™ :smiley:) I’ll fix these.

fifthposition:
Closing. See #471.

PR #471, opened 2016-04-05
Edit reference manual

fifthposition:
(Description copied from #470.)

A few notes:

* There were a couple of headers without content (“Type-Converting Message Channels” and “@StreamListener and Conversion”). I’ve commented those out.
* Images were missing `images/` in their file paths. I’m not sure whether they were still being rendered without that—they weren’t on my machine (locally), and they seem not to render currently in [GitHub’s view of the manual file](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc). I added the directory.
* The image [SCSt-with-binder.png](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-docs/src/main/asciidoc/images/SCSt-with-binder.png) is referenced twice, once with the caption “Spring Cloud Stream Application” and once with the caption “Spring Cloud Stream Application topologies”. (I’ve left those both.)
* The heading “Programming Model” is used twice (as I recall, one is a subsection of the other; I’ve left these both as well).

### DIFF
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -3,20 +3,17 @@
 
 [partintro]
 --
-This section goes into more detail about how you can work with Spring Cloud Stream. It covers topics
-such as creating and running stream applications.
+This section goes into more detail about how you can work with Spring Cloud Stream.
+It covers topics such as creating and running stream applications.
 --
 
 == Introducing Spring Cloud Stream
+Spring Cloud Stream is a framework for building message-driven microservice applications.
+Spring Cloud Stream builds upon Spring Boot and uses Spring Integration to provide connectivity to message brokers.
+It provides opinionated configuration of middleware from several vendors, introducing the concepts of persistent publish-subscribe semantics, consumer groups, and partitions.
 
-Spring Cloud Stream is a framework for building message-driven microservices.
-Spring Cloud Stream builds upon Spring Boot to create DevOps friendly microservice applications and Spring Integration to provide connectivity to message brokers.
-Spring Cloud Stream provides an opinionated configuration of message brokers, introducing the concepts of persistent pub/sub semantics, consumer groups and partitions across several middleware vendors.
-This opinionated configuration provides the basis to create stream processing applications.
-
-By adding `@EnableBinding` to your main application, you get immediate connectivity to a message broker and by adding `@StreamListener` to a method, you will receive events for stream processing.
-
-Here's a sample sink application for receiving external messages:
+You can add the `@EnableBinding` annotation to your application to get immediate connectivity to a message broker, and you can add `@StreamListener` to a method to cause it to receive events for stream processing.
+The following is a simple sink application which receives external messages.
 
 [source,java]
 ----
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -40,9 +37,11 @@ public class TimerSource {
 }
 ----
 
-`@EnableBinding` is parameterized by one or more interfaces (in this case a single `Sink` interface), which declares input and/or output channels.
-The interfaces `Source`, `Sink` and `Processor` are provided but you can define others.
-Here's the definition of `Source`:
+The `@EnableBinding` takes one or more interfaces as parameters (in this case, the parameter is a single `Sink` interface).
```
mbogoevici: We can drop the 'The' here, I think.

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -3,20 +3,17 @@
...
 --
-This section goes into more detail about how you can work with Spring Cloud Stream. It covers topics
-such as creating and running stream applications.
+This section goes into more detail about how you can work with Spring Cloud Stream.
+It covers topics such as creating and running stream applications.
 --
 
 == Introducing Spring Cloud Stream
+Spring Cloud Stream is a framework for building message-driven microservice applications.
+Spring Cloud Stream builds upon Spring Boot and uses Spring Integration to provide connectivity to message brokers.
```
mbogoevici: I just think we should have a simple rationale for using Boot. Instead of DevOps we could say: 
'Spring Cloud Stream builds upon Spring Boot to create standalone, production-grade Spring Applications, and uses Spring Integration etc.'

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -76,104 +77,132 @@ public class StreamApplicationTests {
...
-Spring Cloud Stream provides a number of abstractions and primitives that simplify writing message-driven microservices.
-In this section we will provide an overview of:
+Spring Cloud Stream provides a number of abstractions and primitives that simplify the writing of message-driven microservice applications.
+This section gives an overview of the following:
 
-* Spring Cloud Stream application model together with the Binder abstraction
-* Persistent publish-subscribe and consumer group support
-* Partitioning
-* Pluggable Binder API
+* A standard application model
```
mbogoevici: Maybe we can just drop 'standard' (in the sense that there's no non-standard alternative) and just say 'The application mode' ?

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -76,104 +77,132 @@ public class StreamApplicationTests {
...
+A Spring Cloud Stream application consists of a middleware-neutral core.
+The application communicates with the outside world through input and output _channels_ injected into it by Spring Cloud Stream.
+Channels are connected to external brokers through middleware-specific Binder implementations.
 
 .Spring Cloud Stream Application
-image::SCSt-with-binder.png[width=300,scaledwidth="50%"]
+image::images/SCSt-with-binder.png[width=300,scaledwidth="50%"]
```
mbogoevici: This will unfortunately throw off the maven build as the build process inserts the `images` path.

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
fifthposition: I’ve added “annotation”, which is what I suspect I meant to do. :smiley:

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
fifthposition: Oops. I’ve re-removed the directory from the paths.

mbogoevici: Thanks! We need to improve this I am sure (I believe it's something with docbook being on the pipeline) but until we do, it's a bit confusing.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/9a92bda9c121779bb08d725fbab2a0f4701d28d3

PR #472, opened 2016-04-05
Insist Message channel types on input/output

### LINKED ISSUE #469: Make it clear that binding is only for SubscribableChannel (input) and MessageChannel (output)
In initial versions of Spring Cloud Stream, we discussed about the possibility of supporting PollableChannels and kept the door open for it. This never materialized, as the binder implementations don't actually support it, but aggregation does. In the end, we do not intend to support PollableChannels directly (although SCSt users could bridge them to bound channels if necessary).

The intent of this would be to clean things up before 1.1.0.RC2.

ilayaperumalg:
- Add necessary assertions on the type of message channels for input/output
  - Make sure to support message channel of type `MessageChannel` for input
  - support message channel of type `SubscribableChannel` for output and if the type is exactly equal to `MessageChannel` type then create subscribable channel
 - Remove references of creating PollableChannel for binding

This resolves #469

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableChannelFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableChannelFactory.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.stream.binding;
 
-import org.springframework.messaging.PollableChannel;
 import org.springframework.messaging.SubscribableChannel;
 
 /**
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
We can also do some cleanup on using `PollableChannel` in Aggregate application based on [this] (https://github.com/spring-cloud/spring-cloud-stream/issues/469#issuecomment-205656155)

ilayaperumalg:
Pushed a commit to remove `bridging` of shared channels in Aggregate application. I couldn't recollect a reason to do the bridging here. If there is a specific reason, then the commit 4c5a237 can be omitted from this PR.

mbogoevici:
The bridging was done because the interfaces could potentially have different types of channels on each end. This is not possible anymore, so the PR is correct in that respect.

mbogoevici:
Closed via: https://github.com/spring-cloud/spring-cloud-stream/commit/a0cdbdc1486521c4c15b923f75b1e2e03532cf3d w/additional changes in https://github.com/spring-cloud/spring-cloud-stream/commit/cd81f244511ea12792c1bba9ce02dea11bbd9d11

PR #473, opened 2016-04-05
Use Jackson ObjectMapper from AutoConfiguration

### LINKED ISSUE #466: Make Objectmapper user in PojoToJsonMessageConverter and JsonToPojoMessageConverter configurable.
Hi,

We'd like to user LocalDateTime in our object that's being sent to Kafka. We do publish the messages in JSON. We cannot use converters defined in jackson-datatype-jsr310 because the ObjectMapper used within mentioned classes isn't configurable.

Kind regards

ilayaperumalg:
- For the message converters that use Jackson ObjectMapper, autowire the ObjectMapper from JacksonAutoConfiguration
   - This will allow configuration of ObjectMapper using boot configuration properties `JacksonProperties`

 - Add test

This resolves #466

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
@@ -16,6 +16,7 @@
 package org.springframework.cloud.stream.config;
 
 import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
@@ -39,6 +40,9 @@ import org.springframework.messaging.MessagingException;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 import org.springframework.tuple.Tuple;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+
 /**
  * @author Ilayaperumal Gopinathan
  */
@@ -48,6 +52,9 @@ public class MessageChannelConfigurerTests {
 
 	@Autowired @Bindings(TestSink.class)
 	private Sink testSink;
+
+	@Autowired
+	private ObjectMapper objectMapper;
 
 	@Test
 	public void testMessageConverterConfigurer() throws Exception {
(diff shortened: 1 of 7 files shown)


spring-cloud-stream-integration-tests/src/test/resources/org/springframework/cloud/stream/config/channel/sink-channel-configurers.properties
```diff
@@ -1,2 +1,3 @@
 spring.cloud.stream.bindings.input.destination=configure1
 spring.cloud.stream.bindings.input.contentType=application/x-spring-tuple
+spring.jackson.serialization.WRITE_DATES_AS_TIMESTAMPS:false
```
mbogoevici: Nitpick, but we should use the `=` sign throughout. Will fix on merge.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.java
```diff
@@ -83,6 +85,9 @@
 	@Autowired
 	private MessageBuilderFactory messageBuilderFactory;
 
+	@Lazy @Autowired(required = false)
```
mbogoevici: Why should this be `@Lazy` ?

ilayaperumalg: This is because of the `CompositeMessageConverterFactory` which is lazily initialised as well as autowired `ObjectMapper`.

mbogoevici: But a lazily referenced `CompositeMessageConverterFactory ` doesn't necesarily mean that the `ObjectMapper` it is injected with should be a lazy reference as well.

ilayaperumalg: I just tried removing `@Lazy` on the ObjectMapper and `SpelExpressionConverterConfigurationTests` fails. I couldn't figure out what exactly causes this but I guess something is going on with the BeanPostProcessor.

mbogoevici: Ok, will check. It may be necessary after all, but I'm trying to understand why.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
```diff
@@ -66,6 +73,12 @@ public void handleMessage(Message<?> message) throws MessagingException {
 		assertTrue(latch.await(10, TimeUnit.SECONDS));
 		testSink.input().unsubscribe(messageHandler);
 	}
+
+	@Test
+	public void testObjectMapperConfig() {
+		assertNotNull( "ObjectMapper should exist", objectMapper);
```
mbogoevici: The autowiring would fail if the bean didn't exist.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.java
mbogoevici: Yes, it seems like we're triggering the early initialization of an entire tree of beans, and the postprocessor that installs the tuple and json property accessor doesn't run when it should. Leaving it as is, but we should clean up `ChannelBindingServiceConfiguration` before GA.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/277d896e1d1d54f4f739956764ff10efe7cb4fc2

PR #479, opened 2016-04-08
Make AggregateApplication package local

### LINKED ISSUE #474: Create AggregateApplication via AggregateApplicationBuilder only
Currently, we provide a couple of ways to create an aggregate application

1) Using AggregateApplicationBuilder
2) Directly calling AggregateApplication.run()

The second approach seems to have limitations in terms of customising the apps. Also, having two different approaches to create aggregate application can likely diverge. It would be nice if we can have only the builder approach for creating aggregate application.

mbogoevici:
Fixes #474

- Now the only API for aggregating is AggregateApplicationBuilder
- add namespace tests


ilayaperumalg:
LGTM, merging.

PR #484, opened 2016-04-15
BinderAwareChannelResolver to check for ExtendedProperties

### LINKED ISSUE #483: Use BinderAwareChannelResolver with Kafka Binder encountered a ClassCastException
I'am using spring cloud stream with kafka.  

I wan't to use `BinderAwareChannelResolver ` to create message channel in code.  

here is my code:
```Java
public class EventActivator {

    @Autowired
    private BinderAwareChannelResolver binderAwareChannelResolver;

    public boolean sendMessage(String message, String destination) {

        MessageChannel messageChannel = binderAwareChannelResolver.resolveDestination(destination);
        //...
    }

}

```

my configration:
```Java
    @Bean
    public BinderAwareChannelResolver binderAwareChannelResolver(BinderFactory<MessageChannel> binderFactory,
                                                                 ChannelBindingServiceProperties channelBindingServiceProperties,
                                                                 DynamicDestinationsBindable dynamicDestinationsBindable,
                                                                 BindableChannelFactory bindableChannelFactory) {

        return new BinderAwareChannelResolver(binderFactory, channelBindingServiceProperties,
                dynamicDestinationsBindable, bindableChannelFactory);
    }

```

If I call `binderAwareChannelResolver.resolveDestination(destination);`, a ClassCastException occurred.  Here is the stack trace:

```
java.lang.ClassCastException: org.springframework.cloud.stream.binder.ProducerProperties cannot be cast to org.springframework.cloud.stream.binder.ExtendedProducerProperties
	at org.springframework.cloud.stream.binder.kafka.KafkaMessageChannelBinder.doBindProducer(KafkaMessageChannelBinder.java:104) ~[spring-cloud-stream-binder-kafka-1.0.0.RC1.jar:1.0.0.RC1]
	at org.springframework.cloud.stream.binder.AbstractBinder.bindProducer(AbstractBinder.java:163) ~[spring-cloud-stream-1.0.0.RC1.jar:1.0.0.RC1]
	at org.springframework.cloud.stream.binding.BinderAwareChannelResolver.resolveDestination(BinderAwareChannelResolver.java:115) ~[spring-cloud-stream-1.0.0.RC1.jar:1.0.0.RC1]
	at com.akkafun.common.event.service.EventActivator.sendMessage(EventActivator.java:49) ~[classes/:na]
```

I look into the `BinderAwareChannelResolver` source code,  found the exception was raise by this code:
```Java
					this.dynamicDestinationsBindable.addOutputBinding(beanName,
							binder.bindProducer(destinationName, channel, producerProperties));				
```
The `binder` is `KafkaMessageChannelBinder`, `binder.bindProducer` method need a `ExtendedProducerProperties`, but the type of `producerProperties` is `ProducerProperties`.
So a `ClassCastException` was thrown.

To solve the problem, I have to use a custom `BinderAwareChannelResolver `. I find the code in `ChannelBindingService` to obtain `ExtendedPropertiesBinder`:
```Java
		if (binder instanceof ExtendedPropertiesBinder) {
			ExtendedPropertiesBinder extendedPropertiesBinder = (ExtendedPropertiesBinder) binder;
			Object extension = extendedPropertiesBinder.getExtendedProducerProperties(outputChannelName);
			ExtendedProducerProperties extendedProducerProperties = new ExtendedProducerProperties<>(extension);
			BeanUtils.copyProperties(producerProperties, extendedProducerProperties);
			producerProperties = extendedProducerProperties;
		}

```

I think it is a general purpose.  

Are there another more elegant way to solve the problem?

ilayaperumalg:
- When resolving the destination names via BinderAwareChannelResolver, the underlying binder needs to be checked if it is of type `ExtendedPropertiesBinder` so that any ExtendedProducerProperties will be considered based on the binder
 - Also move the validation and retrieval of producer/consumer properties to a util
 - Update BinderAwareChannelResolverTests to use ExtendedPropertiesBinder

This resolves #483


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/BinderAwareChannelResolverTests.java
```diff
@@ -74,7 +74,7 @@
 
 	private volatile BinderAwareChannelResolver resolver;
 
-	private volatile Binder<MessageChannel, ConsumerProperties, ProducerProperties> binder;
+	private volatile ExtendedPropertiesBinder<MessageChannel, ExtendedConsumerProperties, ExtendedProducerProperties> binder;
```
mbogoevici: While we need to test the new change, we should create a separate test for it - now we don't have a test for a normal (non-extended) binder.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -78,13 +73,10 @@ public ChannelBindingService(ChannelBindingServiceProperties channelBindingServi
 		ConsumerProperties consumerProperties =
 				this.channelBindingServiceProperties.getConsumerProperties(inputChannelName);
 		if (binder instanceof ExtendedPropertiesBinder) {
-			ExtendedPropertiesBinder extendedPropertiesBinder = (ExtendedPropertiesBinder) binder;
-			Object extension = extendedPropertiesBinder.getExtendedConsumerProperties(inputChannelName);
-			ExtendedConsumerProperties extendedConsumerProperties = new ExtendedConsumerProperties(extension);
-			BeanUtils.copyProperties(consumerProperties, extendedConsumerProperties);
-			consumerProperties = extendedConsumerProperties;
+			consumerProperties = ProducerConsumerPropertiesUtil.getExtendedConsumerProperties((ExtendedPropertiesBinder) binder,
```
mbogoevici: Instead of making this a static method on an external class, can we just encapsulate this as part of `ChannelBindingService` ?

mbogoevici: And delegate the entire binding process to it ?

ilayaperumalg: The intention to move the producer/consumer properties extraction to a external class is to make a way for the `BinderAwareChannelResolver` to invoke them explicitly. Since both the `binder` isn't obtained differently in `BinderAwareChannelResolver` and `ChannelBindingService`'s bindProducer/Consumer, encapsulating inside the `ChannelBindingService` may not help BinderAwareChannelResolver. What do you think?

ilayaperumalg:
Updated the PR to have separate binder aware channel resolver test class that uses `ExtendedPropertiesBinder`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: In the end, in both cases we are binding a channel based on its name, to a given destination. So at a first glance, I think that a call to `channelBindingService.bindProducer(channel, channelName);` should be enough, instead of copying in the code and calling the utility method.

ilayaperumalg: I thought the same when I first started with this. But, the way binder is extracted is different in both the cases. Either we need to refactor that part or we can copy the same code twice (that extended properties extraction part) or have the code in common place like this PR has. I am ok with any of these options.

ilayaperumalg:
Updated the PR to remove support of `binderType` from BinderAwareChannelResolver so that we can refactor it to use `ChannelBindingService` to bind producer.

May be it is worth to rename the class `BinderAwareChannelResolver` to `DynamicDestinationResolver` or something. Given the apps would directly use the class `BinderAwareChannelResolver` (for instance the router app), we need to make it clear we update it there and a documentation for this as well.

mbogoevici:
I'm not very keen on renaming it. It's pretty late for that, and the relationship with the binder is clearer expressed in the current naming. @garyrussell - wdyt?

garyrussell:
I agree - best to leave the name I think. An additional complication is 'destination' has two meanings in the context of channel resolvers.

Pre SI 4.0, we had `ChannelResolver`, when the common code was factored out into Spring Messaging it was renamed and a generic parameter added `DestinationResolver<MessageChannel>`.

So, in this case, we have the spring-messaging destination and the SCSt destination so the rename could be confusing; plus, from the perspective of the user (the router sink in particular) it really is resolving the channel to which the message is to be routed - the fact that that channel is then bound to a destination is not relevant.

mbogoevici:
I don't get the removal of the `DynamicDestinationBindable` - we used it to keep track of the dynamically bound channels so that we can unbind them on context `stop()` - was there any motivation for that?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: We should get rid of the utility class now that it's used only in one place.

mbogoevici:
Added some additional changes to the PR here: https://github.com/mbogoevici/spring-cloud-stream/commit/f1645528ec9f70406559b904a1c3588753e0d509 

@ilayaperumalg let me know what you think

ilayaperumalg:
@mbogoevici your changes look good to me. thank you!

mbogoevici:
Squashed, rebased, merged as https://github.com/spring-cloud/spring-cloud-stream/commit/f029249e1ba0502366b146b1ee44bc588e3a2b4d with additional changes https://github.com/spring-cloud/spring-cloud-stream/commit/b4dc6bbd9596612d25b1545e037e941873fc9955

PR #487, opened 2016-04-15
Reinstate checkstyle running by default

mbogoevici:
- Rely on `checkstyle.skip` instead of a custom property
- Remove configuration from plugin definition, rely on execution only to set project-specific dependencies. This prevents inheritors from outside the project to fail on not finding the configuration. Note: on the long run, we will want to create a `build-tools` submodule that hold the configurations.

Note: when merging this, https://github.com/spring-cloud/spring-cloud-stream-modules/pull/253 must be merged as well (ideally beforehand) to prevent Spring Cloud Stream Modules to fail on the checkstyle plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,6 @@
 	</scm>
 	<properties>
 		<java.version>1.7</java.version>
-		<disable.checks>true</disable.checks>
 	</properties>
 	<dependencyManagement>
 		<dependencies>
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
With the changes, `mvn checkstyle:checkstyle` runs fine. But `mvn checkstyle:check` doesn't seem to use the explicit checkstyle config and that results in build failure.

mbogoevici:
That's okay for a first pass, I believe. This is not so much about running checkstyle independently, as it is about making it part of the build process (i.e. executing on `compile`). Spring Boot for instance does the same thing. We will be able to fix this once we create a separate artifact for storing the rules, but not keen on doing that before GA.

ilayaperumalg:
@mbogoevici I agree. will merge this soon.

ilayaperumalg:
LGTM, merged as bfff729

PR #501, opened 2016-04-26
Use a fixed-size daemon thread pool for the Kafka Binder

### LINKED ISSUE #500: Applications using the Kafka binder do not shut down on context close()
When the context is closed (e.g. via actuator endpoint), Reactor keeps running on the thread pool of `QueueingMesageListenerInvoker`, preventing the application to shutdown. This is related to https://jira.spring.io/browse/INTEXT-214.

mbogoevici:
Fix #500

- Initialize a fixed-size ExecutorService of the size of the concurrency setting of the binder
- Clean up the ExecutorService on unbind
- Remove unused `spyOn` support

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -27,6 +27,9 @@ import java.util.LinkedList;
 import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import kafka.admin.AdminUtils;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -460,9 +465,28 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
 		final FixedSubscriberChannel bridge = new FixedSubscriberChannel(rh);
 		bridge.setBeanName("bridge." + name);
 
-		final KafkaMessageListenerContainer messageListenerContainer =
-				createMessageListenerContainer(properties, group, null, listenedPartitions, referencePoint);
+		Assert.isTrue(StringUtils.hasText(null) ^ !CollectionUtils.isEmpty(listenedPartitions),
```
markfisher: is this `StringUtils.hasText(null)` intentional?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -460,9 +465,28 @@ public Object doWithRetry(RetryContext context) throws RuntimeException {
...
+				"Exactly one of topic or a list of listened partitions must be provided");
+		final KafkaMessageListenerContainer messageListenerContainer = new KafkaMessageListenerContainer(connectionFactory,
+				listenedPartitions.toArray(new Partition[listenedPartitions.size()]));
+
+		if (logger.isDebugEnabled()) {
+			logger.debug("Listening to topic " + null);
```
markfisher: is `null` intentional here also?

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
mbogoevici: Good point. This is an artifact of inlining the old code. Initially we had support for passing both a topic or a list of partitions (for the never used `spyOn`),  so now it's always null - will fix.

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/0bd7899a77977cad29fbdd67e1a908b552c77e32

PR #502, opened 2016-04-27
Use the binder `minPartitionCount` property for consumers

### LINKED ISSUE #495: Kafka consumers should use `minPartitionCount` configured in the binder

mbogoevici:
Fixes #495

Currently, `minPartitionCount` is available both as a binder default and as a consumer property.
It would be simpler if it only was a binder setting (seeing as it has effect only as a default).


ilayaperumalg:
LGTM, rebased and merged as 1d6ab21067b0dca9d8ac8be29ef0f63bf398b735

PR #507, opened 2016-05-02
[Proposal - review only] Support registering `MessageConverter` beans directly

### LINKED ISSUE #505: Allow the registration of regular `MessageConverters` as beans
Currently, all custom message converters must extend `AbstractFromMessageConverter` and must use the specific class of the bean or at least `AbstractFromMessageConverter` for being added to the configuration. This is inconvenient as it exposes too much implementation detail to the end user.

Proposed solution:

- allow the registration of ordinary `MessageConverter` beans and allow them to participate in the `StreamListener` argument conversion as well as bound message channel conversion (see below);
- introduce an extended interface e.g. `TargetTypeMessageConverter` to capture the additional contract of converters that can be added on the input and output channels, i.e. that they advertise specific target content-types and target classes - so only `MessageConverters` that are also `TargetTypeMessageConverters` can be used for setting up conversion on input and output channels;
- make `AbstractFromMessageConverter` a utility base implementation of `TargetTypeMessageConverter`;

mbogoevici:
Fixes #505

- introduces the `TargetTypeMessageConverter` extension to `MessageConverter`
   for converters that advertise target MimeTypes and/or target classes
- beans of the type `TargetTypeMessageConverter` are used to register bound channel converters

Note: if accepted, will add documentation

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
@@ -19,13 +19,17 @@ package org.springframework.cloud.stream.converter;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.springframework.messaging.converter.CompositeMessageConverter;
 import org.springframework.messaging.converter.MessageConverter;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.MimeType;
+import org.springframework.util.ObjectUtils;
 
 
 /**
@@ -80,9 +84,9 @@ public class CompositeMessageConverterFactory {
 	public CompositeMessageConverter getMessageConverterForTargetType(MimeType targetMimeType) {
 		List<MessageConverter> targetMimeTypeConverters = new ArrayList<MessageConverter>();
 		for (MessageConverter converter : converters) {
-			if (converter instanceof TargetTypeMessageConverter) {
-				TargetTypeMessageConverter targetTypeMessageConverter = (TargetTypeMessageConverter) converter;
-				if (targetTypeMessageConverter.supportsTargetMimeType(targetMimeType)) {
+			if (converter instanceof TargetMimeTypeMessageConverter) {
+				TargetMimeTypeMessageConverter targetTypeMessageConverter = (TargetMimeTypeMessageConverter) converter;
+				if (!ObjectUtils.isEmpty(targetTypeMessageConverter.getSupportedJavaTypes(targetMimeType))) {
 					targetMimeTypeConverters.add(converter);
 				}
 			}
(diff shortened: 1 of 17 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TargetTypeMessageConverter.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.converter;
+
+import org.springframework.util.MimeType;
+
+/**
+ * A {@link org.springframework.messaging.converter.MessageConverter} that supports a
+ * target {@link MimeType}/{@link Class} combination.
```
markfisher: this javadoc makes it sound as though there is a single mime-type/class combination that is supported, but according to the interface methods, the contract is such that an arbitrary number of mime-types could be supported, as well as an arbitrary number of target *classes* without any relationship between them (no "combination"... i.e. it doesn't appear to be a 1:1 mapping)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
```diff
@@ -81,11 +81,14 @@ private void initDefaultConverters() {
 	 * @param targetMimeType the target MIME type
 	 * @return a converter for the target MIME type
 	 */
-	public CompositeMessageConverter getMessageConverterForType(MimeType targetMimeType) {
+	public CompositeMessageConverter getMessageConverterForTargetType(MimeType targetMimeType) {
```
markfisher: should rename: getMessageConverterForTarget*Mime*Type

mbogoevici:
Pushed a few more changes, with interface modifications.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
```diff
@@ -51,7 +51,7 @@ public CompositeMessageConverterFactory() {
 	/**
 	 * @param customConverters a list of {@link AbstractFromMessageConverter}
 	 */
-	public CompositeMessageConverterFactory(List<? extends AbstractFromMessageConverter> customConverters,
+	public CompositeMessageConverterFactory(List<? extends MessageConverter> customConverters,
```
ilayaperumalg: With the introduction of `TargetMimeTypeMessageConverter`, are we enforcing any custom message converters to implement this interface instead of extending `AbstractFromMessageConverter`? Also, I think the `ChannelBindingServiceConfiguration` requires this change to have 
```
@Autowired(required = false)
	private List<TargetMimeTypeMessageConverter> customMessageConverters;

```
If I understand it right, we aren't injecting any `MessageConverter` beans in the context instead only the `TargetMimeTypeMessageConverter` beans?

mbogoevici: Generic converters (i.e. `MessageConverter` implementations) could be used by the StreamListener annotation if they are registered by the user.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TargetMimeTypeMessageConverter.java
```diff
@@ -0,0 +1,36 @@
...
+package org.springframework.cloud.stream.converter;
+
+import org.springframework.util.MimeType;
+
+/**
+ * A {@link org.springframework.messaging.converter.MessageConverter} that support
+ *
+ * @author Marius Bogoevici
+ */
+public interface TargetMimeTypeMessageConverter {
```
mbogoevici: Should extend `MessageConverter`.

mbogoevici:
Deferring this to post-1.0 (1.1.0.M1) as the overall set of resulting changes is too big for a RC->GA transition.

PR #510, opened 2016-05-03
Rework topic autoconfiguration

### LINKED ISSUE #509: Rework topic creation on the Kafka binder so it avoids reducing the partition count or the replication factor
The logic that manages the topic configuration relies on `AdminUtils.createOrUpdateTopicPartitionAssignmentPathInZK`, which will force the reduction of partition count as well as the replication factor of the topic on the cluster, even that is not normally supported by Kafka.  (http://kafka.apache.org/082/documentation.html#basic_ops_modify_topic).

mbogoevici:
Fixes #509

- prevent autoconfiguration from reducing the existing partition count and/or replication factor

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -156,7 +156,7 @@ public class KafkaMessageChannelBinder
 
 	private int fetchSize = 1024 * 1024;
 
-	private int defaultMinPartitionCount = 1;
+	private int minPartitionCount = 1;
 
 	private ConnectionFactory connectionFactory;
 
@@ -302,8 +302,8 @@ public class KafkaMessageChannelBinder
 		this.fetchSize = fetchSize;
 	}
 
-	public void setDefaultMinPartitionCount(int defaultMinPartitionCount) {
-		this.defaultMinPartitionCount = defaultMinPartitionCount;
+	public void setMinPartitionCount(int minPartitionCount) {
+		this.minPartitionCount = minPartitionCount;
 	}
 
 	public void setMaxWait(int maxWait) {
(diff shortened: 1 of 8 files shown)


mbogoevici:
On a second thought, I am strongly considering to add a separate configuration for adding new partitions (off by default) and rename `autoConfigure` to `autoCreate`. SCDF applications may turn repartitioning on. So the default behaviour becomes:

- if the topic does not exist - autoCreateTopic with the number of partitions required by the application;
- if the topic does exist - use it as is and check if the partition count is the one required by the application (e.g. an application with an `instanceCount` of 10 will expect 10 partitions at a minimum) - fail if not;

Then both behaviours can be turned on and off as needed. I think this is a decent compromise as it mimics what we do with Rabbit: auto-create the destination but not reconfigure it and it allows an option to reconfigure it later.

garyrussell:
Yes, makes sense to me.

mbogoevici:
Pushed

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -429,28 +448,53 @@ protected void doStop() {
...
-					@Override
-					public Object doWithRetry(RetryContext context) throws RuntimeException {
-						AdminUtils.createOrUpdateTopicPartitionAssignmentPathInZK(zkClient, topicName, replicaAssignment,
-								topicConfig, true);
-						return null;
+			TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topicName, zkClient);
+			if (topicMetadata.errorCode() == ErrorMapping.NoError()) {
+				// only consider minPartitionCount for resizing if autoAddPartitions is true
+				int effectivePartitionCount = isAutoAddPartitions() ? Math.max(defaultMinPartitionCount, partitionCount)
+						: partitionCount;
```
garyrussell: It's not clear why `defaultMinPartitionCount` wins if it's larger than the requested partition count, but I see from the doc changes that this was your intention.

That said, the code in the configuration...
    		kafkaMessageChannelBinder.setDefaultMinPartitionCount(kafkaBinderConfigurationProperties.getMinPartitionCount());

... and the docs talking about `minPartitionCount`, tell me that this property should really be called `minPartitionCount` if it truly is an absolute minimum rather than a default minimum.

mbogoevici: Will rename. The reason why it ends up winning is that its role is to override the inferred partition count (which in most cases would be 1 or the instance county) and allow overpartitioning. I am not fundamentally opposed to the notion of letting `partitionCount` take precedence - but I think it is a bad idea to combine both in the same application (as I tried to explained in the documentation).

garyrussell: I am fine with the logic and explanation in the docs, just that the `default` here is confusing.

mbogoevici:
Pushed

garyrussell:
Merged as 02ee8d4d88fc3bebfcb6a013ec72b16cbf5513ca

PR #513, opened 2016-05-06
Fixing partition section where 'producer' key was missing from example'

viniciusccarvalho:
Docs is missing the 'producer' key. It was a honest mistake to copy and paste on a sample and not get the partitioning configured for the producer.


ilayaperumalg:
@viniciusccarvalho thank you! LGTM, merged.

PR #524, opened 2016-05-10
Upgrade RxJava to 1.1.5

### LINKED ISSUE #522: Update to RxJava 1.1.5
There are some big bug fixes in RxJava that we rely on in spring-cloud-netflix. Stream being tied to a different version makes dependency management difficult.

ilayaperumalg:
This resolves #522

### DIFF
--- a/spring-cloud-stream-dependencies/pom.xml
+++ b/spring-cloud-stream-dependencies/pom.xml
@@ -13,7 +13,7 @@
 	<name>spring-cloud-stream-dependencies</name>
 	<description>Spring Cloud Stream Dependencies</description>
 	<properties>
-		<rxjava.version>1.0.14</rxjava.version>
+		<rxjava.version>1.1.5</rxjava.version>
 		<spring.tuple.version>1.0.0.RELEASE</spring.tuple.version>
 		<spring.integration.tuple.version>1.0.0.RELEASE</spring.integration.tuple.version>
 	</properties>


ilayaperumalg:
This will also use `1.1.5` rxjava in `spring-integration-kafka` as well.

artembilan:
@ilayaperumalg 

> This will also use 1.1.5 rxjava in spring-integration-kafka as well.

What do you mean? Spring Integration Kafka has never used RxJava. Having the `Flux` and `Mono` foundation in the latest Reactor I even don't think that we will never take a look into RxJava 1.x at all there.

ilayaperumalg:
@artembilan RxJava is being used by the `WindowingOffsetManager` of the Kafka binder. I thought this offset manager is in `spring-integration-kafka` and now realized it is part of kafka binder code only.
So, RxJava `1.1.5` is used by the Kafka binder.

ilayaperumalg:
> Having the Flux and Mono foundation in the latest Reactor I even don't think that we will never take a look into RxJava 1.x at all there.

Yeah, I agree. But I think this fix is needed only for the concern raised by @dsyer in #522

artembilan:
OK. Got it @ilayaperumalg .
Thank you for explanation!

Anyway it isn't clear why you still use RxJava there since we already have `WindowingOffsetManager` based on Reactor 2.0 in SIK since its 1.3: https://github.com/spring-projects/spring-integration-kafka/blob/1.3.x/src/main/java/org/springframework/integration/kafka/listener/WindowingOffsetManager.java

artembilan:
OK-OK. I understand that upgrade concern here and suppose my comments are a bit out of the purpose of the PR. Just wanted to place all the dots under "i" do not confuse other peoples like I was seeing such a comment in my mail box.

Thank you for the attention!

mbogoevici:
@artembilan Your concern is well-founded, but unfortunately there's been no release of Spring Integration Kafka 1.3.1 and I'm a bit wary of making additional changes at this point. That being said, we will remove the need for this relatively soon as we start upgrading the binder to 0.9 and also begin the overhaul of the binder system to support reactive binders.

artembilan:
Doh...
Thank you, @mbogoevici , for pointing that out. I really missed that we didn't release that Reactor support in the SIK :smile: 
Although there are a couple issues fixed that I think we can release `1.3.1` together with the `2.0.0`.
But that is a difference story from this one of course.
Never mind.

Issue #525, opened 2016-05-10
Do not start the context automatically


mbogoevici:
In most cases it is a good thing to start the other beans as part of the binding process, but in some cases there are undesirable side effects.

See: https://github.com/spring-projects/spring-integration-java-dsl/issues/80#issuecomment-218245617

mbogoevici:
After a discussion with the Boot team, we will want to rely on Spring Boot/Spring Core controlling that behaviour. Keeping this open as once this feature is available in Boot/Core, we will need to update our approach by removing the autostart code.

mbogoevici:
Boot issue in discussion is: https://github.com/spring-projects/spring-boot/issues/6021

mbogoevici:
Based on the outcome of the discussion in the Spring Boot project, we will not want to do this at all. 

If necessary, end users can start the context themselves. The current behaviour should be changed in v1.1 (which breaks backwards compatibility to some extent and we will need to be properly documented).

PR #527, opened 2016-05-11
Make Starting context autostart optional

ilayaperumalg:
- Disable autostarting the main application context if the property `spring.cloud.stream.autoStart` is set to false
 - By default the main application context is autoStarted

This addresses #525

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
@@ -19,7 +19,6 @@ package org.springframework.cloud.stream.binding;
 import java.util.Map;
 
 import org.springframework.beans.BeansException;
-import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
@@ -46,12 +45,10 @@ public class InputBindingLifecycle implements SmartLifecycle, ApplicationContext
 	@Override
 	public void start() {
 		if (!running) {
-			ChannelBindingServiceProperties channelBindingServiceProperties = null;
 			// retrieve the ChannelBindingService lazily, avoiding early initialization
 			try {
 				ChannelBindingService channelBindingService = this.applicationContext
 						.getBean(ChannelBindingService.class);
-				channelBindingServiceProperties = channelBindingService.getChannelBindingServiceProperties();
 				Map<String, Bindable> bindables = this.applicationContext.getBeansOfType(Bindable.class);
 				for (Bindable bindable : bindables.values()) {
 					bindable.bindInputs(channelBindingService);
@@ -61,9 +58,7 @@ public class InputBindingLifecycle implements SmartLifecycle, ApplicationContext
 				throw new IllegalStateException(
 						"Cannot perform binding, no proper implementation found", e);
 			}
-			if (channelBindingServiceProperties != null && channelBindingServiceProperties.isAutoStartup()) {
-				this.running = true;
-			}
+			this.running = true;
 		}
 	}
 
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.stream.binding;
 import java.util.Map;
 
 import org.springframework.beans.BeansException;
+import org.springframework.cloud.stream.config.ChannelBindingServiceProperties;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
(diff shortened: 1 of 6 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
```diff
@@ -45,12 +46,13 @@ public void setApplicationContext(ApplicationContext applicationContext)
...
 			// retrieve the ChannelBindingService lazily, avoiding early initialization
 			try {
 				ChannelBindingService channelBindingService = this.applicationContext
 						.getBean(ChannelBindingService.class);
-				Map<String, Bindable> bindables = this.applicationContext
-						.getBeansOfType(Bindable.class);
+				channelBindingServiceProperties = channelBindingService.getChannelBindingServiceProperties();
+				Map<String, Bindable> bindables = this.applicationContext.getBeansOfType(Bindable.class);
 				for (Bindable bindable : bindables.values()) {
 					bindable.bindInputs(channelBindingService);
```
ilayaperumalg: TODO: Controlling binding as part of `autoStartup`. 
This needs to be done somehow by calling explicitly to iterate over `Bindables` and perform binding.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
```diff
@@ -59,7 +61,9 @@ public void start() {
 				throw new IllegalStateException(
 						"Cannot perform binding, no proper implementation found", e);
 			}
-			this.running = true;
+			if (channelBindingServiceProperties != null && channelBindingServiceProperties.isAutoStartup()) {
```
mbogoevici: we shouldn't make the setting of the flag conditional on `autostartup`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/OutputBindingLifecycle.java
```diff
@@ -61,8 +62,10 @@ public void start() {
 				throw new IllegalStateException(
 						"Cannot perform binding, no proper implementation found", e);
 			}
-			this.running = true;
-			this.applicationContext.start();
+			if (channelBindingServiceProperties != null && channelBindingServiceProperties.isAutoStartup()) {
```
mbogoevici: Same as https://github.com/spring-cloud/spring-cloud-stream/pull/527/files#r62900470.

Only the context autostart may be optional.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/OutputBindingLifecycle.java
```diff
@@ -46,13 +47,13 @@ public void setApplicationContext(ApplicationContext applicationContext)
...
 		if (!running) {
-
+			ChannelBindingServiceProperties channelBindingServiceProperties = null;
 			// retrieve the ChannelBindingService lazily, avoiding early initialization
 			try {
 				ChannelBindingService channelBindingService = this.applicationContext
 						.getBean(ChannelBindingService.class);
-				Map<String, Bindable> bindables = this.applicationContext
-						.getBeansOfType(Bindable.class);
+				channelBindingServiceProperties = channelBindingService.getChannelBindingServiceProperties();
```
mbogoevici: We should inject the `ChannelBindingServiceProperties` bean directly, not retrieve it from the service.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
mbogoevici: @ilayaperumalg - can you clarify ? I believe this is what we intend to do (and currently do): perform the binding/unbinding when the `*BindingLifecycle` beans are started/stopped.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -69,6 +69,8 @@
 
 	private boolean ignoreUnknownProperties = true;
 
+	private boolean autoStartup = true;
```
mbogoevici: maybe the property should be called `autoStartContext` ?

ilayaperumalg:
Updated the PR with test and doc update. Also, addressed the review comments.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
ilayaperumalg: My concern was with performing the binding (starting the producer/consumer adapter contexts) even before the main application context is started. I guess that is fine since it is just the binding that is established.

mbogoevici: That is the expected behaviour for auto-started `SmartLifecycles` so we should be fine.

mbogoevici:
LGTM, just checking if this makes more sense as a general Boot feature.

mbogoevici:
Closing this, as we intend to resolve it at a higher level via https://github.com/spring-projects/spring-boot/issues/6021

Once that is available, we will follow up with another PR removing the autostart code.

PR #549, opened 2016-05-25
Use `KafkaBinderConfigurationProperties` directly

mbogoevici:
Fixes #548

- Remove redundant property copying
- Make `socketBufferSize` configurable (it wasn't provided as a configuration option)

PR #550, opened 2016-05-25
GH-542 Add setting for `autoCommitOnError`

### LINKED ISSUE #542: kafka binder should set autoCommitOnError to true on kafka listener container
When a kafka consumer throws an error, s-c-s retries 3 times then either discards the message (enableDlq = false) or publishes the message to error.\<destination\>.\<group\> (enableDlq = true), but because the message listener container used has the default autoCommitOnError value of false, the consumer offset does not get updated, which causes the consumer to reprocess the message on restart, until another message is processed on the same topic successfully and the offset is updated past the failed message. 

Given the consumer processing expectations here, the listener container autoCommitOnError should be set to true so that after retries, the offset will be updated.

mbogoevici:
(Only merge after #549 as it relies on the cleanup done there, or merge only this PR with both changes)

Add setting for `autoCommitOnError`
Fixes #542

- add option for `autoCommitOnError`
- make `autoCommitOnError` follow the state of `enableDlq` - no need to suppress commits if messages will be sent elsewhere

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaConsumerProperties.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaConsumerProperties.java
@@ -23,6 +23,8 @@ public class KafkaConsumerProperties {
 
 	private boolean autoCommitOffset = true;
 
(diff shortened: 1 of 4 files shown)


mbogoevici:
Rebased.

garyrussell:
Merged as e2f91417c368c9973b48727c80516f0ca775e5e5 with minor doc tweak:

    s/it effectively has the same value `enableDlq`/it effectively has the same value as `enableDlq`/

cherry-picked to 1.0.x

PR #552, opened 2016-05-26
Clarify channel names in the doc

### LINKED ISSUE #551: Suggested change in documentation for 2.4 Consumer Groups
This is a suggestion based on my own awkward experience, my bad obviously:

I have started using Spring Cloud Stream recently and struggled to configure a consumer group following the (first chapters of the) documentation.

The point is that instead of using the standard Source and Input interfaces I defined my own:

	public interface LogSource{
		String OUTPUT = "logMessages";
		@Output(LogSource.OUTPUT)
		MessageChannel output();
	}

	public interface LogSink {
		String INPUT = "logMessages";
		@Input(LogSink.INPUT)
		SubscribableChannel input();
	}

The documentation states:

> Each consumer binding can use the `spring.cloud.stream.bindings.input.group` property to specify a group name.

I wanted to create a consumer group called `logProcessor`, so I set up:
```
spring:
  cloud:
    stream:
      bindings:
        input:
          group: logProcessor
```
The problem is that this way both instances of the consumer got the messages, that is to say, the group wasn't working.

After the initial "what the...?" stage I came up with the idea of replacing `input` with logMessages in my config and it solved it:

`spring.cloud.stream.bindings.logMessages.group=logProcessor`
instead of 
`spring.cloud.stream.bindings.input.group=logProcessor`

Later I found that in `5.2 Binding Properties` it is correctly defined how to use those properties. Maybe eager people like me wouldn't have a headache if 2.4 were like this: 

> Each consumer binding can use the `spring.cloud.stream.bindings.<channelName>.group` property to specify a group name.

Just a suggestion ;-)

ilayaperumalg:
Fixes #551

### DIFF
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -143,8 +143,8 @@ When doing this, different instances of an application are placed in a competing
 
 Spring Cloud Stream models this behavior through the concept of a _consumer group_.
 (Spring Cloud Stream consumer groups are similar to and inspired by Kafka consumer groups.)
-Each consumer binding can use the `spring.cloud.stream.bindings.input.group` property to specify a group name.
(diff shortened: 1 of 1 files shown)


sobychacko:
merged.

PR #556, opened 2016-05-31
GH-541 Make reconnect time for Kafka/Rabbit configurable

### LINKED ISSUE #541: Configuration options for Kafka/RabbitMQ reconnect
The reconnect to Kafka works fine, but it is attempted with 'full speed' (about 10 per second on my machine). This behaviour should be configurable (for all binders). At least there should be a parameter for the intervall between two attempts and maybe some sort of backoff makes sense, too.

Furthermore, in the case of connection loss, an exception is logged. This should be replaced by dedicated logging.

mbogoevici:
Fixes #541


garyrussell:
Rebased, merged as 99db7dff9b01cc7533c36b51a009f49bb3a1d61b

PR #559, opened 2016-06-01
Simplify marshalling of JSON messages

### LINKED ISSUE #558: Simplify encoding of JSON messages
For JSON messages (i.e. having a String payload and a `contentType` header set to `application/json`) , Spring Cloud Stream can simply set the `contentType` header to `application/json` instead of the default behaviour of setting `contentType` to `text/plain` and `originalContentType` to `application/json`, as it handles `content-type=application/json` messages automatically already.

mbogoevici:
Fixes #558

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -46,6 +46,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.MimeType;
 import org.springframework.util.MimeTypeUtils;
+import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -339,13 +341,14 @@ public void doManualAck(LinkedList<MessageHeaders> messageHeaders) {
...
-		static MimeType mimeTypeFromObject(Object payload) {
+		static MimeType mimeTypeFromObject(Object payload, String originalContentType) {
 			Assert.notNull(payload, "payload object cannot be null.");
 			if (payload instanceof byte[]) {
 				return MimeTypeUtils.APPLICATION_OCTET_STREAM;
 			}
 			if (payload instanceof String) {
-				return MimeTypeUtils.TEXT_PLAIN;
+				return MimeTypeUtils.APPLICATION_JSON_VALUE.equals(originalContentType) ? MimeTypeUtils.APPLICATION_JSON
+						: MimeTypeUtils.TEXT_PLAIN;
```
artembilan: I think I'm fine with the solution, but I wonder if we should do the same for some other general String-based content types, e.g. `application/xml`.
Feel free to merge, if that.
That was just an info to share and think about.

mbogoevici: Yeah, I agree. We don't have an OOTB converter for XML though, and JSON is fairly popular . One of the changes intended for 1.1 would be https://github.com/spring-cloud/spring-cloud-stream/issues/505 and we can explore adressing this in a more systematic fashion.

ilayaperumalg:
LGTM, rebased and merged as 53d86f55236e7cb25138300c34d332bf853296c7

ilayaperumalg:
Cherry-picked and pushed to 1.0.x 78487442a8161490f9e756879742602eb57be773

PR #560, opened 2016-06-01
Allow for binder configurations that do not affect the defaulting process

### LINKED ISSUE #557: Allow `named` binders and binders using the default Boot settings to coexist
Currently, we are disabling the default binder configuration if named binders are present in the application. The intent is to avoid ambiguities inherent in a mix of implicit and explicit configuration and force the user to acknowledge that.

There are cases, however, when binder configurations are created transparently (e.g. Hystrix) and they may inadvertently turn off the expected default binder configuration for users. We need to be able to add configurations without disrupting the normal user experience. One possibility is to relax the restrictions around default binder configurations. The other option is to mark the framework-provided configurations so that they remain transparent to the application - and they aren't counted towards the 'user-defined' configurations.

mbogoevici:
Fixes #557

Allow for a special `applicationProvided` flag on a binder configuration that
makes it transparent to the default configuration process.

### DIFF
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -729,6 +729,9 @@ If your application should connect to more than one broker of the same type, you
 ====
 Turning on explicit binder configuration will disable the default binder configuration process altogether.
 If you do this, all binders in use must be included in the configuration.
+Frameworks that intend to use Spring Cloud Stream transparently may create binder configurations that will be referenceable by name, but will not affect the default binder configuration.
+In order to do so, a binder configuration may have its the `applicationProvided` flag set to false, e.g. `spring.cloud.stream.binders.<configurationName>.applicationProvided=false`.
+This denotes a configuration that will exist independently of the default binder configuration process.
 ====
 
 For example, this is the typical configuration for a processor application which connects to two RabbitMQ broker instances:
@@ -759,6 +762,32 @@ spring:
               rabbitmq:
                 host: <host2>
 ----
+
+=== Binder configuration properties
+
+The following properties are available when creating custom binder configurations.
+They must be prefixed with `spring.cloud.stream.binder.<configurationName>`.
+
+type::
+  The binder type.
+It typically references one of the binders found on the classpath, in particular a key in a `META-INF/spring.binders` file.
++
+By default, it has the same value as the configuration name.
+inheritEnvironment::
+  Whether the configuration will inherit the environment of the application itself.
++
+Default `true`.
+environment::
+  Root for a set of properties that can be used to customize the environment of the binder.
+When this is configured, the context in which the binder is being created is not a child of the application context.
+This allows for complete separation between the binder components and the application components.
++
+Default `empty`.
+applicationProvided::
+  Whether the binder configuration is provided by the application (the default case) or a framework included by an application.
+This allows framework authors to depend on Spring Cloud Stream and provide their own infrastructure without interfering with the application.
++
+Default `true`.
 
 
 === Implementation strategies
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -729,8 +729,8 @@ If your application should connect to more than one broker of the same type, you
 ====
 Turning on explicit binder configuration will disable the default binder configuration process altogether.
 If you do this, all binders in use must be included in the configuration.
-Frameworks that intend to use Spring Cloud Stream transparently may create binder configurations that will be referenceable by name, but will not affect the default binder configuration.
-In order to do so, a binder configuration may have its the `applicationProvided` flag set to false, e.g. `spring.cloud.stream.binders.<configurationName>.applicationProvided=false`.
+Frameworks that intend to use Spring Cloud Stream transparently may create binder configurations that can be referenced by name, but will not affect the default binder configuration.
+In order to do so, a binder configuration may have its the `defaultCandidate` flag set to false, e.g. `spring.cloud.stream.binders.<configurationName>.defaultCandidate=false`.
 This denotes a configuration that will exist independently of the default binder configuration process.
 ====
 
@@ -783,9 +783,9 @@ When this is configured, the context in which the binder is being created is not
 This allows for complete separation between the binder components and the application components.
 +
 Default `empty`.
-applicationProvided::
-  Whether the binder configuration is provided by the application (the default case) or a framework included by an application.
-This allows framework authors to depend on Spring Cloud Stream and provide their own infrastructure without interfering with the application.
+defaultCandidate::
+  Whether the binder configuration is a candidate for being considered a default binder, or can be used only when explicitly referenced.
+This allows adding binder configurations without interfering with the default processing.
 +
 Default `true`.
 
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderConfiguration.java
@@ -34,15 +34,21 @@ public class BinderConfiguration {
 
 	private final boolean inheritEnvironment;
 
+	private final boolean applicationProvided;
+
 	/**
 	 * @param binderType the binder type used by this configuration
 	 * @param properties the properties for setting up the binder
-	 * @param inheritEnvironment whether the binder should inherit the environment of the module
+	 * @param inheritEnvironment whether the binder should inherit the environment of the
+	 * module
+	 * @param applicationProvided whether the binder is user defined
 	 */
-	public BinderConfiguration(BinderType binderType, Properties properties, boolean inheritEnvironment) {
+	public BinderConfiguration(BinderType binderType, Properties properties, boolean inheritEnvironment,
+			boolean applicationProvided) {
 		this.binderType = binderType;
 		this.properties = properties;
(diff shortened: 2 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderConfiguration.java
```diff
@@ -34,15 +34,21 @@
...
 	private final boolean inheritEnvironment;
 
+	private final boolean applicationProvided;
+
 	/**
 	 * @param binderType the binder type used by this configuration
 	 * @param properties the properties for setting up the binder
-	 * @param inheritEnvironment whether the binder should inherit the environment of the module
+	 * @param inheritEnvironment whether the binder should inherit the environment of the
+	 * module
```
markfisher: should "module" be changed to "application" here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderConfiguration.java
```diff
@@ -34,15 +34,21 @@
...
+	 * @param applicationProvided whether the binder is user defined
```
markfisher: I think this would be more clear if it said something like:

```
whether the binder is provided by an application, as opposed to being user defined
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -95,8 +97,31 @@ public void destroy() throws Exception {
...
+					this.defaultBinder = applicationProvidedConfigurations.iterator().next();
+					configurationName = this.defaultBinder;
+				}
+				else {
+					if (applicationProvidedConfigurations.size() > 1) {
+						throw new IllegalStateException(
+								"A default binder has been requested, but there is more than one binder available: "
+										+ StringUtils
+												.collectionToCommaDelimitedString(applicationProvidedConfigurations)
+										+ ", and" + " no default binder has been set.");
```
markfisher: excessive string concatenation here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -55,30 +55,34 @@
...
+			BinderProperties binderProperties = binderEntry.getValue();
+			if (binderTypeRegistry.get(binderEntry.getKey()) != null) {
+				binderConfigurations.put(binderEntry.getKey(),
+						new BinderConfiguration(binderTypeRegistry.get(binderEntry.getKey()),
+								binderProperties.getEnvironment(), binderProperties.isInheritEnvironment(),
+								binderProperties.isApplicationProvided()));
+			}
+			else {
+				Assert.hasText(binderProperties.getType(),
+						"No 'type' property present for custom " + "binder " + binderEntry.getKey());
```
markfisher: excessive string concatenation

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -729,6 +729,9 @@ If your application should connect to more than one broker of the same type, you
 ====
 Turning on explicit binder configuration will disable the default binder configuration process altogether.
 If you do this, all binders in use must be included in the configuration.
+Frameworks that intend to use Spring Cloud Stream transparently may create binder configurations that will be referenceable by name, but will not affect the default binder configuration.
+In order to do so, a binder configuration may have its the `applicationProvided` flag set to false, e.g. `spring.cloud.stream.binders.<configurationName>.applicationProvided=false`.
```
markfisher: remove 'its' from "may have its the"

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -55,30 +55,34 @@
...
+						new BinderConfiguration(binderType, binderProperties.getEnvironment(),
+								binderProperties.isInheritEnvironment(), binderProperties.isApplicationProvided()));
 			}
 		}
-		else {
+		if (!hasUserDefinedBinders) {
 			for (Map.Entry<String, BinderType> entry : binderTypeRegistry.getAll().entrySet()) {
 				binderConfigurations.put(entry.getKey(),
-						new BinderConfiguration(entry.getValue(), new Properties(), true));
+						new BinderConfiguration(entry.getValue(), new Properties(), true, true));
```
markfisher: I'm confused why the last boolean parameter (which is `applicationProvided`) is set to `true` here, given that we are within the scope of an if condition that indicates there are no user-defined binders, which IIUC user-defined is equivalent to `applicationProvided`

mbogoevici: The variable needs to be renamed to `containsApplicationProvidedBinders`, but the idea is that:
a) only binders having `applicationProvided` set to `true` are candidates for being the `defautBinder` (if none is specified) - this allows framework providers to include an arbitrary number of binder configurations in the environment, transparent to the user.
b) this logic preserves the original behaviour, where if there are no user-defined (i.e. `applicationProvided`) binders then we automatically create one configuration for each binder found on the classpath - the flag is set to true because of a) - so that these automatically created configurations can be considered for being the `defaultBinder`

markfisher: now that I think I do understand, maybe the boolean property's name should simply be `defaultCandidate`, and a framework provided binder that should not interfere with the default determination would set that value to FALSE (otherwise it is TRUE by default)

mbogoevici: I like the name.

mbogoevici: How about 'hidden' or 'private' ?

markfisher: I guess that depends on how "hidden" it really is. If a user knew the name, they could still use it right?

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -760,6 +763,32 @@ spring:
                 host: <host2>
 ----
 
+=== Binder configuration properties
+
+The following properties are available when creating custom binder configurations.
+They must be prefixed with `spring.cloud.stream.binder.<configurationName>`.
```
ilayaperumalg: In prefix: `binder` -> `binders`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
mbogoevici: Yes. I'm just struggling with `defaultCandidate` - since it's having two meanings "is the *default* candidate" or "is this a candidate for defaulting" ?

mbogoevici: And yes - it's not so much hidden - as it is "just not interfering with the rest of the process"

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/BinderFactoryConfigurationTests.java
```diff
@@ -148,6 +148,38 @@ public void loadBinderTypeRegistryWithTwoBinders() throws Exception {
 	}
 
 	@Test
+	public void loadBinderTypeRegistryWithUserDefinedBinder() throws Exception {
+
+		ConfigurableApplicationContext context = createBinderTestContext(
+				new String[] { "binder1"},
+				"spring.cloud.stream.binders.custom.type=binder1",
+				"spring.cloud.stream.binders.custom.environment.binder1.name=foo",
+				"spring.cloud.stream.binders.custom.applicationProvided=false",
```
ilayaperumalg: Is this supposed to use `true` as this test case suggests `user-defined` binder? or may be we just need another test case to test `user-defined` binder to be chosen as the `defaultBinder`?

mbogoevici: This is intentional set to `false`, as `binder1` will still be elected as the default binder.

mbogoevici:
Pushed

spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -729,6 +729,9 @@ If your application should connect to more than one broker of the same type, you
 ====
 Turning on explicit binder configuration will disable the default binder configuration process altogether.
 If you do this, all binders in use must be included in the configuration.
+Frameworks that intend to use Spring Cloud Stream transparently may create binder configurations that can be referenced by name, but will not affect the default binder configuration.
+In order to do so, a binder configuration may have its the `defaultCandidate` flag set to false, e.g. `spring.cloud.stream.binders.<configurationName>.defaultCandidate=false`.
```
markfisher: remove one of: "its" or "the"

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/6eddac9056aa2e7caa28e2e107a163cde95752d6
along with https://github.com/spring-cloud/spring-cloud-stream/commit/4d8ced734b00770d33d67f67f781c05d285627e2

Issue #546, opened 2016-05-24
Move binder implementations to their own repo


sabbyanandan:
As a developer, I'd like to move rabbit and kafka binders from SCSt core repo to separate GH repos, so I can build the Stream project with relevant binders and on-demand.

**Acceptance:**
- rabbit and kafka binders are moved to its own repos
- test infrastructure is moved and the binder tests pass successfully
- ci plans established for the new repos
- docs updated

PR #561, opened 2016-06-02
Move binders out of the main repo

ericbottard:
Fixes spring-cloud/spring-cloud-stream#546

Repos from https://github.com/ericbottard/spring-cloud-stream-binder-rabbit
and https://github.com/ericbottard/spring-cloud-stream-binder-kafka
should also be forked/copied to the the spring-cloud org.

After that, CI plans will need to be set up


sobychacko:
This PR is manually merged.

PR #563, opened 2016-06-03
Fix mainApplicationClass for child contexts

### LINKED ISSUE #562: Set `mainApplicationClass` correctly for aggregate child contexts
Currently it defaults to the parent, which causes multiple `Started ` messages to be displayed with the same class name.

ilayaperumalg:
- Make sure to set the appropriate mainApplicationClass name for the child applications

This resolves #562


markfisher:
LGTM; merged

PR #566, opened 2016-06-04
Align checkstyle with other projects in Spring Cloud

dsyer:
Do not merge. There are still some violations to do with unused imports (they are used in javadocs).


mbogoevici:
The differences between that you see are due to different versions of Puppycrawl being used (v 6.17 which got from Boot) is by default more lenient around `this` and imports in JavaDoc. That's an easy fix.

mbogoevici:
Thanks for the cleanup! However, I've noticed that a large of number of lines here are just reformatting around shortening the lines. Can we revert those?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
```diff
@@ -73,8 +75,8 @@ static ConfigurableApplicationContext run(Class<?>... apps) {
 	 * @param apps a list of classes, representing root context definitions for apps
 	 * @param args arguments for the apps
 	 */
-	static void runEmbedded(ConfigurableApplicationContext parentContext,
-			Class<?>[] apps, String[][] args) {
+	static void runEmbedded(ConfigurableApplicationContext parentContext, Class<?>[] apps,
```
mbogoevici: This is an example where keeping things one line would be perfectly acceptable and more readable. I don't think we want to enforce wrapping at 80 characters.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -133,82 +135,97 @@ public void setIntegrationEvaluationContext(EvaluationContext evaluationContext)
 
 	@Override
 	public final void afterPropertiesSet() throws Exception {
-		Assert.notNull(this.applicationContext, "The 'applicationContext' property must not be null");
+		Assert.notNull(this.applicationContext,
+				"The 'applicationContext' property must not be null");
 		if (this.evaluationContext == null) {
-			this.evaluationContext = ExpressionUtils.createStandardEvaluationContext(getBeanFactory());
+			this.evaluationContext = ExpressionUtils
```
mbogoevici: This is one example where I don't think we want to wrap. Generally speaking, I don't agree with forcibly wrapping lines at 80 chars.

garyrussell: We (SI) use 120 for code, 90 for javadocs.

src/checkstyle/checkstyle.xml
```diff
@@ -1,152 +0,0 @@
-<?xml version="1.0"?>
-<!DOCTYPE module PUBLIC
```
mbogoevici: My strong preference here is to keep the rules and correct the behaviour for `this` as well as unused imports. Also, wrap them into an artifact so that the reported issues around reactor and subproject builds are avoided.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
mbogoevici: Yes, that's something we should enforce as well I think. Spring Framework rules are 120 max for lines and 80 max for Javadoc.

dsyer:
I have a maximum line width set to 90. Should we go beyond that (see 5cb2e8e)?

mbogoevici:
I think we should use 120 for code and 80/90 for javadoc and wrap manually if we feel that it provides a better flow (which is pretty much the standard).

mbogoevici:
Closing as we've adopted the checkstyle settings from 'spring-cloud-tools'

PR #569, opened 2016-06-07
Rabbit Configuration Corrections

### LINKED ISSUE #564: Issues creating multiple binder configurations in Cloud Foundry
Identified with version 1.0.1.RELEASE:

- on multibinder configurations, when the cloud profile is active, and no cloud connector libraries are present on the classpath, and the binder is set not to inherit its connection factory from parent, the logic does not provide the path for the creation of a connection factory;
- if a service is bound and the connector libraries are present on the classpath,  it not possible to create a customized binder (i.e. using an arbitrary set of credentials) as the connection factory will either be created by the connector or will be overridden by autoreconfiguration;

Solution: while there are workarounds for specific applications, but in the general case (e.g. when embedding  hystrix-stream into the application) it is desirable to provide more flexibility. So the proposed solution is to add configuration that allows to override the use of the cloud connectors and always fall back to the configuration properties.

mbogoevici:
Fixes #564 

To be cherry-picked to master.

- Correct an issue where if the `cloud` profile is enabled but no Connectors are available,
no `ConnectionFactory` will be created;
- Introduce a property named `spring.cloud.stream.overrideCloudConnectors` that allows
to suppress the use of Cloud connectors in the application and fall back to Spring Boot
configuration. This is provided so that specific binders can be created with arbitrary
connectivity parameters if needed.

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitServiceAutoConfiguration.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitServiceAutoConfiguration.java
@@ -21,10 +21,12 @@ import org.springframework.amqp.rabbit.core.RabbitTemplate;
 import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.boot.actuate.health.RabbitHealthIndicator;
 import org.springframework.boot.autoconfigure.amqp.RabbitAutoConfiguration;
+import org.springframework.boot.autoconfigure.amqp.RabbitProperties;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binders/spring-cloud-stream-binder-rabbit/src/main/java/org/springframework/cloud/stream/binder/rabbit/config/RabbitServiceAutoConfiguration.java
```diff
@@ -41,38 +43,95 @@
...
+			// configuration is not triggered.
+			@EnableConfigurationProperties(RabbitProperties.class)
+			protected static class UseCloudConnectors {
+
+				/**
+				 * Creates a {@link ConnectionFactory} using the singleton service
+				 * connector.
+				 *
+				 * @param cloud {@link Cloud} instance to be used for accessing services.
+				 * @return the @{@link ConnectionFactory} used by the binder.
```
markfisher: one extra `@` here

markfisher:
squashed and merged (with the one minor fix mentioned above) via https://github.com/spring-cloud/spring-cloud-stream/commit/228f24e03440aab3ac4467c7cab1b7a944b1ad2e

Issue #587, opened 2016-07-11
Need additional Kafka Producer/ConsumerConfig parameters exposed


kdunn926:
I'm building a custom splitter processor module with reasonably large records and am hitting this exception:
```
org.apache.kafka.common.errors.RecordTooLargeException: The message is 1125040 bytes when serialized which is larger than
the maximum request size you have configured with the max.request.size configuration.
```

I've adjusted the `max.request.size` parameter on the Kafka server configuration although it looks like the Kafka Binder can override this setting on the client side. In my stdout logs below, I see `max.request.size = 1048576` regardless of the server setting.
```
2016-07-11 16:17:21.077  INFO 83422 --- [           main] o.s.i.kafka.support.ProducerFactoryBean  : Using producer proper
ties => {bootstrap.servers=localhost:9092, linger.ms=0, acks=1, compression.type=none, batch.size=16384}
2016-07-11 16:17:21.085  INFO 83422 --- [           main] o.a.k.clients.producer.ProducerConfig    : ProducerConfig values:
        compression.type = none
        metric.reporters = []
        metadata.max.age.ms = 300000
        metadata.fetch.timeout.ms = 60000
        acks = 1
        batch.size = 16384
        reconnect.backoff.ms = 10
        bootstrap.servers = [localhost:9092]
        receive.buffer.bytes = 32768
        retry.backoff.ms = 100
        buffer.memory = 33554432
        timeout.ms = 30000
        key.serializer = class org.apache.kafka.common.serialization.ByteArraySerializer
        retries = 0
        max.request.size = 1048576
        block.on.buffer.full = true
        value.serializer = class org.apache.kafka.common.serialization.ByteArraySerializer
        metrics.sample.window.ms = 30000
        send.buffer.bytes = 131072
        max.in.flight.requests.per.connection = 5
        metrics.num.samples = 2
        linger.ms = 0
        client.id =
```

Any chance we can expose this parameter to the Producer and Consumer clients, similar to this:
`spring.cloud.stream.kafka.binder.fetchSize`

Thanks!

-Kyle

PR #588, opened 2016-07-13
Exposing max.request.size for producers with large payloads

kdunn-pivotal:
Addressing issue #587 .

### DIFF
--- a/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaProducerProperties.java
+++ b/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaProducerProperties.java
@@ -27,6 +27,8 @@ public class KafkaProducerProperties {
 
 	private int bufferSize = 16384;
 
+	private int maxRequestSize = 1048576;
+
 	private ProducerMetadata.CompressionType compressionType = ProducerMetadata.CompressionType.none;
 
 	private boolean sync;
@@ -39,6 +41,14 @@ public class KafkaProducerProperties {
 
 	public void setBufferSize(int bufferSize) {
 		this.bufferSize = bufferSize;
+	}
+
+	public int getMaxRequestSize() {
+		return maxRequestSize;
+	}
+
+	public void setMaxRequestSize(int maxRequestSize) {
+		this.maxRequestSize = maxRequestSize;
 	}
 
 	@NotNull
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@kdunn-pivotal Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=588)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=588) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@kdunn-pivotal Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=588)!

spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaProducerProperties.java
```diff
@@ -41,6 +43,14 @@ public void setBufferSize(int bufferSize) {
 		this.bufferSize = bufferSize;
 	}
 
+	public int getMaxRequestSize() {
```
mbogoevici: Besides adding the parameter, we will need to add code in the binder to act on it.

kdunn-pivotal: The definitely makes sense. I was attempting to follow bufferSize as a pattern but I'm not seeing that parameter referenced outside this source file - just assumed there was Spring magic in play here. Any pointers you can provide would be helpful and I'll try to follow up with the rest of the implementation.

mbogoevici: For the sake of the previous example: bufferSize is set here https://github.com/spring-cloud/spring-cloud-stream/blob/1.0.x/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L312

However, since ProducerMetadata doesn't have a field for it, it will need to be set the `additional` set of properties, eg.

https://github.com/spring-cloud/spring-cloud-stream/blob/1.0.x/spring-cloud-stream-binders/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L314

mbogoevici: Thanks for the contribution!

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/3be5ae933d2cbe75cd14459c0bdefa2fad1ea2d8

prashantsingh99:
I cant find this change in Maven Repository:  Camden.SR6. Has this been rolled back?

ilayaperumalg:
Irrespective of this, you can set any producer config via `configuration` property in Kafka producer config like: `spring.cloud.stream.kafka.bindings.<channelName>.producer.configuration.max.request.size`. This property is supported since 1.1.0.M1 or 1.2.x Kafka binder releases.

Also, any Kafka binder related issues, please open GH issue [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues)

mbogoevici:
@prashantsingh99 Also, we discourage commenting on closed issues. Rather, we prefer that a new question/issue is raised, referencing the old one. Aside from that, @ilayaperumalg has mentioned, the intent is to use the `configuration` property for passing native consumer/producer properties.

Issue #493, opened 2016-04-20
Partitioning should be processed before serialization


gadams00:
I'm using s-c-s with kafka and have noticed that if you combine 
`s.c.s.bindings.<channel name>.content-type: application/json `
and 
`s.c.s.bindings.<channel name>.producer.partitionKeyExpression`
your partitionKeyExpression is applied after your pojo has already been transformed to json, so expressions like payload.getName() don't work. Is there a way to combine partitioning based on pojo expressions with the portability of json-encoded messages?

mbogoevici:
Hi @gadams00,

Thanks for pointing that out. Ideally we should apply the partitioning logic before the conversion.

However, in your case  `partition.name` as an expression will work as of now (the expression is evaluated against the JSON). 

Cheers,
Marius

gadams00:
You mean payload.name as s.c.s.bindings.\<channel name\>.producer.partitionKeyExpression? That seems to be working fine. Thanks.

gadams00:
On second thought, I probably shouldn't have closed this, since mbogoevici indicated that it could be a potential enhancement at some point.

mbogoevici:
Thanks @gadams00.

mbogoevici:
One thing that we will want to do for 1.1.0.M1 is to ensure that the partitioning logic is applied before the final conversion - this will allow to use the partitioning logic on seralized POJOs (eg. Avro).

ilayaperumalg:
This can be closed as the PR #589 addressed it.

PR #589, opened 2016-07-14
Add AbstractMessageChannelBinder to handle common aspects of message-…

mbogoevici:
…channel binders

- Added generic, customizable ReceivingHandler and SendingHandler
- made doBindProducer delegate to a series of template methods
- made doBindConsumer delegate to a series of template methods
- moved partitioning to an interceptor, thus fixing #493

Removed unused manual ack handling


mbogoevici:
Applying this also requires https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/11 and https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/pull/3 to be applied.

garyrussell:
`AbstractBinder` and `AbstractMessageChannelBinder` "feel" like they should be `BinderSupport` and `MessageChannelBinderSupport` - the general convention is to name them like that if subclasses invoke "support" methods on the superclass (e.g. `buildRetryTemplateIfRetryEnabled()`).

(Hence `MessageBusSupport` in XD).

markfisher:
I think it's ok to keep those 2 as `Abstract*` for two reasons:

1. they do implement the actual `Binder

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -0,0 +1,334 @@
...
+	public final Binding<MessageChannel> doBindProducer(final String destination, MessageChannel outputChannel,
+			final P producerProperties) {
+		Assert.isInstanceOf(SubscribableChannel.class, outputChannel,
+				"Binding is supported only for SubscribableChannel instances");
+		createProducerDestinationIfNecessary(destination, producerProperties);
+		final MessageHandler producerMessageHandler;
+		try {
+			producerMessageHandler = createProducerMessageHandler(destination, producerProperties);
+		}
+		catch (Exception e) {
```
markfisher: should add a `throws BinderException` clause to the method definition for clarity (and corresponding `@throws` in JavaDoc)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -0,0 +1,334 @@
...
+			return new DefaultBinding<MessageChannel>(name, group, inputChannel,
+					endpoint instanceof Lifecycle ? (Lifecycle) endpoint : null) {
+
+				@Override
+				protected void afterUnbind() {
+					AbstractMessageChannelBinder.this.afterUnbindConsumer(this.name, this.group, properties);
+				}
+			};
+		}
+		catch (Exception e) {
```
markfisher: same comment (re: `throws` clause + JavaDoc) applies here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -0,0 +1,334 @@
...
+	 * Creates target destinations for outbound channels. The implementation is middleware-specific.
+	 * @param name the name of the producer destination
+	 * @param properties producer properties
+	 */
+	protected abstract void createProducerDestinationIfNecessary(String name, P properties);
+
+	/**
+	 * Creates a {@link MessageHandler} with the ability to send data to the
+	 * target middleware. If the returned instance is also a {@link Lifecycle},
+	 * it must be started by the method and it will be stopped automatically
```
markfisher: what's the rationale for requiring it to be started before returned by the implementation, rather than starting it from the common superclass IF it's `Lifecycle` (could even check `isRunning()` so that it's optional)

mbogoevici: This is an oversight on my end. I've changed the consumer endpoint to be handled by the superclass, but wasn't consistent here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -168,12 +159,78 @@ public Object fromMessage(Message<?> message, Class<?> targetClass, Object conve
...
+			this.contentType = contentType;
+			this.mimeType = MessageConverterUtils.getMimeType(contentType);
+		}
+
+		@Override
+		public Message<?> preSend(Message<?> message, MessageChannel channel) {
+			Class<?>[] classes =
+					MessageConverterConfigurer.this.compositeMessageConverterFactory.supportedDataTypes(
+							this.mimeType);
+			MessageWrappingMessageConverter messageConverter =
```
markfisher: could defer creation of this until the last `else` block below since that's the only place it's used? (no need to create here if the method might `return` without using it)

mbogoevici: Yeah, I want to redo this entire part as an effort to overhaul conversion. So I just mashed together the existing behaviour until this gets refactored.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -168,12 +159,78 @@ public Object fromMessage(Message<?> message, Class<?> targetClass, Object conve
...
+		@Override
+		public Message<?> preSend(Message<?> message, MessageChannel channel) {
+			EvaluationContext evaluationContext = ExpressionUtils.createStandardEvaluationContext(
+					MessageConverterConfigurer.this.beanFactory);
+			PartitionHandler partitionHandler = new PartitionHandler(MessageConverterConfigurer.this
+					.beanFactory, evaluationContext, null,
+					this.bindingProperties.getProducer());
+			int partition = partitionHandler.determinePartition(message);
+			return new MutableMessageBuilderFactory().fromMessage(message)
+					.setHeader(BinderHeaders.PARTITION_HEADER, partition).build();
```
ilayaperumalg: Does this mean, the message will always have the `BinderHeaders.PARTITION_HEADER` with the `partition` value if the producer has `partitioned` consumer and will never have to determine partition at runtime (while that is still possible right before serialization) ?

mbogoevici: Hi Ilaya, can you clarify the question? Yes, the header will be set if `partitioned` is true on the producer, and the value is computed at runtime based on the producer settings. The binders do not determine the partition anymore (that has become the responsibility of the core).

ilayaperumalg: > The binders do not determine the partition anymore (that has become the responsibility of the core).

This is the answer to my intended question - While the binder can still determine the partition at runtime, it is left to the responsibility to core to `always` update the header with the `partition` value. May be we can document this somewhere in the Javadoc.

mbogoevici: Sure, we can document this on `createProducerMessageHandler`.

spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/PartitionCapableBinderTests.java
```diff
@@ -51,6 +52,7 @@
 	protected static final SpelExpressionParser spelExpressionParser = new SpelExpressionParser();
 
 	@Test
+	@Ignore
```
markfisher: did you intentionally add this?

mbogoevici: Should be removed, was added by mistake while debugging on a different branch.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/bd92af47840e861afccb3a1a263216311e400743

PR #596, opened 2016-07-20
Allow the use of regular MessageConverters 

mbogoevici:


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -174,41 +187,81 @@ protected Object build(Object payload, MessageHeaders headers) {
...
+				this.klazz = Tuple.class;
+			}
+			else if (this.mimeType.getType().equals("text") || this.mimeType.getSubtype().equals(
+					"json") || this.mimeType.getSubtype().equals("xml")) {
+				this.klazz = String.class;
+			}
+			else {
+				this.klazz = byte[].class;
+			}
+			if (input) {
```
markfisher: the if and else branches appear identical

mbogoevici: Yes, I thought that the two would be different, but at this point they don't seem to.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -102,25 +102,26 @@ public synchronized Object invoke(MethodInvocation invocation) throws Throwable
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-		ReflectionUtils.doWithMethods(type, new ReflectionUtils.MethodCallback() {
+		ReflectionUtils.doWithMethods(this.type, new ReflectionUtils.MethodCallback() {
 			@Override
 			public void doWith(Method method) throws IllegalArgumentException {
-				Assert.notNull(channelFactory, "Channel Factory cannot be null");
+				Assert.notNull(BindableProxyFactory.this.channelFactory, "Channel Factory cannot be null");
```
garyrussell: This can be done once, before calling `doWith()` rather than for each method.

mbogoevici: Yes, wrong placement :)

mbogoevici:
Closing and reopening

mbogoevici:
Closed in favour of https://github.com/spring-cloud/spring-cloud-stream/pull/598

PR #597, opened 2016-07-20
Aggregate application doesn't need binder in its classpath

### LINKED ISSUE #576: AggregateApplication's should support fully contained configurations
This issue follows on from the discussion started on Gitter [here](https://gitter.im/spring-cloud/spring-cloud?at=57680d232a4cd63745eb2f88) and following the advise of @dsyer [here](https://gitter.im/spring-cloud/spring-cloud?at=576812526577f032450ce70a).

Given an AggregateApplication with a Source (like the [TimeSource](https://github.com/spring-cloud/spring-cloud-stream-samples/blob/master/source/src/main/java/demo/TimeSource.java example)), a Processor and a Sink, the AggregateApplication does not require any external broker (fully contained) and therefore should not require a binder like the Rabbit or Kafka binder to be on the classpath.

ilayaperumalg:
- BinderTypeRegistry will have empty binder types for aggregate applications
  - Check if `SharedChannelRegistry` bean is in the context, then proceed with no binder type

This resolves #576

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -21,9 +21,6 @@ import java.util.Map.Entry;
 
 import org.springframework.boot.Banner.Mode;
 import org.springframework.boot.builder.SpringApplicationBuilder;
-import org.springframework.cloud.stream.messaging.Processor;
-import org.springframework.cloud.stream.messaging.Sink;
-import org.springframework.cloud.stream.messaging.Source;
 import org.springframework.context.ConfigurableApplicationContext;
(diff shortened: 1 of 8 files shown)


mbogoevici:
I am afraid that the solution is not entirely correct. What the original issue says is that we should not validate for the existence of a binder only if the application is entirely contained, i.e. it starts with a source and ends with a sink. With this approach, we skip binder validation even if the application has bindable channels (e.g start and/or ends with a processor). We should do a more thorough inspection of the aggregated pipeline.

ilayaperumalg:
Pushed a change that explicitly checks if aggregator application is self contained. Since the `AggregatorParentContext` itself brings in the `BinderFactorConfiguration`, the `AggregatorBuilderConfigurer` needs to handle it earlier in the lifecycle.

sobychacko:
LGTM. Reviewed/rebased and merged.

PR #598, opened 2016-07-20
Rework the converter system

### LINKED ISSUE #505: Allow the registration of regular `MessageConverters` as beans

mbogoevici:
Fixes #505

The goals are twofold: to simplify the registration of new converters and
to add some consistency to the conversion process, and align them with the
way converters are used in generic Spring Messaging listeners, by describing
two possible transformations: inbound, message containing a serialized format (including contentType)->targetClass,
and outbound payload+headers -> message. The idea is for the two
transformations to match the input and output directions of the bound channels.

List of changes:

- Create and configure input/output channels distinctly - reflected
  in the definitions of `BindableChannelFactory` and `MessageChannelConfigurer`;
- Replace AbstractFromMessageConverter with bidirectional converters;
- Use the channel direction (input/output) to determine whether
  `toMessage` or `fromMessage` will be invoked;
- Use contentType support from `AbstractMessageConverter` to map
  converters to mime types instead;


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -174,41 +187,74 @@ protected Object build(Object payload, MessageHeaders headers) {
...
-					if (converted != null) {
-						return (Message<?>) converted;
+					sentMessage = message;
+				}
+			}
+			else {
+				Object converted = this.input ? this.messageConverter.fromMessage(message, this.klazz)
+						: this.messageConverter.toMessage(message.getPayload(), message.getHeaders(), null);
+				if (converted != null) {
+					Message<?> convertedMessage = (Message<?>) converted;
```
markfisher: won't this fail after the `fromMessage` calls?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -174,41 +187,74 @@ protected Object build(Object payload, MessageHeaders headers) {
...
+					Object contentType = convertedMessage.getHeaders().get(MessageHeaders.CONTENT_TYPE);
+					if (contentType instanceof MimeType) {
```
markfisher: could the content-type header be a `String` here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/JavaSerializationMessageConverter.java
```diff
@@ -16,44 +16,51 @@
...
-		return new Class<?>[] { byte[].class };
+	protected boolean supports(Class<?> clazz) {
+		return true;
 	}
 
 	@Override
-	protected Class<?>[] supportedPayloadTypes() {
-		return new Class<?>[] { Serializable.class };
+	public Object convertFromInternal(Message<?> message, Class<?> targetClass, Object conversionHint) {
+		ByteArrayInputStream bis = new ByteArrayInputStream((byte[]) (message.getPayload()));
```
markfisher: it doesn't look like we can assume the payload is a byte array here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TupleJsonMessageConverter.java
```diff
@@ -79,4 +74,16 @@ public Object convertFromInternal(Message<?> message, Class<?> targetClass, Obje
 		return json;
 	}
 
+	@Override
+	public Object convertFromInternal(Message<?> message, Class<?> targetClass, Object conversionHint) {
+		String source;
+		if (message.getPayload() instanceof byte[]) {
+			source = new String((byte[]) message.getPayload());
+		}
+		else {
```
markfisher: shouldn't we at least protect this with an assert before the cast to `String`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
mbogoevici: I think it could - generally speaking I think it would be safer to assume so.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TupleJsonMessageConverter.java
mbogoevici: `convertFromInternal` is supposed to do the conversion from the serialized/wire format - which can be either `String` or `byte[]`. I believe we should assume in general that messages bearing a `contentType` header should be either one of the two (perhaps with the sole exception of already deserialized objects, which can have the `application/java-object;type=xyz` header, which is sort of an anomaly).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/JavaSerializationMessageConverter.java
mbogoevici: See comment here: https://github.com/spring-cloud/spring-cloud-stream/pull/598#discussion_r71621110

We can do an assert, but a message bearing this specific content type header which has not a byte[] payload is an anomaly.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
mbogoevici: No, the wrapper converter wraps it a `Message` of its own ... we could probably remove that and move the code here (the wrapper was required for DataType channels).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TupleJsonMessageConverter.java
mbogoevici: Probably it would be better to use `toString()` instead.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -126,9 +136,12 @@ public Object fromMessage(Message<?> message, Class<?> targetClass) {
 			if (converted instanceof Message) {
 				return converted;
 			}
-			else {
+			else if (converted != null) {
 				return build(converted, message.getHeaders());
 			}
+			else {
```
dturanski: Do you need else here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
```diff
@@ -65,56 +65,46 @@ public CompositeMessageConverterFactory(List<? extends AbstractFromMessageConver
...
 	public CompositeMessageConverter getMessageConverterForAllRegistered() {
-		return new CompositeMessageConverter(new ArrayList<MessageConverter>(converters));
+		return new CompositeMessageConverter(new ArrayList<>(this.converters));
 	}
 
-	public Class<?>[] supportedDataTypes(MimeType targetMimeType) {
-		Set<Class<?>> supportedDataTypes = new HashSet<>();
-		// Make sure to check if the target type is of explicit java object type.
-		if (MessageConverterUtils.X_JAVA_OBJECT.includes(targetMimeType)) {
-			supportedDataTypes.add(MessageConverterUtils.getJavaTypeForJavaObjectContentType(targetMimeType));
```
dturanski: This code supported an XD feature where you could say --outputType=com.my.Object  as short cut for x-java-object;type=com.my.Object . Probably gold plating, but I thought it was a nice touch.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/JavaSerializationMessageConverter.java
```diff
@@ -16,44 +16,51 @@
...
-		super(MessageConverterUtils.X_JAVA_OBJECT, MessageConverterUtils.X_JAVA_SERIALIZED_OBJECT);
+	public JavaSerializationMessageConverter() {
+		super(Arrays.asList(MessageConverterUtils.X_JAVA_SERIALIZED_OBJECT, MessageConverterUtils.X_JAVA_OBJECT));
 	}
 
 	@Override
-	protected Class<?>[] supportedTargetTypes() {
-		return new Class<?>[] { byte[].class };
+	protected boolean supports(Class<?> clazz) {
+		return true;
```
dturanski: Why do we not require Serializable?

mbogoevici: I subsequently corrected that. https://github.com/spring-cloud/spring-cloud-stream/pull/598/commits/94678ab17881ec711688eed8ddb3d97210393fd4#diff-3ee67112068a7c2632f85b9d9e290d5aR42

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -81,23 +82,32 @@ public void afterPropertiesSet() throws Exception {
...
-	@Override
-	public void configureMessageChannel(MessageChannel channel, String channelName) {
+	private void configureMessageChannel(MessageChannel channel, String channelName, boolean input) {
 		Assert.isAssignable(AbstractMessageChannel.class, channel.getClass());
 		AbstractMessageChannel messageChannel = (AbstractMessageChannel) channel;
 		final BindingProperties bindingProperties = this.channelBindingServiceProperties.getBindingProperties(
 				channelName);
 		final String contentType = bindingProperties.getContentType();
-		if (bindingProperties.getProducer() != null && bindingProperties.getProducer().isPartitioned()) {
+		if (!input && bindingProperties.getProducer() != null && bindingProperties.getProducer().isPartitioned()) {
```
ilayaperumalg: I guess adding `!input` is just extra check here for the output channel as the `producer` property `partitioned` is only set for `output` channel. or Am I missing something here?

mbogoevici: No, it's redundant, but I believe it helps focusing on the different code paths. Eventually we might want to refactor this further, I believe.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
mbogoevici: Not specifically, just highlighting that it's a different branch.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
markfisher: Then why not make the `converted` var a `Message` type?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
markfisher: so we need to parse `String` to `MimeType` in that case?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -174,41 +187,74 @@ protected Object build(Object payload, MessageHeaders headers) {
...
+										MessageHeaders.CONTENT_TYPE,
+										contentType.toString()).build();
+					}
+					else {
+						sentMessage = convertedMessage;
 					}
 				}
 			}
-			throw new MessageConversionException("Cannot convert " + message + " to " + this.contentType);
+			if (sentMessage != null) {
```
markfisher: I think this would read better as:

```
if (sentMessage == null) {
    throw new MessageConversionException("Cannot convert " + message + " to " + this.contentType);
 }
return sentMessage;
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
mbogoevici: We can do the cast there, I guess, but I want to remove the wrapper.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
mbogoevici: We can restore that behaviour but I wonder if it wouldn't be better to use `contentType` as a real `MimeType` and perhaps have a separate alternate parameter for `javaType`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -81,134 +80,104 @@ public void afterPropertiesSet() throws Exception {
...
+
+		private final Class<?> klazz;
+
+		private final MessageConverter messageConverter;
+
+		private ContentTypeConvertingInterceptor(String contentType, boolean input) {
 			this.contentType = contentType;
 			this.mimeType = MessageConverterUtils.getMimeType(contentType);
+			this.input = input;
+			if (MessageConverterUtils.X_JAVA_OBJECT.includes(this.mimeType)) {
```
markfisher: is this backwards? i.e. should it be:

```
if (this.mimeType.includes(MessageConverterUtils.X_JAVA_OBJECT)) {...
```
?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
```diff
@@ -42,16 +42,16 @@
...
-		this(Collections.<AbstractFromMessageConverter>emptyList(), new ObjectMapper());
+		this(Collections.<MessageConverter>emptyList(), new ObjectMapper());
 	}
 
 	/**
-	 * @param customConverters a list of {@link AbstractFromMessageConverter}
+	 * @param customConverters a list of {@link AbstractMessageConverter}
 	 */
-	public CompositeMessageConverterFactory(List<? extends AbstractFromMessageConverter> customConverters,
+	public CompositeMessageConverterFactory(List<? extends MessageConverter> customConverters,
```
markfisher: why not keep `AbstractMessageType` as the type within the list considering anything else would be ignored based on line 91 (why permit anything else to be added?)

mbogoevici: Because `getMessageConverterForAllRegistered()` is used for setting up message conversion in `StreamListener`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TupleJsonMessageConverter.java
```diff
@@ -79,4 +74,16 @@ public Object convertFromInternal(Message<?> message, Class<?> targetClass, Obje
 		return json;
 	}
 
+	@Override
+	public Object convertFromInternal(Message<?> message, Class<?> targetClass, Object conversionHint) {
+		String source;
+		if (message.getPayload() instanceof byte[]) {
+			source = new String((byte[]) message.getPayload());
```
markfisher: do we need to specify charset here? (UTF-8?)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
mbogoevici: We could for example have converters that use a `ConversionService` for simple payload->argumentType transformation.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TupleJsonMessageConverter.java
mbogoevici: `UTF-8` is the default. We can extract the encoding from the content type I suppose.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
markfisher: I see. Maybe we need an `else` block after the line 91 `if` that logs if a registered converter is not relevant (probably at `DEBUG` level).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
mbogoevici: No, because then `application/*` would be a match. I guess we should use `equals` instead anyway.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TupleJsonMessageConverter.java
markfisher: wouldn't it be the platform default charset here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
mbogoevici: Maybe .. it's not a bug or an incorrect condition, it's just not relevant for that particular type of query (by MimeType).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/TupleJsonMessageConverter.java
mbogoevici: Yes.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CompositeMessageConverterFactory.java
mbogoevici: Added a debugging note.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/55b9aa75ddc3cea073855eb4cb9096ba81da8cc5

PR #600, opened 2016-07-22
Move calling afterProperties on consumer endpoints from binder implem…

sobychacko:
…entations to core

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -21,6 +21,7 @@ import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.http.MediaType;
 import org.springframework.integration.channel.FixedSubscriberChannel;
+import org.springframework.integration.context.IntegrationObjectSupport;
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -179,6 +180,9 @@ protected void afterUnbindProducer(String destination, P producerProperties) {
 			bridge.setBeanName("bridge." + name);
 			consumerEndpoint = createConsumerEndpoint(name, group, destination, properties);
 			consumerEndpoint.setOutputChannel(bridge);
+			if(consumerEndpoint instanceof IntegrationObjectSupport) {
+				((IntegrationObjectSupport) consumerEndpoint).afterPropertiesSet();
```
mbogoevici: This should be `InitializingBean`

mbogoevici: Also, we need to do the same for the producer handler and documented the contract in the Javadoc.

sobychacko:
@mbogoevici PR comments are addressed.

mbogoevici:
Squashed and merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/8dd22ebca0bd6439d3f4d75ce8674f07d09eca70

PR #601, opened 2016-07-26
Initial support for Reactive API

### LINKED ISSUE #458: Introduce full Reactive support in SCS 1.1
Since we are in the process of providing a comprehensive Reactive support in various Spring projects based on [Reactive Streams](http://www.reactive-streams.org/) defacto standard, and supporting both Reactor Core and RxJava API out of the box, I think that would make sense for Spring Cloud Stream  to refactor RxJava processor support before `1.0.0.GA` to a more generic Reactive support that will be suitable for Reactive Streams, Reactor Core 2.5+ ([Flux](http://projectreactor.io/core/docs/api/reactor/core/publisher/Flux.html) and [Mono](http://projectreactor.io/core/docs/api/reactor/core/publisher/Mono.html) types implements Reactive Streams `Publisher`) and RxJava 2.x (it is using different packages and is Reactive Streams compliant). 

RxJava 1.x support would be provided as one of the composition library supported, and @smaldini and I could contribute a draft PR for Reactive Streams generic support (`Publisher<O> process(Publisher<I> input)` processor) based on Reactor Core 2.5+ to validate the design (this PR could be integrated when Reactor Core 2.5 GA will be available maybe it could be available before Spring Cloud Stream `1.0.0.GA` release).

My initial proposal to be discussed is the following:
 - Rename `spring-cloud-stream-rxjava` module to `spring-cloud-stream-reactive`
 - Rename `org.springframework.cloud.stream.annotation.rxjava` to `org.springframework.cloud.stream.annotation.reactive`
 - Make `@EnableRxJavaProcessor` usable for any kind of Reactive library by renaming it to `@EnableReactiveProcessor` (that will also help for the upcoming RxJava 2.x)
 - Clarify, with proper package name, class name or Javadoc, that `RxJavaProcessor` is only for RxJava 1.x

@smaldini and I are available to work on this with you. Any thoughts?

### LINKED ISSUE #520: Add support for @Input and @Output annotated arguments
Depends on spring-cloud/spring-cloud-stream#521

mbogoevici:
Fixes #520
Fixes #458 (for non-reactive binders)

Adds spring-cloud-stream-reactive module.
Introduces support for declarative @StreamListener and @Input and @Output annotated parameters.
Add StreamListenerArgumentAdapter and StreamListenerResultAdapter for wrapping bindable inputs
and outputs when passing arguments to declarative @StreamListener.
Adds support for using reactive types (Flux/Observable) with traditional binders.
Introduce FluxSender and ObservableSender for handling multiple streaming outputs per method.
Ensure that errors are caught and logged.

### DIFF
--- a/spring-cloud-stream-integration-tests/pom.xml
+++ b/spring-cloud-stream-integration-tests/pom.xml
@@ -33,6 +33,12 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-java-dsl</artifactId>
+			<version>1.1.3.RELEASE</version>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 </project>
--- a/spring-cloud-stream-integration-tests/pom.xml
+++ b/spring-cloud-stream-integration-tests/pom.xml
@@ -33,12 +33,6 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
-		<dependency>
-			<groupId>org.springframework.integration</groupId>
-			<artifactId>spring-integration-java-dsl</artifactId>
-			<version>1.1.3.RELEASE</version>
-			<scope>test</scope>
-		</dependency>
 	</dependencies>
 
 </project>
--- a/spring-cloud-stream-reactive/pom.xml
+++ b/spring-cloud-stream-reactive/pom.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-stream-parent</artifactId>
+		<version>1.1.0.BUILD-SNAPSHOT</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-stream-reactive</artifactId>
+
+	<properties>
+		<java.version>1.8</java.version>
+	</properties>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<version>3.0.0.BUILD-SNAPSHOT</version>
+		</dependency>
+		<dependency>
+			<groupId>io.reactivex</groupId>
+			<artifactId>rxjava</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-test-support</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream-test-support-internal</artifactId>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
+
+</project>
--- a/spring-cloud-stream-reactive/pom.xml
+++ b/spring-cloud-stream-reactive/pom.xml
@@ -23,7 +23,6 @@
 		<dependency>
 			<groupId>io.projectreactor</groupId>
 			<artifactId>reactor-core</artifactId>
-			<version>3.0.0.BUILD-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>io.reactivex</groupId>
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToFluxSenderArgumentAdapter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToFluxSenderArgumentAdapter.java
@@ -18,7 +18,6 @@ package org.springframework.cloud.stream.reactive;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import reactor.core.publisher.Flux;
 import reactor.core.publisher.MonoProcessor;
 
 import org.springframework.cloud.stream.binding.StreamListenerArgumentAdapter;
@@ -36,7 +35,7 @@ import org.springframework.messaging.MessageChannel;
 public class MessageChannelToFluxSenderArgumentAdapter
 		implements StreamListenerArgumentAdapter<FluxSender, MessageChannel> {
 
-	private Log log = LogFactory.getLog(FluxToMessageChannelResultAdapter.class);
+	private Log log = LogFactory.getLog(MessageChannelToFluxSenderArgumentAdapter.class);
 
 	@Override
 	public boolean supports(Class<?> boundElementType, MethodParameter methodParameter) {
@@ -48,22 +47,16 @@ public class MessageChannelToFluxSenderArgumentAdapter
 	@Override
 	public FluxSender adapt(MessageChannel boundElement, MethodParameter parameter) {
 		return resultPublisher -> {
-			MonoProcessor<Void> sendError = MonoProcessor.create();
-			Flux<?> outboundFlux;
-			if (resultPublisher instanceof Flux) {
-				outboundFlux = ((Flux<?>) resultPublisher);
-			}
-			else {
-				outboundFlux = (Flux.merge(resultPublisher));
-			}
-			outboundFlux
+			MonoProcessor<Void> sendResult = MonoProcessor.create();
+			// add error handling and reconnect in the event of an error
+			resultPublisher
 					.doOnError(e -> this.log.error("Error during processing: ", e))
 					.retry()
 					.subscribe(
 							result -> boundElement.send(result instanceof Message<?> ? (Message<?>) result :
-									MessageBuilder.withPayload(result).build()), e -> sendError.onError(e),
-							() -> sendError.onComplete());
-			return sendError;
+									MessageBuilder.withPayload(result).build()), e -> sendResult.onError(e),
+							() -> sendResult.onComplete());
+			return sendResult;
 		};
 	}
 }
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToFluxSenderArgumentAdapter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToFluxSenderArgumentAdapter.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.reactive;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.MonoProcessor;
+
+import org.springframework.cloud.stream.binding.StreamListenerArgumentAdapter;
+import org.springframework.core.MethodParameter;
+import org.springframework.core.ResolvableType;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+
+/**
+ * Adapts an {@link org.springframework.cloud.stream.annotation.Output} annotated
+ * {@link FluxSender} to an outbound {@link MessageChannel}.
+ * @author Marius Bogoevici
+ */
+public class MessageChannelToFluxSenderArgumentAdapter
+		implements StreamListenerArgumentAdapter<FluxSender, MessageChannel> {
+
+	private Log log = LogFactory.getLog(FluxToMessageChannelResultAdapter.class);
+
+	@Override
+	public boolean supports(Class<?> boundElementType, MethodParameter methodParameter) {
+		ResolvableType type = ResolvableType.forMethodParameter(methodParameter);
+		return MessageChannel.class.isAssignableFrom(boundElementType) && FluxSender.class.isAssignableFrom(
+				type.getRawClass());
+	}
+
+	@Override
+	public FluxSender adapt(MessageChannel boundElement, MethodParameter parameter) {
+		return resultPublisher -> {
+			MonoProcessor<Void> sendError = MonoProcessor.create();
+			Flux<?> outboundFlux;
+			if (resultPublisher instanceof Flux) {
+				outboundFlux = ((Flux<?>) resultPublisher);
+			}
+			else {
+				outboundFlux = (Flux.merge(resultPublisher));
+			}
+			outboundFlux
+					.doOnError(e -> this.log.error("Error during processing: ", e))
+					.retry()
+					.subscribe(
+							result -> boundElement.send(result instanceof Message<?> ? (Message<?>) result :
+									MessageBuilder.withPayload(result).build()), e -> sendError.onError(e),
+							() -> sendError.onComplete());
+			return sendError;
+		};
+	}
+}
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxArgumentAdapter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxArgumentAdapter.java
@@ -27,6 +27,7 @@ import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.messaging.SubscribableChannel;
 import org.springframework.messaging.converter.CompositeMessageConverter;
+import org.springframework.util.Assert;
 
 /**
  * Adapts an {@link org.springframework.cloud.stream.annotation.Input} annotated
@@ -36,9 +37,10 @@ import org.springframework.messaging.converter.CompositeMessageConverter;
 public class MessageChannelToInputFluxArgumentAdapter
 		implements StreamListenerArgumentAdapter<Flux<?>, SubscribableChannel> {
 
-	private CompositeMessageConverter messageConverter;
+	private final CompositeMessageConverter messageConverter;
 
 	public MessageChannelToInputFluxArgumentAdapter(CompositeMessageConverter messageConverter) {
+		Assert.notNull(messageConverter, "cannot not be null");
 		this.messageConverter = messageConverter;
 	}
 
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxArgumentAdapter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxArgumentAdapter.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.reactive;
+
+import reactor.core.publisher.Flux;
+
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.binding.StreamListenerArgumentAdapter;
+import org.springframework.core.MethodParameter;
+import org.springframework.core.ResolvableType;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.SubscribableChannel;
+import org.springframework.messaging.converter.CompositeMessageConverter;
+
+/**
+ * Adapts an {@link org.springframework.cloud.stream.annotation.Input} annotated
+ * {@link MessageChannel} to a {@link Flux}.
+ * @author Marius Bogoevici
+ */
+public class MessageChannelToInputFluxArgumentAdapter
+		implements StreamListenerArgumentAdapter<Flux<?>, SubscribableChannel> {
+
+	private CompositeMessageConverter messageConverter;
+
+	public MessageChannelToInputFluxArgumentAdapter(CompositeMessageConverter messageConverter) {
+		this.messageConverter = messageConverter;
+	}
+
+	@Override
+	public boolean supports(Class<?> boundElementType, MethodParameter methodParameter) {
+		return SubscribableChannel.class.isAssignableFrom(boundElementType)
+				&& methodParameter.getParameterAnnotation(Input.class) != null
+				&& Flux.class.isAssignableFrom(methodParameter.getParameterType());
+	}
+
+	@Override
+	public Flux<?> adapt(final SubscribableChannel boundElement, MethodParameter parameter) {
+		ResolvableType resolvableType = ResolvableType.forMethodParameter(parameter);
+		Class<?> argumentClass = resolvableType.getGeneric(0).getRawClass();
+		final Object monitor = new Object();
+		if (Message.class.isAssignableFrom(argumentClass)) {
+			return Flux.create(emitter -> {
+				MessageHandler messageHandler = message -> {
+					synchronized (monitor) {
+						emitter.next(message);
+					}
+				};
+				boundElement.subscribe(messageHandler);
+				emitter.setCancellation(() -> boundElement.unsubscribe(messageHandler));
+			});
+		}
+		else {
+			return Flux.create(emitter -> {
+				MessageHandler messageHandler = message -> {
+					synchronized (monitor) {
+						if (argumentClass.isAssignableFrom(message.getPayload().getClass())) {
+							emitter.next(message.getPayload());
+						}
+						else {
+							emitter.next(this.messageConverter.fromMessage(message, argumentClass));
+						}
+					}
+				};
+				boundElement.subscribe(messageHandler);
+				emitter.setCancellation(() -> boundElement.unsubscribe(messageHandler));
+			});
+		}
+	}
+}
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputObservableArgumentAdapter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputObservableArgumentAdapter.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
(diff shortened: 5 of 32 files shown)


spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ObservableSender.java
```diff
@@ -0,0 +1,37 @@
...
+ * org.springframework.cloud.stream.annotation.Output}.
+ * @author Marius Bogoevici
+ */
+public interface ObservableSender {
+
+	/**
+	 * Streams the {@link Observable} through the bound
+	 * element corresponding to the {@link org.springframework.cloud.stream.annotation.Output} annotation of the
+	 * argument.
+	 * @param observable an {@link Observable} that will be red
```
ilayaperumalg: typo: red

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ObservableSender.java
```diff
@@ -0,0 +1,37 @@
...
+	 * @return a single representing the result of an operation
```
ilayaperumalg: Might be good to have {@link Single}

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToFluxSenderArgumentAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Adapts an {@link org.springframework.cloud.stream.annotation.Output} annotated
+ * {@link FluxSender} to an outbound {@link MessageChannel}.
+ * @author Marius Bogoevici
+ */
+public class MessageChannelToFluxSenderArgumentAdapter
+		implements StreamListenerArgumentAdapter<FluxSender, MessageChannel> {
+
+	private Log log = LogFactory.getLog(FluxToMessageChannelResultAdapter.class);
```
ilayaperumalg: The logger class name: MessageChannelToFluxSenderArgumentAdapter

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ObservableToMessageChannelResultAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+
+import reactor.adapter.RxJava1Adapter;
+import reactor.core.publisher.Flux;
+import rx.Observable;
+
+import org.springframework.cloud.stream.binding.StreamListenerResultAdapter;
+import org.springframework.messaging.MessageChannel;
+
+/**
+ * A {@link StreamListenerResultAdapter} from a {@link Flux}
```
ilayaperumalg: Should it be `Observable` instead of `Flux` as it is `ObservableToMessageChannelResultAdapter` here.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+
+		@Bean
+		public MessageChannelToObservableSenderArgumentAdapter messageChannelToObservableSenderArgumentAdapter(
+				MessageChannelToFluxSenderArgumentAdapter messageChannelToFluxSenderArgumentAdapter) {
+			return new MessageChannelToObservableSenderArgumentAdapter(messageChannelToFluxSenderArgumentAdapter);
+		}
+
+		@Bean
+		ObservableToMessageChannelResultAdapter
+		observableMessageChannelObservableToMessageChannelResultAdapter(
```
ilayaperumalg: The method name (and the bean name) can be `observableToMessageChannelResultAdapter`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -60,7 +66,18 @@
 
 	private ConfigurableApplicationContext applicationContext;
 
-	public StreamListenerAnnotationBeanPostProcessor(DestinationResolver<MessageChannel> binderAwareChannelResolver, MessageHandlerMethodFactory messageHandlerMethodFactory) {
+	private final List<StreamListenerArgumentAdapter<?, Object>> streamListenerArgumentAdapters = new ArrayList<>();
+
+	private final List<StreamListenerResultAdapter<?, ?>> streamListenerResultAdapters = new ArrayList<>();
+
+	private int phase;
```
ilayaperumalg: The variables `phase`, `running` and `lifecycleMonitor` don't seem to be in use.

spring-cloud-stream-integration-tests/pom.xml
```diff
@@ -33,6 +33,12 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-java-dsl</artifactId>
+			<version>1.1.3.RELEASE</version>
```
ilayaperumalg: Any specific reason to explicitly specifying the version here?

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactorTests.java
```diff
@@ -0,0 +1,272 @@
...
+		context.close();
+	}
+
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class TestInputOutputArgs {
+
+		@StreamListener
+		public void receive(@Input("input") Flux<String> input, @Output("output") FluxSender output) {
```
ilayaperumalg: Might be good to use `Processor.INPUT` and `Processor.OUTPUT` to refer the message channel names`input` and `output`. If agreed, the same can be applied in all the test methods.

spring-cloud-stream-reactive/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-stream</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<version>3.0.0.BUILD-SNAPSHOT</version>
```
markfisher: should probably extract this to a property

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToFluxSenderArgumentAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+	public boolean supports(Class<?> boundElementType, MethodParameter methodParameter) {
+		ResolvableType type = ResolvableType.forMethodParameter(methodParameter);
+		return MessageChannel.class.isAssignableFrom(boundElementType) && FluxSender.class.isAssignableFrom(
+				type.getRawClass());
+	}
+
+	@Override
+	public FluxSender adapt(MessageChannel boundElement, MethodParameter parameter) {
+		return resultPublisher -> {
+			MonoProcessor<Void> sendError = MonoProcessor.create();
```
markfisher: maybe renamed the `sendError` var since it's used for completion or error

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxArgumentAdapter.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Adapts an {@link org.springframework.cloud.stream.annotation.Input} annotated
+ * {@link MessageChannel} to a {@link Flux}.
+ * @author Marius Bogoevici
+ */
+public class MessageChannelToInputFluxArgumentAdapter
+		implements StreamListenerArgumentAdapter<Flux<?>, SubscribableChannel> {
+
+	private CompositeMessageConverter messageConverter;
```
markfisher: should change to one of the following:

1) switch this to a constructor arg (with an assert that it's not null)
2) set a default instance here
3) assert that it's not null when it's used below (line 76)

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputObservableArgumentAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.binding.StreamListenerArgumentAdapter;
+import org.springframework.core.MethodParameter;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.SubscribableChannel;
+
+/**
+ * Adapts an {@link org.springframework.cloud.stream.annotation.Input} annotated
+ * {@link MessageChannel} to an {@link Observable}.
```
markfisher: could mention here that it does so by wrapping the Flux adapter and using Reactor's own adapter support

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputObservableArgumentAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Marius Bogoevici
+ */
+public class MessageChannelToInputObservableArgumentAdapter
+		implements StreamListenerArgumentAdapter<Observable<?>, SubscribableChannel> {
+
+	private final MessageChannelToInputFluxArgumentAdapter messageChannelToFluxArgumentAdapter;
+
+	public MessageChannelToInputObservableArgumentAdapter(
+			MessageChannelToInputFluxArgumentAdapter messageChannelToFluxArgumentAdapter) {
+		this.messageChannelToFluxArgumentAdapter = messageChannelToFluxArgumentAdapter;
```
markfisher: assert not null to be safe

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToObservableSenderArgumentAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+import rx.Single;
+
+import org.springframework.cloud.stream.binding.StreamListenerArgumentAdapter;
+import org.springframework.core.MethodParameter;
+import org.springframework.core.ResolvableType;
+import org.springframework.messaging.MessageChannel;
+
+/**
+ * Adapts an {@link org.springframework.cloud.stream.annotation.Output} annotated
+ * {@link ObservableSender} to an outbound {@link MessageChannel}.
```
markfisher: likewise here - could mention that it delegates to the Flux adapter and uses Reactor's own Rx adapter

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToObservableSenderArgumentAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+ * @author Marius Bogoevici
+ */
+public class MessageChannelToObservableSenderArgumentAdapter implements
+		StreamListenerArgumentAdapter<ObservableSender, MessageChannel> {
+
+	private final MessageChannelToFluxSenderArgumentAdapter messageChannelToFluxSenderArgumentAdapter;
+
+	public MessageChannelToObservableSenderArgumentAdapter(
+			MessageChannelToFluxSenderArgumentAdapter messageChannelToFluxSenderArgumentAdapter) {
+		this.messageChannelToFluxSenderArgumentAdapter = messageChannelToFluxSenderArgumentAdapter;
```
markfisher: assert not null to be safe

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ObservableToMessageChannelResultAdapter.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author Marius Bogoevici
+ */
+public class ObservableToMessageChannelResultAdapter
+		implements StreamListenerResultAdapter<Observable<?>, MessageChannel> {
+
+	private FluxToMessageChannelResultAdapter fluxToMessageChannelResultAdapter;
+
+	public ObservableToMessageChannelResultAdapter(
+			FluxToMessageChannelResultAdapter fluxToMessageChannelResultAdapter) {
+		this.fluxToMessageChannelResultAdapter = fluxToMessageChannelResultAdapter;
```
markfisher: assert not null to be safe

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxArgumentAdapter.java
mbogoevici: It is a constructor argument. We should add 'final' and the null assertion.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+				compositeMessageConverterFactory.getMessageConverterForAllRegistered());
+	}
+
+	@Bean
+	public MessageChannelToFluxSenderArgumentAdapter messageChannelToStreamSenderArgumentAdapter() {
+		return new MessageChannelToFluxSenderArgumentAdapter();
+	}
+
+	@Bean
+	FluxToMessageChannelResultAdapter subscriberStreamListenerResultHandler() {
```
markfisher: why is this one not public?

mbogoevici: oversight

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * @author Marius Bogoevici
+ */
+@Configuration
+public class ReactiveSupportAutoConfiguration {
+
+	@Bean
```
markfisher: should probably name these beans to match their class names as done with the RxJava beans nested here

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+		return new MessageChannelToFluxSenderArgumentAdapter();
+	}
+
+	@Bean
+	FluxToMessageChannelResultAdapter subscriberStreamListenerResultHandler() {
+		return new FluxToMessageChannelResultAdapter();
+	}
+
+	@Configuration
+	@ConditionalOnClass(Observable.class)
```
markfisher: since this is an optional dependency, would it be better to use the fully-qualified class here rather than an `import` above?

mbogoevici: Is that done elsewhere?

mbogoevici: We can use the qualified class name as a string though.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/StreamListener.java
```diff
@@ -43,10 +52,17 @@
  * <li>The value set on the {@link org.springframework.messaging.handler.annotation.SendTo} annotation, if present</li>
  * </ul>
  *
+ * <h3>Declarative mode</h3>
+ *
+ * The annotation may also have an empty value (the default), case in which the method is a
```
markfisher: something not quite right with ", case in which" (maybe you meant "for the case" or something similar?)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageChannelStreamListenerResultAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+	public boolean supports(Class<?> resultType, Class<?> boundElement) {
+		return MessageChannel.class.isAssignableFrom(resultType) && MessageChannel.class.isAssignableFrom(
+				boundElement);
+	}
+
+	@Override
+	public void adapt(MessageChannel streamListenerResult, MessageChannel boundElement) {
+		BridgeHandler handler = new BridgeHandler();
+		handler.setOutputChannel(boundElement);
+		((SubscribableChannel) streamListenerResult).subscribe(handler);
```
markfisher: do any lifecycle methods need to be invoked on the handler?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -70,6 +81,19 @@ public StreamListenerAnnotationBeanPostProcessor(DestinationResolver<MessageChan
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
 		this.applicationContext = (ConfigurableApplicationContext) applicationContext;
+		Map<String, StreamListenerArgumentAdapter> streamListenerArgumentHandlerMap =
+				this.applicationContext.getBeansOfType(
+						StreamListenerArgumentAdapter.class);
+		for (Map.Entry<String, StreamListenerArgumentAdapter> entry : streamListenerArgumentHandlerMap.entrySet()) {
```
markfisher: nitpick: why not just call `.values()` like you did in the `for` loop below?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageChannelStreamListenerResultAdapter.java
mbogoevici: As far as I can tell, the BridgeHandler is not Lifecycle. If we need to do that, I'd rather just create a vanilla forwarding MessageHandler. There's nothing active to this component. Maybe we should avoid returning MessageChannel completely, it is a bad idea anyway.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -85,48 +109,148 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
-								" mapping for '" + streamListener.value() + "' on " + invocableHandlerMethod.getShortLogMessage() +
-								" already existing for " + mappedBindings.get(streamListener.value()).getShortLogMessage());
+					Class<?>[] parameterTypes = method.getParameterTypes();
+					boolean hasInputOrOutputAnnotatedParameters = false;
+					for (int parameterIndex = 0; parameterIndex < parameterTypes.length; parameterIndex++) {
+						MethodParameter methodParameter = MethodParameter.forMethodOrConstructor(method,
+								parameterIndex);
+						boolean currentParameterIsAnnotated = methodParameter.getParameterAnnotation(
+								Input.class) != null || methodParameter.getParameterAnnotation(Output.class) != null;
+						if (hasInputOrOutputAnnotatedParameters) {
```
markfisher: I don't believe this would catch the case where the 1st param is *not* annotated even if all subsequent params are... instead maybe it should just iterate a counter and at the end of the loop assert that the count is either 0 or equal to the `parameterTypes.length`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
mbogoevici: We can do that.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
mbogoevici: ok.

mbogoevici: We should clean that up, anyway - as per the JavaDoc, the decision should be made based on whether `@StreamListener` has a value or not.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -85,48 +109,148 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+						}
+					}
+				}
+				if (arguments[parameterIndex] == null) {
+					throw new BeanInitializationException(
+							"Cannot convert argument " + parameterIndex + " of " + method + "from " + targetBean
+									.getClass().toString() + " to " + parameterType.toString());
+				}
+			}
+			else {
```
markfisher: would it be overkill to assert that no param has _both_ `@Input` and `@Output`?

mbogoevici: Probably not.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -85,48 +109,148 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
 		return bean;
 	}
 
+	private void invokeSetupMethodOnListenedChannel(Method method, Object bean) {
+		Object[] arguments = new Object[method.getParameterTypes().length];
+		for (int parameterIndex = 0; parameterIndex < arguments.length; parameterIndex++) {
+			MethodParameter methodParameter = MethodParameter.forMethodOrConstructor(method, parameterIndex);
+			Class<?> parameterType = methodParameter.getParameterType();
+			Input input = methodParameter.getParameterAnnotation(Input.class);
+			if (input != null) {
```
markfisher: I'm stumped as to why the if and else branch appear identical, since the argument adapters themselves must be directional

mbogoevici: Initially, they weren't. We can clean this up.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -85,48 +109,148 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+				Output output = AnnotationUtils.getAnnotation(method, Output.class);
+				if (output == null) {
+					throw new BeanInitializationException("A method returning a result must be annotated with " +
+							"@Output");
+				}
+				Object targetBean = this.applicationContext.getBean(output.value());
+				for (StreamListenerResultAdapter streamListenerResultAdapter : this
+						.streamListenerResultAdapters) {
+					if (streamListenerResultAdapter.supports(result.getClass(), targetBean.getClass())) {
+						streamListenerResultAdapter.adapt(result, targetBean);
```
markfisher: should this break out of the `for` loop once a supporting adapter is invoked?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -85,48 +109,148 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+			}
+		}
+		try {
+			if (method.getReturnType().equals(Void.TYPE)) {
+				method.invoke(bean, arguments);
+			}
+			else {
+				Object result = method.invoke(bean, arguments);
+				Output output = AnnotationUtils.getAnnotation(method, Output.class);
+				if (output == null) {
```
markfisher: since this is already wrapped as a `BeanInitializationException` in the `catch` block below, this could just be an `Assert.notNull`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerArgumentAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.springframework.cloud.stream.binding;
+
+import org.springframework.core.MethodParameter;
+
+/**
+ * Strategy for adapting a method argument type annotated with
+ * {@link org.springframework.cloud.stream.annotation.Input} or
+ * {@link org.springframework.cloud.stream.annotation.Output} from a bound element
+ * (e.g. {@link org.springframework.messaging.MessageChannel} supported by an
```
markfisher: nitpick: no closing parens

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerArgumentAdapter.java
```diff
@@ -0,0 +1,51 @@
...
+ * existing binder.
+ *
+ * This is a framework extension and is not primarily intended for use by end-users.
+ * @author Marius Bogoevici
+ */
+public interface StreamListenerArgumentAdapter<A, B> {
```
markfisher: based on the comments and usage, I'm actually wondering if this should be renamed to `StreamListenerParameterAdapter` ("parameter" instead of "argument")

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
mbogoevici: Also, the annotation where we get the target bean from is different - we can extract the common code from the branches though.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+				compositeMessageConverterFactory.getMessageConverterForAllRegistered());
+	}
+
+	@Bean
+	public MessageChannelToFluxSenderArgumentAdapter messageChannelToFluxSenderArgumentAdapter() {
+		return new MessageChannelToFluxSenderArgumentAdapter();
+	}
+
+	@Bean
+	FluxToMessageChannelResultAdapter fluxToMessageChannelResultAdapter() {
```
markfisher: should be public

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		@Bean
+		public MessageChannelToObservableSenderArgumentAdapter messageChannelToObservableSenderArgumentAdapter(
+				MessageChannelToFluxSenderArgumentAdapter messageChannelToFluxSenderArgumentAdapter) {
+			return new MessageChannelToObservableSenderArgumentAdapter(messageChannelToFluxSenderArgumentAdapter);
+		}
+
+		@Bean
+		ObservableToMessageChannelResultAdapter
```
markfisher: should be public

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageChannelStreamListenerResultAdapter.java
markfisher: I think it's ok for now... but the `BridgeHandler` does implement `InitializingBean`

mbogoevici: Ok.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -70,6 +82,20 @@ public StreamListenerAnnotationBeanPostProcessor(DestinationResolver<MessageChan
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
 		this.applicationContext = (ConfigurableApplicationContext) applicationContext;
+		Map<String, StreamListenerParameterAdapter> streamListenerArgumentHandlerMap =
```
markfisher: nitpick: this var still has "argument" in the name

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/08d65a92bb76591e7e88047f58e9ed0fe329584c

PR #617, opened 2016-08-16
Fix copy-paste-o

dturanski:
change STANDARD_REQUEST_HEADERS to STANDARD_REPLY_HEADERS where appropriate


mbogoevici:
This should be applied to the 1.0.x branch only - for 1.1 we have separate binder implementations (and docs) - see https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/pull/5

markfisher:
LGTM; merged

PR #622, opened 2016-08-17
Extract `spring-cloud-stream-dependencies`

### LINKED ISSUE #621: Externalize `spring-cloud-stream-dependencies`
As a developer, I want that `spring-cloud-stream-dependencies` contains binder implementation versions, so I can have all dependencies to my project managed in one place.

mbogoevici:
Fixes #621


pom.xml
```diff
@@ -20,21 +20,82 @@
...
-				<artifactId>spring-cloud-stream-dependencies</artifactId>
+				<artifactId>spring-cloud-stream</artifactId>
 				<version>1.1.0.BUILD-SNAPSHOT</version>
-				<type>pom</type>
-				<scope>import</scope>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream-build-tools</artifactId>
+				<version>1.1.0.BUILD-SNAPSHOT</version>
```
markfisher: should these all use `${project.version}`?

mbogoevici: No, because we'd want to use it for binders so that they get some dependencies managed like that - e.g. as parent. Having `${project.version}` there would mean that if the child has a different version, that value will be used as a placeholder.

mbogoevici: Release plugins replace these values automatically anyway.

markfisher: ok

markfisher:
LGTM; rebased and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/ac443b56bc03d3b84cda5dd9f55b195a82020a1c

PR #623, opened 2016-08-17
Set an explicit context class for the binders

mbogoevici:
Since this is an internal construct, we can safely ignore
the application settings.


markfisher:
LGTM; merged

PR #625, opened 2016-08-18
Correct example in the documentation

### LINKED ISSUE #586: Confusing example code in the reference doc
In here: http://cloud.spring.io/spring-cloud-static/spring-cloud.html#_introducing_spring_cloud_stream

This code example:

```
@SpringBootApplication
public class StreamApplication {

    SpringApplication.run(StreamApplication.class, args);
  }
}

@EnableBinding(Sink.class)
public class TimerSource {

  ...

  @StreamListener(Sink.INPUT)
  public void processVote(Vote vote) {
      votingService.recordVote(vote);
  }
}
```

So what is this? Is it a source? It's name is `TimerSource` which suggests it is some kind of source of time data. But.... its annotated with `@EnableBinding(Sink.class)` suggesting instead it is a sink. And then it has a method there for processing votes. I'm left wondering... what has this to do with either a source or a sink of time data?

That is all really very confusing to someone reading the docs for the very first time.

mbogoevici:
Fixes #586

### DIFF
--- a/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -15,17 +15,12 @@ The following is a simple sink application which receives external messages.
 [source,java]
 ----
 @SpringBootApplication
-public class StreamApplication {
+@EnableBinding(Sink.class)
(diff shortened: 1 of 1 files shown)


markfisher:
LGTM; rebased and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/bac8877a524042cac993402fb8d30dd9c264b82e

PR #632, opened 2016-08-24
Do not start application context automatically

### LINKED ISSUE #525: Do not start the context automatically

ilayaperumalg:
- This will avoid other lifecycle beans inside the application context starting when
binding lifecycle beans are started.

Resolves #525

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -387,6 +387,12 @@ public class SendingBean {
 }
 ----
 
+===== Spring application context for Spring Cloud Stream applications
+
+Spring cloud stream application will not start its application context automatically when the binding lifecycle beans
+are started.This is to avoid any undesirable effects on starting other beans inside the application context when those
+beans are intended to get started in a controlled manner.
+
 ==== Producing and Consuming Messages
 
 You can write a Spring Cloud Stream application using either Spring Integration annotations or Spring Cloud Stream's `@StreamListener` annotation.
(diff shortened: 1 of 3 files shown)


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -387,6 +387,12 @@ public class SendingBean {
 }
 ----
 
+===== Spring application context for Spring Cloud Stream applications
```
mbogoevici: Not sure if we need to add this as a documentation feature. Probably part of the upgrade guide.

mbogoevici:
This looks incomplete - `org.springframework.cloud.stream.binder.LifecycleBinderTests` specifically tests that a regular lifecycle is started. If it passes, it means that the issue is not fixed. If it doesn't pass, it needs to be removed.

ilayaperumalg:
It looks like the test `testNonSmartLifecyclesStarted` is incorrectly asserts [here] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/LifecycleBinderTests.java#L42) and after modifying it to use `assertThat(simpleLifecycle.isRunning()).isTrue();` the test fails as expected :-) 

Do you think we can still keep this test with the assertion on application context not started for the explicit lifecycle bean?

mbogoevici:
Oh, that must have been broken during migration to AssertJ - we might want to keep that around, and modify the assert to False.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
ilayaperumalg: I am not sure where is `upgrade` guide though ;)

mbogoevici:
We should add a section with 'What's new in 1.1' under intro.

sobychacko:
Merged.

PR #634, opened 2016-08-24
Restore backwards compatibility with applications using input.contentType

### LINKED ISSUE #633: Applications specifying java types as input contentType cannot deserialize content
While specifying a Java type as input should be deprecated in favour of `StreamListener`, this breaks backwards compatibility when using:

```
spring.cloud.stream.bindings.input.content-type=application/x-java-object;type=some.Type
```

as no converter is found.

We should document this as deprecated but not break backwards compatibility outright.

mbogoevici:
Fixes #633


sobychacko:
rebased, fixed the checkstyle issues, squashed the commits - merged.

PR #637, opened 2016-08-26
Reinstate no-arg constructor for AggregateApplicationBuilder

### LINKED ISSUE #636: Reinstate no-arg constructor for `AggregateApplicationBuilder`
Removed in 1.1.0.M1.

Current workaround (and recommended approach anyway) is to pass the application class as argument. 

```
new AggregateApplicationBuilder(DoubleApplication.class, args)
				.from(SourceApplication.class).args("--fixedDelay=5000")
				.via(ProcessorApplication.class)
				.to(SinkApplication.class).args("--debug=true").run();
```

mbogoevici:
Fixes #636

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
@@ -49,6 +49,10 @@ public class AggregateApplicationBuilder {
 
 	private ConfigurableApplicationContext parentContext;
 
+	public AggregateApplicationBuilder(String... args) {
+		this(new Object[]{ParentConfiguration.class}, args);
+	}
+
 	public AggregateApplicationBuilder(Object source, String... args) {
 		this(new Object[]{source}, args);
 	}


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -49,6 +49,10 @@
 
 	private ConfigurableApplicationContext parentContext;
 
+	public AggregateApplicationBuilder(String... args) {
```
ilayaperumalg: Since the issue says no-arg constructor, does this PR require changes to include the constructor public AggregateApplicationBuilder() as well?

mbogoevici:
Pushed rebased version.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
mbogoevici: This constructor would also function as a no-arg constructor if we don't pass any `String` arguments.

ilayaperumalg:
LGTM, merging.

PR #643, opened 2016-09-07
Pass aggregator parent sources/args appropriately

### LINKED ISSUE #626: `run(args)` from `AggregateApplicationBuilder` has no effect
The intent for the method was to support passing of arguments to the parent context being created.

Currently, though, all constructors of `AggregateApplicationBuilder` will create the context eagerly so by the time the method is invoked, the arguments passed in are ignored.

One option is to defer the creation of the parent context until the method is invoked and use the supplied arguments (possibly merge with arguments passed from the constructor).

ilayaperumalg:
- The aggregator builder parent sources and cmdline args can be passed via
   - AggregateApplicationBuilder constructor
   - AggregateApplicationBuilder parent() builder method
and, the cmd line args can also be passed via
   - AggregateApplicationBuilder run() method

This PR supports passing all these options. The implementation sets `parentSources` and `parentArgs` by collecting the values from all the above options and use them when creating the parent context.

 - Add test case to verify these scenarios

This resolves #626

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -43,10 +43,11 @@ abstract class AggregateApplication {
 
 	public static final String OUTPUT_CHANNEL_NAME = "output";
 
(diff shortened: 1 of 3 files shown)


sobychacko:
LGTM - rebased & merged.

Issue #571, opened 2016-06-09
Property prefixes for aggregation (via namespaces)


mbogoevici:
Aggregation should be able to define property prefixes either for command-line arguments or for the environment. Right now passing arguments to the applications themselves is a manual process. At a first pass, namespaces (which currently only help with disambiguating channels) can fufill that role.

e.g. when running application `A` expecting property `foo` and application `B` expecting property `bar`, for:

```
return new AggregateApplicationBuilder(Aggregator.class)
				.from(SourceApplication.class).namespace("a")
				.to(SinkApplication.class).namespace("b")
				.run(args);
```

using CLI arguments such as

```
--a.foo=baz --b.bar=qux
```

or environment variables such as:

```
A_FOO=baz
B_BAR=qux
```

should result in the properties `foo` and `bar` being set to `baz` and `qux` on the respective applications.

(The benefit of using of the namespace becomes more obvious when both applications expect the same property, e.g. `--a.foo=bar --b.foo=baz`)

Note: we should take special care for existing applications that do not support this type of behaviour and might be taken by surprise by the introduction of this feature.

mbogoevici:
Fixed w/https://github.com/spring-cloud/spring-cloud-stream/commit/6fc6445c4fa604e7894d7da1b2a7be75992171a5

PR #645, opened 2016-09-08
Support namespace based prefixed properties

ilayaperumalg:
- In AggregateApplication, support specifying commandline and environment properties based on namespace for each containing applications using namespace as prefix
  - As usual, command line properties override the environment as well as the properties specified inside the application

Resolves spring-cloud/spring-cloud-stream#571

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
@@ -18,8 +18,9 @@ package org.springframework.cloud.stream.aggregate;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
This PR is based on top of the PR #643. The commit https://github.com/spring-cloud/spring-cloud-stream/pull/645/commits/8f939c0d3baae99e9075c26985d8c6fcfb1da5ba addresses the story for #571.

ilayaperumalg:
Rebased.

mbogoevici:
Here's an attempt to simplify the processing logic: https://github.com/mbogoevici/spring-cloud-stream/commit/ae5899ade4de53420d269296e91b5097bdbedc8b

ilayaperumalg:
@mbogoevici The above approach of directly using the `RelaxedDataBinder`  would require additional efforts for the following:

1) To be able to override the properties set from various property sources based on their precedence level. With the suggested simplification, we would pass all the properties that are matching the namespace. There is neither filtering nor order of precedence handling.

2) Since we update all the **extracted** properties as `command-line` property args for the child application, we also need to modify the environment property into appropriate lower case as the upper case environment properties that are converted into `command-line` properties wouldn't work always. (for example passing --FIXED_DELAY=<?> wouldn't set @Value annotated with fixed-delay)

In this commit https://github.com/ilayaperumalg/spring-cloud-stream/commit/a1eaf2b2531a33ce501c76519a81e293c441d07e I try to fix the property overriding capability on top of your changes. But, I am still investigating the case sensitiveness issue (from environment -> commandline) with the use of `RelaxedDataBinder`.

ilayaperumalg:
This change https://github.com/ilayaperumalg/spring-cloud-stream/commit/a244c614ea4a8f86ce728f5eb0493119fb48cd8a addresses the issue with handling environment properties as commandline properties.

mbogoevici:
Maybe I am missing something, but for 1) isn't the ordering of the PropertySources already taking care of precedence ?

ilayaperumalg:
The ordering is taken care only within the context of `RelaxedDataBinder`. But, we need to do 1) get the first matching arg that is derived based on the precedence inside the `RelaxedDataBinder` 2) consider the args passed via Child appConfigurer args

mbogoevici:
@ilayaperumalg 

It's still not clear for me what '1) get the first matching arg (not all the matching args) that is derived based on the precedence inside the RelaxedDataBinder' means in this context. Deferring to the `RelaxedDataBinder` should take care of the order of precedence already (because the component takes care of it implicitly). What am I missing?

Number 2) seems like something that we can handle by manipulating the property sources before we pass them to the `RelaxedDataBinder`.

In order to move this forward, can you provide a detailed description of the order in which you think parameters should be bound as well as tests that illustrate that?

ilayaperumalg:
> get the first matching arg (not all the matching args) that is derived based on the precedence inside the RelaxedDataBinder

This happens when the same properties are specified with different relaxed names from multiple property sources. For instance, the command line property `--<namespace>.fooValue=test` and the environment property as `<namespace>_FOO_VALUE=envtest`, then the RelaxedBinder target with the namespace gets both `fooValue=test` and `FOO_VALUE=envtest` as the result.
Hence, we need to get the first matching (which is from high precedence) property.

> seems like something that we can handle by manipulating the property sources before we pass them to the RelaxedDataBinder

Matching the args properties with the RelaxedDataBinder target can be effective as we need to match with the **relaxed** names as well.

Added some tests to depict these use cases. Also, rebased and updated the commits.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/6fc6445c4fa604e7894d7da1b2a7be75992171a5

PR #650, opened 2016-09-15
POM changes to use version name effectively

ilayaperumalg:


mbogoevici:
Let's hold on to this, until we do the RELEASE. The binders use `spring-cloud-stream-parent` parent as parent, and this change will cause them to pull in the wrong dependencies (as they set different `${project.version}` values. Sharing the parent may not be correct, but we shouldn't make a major change there between RELEASE and RC.

ilayaperumalg:
Pushed a change that doesn't use `parent` version inside the sub project.

mbogoevici:
We can reconsider this once https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/pull/21 and https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/61 are merged.

mbogoevici:
Squashed, merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/580e268a97853d26280faf0aa7147350a5322c6d

PR #652, opened 2016-09-16
Minor polishing on message converter doc

### LINKED ISSUE #616: Documentation for MessageConverters
There is a reference to "the MessageConverter mechanism" in the user guide but no details on how to use it, customize it etc. E.g. what if I want to add a new converter?

ilayaperumalg:
This fixes #616

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1109,10 +1109,9 @@ This allows you to send and receive data in a variety of custom formats, includi
 Spring Cloud Stream registers all the beans of type `org.springframework.messaging.converter.MessageConverter` as custom message converters along with the out of the box message converters.
 
 If your message converter needs to work with specific `content-type` and the target class (for both input and output), then the message converter needs to extend `org.springframework.messaging.converter.AbstractMessageConverter`.
-As an example, You can refer link:https://github.com/spring-projects/spring-framework/blob/master/spring-messaging/src/main/java/org/springframework/messaging/converter/StringMessageConverter.java[String message converter].
 For the conversion using @StreamListener, the message converter that implements `org.springframework.messaging.converter.MessageConverter` would be suffice.
 
-Here is an example of creating a message converter bean inside the application:
+Here is an example of creating a message converter bean (with the content-type application/bar) inside the Spring Cloud Stream application:
 
 [source,java]
 ----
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1106,8 +1106,24 @@ For the conversion of inbound messages, especially when the target is a POJO, th
 
 Besides the conversions that it supports out of the box, Spring Cloud Stream also supports registering your own message conversion implementations.
 This allows you to send and receive data in a variety of custom formats, including binary, and associate them with specific `contentTypes`.
-In order to do so, you can create a class that extends `AbstractMessageConverter`
+In order to do so, you can create a class of type `org.springframework.messaging.converter.MessageConverter` and make it available in the application context.
+Spring Cloud Stream registers all the beans of type `org.springframework.messaging.converter.MessageConverter` as custom message converters along with the out of the box message converters.
 
+Here is an example of creating message converter bean inside the application:
```
markfisher: "creating *a* message converter bean.."

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1106,8 +1106,24 @@ For the conversion of inbound messages, especially when the target is a POJO, th
 
 Besides the conversions that it supports out of the box, Spring Cloud Stream also supports registering your own message conversion implementations.
 This allows you to send and receive data in a variety of custom formats, including binary, and associate them with specific `contentTypes`.
-In order to do so, you can create a class that extends `AbstractMessageConverter`
+In order to do so, you can create a class of type `org.springframework.messaging.converter.MessageConverter` and make it available in the application context.
```
ilayaperumalg: ok, I think this also means we need to use `AbstractMessageConverter` to get the list of custom message converters instead of just looking for the type `MessageConverter` list in the application context as [here] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceConfiguration.java#L91)

mbogoevici: Not really. All `MessageConverters` are registered because there can be some that have logic that is not based off `contentType` - e.g. `ConversionService` and so on. Those can be used in `@StreamListener` dispatching. But only `MessageConverters` that extend `AbstractMessageConverters` can participate in scenarios where `contentType` is specified - e.g. `contentType`-based conversions on channels.

markfisher: I believe that itself should be documented clearly (the role of `@StreamListener` as compared-to/combined-with `content-type`)

markfisher:
LGTM; rebased squashed and pushed as https://github.com/spring-cloud/spring-cloud-stream/commit/0d8bbc6665b33893adda6c2c7521b312df4ff6ba along with https://github.com/spring-cloud/spring-cloud-stream/commit/15b562132acd338ebaeeb3d90b28d7bcbecef3ee

PR #657, opened 2016-09-19
Allow instanceIndex and instanceCount to be set per binding

### LINKED ISSUE #654: Cannot set partitioning properties on the consumer
In theory, we should be able to customize the `instanceIndex` and `instanceCount` on the consumer, but in practice, they're always overriden by `spring.cloud.stream.instanceIndex` and `spring.cloud.stream.instanceCount`

mbogoevici:
Fixes #654


garyrussell:
Rebased, merged as d11b3cf08db4a10b8740401b20203f6355bb6b56

Issue #661, opened 2016-09-20
Aggregated applications do not start a web endpoint anymore


mbogoevici:
Seems to be a regression since RC1

PR #662, opened 2016-09-21
Expose channel metrics for aggregate applications

### LINKED ISSUE #660: Aggregate application does not expose metrics of the component apps
We have an aggregate application configured as follows:

```java
new AggregateApplicationBuilder(OurApplication.class, args)
                .from(OurSourceApplication.class).args(args)
                .to(OurDestinationApplication.class).args(args)
                .namespace("our-app")
                .run(args);
```

We would be interested in querying the number of messages consumed and produced by the second app of the aggregate (`OurSourceDestinationApplication`). However, the `/metrics` endpoint of the application deployed is only exposing properties of the aggregate app, ignoring the components, so, for example, we can't see the value for the property `integration.channel.output.sendCount`, which would be extremely useful for debugging.

mbogoevici:
Fixes #660 #661

* Fix an issue where aggregated applications were not enabling web endpoints
* Add spring-integration-jmx as mandatory dependency, so that SCSt applications
  expose metrics by default;
* Remove `AggregatorParentConfiguration` as redundant
* Register a `PublicMetrics` instance for each aggregate, using a customized version
  of SpringIntegrationMetricReader


ilayaperumalg:
LGTM

ilayaperumalg:
Also, the `/metrics` endpoint provides the channel metrics from the applications contained inside the aggregate application.

ilayaperumalg:
Merged as 8108038.

ilayaperumalg:
I believe this change needs to be back ported to `1.0.x` as well. @mbogoevici wdyt?

spring-cloud-stream/pom.xml
```diff
@@ -35,6 +35,10 @@
 			<artifactId>spring-integration-core</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-jmx</artifactId>
```
artembilan: If you don't like this dependency, consider to use `@EnableIntegrationManagement`: http://docs.spring.io/spring-integration/reference/html/system-management-chapter.html#_configuring_metrics_capture

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/NamespaceAwareSpringIntegrationMetricReader.java
```diff
@@ -0,0 +1,104 @@
...
+				new Metric<Double>(name + ".stdev", statistic.getStandardDeviation()));
+		metrics.add(new Metric<Long>(name + ".count", statistic.getCountLong()));
+		return metrics;
+	}
+
+	@Override
+	public long count() {
+		int totalChannelCount = this.exporter.getChannelCount() * 11;
+		int totalHandlerCount = this.exporter.getHandlerCount() * 5;
+		return totalChannelCount + totalHandlerCount + 4;
```
artembilan: What is the logic behind this method?
Thanks

mbogoevici: That's the original method was in `SpringIntegrationMessageReader` and I copied it verbatim.

spring-cloud-stream/pom.xml
mbogoevici: Yes, but the metrics endpoint is conditional on JMX in Boot so for now we need to have it.

mbogoevici: On the long run, I'd prefer if the Boot feature would not require JMX, but that's something we need to address in Boot.

mbogoevici:
@ilayaperumalg Let's backport only if really necessary. I'd like to encourage folks to upgrade to 1.1 sooner rather than later.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/NamespaceAwareSpringIntegrationMetricReader.java
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	@Override
+	public Metric<?> findOne(String metricName) {
+		return null;
+	}
+
+	@Override
+	public Iterable<Metric<?>> findAll() {
+		IntegrationMBeanExporter exporter = this.exporter;
```
artembilan: No-op. I think that was just blink copy-paste: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/integration/SpringIntegrationMetricReader.java#L49 :smile: .

Here is the original change: https://github.com/spring-projects/spring-boot/commit/31d6a0f17a66e5342b9d2337cffdbcd3c6f2b6cd.
I guess that was just quick fix of `RequireThis` Checkstyle rule violation :smile:

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/NamespaceAwareSpringIntegrationMetricReader.java
```diff
@@ -0,0 +1,104 @@
...
+		metrics.add(new Metric<Double>(name + ".min", statistic.getMin()));
+		metrics.add(
+				new Metric<Double>(name + ".stdev", statistic.getStandardDeviation()));
+		metrics.add(new Metric<Long>(name + ".count", statistic.getCountLong()));
+		return metrics;
+	}
+
+	@Override
+	public long count() {
+		int totalChannelCount = this.exporter.getChannelCount() * 11;
```
artembilan: The original code is here: https://github.com/spring-projects/spring-boot/commit/5ceb35473d656e086aa13fbc433df003377a4819#diff-d592286eb2093acb9f86af6d5d0dcdc8R78.

I think there is some shift logic to distinguish different counts in a single value.
Or just classical @dsyer 's magic :smile:

PR #674, opened 2016-10-05
Update doc for Spring Initializr

### LINKED ISSUE #673: Hard-coded Spring boot version in reference manual
_From @snicoll on July 22, 2016 11:29_

I was reading [this section](http://docs.spring.io/spring-cloud-stream/docs/1.0.2.RELEASE/reference/htmlsingle/index.html#_getting_started) of the doc and stumbled upon the following:

> To get started with creating Spring Cloud Stream applications, visit the Spring Initializr and create a new Maven project named "GreetingSource". Select Spring Boot version 1.3.4 SNAPSHOT

`start.spring.io` offers the latest and greatest so that version isn't available anymore. Maybe that section should be rephrased along the lines of "select at least Spring Boot 1.3"?

_Copied from original issue: spring-cloud/spring-cloud-dataflow#803_

ilayaperumalg:
Resolves #673

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1539,7 +1539,7 @@ For Spring Cloud Stream samples, please refer to the https://github.com/spring-c
 == Getting Started
 
 To get started with creating Spring Cloud Stream applications, visit the https://start.spring.io[Spring Initializr] and create a new Maven project named "GreetingSource".
-Select Spring Boot version 1.3.4 SNAPSHOT and search or tick the checkbox for Stream Kafka (we will be using Kafka for messaging).
+Select the Spring Boot version that works for you and search or tick the checkbox for Stream Kafka (we will be using Kafka for messaging).
 
 Next, create a new class, `GreetingSource`, in the same package as the `GreetingSourceApplication` class.
 Give it the following code:


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1539,7 +1539,7 @@ For Spring Cloud Stream samples, please refer to the https://github.com/spring-c
 == Getting Started
 
 To get started with creating Spring Cloud Stream applications, visit the https://start.spring.io[Spring Initializr] and create a new Maven project named "GreetingSource".
-Select Spring Boot version 1.3.4 SNAPSHOT and search or tick the checkbox for Stream Kafka (we will be using Kafka for messaging).
+Select the Spring Boot version that works for you and search or tick the checkbox for Stream Kafka (we will be using Kafka for messaging).
```
markpollack: We should say only 1.4.x and 1.3.x since there are options for 2.0 snapshot and 1.5 snapshot and we may not work with those versions.

markpollack:
Modified the text a bit and commited as 1e8434d32e5584050214c89715e4e9e21a8fd512

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
markpollack: So 
`:spring-boot-supported-versions: 1.3.x and 1.4.x`

and

Note that Spring Cloud Stream applications are known to work with Spring Boot versions {spring-boot-supported-versions}

?

ilayaperumalg: yeah, I will give a try to generate this property value from the current `spring-boot` version that the project uses and say something like `:spring-boot-supported-versions: starting from 1.3.x to <currently-used-version>`

PR #675, opened 2016-10-05
Fix NPE for wildcard/generic type reactive stream listener @Input type

### LINKED ISSUE #664: Align `@StreamListener` mapping for reactive and non-reactive types
Currently there are differences between how `@StreamListener` should be mapped in the event-by-event case vs when bindable types are. At a minimum, we'd want both following declarations to work:

```
@SendTo(Processor.OUTPUT)
public Flux<String> toUpper(Flux<String> input)
```

and

```
@StreamListener
@Output(Processor.OUTPUT)
public String toUpper(@Input(Processor.INPUT) String input)
```

NOTE: as part of the previous discussions, we deemed that the second use case will not be supported.

### LINKED ISSUE #665: Using generic types and wildcards in Reactive StreamListeners yields an NPE
When the concrete type of the argument cannot be inferred - in that case we can safely assume `Object` I believe.

ilayaperumalg:
- Set argumentClass to Object when using wildcard/generic types if the StreamListener method parameter for @Input annotated type uses wildcard/generic type, then set the argumentClass to `Object.class` instead of null.

Resolves #665

### DIFF
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxParameterAdapter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxParameterAdapter.java
@@ -33,6 +33,7 @@ import org.springframework.util.Assert;
  * Adapts an {@link org.springframework.cloud.stream.annotation.Input} annotated
  * {@link MessageChannel} to a {@link Flux}.
  * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
  */
 public class MessageChannelToInputFluxParameterAdapter
 		implements StreamListenerParameterAdapter<Flux<?>, SubscribableChannel> {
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/MessageChannelToInputFluxParameterAdapter.java
```diff
@@ -71,7 +73,9 @@ public boolean supports(Class<?> boundElementType, MethodParameter methodParamet
 			return Flux.create(emitter -> {
 				MessageHandler messageHandler = message -> {
 					synchronized (monitor) {
-						if (argumentClass.isAssignableFrom(message.getPayload().getClass())) {
+						if (argumentClass.equals(Object.class)
```
mbogoevici: `argumentClass` may not always be `Object.class`, but it seems like this condition is already covered by the second part of the expression (i.e. if true, `argumentClass` it would always be assignable from anything)

ericbottard: Err, yeah, sorry, this is what I meant.

sobychacko:
Reviewed, LGTM - rebased, squashed & merged.

PR #681, opened 2016-10-07
Align StreamListener method/parameter mappings

### LINKED ISSUE #664: Align `@StreamListener` mapping for reactive and non-reactive types

ilayaperumalg:
- Use the same mapping of method/parameters for both declarative and message handler Stream Listeners
  - Declarative mode of the StreamListener method is determined if at least one of the the method parameters is annotated with @Input or @Output with either bound elements (e.g. channels) or conversion targets from bound elements via a registered StreamListenerParameterAdapter
  - If the method is non-declarative then it is considered to be in message handler mode
  - Declarative mode now accepts @Output annotation at method level as well along with @SendTo
  - Declarative mode also accepts @Input annotation value at the method level via @StreamListener valuue.
  - Message handler mode accepts @Input annotation value at the method parameter level and @Output annotation at the method annotation level
  - Both @Output and @SendTo annotations are supported to specify the outbound target value while @SendTo is allowed only as a method level annotation
  - Add assertions on allowable use of method and parameter annotations
  - Add parameterized tests to cover all possible cases
    - Message handler tests
    - Reactor and RxJava tests

Resolves #664


spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactorTests.java
```diff
@@ -165,7 +334,17 @@ public void receive(@Input(Processor.INPUT) Flux<String> input, @Output(Processo
 
 	@EnableBinding(Processor.class)
 	@EnableAutoConfiguration
-	public static class TestInputOutputArgsWithMessage {
+	public static class TestInputOutputArgs2 {
+
+		@StreamListener(Processor.INPUT)
```
mbogoevici: I don't believe we should support this - and all the other cases that involve a mix of parameters that are annotated with `@Input`/`@Output` and parameters that are not annotated - either all of them should be annotated, or none (that would really only allow a single parameter in practice). In my opinion, having a mix of parameters is too hard to follow.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/StreamListener.java
```diff
@@ -33,12 +33,17 @@
  * the method is invoked and how their return results are processed. This annotation
  * can be applied for two separate classes of methods.
  *
+ * <h3>Declarative mode</h3>
+ *
+ * Method is considered as declarative if at least one of the method parameters is annotated with {@link Input}
```
mbogoevici: See my other comment about annotating parameters: I don't think we should accept cases when there is a mix of annotated and not annotated parameters.

mbogoevici: Also, as a writing style, use  "Methods that"  or "A method is considered ...".

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactorTests.java
ilayaperumalg: ok, that would simplify things. My intention was to support all possible combinations of method annotations and parameter annotations.

ilayaperumalg: On a second thought, I think it would be a convenient for the users to specify the binding targets in whatever places that it actually matters. For instance, user knows either `@StreamListener` or the method parameter with `@Input` accepts the inbound target and they can specify wherever. We already cover some checks if they specify in both the places etc., This way, we don't have to restrict the user to follow a specific path to specify the targets while those way of specifying would still look correct logically.  What do you think?

ilayaperumalg: I will have some changes that avoid the ambiguities you pointed out.

mbogoevici: Can you clarify your comment a bit?

>  I think it would be a convenient for the users to specify the binding targets in whatever places that it actually matters.

I specifically disagree with:

>For instance, user knows either @StreamListener or the method parameter with @Input accepts the inbound target and they can specify wherever.

I am specifically concerned about cases where inputs and outputs have the same type, e.g. `MessageChannel` - so it is impossible to tell whether the parameter is an input or an output - ordering shouldn't be a criterion. 

To me, it would be much clearer to the user if we supported only this:

a) `@StreamListener`/`@SendTo` without any `@Input`/`@Output` annotations wherever there's a single input/single or no output use case. This would align individual messages and bindable types.

b) `@StreamListener` + `@Input`/`@Output` to cover all the cases where we use bindable types. We could use `SendTo` as an alias for `@Output` in that case. 

I don't see any feature loss, but a lot fewer corner cases and idiomatic subtleties for us to deal with in the future.

ilayaperumalg:
@mbogoevici Pushed changes that exclude the patterns that cause ambiguity (especially those that are not distinguishable when using inside the method parameters). Added tests for error scenarios. The tests are separated out into multiple classes now.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,190 @@
...
+		catch (BeanCreationException e) {
+			assertThat(e.getCause().getMessage()).contains("@Input annotation can not be used in multiple method parameters");
+		}
+	}
+
+	@Test
+	public void testMethodWithOutputAsMethodAndParameter() throws Exception {
+		try {
+			SpringApplication.run(TestMethodWithOutputAsMethodAndParameter.class, "--server.port=0");
+			fail("Exception expected on using Input element as method and parameter");
```
mbogoevici: The error message is referring to input, but it is testing output.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,190 @@
...
+		}
+		catch (BeanCreationException e) {
+			assertThat(e.getCause().getMessage()).startsWith("Cannot set both Output (@Output/@SendTo) method annotation value ");
```
mbogoevici: This is a confusing error message.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,190 @@
...
+			SpringApplication.run(TestMethodWithoutInput.class, "--server.port=0");
+			fail("Exception expected when inbound target is not set");
+		}
+		catch (BeanCreationException e) {
+			assertThat(e.getCause().getMessage()).startsWith("Either StreamListener or a method parameter should be set with ");
+		}
+	}
+
+	@Test
+	public void testMethodWithMultipleInputParameters() throws Exception {
```
mbogoevici: This test is wrong altogether. We should allow multiple input parameters.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,190 @@
...
+			SpringApplication.run(TestMethodWithMultipleInputParameters.class, "--server.port=0");
+			fail("Exception expected on multiple @Input method parameters");
+		}
+		catch (BeanCreationException e) {
+			assertThat(e.getCause().getMessage()).startsWith("@Input annotation can not be used in multiple method parameters");
+		}
+	}
+
+	@Test
+	public void testMethodWithMultipleOutputParameters() throws Exception {
```
mbogoevici: Another wrong test. A method using `@Input`/`@Output` can have an arbitrary number of inputs and outputs. What prompted this change?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,190 @@
...
+	@EnableBinding({Sink.class, AnotherSink.class})
+	@EnableAutoConfiguration
+	public static class TestMethodWithMultipleInputParameters {
+
+		@StreamListener
+		public void receive(@Input(Processor.INPUT) FooPojo fooPojo, @Input(AnotherSink.INPUT) FooPojo fooPojo1) {
+		}
+	}
+
+	public interface AnotherSink {
```
mbogoevici: We should consolidate around a set of classes that are used consistently during the test.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,190 @@
...
+
+
+	@Test
+	public void testMethodWithInputAsMethodAndParameter() throws Exception {
+		try {
+			SpringApplication.run(TestMethodWithInputAsMethodAndParameter.class, "--server.port=0");
+			fail("Exception expected on using Input element as method and parameter");
+		}
+		catch (BeanCreationException e) {
+			assertThat(e.getCause().getMessage()).contains("@Input annotation can not be used in multiple method parameters");
```
mbogoevici: This needs to be tested as a supported use case.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
mbogoevici: This needs to be tested as a supported use case.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
mbogoevici: This needs to be tested as a supported use case.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestAnnotatedArguments.java
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class TestPojoWithAnnotatedArguments2 extends TestPojoWithAnnotatedArguments {
+
+		@StreamListener
+		public void receive(@Input(Processor.INPUT) @Payload FooPojo fooPojo,
+				@Headers Map<String, Object> headers,
```
mbogoevici: Why is this a valid use case? How does the mapping what message it should get the headers from? I think we might want to ditch supporting `@Input` on messages - due to the explosion of things we should support afterwards. The initial idea of having it supported was ill-considered IMO.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
ilayaperumalg: The actual message is `"Cannot set both Output (@Output/@SendTo) method annotation value " + outboundBindingName +
					" and @Output annotation as a method parameter"`.

For testing purpose, I added the content with `startsWith`

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	@Test
+	public void testMethodWithoutInput() throws Exception {
+		try {
+			SpringApplication.run(TestMethodWithoutInput.class, "--server.port=0");
+			fail("Exception expected when inbound target is not set");
+		}
+		catch (BeanCreationException e) {
+			assertThat(e.getCause().getMessage()).startsWith("Either StreamListener or a method parameter should be set with ");
```
ilayaperumalg: this as well was added with `startsWith`. The actual message is: `Either StreamListener or a method parameter should be set with an inbound element`. Can add your suggestion there.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestAnnotatedArguments.java
ilayaperumalg: > How does the mapping know what message it should get the headers from? 

Isn't the `headers` resolved for each `message` by the `HeadersMethodArgumentResolver`?

ilayaperumalg: Given the `Payload`, `Headers` and `Header` annotated method parameters would get resolved by their corresponding method argument resolvers, I don't see any issues supporting this type of mapping. What am I missing?

markpollack: I guess there can be a method with both @Input and @Output arguments... so then the method level annotation is needed to disambiguate.

mbogoevici: @markpollack https://github.com/spring-cloud/spring-cloud-stream/pull/681#discussion_r84966975

This is why I don't think we should have `@Input`/`@Output` for the 'classic' use case at all and reserve it for bindable types. See my comment above. The only concession in the name of consistency that we can make is to allow:

```
@StreamListener("foo") @SendTo("bar")
public Flux<String> toUpper(Flux<String> input) {
}
```

so that there a 1:1 match in the 1-input 0..1-output case, since that's fairly intuitive, and we could stick to the current approach where `@Input` is used only on method arguments, `@Output` only on arguments and return types.

@ilayaperumalg  when it comes to header extraction, the use of `@Headers` is a valid use case, but I don't find that  the use of `@Headers` with another parameter annotated with `@Input` is.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
```diff
@@ -0,0 +1,399 @@
...
+/**
+ * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
+ */
+public class StreamListenerMethodTests {
+
+	@Test
+	public void testMethodWIthInputOnStreamListener() throws Exception {
+		try {
+			SpringApplication.run(TestMethodWIthInputOnStreamListener.class, "--server.port=0");
```
markfisher: It would be easier to follow the testing logic if the invalid example classes were named as such. Perhaps it's even worth interleaving their definitions with test methods rather than having them all at the end of the file. Although maybe the bigger issue is how complex the rules are for these handler methods in the first place; are we possibly trying to cram too many options into a single annotation? i.e. should we consider making an explicit distinction between "declarative" and "message-at-a-time" at the annotation level? cc @mbogoevici

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestHandlerBean.java
```diff
@@ -0,0 +1,140 @@
...
+	@EnableAutoConfiguration
+	public static class TestHandlerBean2 {
+
+		@Bean
+		public HandlerBean2 handlerBean() {
+			return new HandlerBean2();
+		}
+	}
+
+	public static class HandlerBean1 extends HandlerBean {
```
markfisher: are HandlerBean1 and HandlerBean2 intentionally identical?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMessageArgument.java
```diff
@@ -0,0 +1,114 @@
...
+import org.springframework.messaging.handler.annotation.SendTo;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
+ */
+@RunWith(Parameterized.class)
+public class StreamListenerTestMessageArgument {
```
markfisher: shouldn't this classname end with "Tests"?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMessageArgument.java
```diff
@@ -0,0 +1,114 @@
...
+		Message<StreamListenerTestInterfaces.BarPojo> message = (Message<StreamListenerTestInterfaces.BarPojo>) collector
+				.forChannel(processor.output()).poll(1, TimeUnit.SECONDS);
+		assertThat(message).isNotNull();
+		assertThat(message.getPayload().getBar()).isEqualTo("barbar" + id);
+		context.close();
+	}
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class TestPojoWithMessageArgument1 extends TestPojoWithMessageArgument {
```
markfisher: again TestPojoWithMessageArgument1 and TestPojoWithMessageArgument2 appear to be identical. What's the point?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMethodReturnWithConversion.java
```diff
@@ -0,0 +1,163 @@
...
+					TimeUnit.SECONDS);
+			assertThat(message).isNotNull();
+			assertThat(message.getPayload().getBar()).isEqualTo("barbar" + id);
+			context.close();
+		}
+	}
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class TestPojoWithMimeType1 extends TestPojoWithMimeType {
```
markfisher: same thing here

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMethodReturnWithConversion.java
```diff
@@ -0,0 +1,163 @@
...
+			String id = UUID.randomUUID().toString();
+			processor.input().send(MessageBuilder.withPayload("{\"foo\":\"barbar" + id + "\"}")
+					.setHeader("contentType", "application/json").build());
+			TestPojoWithMimeType testPojoWithMimeType = context.getBean(TestPojoWithMimeType.class);
+			assertThat(testPojoWithMimeType.receivedPojos).hasSize(1);
+			assertThat(testPojoWithMimeType.receivedPojos.get(0)).hasFieldOrPropertyWithValue("foo", "barbar" + id);
+			Message<StreamListenerTestInterfaces.BarPojo> message = (Message<StreamListenerTestInterfaces.BarPojo>) collector.forChannel(processor.output()).poll(1,
+					TimeUnit.SECONDS);
+			assertThat(message).isNotNull();
+			assertThat(message.getPayload().getBar()).isEqualTo("barbar" + id);
```
markfisher: should add an assert (even if negative) for the contentType here?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMethodWithReturnMessage.java
```diff
@@ -0,0 +1,115 @@
...
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class TestPojoWithMessageReturn2 extends TestPojoWithMessageReturn {
+
+		@StreamListener(Processor.INPUT)
+		@Output(Processor.OUTPUT)
+		public Message<?> receive(StreamListenerTestInterfaces.FooPojo fooPojo) {
+			this.receivedPojos.add(fooPojo);
+			StreamListenerTestInterfaces.BarPojo bazPojo = new StreamListenerTestInterfaces.BarPojo();
```
markfisher: identical in this case as well except for the name of the `bazPojo` var... was this intended to be something else?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMethodWithReturnValue.java
```diff
@@ -0,0 +1,111 @@
...
+import org.springframework.messaging.handler.annotation.SendTo;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
+ */
+@RunWith(Parameterized.class)
+public class StreamListenerTestMethodWithReturnValue {
```
markfisher: all test classes should end in "Tests" as a project convention (there are others I have not commented on here)

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactiveTestInputOutputArgsWIthSenderAndFailure.java
```diff
@@ -0,0 +1,138 @@
...
+
+	private Class<?> configClass;
+
+	public StreamListenerReactiveTestInputOutputArgsWIthSenderAndFailure(Class<?> configClass) {
+		this.configClass = configClass;
+	}
+
+	@Parameterized.Parameters
+	public static Collection InputConfigs() {
+		return Arrays.asList(new Class[] {TestInputOutputArgsWithFluxSenderAndFailure1.class});
```
markfisher: only one class? is the intention to add the RxJava version as well?

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactiveTestInputOutputArgsWIthSenderAndFailure.java
```diff
@@ -0,0 +1,138 @@
...
+							throw new RuntimeException();
+						}
+					})
+					.map(o -> MessageBuilder.withPayload(o).build()));
+		}
+	}
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class TestInputOutputArgsWithFluxSenderAndFailure2 {
```
markfisher: wasn't this case already tested in the preceding test class?

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactiveTestInputOutputArgsWithMessage.java
```diff
@@ -0,0 +1,149 @@
...
+		public void receive(@Input(Processor.INPUT) Flux<Message<?>> input,
+				@Output(Processor.OUTPUT) FluxSender output) {
+			output.send(input.map(m -> MessageBuilder
+					.withPayload(m.getPayload().toString().toUpperCase()).build()));
+		}
+	}
+
+	@EnableBinding(Processor.class)
+	@EnableAutoConfiguration
+	public static class ReactorTestInputOutputArgsWithMessage2 {
```
markfisher: why keep testing this same invalid use-case? one test should be sufficient

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerTestGenericFluxInputOutputArgsWithMessage.java
```diff
@@ -0,0 +1,171 @@
...
+
+	private Class<?> configClass;
+
+	public StreamListenerTestGenericFluxInputOutputArgsWithMessage(Class<?> configClass) {
+		this.configClass = configClass;
+	}
+
+	@Parameterized.Parameters
+	public static Collection InputConfigs() {
+		return Arrays.asList(new Class[] {TestGenericStringFluxInputOutputArgsWithMessageImpl1.class});
```
markfisher: why parameterized if only one class?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
 			public void doWith(final Method method) throws IllegalArgumentException, IllegalAccessException {
 				StreamListener streamListener = AnnotationUtils.findAnnotation(method, StreamListener.class);
 				if (streamListener != null) {
+					Assert.isTrue(method.getAnnotation(Input.class) == null,
+							"A @StreamListener may never be annotated with @Input. If it should listen to a specific input, " +
+									"use the value of @StreamListener instead.");
+					String methodAnnotatedInboundName = getInboundElementNameFromMethodAnnotation(streamListener);
+					String methodAnnotatedOutboundName = getOutboundElementNameFromMethodAnnotation(method);
```
markfisher: seems like the method being called here should be named: `getOutboundElementNameFromMethod(..)` since it takes the `Method` instance and *not* an annotation

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+								targetBean = getBindableBean(outboundElementName);
+								declarativeArguments[parameterIndex] = getDeclarativeArgument(targetBean, methodParameter);
+							}
+							else if (declarativeArguments.length == 1 && StringUtils.hasText(methodAnnotatedInboundName)) {
+								targetBean = getBindableBean(methodAnnotatedInboundName);
+								declarativeArguments[parameterIndex] = getDeclarativeArgument(targetBean, methodParameter);
+							}
+							Assert.isTrue(declarativeArguments[parameterIndex] != null, "Declarative StreamListener method should only have " +
+									"inbound or outbound targets as method parameters");
+							Assert.notNull(declarativeArguments[parameterIndex],
```
markfisher: this would never be invoked given that the assertion on the previous line would have already failed

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+	private boolean isDeclarativeStreamListenerMethod(Method method, String methodAnnotatedinboundName, String methodAnnotatedOutboundName) {
+		int methodArgumentsLength = method.getParameterTypes().length;
+		for (int parameterIndex = 0; parameterIndex < methodArgumentsLength; parameterIndex++) {
+			MethodParameter methodParameter = MethodParameter.forMethodOrConstructor(method, parameterIndex);
+			if (methodParameter.hasParameterAnnotation(Input.class)) {
+				String inboundName = (String) AnnotationUtils.getValue(methodParameter.getParameterAnnotation(Input.class));
+				Assert.isTrue(StringUtils.hasText(inboundName), "@Input annotation should always be associated with a valid inbound name");
+				return isDeclarativeMethodParameter(getBindableBean(inboundName), methodParameter);
+			}
+			else if (methodParameter.hasParameterAnnotation(Output.class)) {
```
markfisher: `else` is not necessary for any of these given the `return` statements at each level... could consider refactoring to a single `return` at the end with a single call to `isDeclarativeMethodParameter` (where the various if conditions lead to assignment of the name to check)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+		if (sendTo != null) {
+			Assert.isTrue(!ObjectUtils.isEmpty(sendTo.value()), "At least one output must be specified");
+			Assert.isTrue(sendTo.value().length == 1, "Multiple destinations cannot be specified");
+			Assert.hasText(sendTo.value()[0], "An empty destination cannot be specified");
+			return StringUtils.hasText(sendTo.value()[0]) ? sendTo.value()[0] : null;
+		}
+		Output output = AnnotationUtils.findAnnotation(method, Output.class);
+		if (output != null) {
+			Assert.isTrue(StringUtils.hasText(output.value()), "At least one output must be specified");
+			return StringUtils.hasText(output.value()) ? output.value() : null;
```
markfisher: another redundant ternary expression

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+					"method parameters when both inbound and outbound values are set as method annotated values");
+		}
+		if (StringUtils.hasText(methodAnnotatedInboundName)) {
+			Assert.isTrue(inputAnnotationCount == 0, "Cannot set both StreamListener value " + methodAnnotatedInboundName +
+					" and @Input annotation as method parameter");
+			Assert.isTrue(outputAnnotationCount == 0, "Cannot set StreamListener value '" + methodAnnotatedInboundName + "' when using" +
+					" @Output annotation as method parameter. Use @Input method parameter annotation to specify inbound value instead");
+		}
+		else {
+			Assert.isTrue(inputAnnotationCount >= 1, "No input destination is configured. Use either a @StreamListener argument or @Input");
```
markfisher: I believe technically it should be referred to as a "@StreamListener *attribute*"

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+					numAnnotatedMethodParameters++;
+				}
+				if (methodParameter.hasParameterAnnotation(Payload.class)) {
+					numPayloadAnnotations++;
 				}
 			}
-			Assert.notNull(arguments[parameterIndex],
-					"Cannot convert argument " + parameterIndex + " of " + method + "from " + targetBean.getClass()
-							.toString() + " to " + parameterType.toString());
+			Assert.isTrue(methodArgumentsLength <= numAnnotatedMethodParameters && numPayloadAnnotations <= 1,
```
markfisher: how can `methodArgumentsLength` ever be less than `numAnnotatedMethodParameters`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+
+	private Object getBindableBean(String boundElementName) {
+		try {
+			if (!this.boundElements.containsKey(boundElementName)) {
+				this.boundElements.put(boundElementName, this.applicationContext.getBean(boundElementName));
+			}
+			return this.boundElements.get(boundElementName);
+		}
+		catch (NoSuchBeanDefinitionException e) {
+			throw new IllegalStateException("Target bean doesn't exist for the bound element name: " + boundElementName);
```
markfisher: should we add the `NoSuchBeanDefinitionException` as the `cause`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+	private Object getBindableArgument(Object targetBean, MethodParameter methodParameter) {
+		if (methodParameter.getParameterType().isAssignableFrom(targetBean.getClass())) {
+			return targetBean;
+		}
+		for (StreamListenerParameterAdapter<?, Object> streamListenerParameterAdapter : streamListenerParameterAdapters) {
+			if (streamListenerParameterAdapter.supports(targetBean.getClass(), methodParameter)) {
+				return streamListenerParameterAdapter.adapt(targetBean, methodParameter);
+			}
+		}
+		throw new IllegalStateException("Bindable target is not found for the target " + targetBean.getClass() + " and" +
```
markfisher: should be "Bindable *argument* is not found..." ?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
ilayaperumalg: It would be much simpler if we go with an `explicit` annotation that specifies `declarative` but I think the intention was not to add such (one more) annotation  as this can be **derived**. 

> are we possibly trying to cram too many options into a single annotation

I think the idea is to allow users to specify all possible options that make sense when it comes to specifying method annotation/parameters as close as possible for both `declarative` and `non-declarative`  `@StreamListener` method.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestHandlerBean.java
ilayaperumalg: They are identical except one uses `@Output` while the other uses `@SendTo` when specifying the `outbound` channel name :-)

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMessageArgument.java
ilayaperumalg: Same here. Testing `@Output` and `@SendTo` variations.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMethodWithReturnMessage.java
ilayaperumalg: `@Output` vs `@SendTo` here as well.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithAnnotatedInputOutputArgsTests.java
```diff
@@ -37,18 +37,49 @@
...
+public class StreamListenerWithAnnotatedInputOutputArgsTests {
 
 	@Test
 	public void testInputOutputArgs() throws Exception {
 		ConfigurableApplicationContext context = SpringApplication.run(TestInputOutputArgs.class, "--server.port=0");
 		sendMessageAndValidate(context);
 	}
 
+	@Test
+	public void testInputOutputArgsWithMoreParameters() throws Exception {
```
ilayaperumalg: No, sorry. copy-paste error. will fix it.

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactiveTestInputOutputArgsWithSender.java
```diff
@@ -0,0 +1,155 @@
...
+				"--server.port=0");
+		// send multiple message
+		sendMessageAndValidate(context);
+		sendMessageAndValidate(context);
+		sendMessageAndValidate(context);
+		context.close();
+	}
+
+	@Test
+	public void testIncorrectUsage() {
```
ilayaperumalg: yeah, this was the result of the initial tests added when all possible `@StreamListener` method parameter/method annotations support was added and all the tests were inside a single test class. Then, I had parameterisation for all the test methods. After refactoring  and removing some of the supported options, the parameterisation diverged. will fix this.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
ilayaperumalg: I can change the name to `getOutboundElementNameFromMethod`. The initial reasoning was that Inside the method it is still the `@Output` or `@SendTo` annotations what we check to get the `outbound` element name.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
ilayaperumalg: Since `getBindableBean` is an expensive operation, I thought it is better to return `true` as soon as we find the `declarative` method parameter and so we don't have to loop through all the method parameters and then `return`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,89 +113,251 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+
+	private String getInboundElementNameFromMethodAnnotation(StreamListener streamListener) {
+		return StringUtils.hasText(streamListener.value()) ? streamListener.value() : null;
+	}
+
+	private String getOutboundElementNameFromMethodAnnotation(Method method) {
+		SendTo sendTo = AnnotationUtils.findAnnotation(method, SendTo.class);
+		if (sendTo != null) {
+			Assert.isTrue(!ObjectUtils.isEmpty(sendTo.value()), "At least one output must be specified");
+			Assert.isTrue(sendTo.value().length == 1, "Multiple destinations cannot be specified");
```
ilayaperumalg: At least from SCSt, it looks ambiguous when specifying multiple outbound names as part of `@SendTo` while multiple output is only achieved using `@Output` annotation in `declarative` mode.

mbogoevici: @markfisher Similar things exist elsewhere where even if the annotation supports it, the framework doesn't - e.g. https://github.com/spring-projects/spring-amqp/blob/master/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/listener/MethodRabbitListenerEndpoint.java#L134

mbogoevici: @ilayaperumalg if we were to support multiple values (which I don't think we should), it would be in any case semantically different from using different parameters ('broadcast to all specified destinations' which is not what the distinct parameters imply)

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
mbogoevici: It shouldn't be so complex. All we eventually wanted was to be able to support this (with or without return argument) in addition to what we already supported: 

```
@SendTo(Processor.OUTPUT)
public Flux<String> toUpper(Flux<String> input)
```

See my last comment : https://github.com/spring-cloud/spring-cloud-stream/pull/681/files/c05266a3f2471da32159f77553e9cdc1b84f5581#r84973574

Thinking again through all this, the initial test was:

- does StreamListener have an argument ?
a) if not, handle it as a declarative method
b) if yes, handle it as a message handler method;

This becomes:
- does StreamListener have an argument ? 
a) if not, handle it as a declarative method;
b) if yes:
     - if the method doesn't have a single parameter, fail automatically (this should currently happen anyway)
     b1) if the argument is bindable or convertible to bindable, handle it as a declarative method;
     b2) if not, handle it as a message handler method;

We already have a) and b2) implemented (including validations) so all we need is to add an additional test in the branch that tests StreamListener where we validate that we're not in b1 and handle accordingly. So these would be the three variants that work:

```
@SendTo(Processor.OUTPUT)
public String> toUpper(String input)
```

```
@SendTo(Processor.OUTPUT)
public Flux<String> toUpper(Flux<String> input)
```

```
@StreamListener
@Output(Processor.OUTPUT)
public Flux<String> toUpper(@Input(Processor.Input) Flux<String> input)
```

Where the latter is a particular case of the already supported idiom where you can have an arbitrary number of params annotated with Input and Output and maybe a return value annotated with Output. Would this be too confusing? I think that there is value in aligning the two variants just as the reactive web model allows for @GetMapping and so on.

ilayaperumalg: @mbogoevici My initial approach was using a similar logic you mentioned above. Further refactoring from there lead to the current state of the PR as few other methods also required some changes (ex: The current `invokeSetupMethodOnListenedChannel` also required refactoring (as it depended upon either `@Input` or `@Output` only while we still wanted to propagate  for the `inbound` and `outbound` names via method annotation values as well)

Also, as a side note:
In case StreamListener attribute set with an `inbound` name, we can't fail the method when it has multiple parameters as we still need to support arguments with `@Payload`, `@Headers` etc.,

mbogoevici: Yes, I see what you mean - I was thinking on whether there is more than one input destination specified as argument. So I guess b1 would be 'if the method has a single bindable or convertible argument, then treat it as a declarative method'. I do not understand this very clearly, though, sorry:

>The current invokeSetupMethodOnListenedChannel also required refactoring (as it depended upon either @Input or @Output only while we still wanted to propagate for the inbound and outbound names via method annotation values as well)

We don't necessarily have to do it all in one method for declarative methods  - we can a structure that holds information about the return type and the argument bindings and have different parsing strategies for the two syntax variants: @StreamListener/@SendTo vs @StreamListener+@Input+@Output.

ilayaperumalg: > The current invokeSetupMethodOnListenedChannel also required refactoring (as it depended upon either @Input or @Output only while we still wanted to propagate for the inbound and outbound names via method annotation values as well)

I meant extracting `outbound` name (especially) from method annotation and use it for `single` argument declarative method. 

> we can a structure that holds information about the return type and the argument bindings and have different parsing strategies for the two syntax variants: @StreamListener/@SendTo vs @StreamListener+@Input+@Output.

I didn't mean to perform all these operations on the same operations. In fact I thought this PR goes step by step approach of doing: 

1) `inbound` and `outbound` element names from method annotations
2)  asserting these along with the method parameter names
3) detecting if the StreamListener method is declarative or not
4) execute appropriate code based on it

with the refactoring on top of what is already there (with the slightly modified logic)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
markfisher: it just bothered me that the names (for inbound vs. outbound) implied consistency but the arguments to those methods were not consistent

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
markfisher: well, my point was to do one of 2 things: either drop the *else*s (they are not necessary in the current impl as far as I can tell) OR refactor to have a single return value after getting the *name* to pass (in which case you keep the *else*s) - and in the latter case, you would defer the call to `getBindableBean` until after the name has been determined (if at all). Does that make sense?

ilayaperumalg: oh, yeah, will drop the `elses` as the latter case of determining `name` is conditional based on what satisfies the `if` conditions.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestMethodReturnWithConversion.java
ilayaperumalg: Testing @Output and @SendTo variations here as well.

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactiveTestInputOutputArgsWIthSenderAndFailure.java
ilayaperumalg: yeah, that was the intention but missed out.

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerReactiveTestInputOutputArgsWIthSenderAndFailure.java
ilayaperumalg: yeah, will remove duplicate test methods.

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamListenerTestGenericFluxInputOutputArgsWithMessage.java
ilayaperumalg: will remove parameterization here. this was the result of previous commit that used multiple supported options.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
ilayaperumalg: Sorry, this was intended to be `methodArgumentsLength >= numAnnotatedMethodParameters`

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerMethodTests.java
ilayaperumalg: I will give it a shot on using the initial `if/else` logic approach and update the PR

ilayaperumalg: Here are the changes I have made so far: https://github.com/ilayaperumalg/spring-cloud-stream/commit/34b09c935ee89d8afef70b361dddc562a76d2242#diff-e04acbd3c15fbd6eac0b0af108694f2aR201. It turns out that it is a convenient to `find` if the StreamListener method is `declarative` well before so as to perform `assertions` and finding the `target` beans (ex: get `outbound` target bean either from method annotation or from method parameters). Let me know what do you think. I am OK with going ahead and make necessary changes and that would look similar to the current state PR except the approach being different one.

markfisher: I added a few comments directly on that commit, but in general, it does seem like a clearer implementation. Interested to hear what @mbogoevici thinks.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,84 +120,138 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
 	}
 
+	private boolean isDeclarativeStreamListenerMethod(Method method, String methodAnnotatedInboundName, String methodAnnotatedOutboundName) {
+		int methodArgumentsLength = method.getParameterTypes().length;
+		for (int parameterIndex = 0; parameterIndex < methodArgumentsLength; parameterIndex++) {
+			MethodParameter methodParameter = MethodParameter.forMethodOrConstructor(method, parameterIndex);
+			if (methodParameter.hasParameterAnnotation(Input.class)) {
+				String inboundName = (String) AnnotationUtils.getValue(methodParameter.getParameterAnnotation(Input.class));
+				Assert.isTrue(StringUtils.hasText(inboundName), INVALID_INBOUND_NAME);
+				return isDeclarativeMethodParameter(getBindableBean(inboundName), methodParameter);
```
mbogoevici: A parameter annotated with `@Input` must be declarative so here we could just throw an error if it isn't.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,84 +120,138 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+			}
+			if (methodParameter.hasParameterAnnotation(Output.class)) {
+				String outboundName = (String) AnnotationUtils.getValue(methodParameter.getParameterAnnotation(Output.class));
+				Assert.isTrue(StringUtils.hasText(outboundName), INVALID_OUTBOUND_NAME);
+				return isDeclarativeMethodParameter(getBindableBean(outboundName), methodParameter);
```
mbogoevici: A parameter annotated with `@Output` must be declarative so here we could just throw an error if it isn't.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,84 +120,138 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+			for (StreamListenerParameterAdapter<?, Object> streamListenerParameterAdapter : this.streamListenerParameterAdapters) {
+				if (streamListenerParameterAdapter.supports(targetBean.getClass(), methodParameter)) {
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
+	private Object getBindableBean(String boundElementName) {
```
mbogoevici: Can be just 'getBean' - we check later on if it is actually bindable.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -43,18 +43,27 @@
...
 import org.springframework.messaging.core.DestinationResolver;
-import org.springframework.messaging.handler.annotation.SendTo;
 import org.springframework.messaging.handler.annotation.support.MessageHandlerMethodFactory;
 import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
 import org.springframework.util.Assert;
-import org.springframework.util.ObjectUtils;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 
+import static org.springframework.cloud.stream.binding.StreamListenerErrorMessages.INPUT_AT_STREAM_LISTENER;
```
mbogoevici: No static imports in production code.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -109,84 +120,138 @@ public Object postProcessAfterInitialization(final Object bean, String beanName)
...
+					if (!method.getReturnType().equals(Void.TYPE)) {
+						if (!StringUtils.hasText(methodAnnotatedOutboundName)) {
+							if (outputAnnotationCount == 0) {
+								throw new IllegalArgumentException(RETURN_TYPE_NO_OUTBOUND_SPECIFIED);
+							}
+							Assert.isTrue((outputAnnotationCount == 1), RETURN_TYPE_MULTIPLE_OUTBOUND_SPECIFIED);
 						}
-						registerHandlerMethodOnListenedChannel(method, streamListener, bean);
+					}
+					if (isDeclarative && (!StringUtils.hasText(methodAnnotatedInboundName) || parameterTypes.length == 1)) {
```
mbogoevici: Logic needs to be cleaned up here - it seems like `isDeclarative` being true isn't a complete indication that the method is actually declarative.

mbogoevici: Actually, at a second glance, this seems wrong - if a method would normally be declarative (i.e. has the right arguments/annotations) but has either more than one argument or declares an input via annotation (e.g. `@StreamListener("input"))` we set up the handler - we should never set up a handler for a declarative method.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerMethodUtils.java
```diff
@@ -0,0 +1,151 @@
...
+				}
+			}
+			if (numPayloadAnnotations > 0) {
+				Assert.isTrue(methodArgumentsLength == numAnnotatedMethodParameters && numPayloadAnnotations <= 1,
+						AMBIGUOUS_MESSAGE_HANDLER_METHOD_ARGUMENTS);
+			}
+		}
+	}
+
+	protected static String getInboundElementNameFromMethod(StreamListener streamListener) {
```
mbogoevici: We can inline this as it is simply getting the value from `StreamListener` - and returning `null` if empty.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
mbogoevici: Actually, it is not clear for me why can't we just use `applicationContext.getBean()` instead and skip the local caching - the bean location process itself is very efficient.

mbogoevici:
Will address the comments as a separate commit.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerHandlerBeanTests.java
```diff
@@ -0,0 +1,140 @@
...
+			barPojo.setBar(fooMessage.getFoo());
+			return barPojo;
+		}
+	}
+
+	public static class HandlerBean2 extends HandlerBean {
+
+		@StreamListener(Processor.INPUT)
+		@Output(Processor.OUTPUT)
+		public StreamListenerTestInterfaces.BarPojo receive(StreamListenerTestInterfaces.FooPojo fooMessage) {
```
mbogoevici: I thought that we agreed that `@Input`/`@Output` should be reserved for bindable types ?

mbogoevici: We should support `@SendTo`  for both bindable and POJO/Message but `@Output` only for bindable types.

mbogoevici: In retrospect, I see what this is trying to do. While I am not a huge fan of it, I suppose we can use `@SendTo` and `@Output` interchangeably at the method level.

mbogoevici:
@ilayaperumalg I've squashed and rebased the commit in my fork, and added some additional polishing: https://github.com/mbogoevici/spring-cloud-stream/tree/ilaya-align-listener - please take a look and we can merge this to master if it's okay.

ilayaperumalg:
@mbogoevici your changes look great! thank you. will merge this PR now.

ilayaperumalg:
Merged as 453a9b0c100f32c8dbe4dd791730584cd53c3e38 along with c6e0c74626e064883a02032b0f03fd598076bb46

PR #684, opened 2016-10-12
Support valid multiple @StreamListener handler methods

### LINKED ISSUE #682: Support of Multiple @StreamListener methods on the same bound element
Hi,

I have a case where I want to have multiple @StreamLinstner methods that are listening on the same bound element. The methods signature are different depending the type of message being received. Here is the example:

```
@StreamListener(Sink.INPUT)
public void on(PostUpvotedEvent event) {

}

@StreamListener(Sink.INPUT)
public void on(PostDownvotedEvent event) {

}
```

Currently, this is not supported and there is an exception with this message `Duplicate @StreamListener mapping for 'input'`.

I think it will be convenient to have this feature.
Thanks

ilayaperumalg:
- Add all the @StreamListener message handler methods that have unique method parameter to receive and process the message payload received.
  - Throw exception when duplicate @StreamListener message handler methods are used

Resolves #682

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithHandlerTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithHandlerTests.java
@@ -50,6 +50,7 @@ import static org.junit.Assert.fail;
 
 /**
  * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
  */
 public class StreamListenerWithHandlerTests {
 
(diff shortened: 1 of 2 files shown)


mbogoevici:
We should hold on to this until we decide we actually want to support this feature and how? There is much ambiguity left in how the arguments are resolved. Specifically, for type-agnostic payloads such as JSON or even Avro, it's not clear which method should we invoke as the conversion happens at runtime and is driven by the parameter type. The current test works because it relies on JSON to POJO conversion failure, but it would fail with more lenient conversion settings.

ilayaperumalg:
ok, the main idea with the current approach is to consider only those parameter types which are not incompatible types. While the message conversion still holds the key as it might overlap with the methods, wanted to give this a shot as someone wouldn't deliberately write multiple methods while message type conversion could happen in one or other way :-)

mbogoevici:
>someone wouldn't deliberately write multiple methods while message type conversion could happen in one or other way :-)

They could write it by mistake and that's an additional problem to deal with. If we become more smart/lenient we should make sure that we don't open the door for subtle ambiguities which would be hard to debug (e.g. the wrong method receiving the message, etc.).

mbogoevici:
Closing this until we come up with a better design and understand the tradeoffs.

PR #687, opened 2016-10-14
add entity scan for org.springframework.cloud.stream.schema.server.mo…

mnadelson:
…del.Schema so that it can be reconginized as a hibernate entity

### DIFF
--- a/spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/config/SchemaServerConfiguration.java
+++ b/spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/config/SchemaServerConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.stream.schema.server.config;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


mbogoevici:
Hi @mnadelson,

Thanks for the contribution - there is indeed an issue with the current implementation where we assume that Schema would be automatically scanned - because the out of the box implementation of the schema registry server is in a parent package. If the application's base package is not a parent of it, then the `Schema` entity class will not be scanned.  The only issue with the approach that you are suggesting is that it sets the default entity scan package for any application that adds the `@EnableSchemaRegistryServer` annotation to the one of `Schema`. This is rather unexpected to the user, who would then have to register their own packages. 

See my suggested approach here: https://github.com/spring-cloud/spring-cloud-stream/pull/689 - it uses a `BeanFactoryPostprocessor` for configuring the Schema package transparently. Based on the test, this should satisfy your use case.

ilayaperumalg:
Hi @mnadelson,  
Given the PR #689 addresses this issue in a broader way, we will close this PR.
We really appreciate your contribution. Thank you!

PR #692, opened 2016-10-19
Support native serialization in MessageChannelBinder

### LINKED ISSUE #691: Allow message channel binder to support native serialization/de-serialization
When the broker supports native serialization/de-serialization (like kafka), the binder should not perform serialization/de-serialization.

ilayaperumalg:
- If the broker supports native serialization, then the serialization of the outbound payload is not done by the binder's producer binding outbound message handler.
  Also, when native serialization is enabled, embedding of headers (if enabled) is allowed only when the outbound payload type is byte[]
- At the consumer binding's inbound message handler, the extraction of embedded headers is performed only for the byte[] payload (given the deserialization would have already occurred at the broker level)
- Add correspoinding tests in Kafka binder implementation

Resolves #691

### DIFF
--- a/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
+++ b/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
@@ -122,7 +122,7 @@ public abstract class AbstractBinderTests<B extends AbstractTestBinder<? extends
 		// Let the consumer actually bind to the producer before sending a msg
 		binderBindUnbindLatency();
 		moduleOutputChannel.send(message);
-		Message<?> inbound = receive(moduleInputChannel);
+		Message<?> inbound = receive(moduleInputChannel, 5);
 		assertThat(inbound).isNotNull();
 		assertThat(inbound.getPayload()).isEqualTo("foo");
 		assertThat(inbound.getHeaders().get(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE)).isNull();
@@ -164,8 +164,8 @@ public abstract class AbstractBinderTests<B extends AbstractTestBinder<? extends
 
 
 		Message<?>[] messages = new Message[2];
-		messages[0] = receive(moduleInputChannel);
-		messages[1] = receive(moduleInputChannel);
+		messages[0] = receive(moduleInputChannel, 5);
+		messages[1] = receive(moduleInputChannel, 5);
 
 		assertThat(messages[0]).isNotNull();
 		assertThat(messages[1]).isNotNull();
@@ -194,7 +194,7 @@ public abstract class AbstractBinderTests<B extends AbstractTestBinder<? extends
 
 		Message<?> message = MessageBuilder.withPayload("foo").build();
 		moduleOutputChannel.send(message);
-		Message<?> inbound = receive(moduleInputChannel);
+		Message<?> inbound = receive(moduleInputChannel, 5);
 		assertThat(inbound).isNotNull();
 		assertThat(inbound.getPayload()).isEqualTo("foo");
 		assertThat(inbound.getHeaders().get(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE)).isNull();
--- a/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
+++ b/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
@@ -122,7 +122,7 @@ public abstract class AbstractBinderTests<B extends AbstractTestBinder<? extends
 		// Let the consumer actually bind to the producer before sending a msg
 		binderBindUnbindLatency();
 		moduleOutputChannel.send(message);
-		Message<?> inbound = receive(moduleInputChannel, 5);
+		Message<?> inbound = receive(moduleInputChannel);
 		assertThat(inbound).isNotNull();
 		assertThat(inbound.getPayload()).isEqualTo("foo");
 		assertThat(inbound.getHeaders().get(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE)).isNull();
@@ -164,8 +164,8 @@ public abstract class AbstractBinderTests<B extends AbstractTestBinder<? extends
 
 
 		Message<?>[] messages = new Message[2];
-		messages[0] = receive(moduleInputChannel, 5);
-		messages[1] = receive(moduleInputChannel, 5);
+		messages[0] = receive(moduleInputChannel);
+		messages[1] = receive(moduleInputChannel);
 
 		assertThat(messages[0]).isNotNull();
 		assertThat(messages[1]).isNotNull();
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -261,31 +262,36 @@ protected void afterUnbindConsumer(String destination, String group, C consumerP
...
 
 		@Override
 		@SuppressWarnings("unchecked")
 		protected Object handleRequestMessage(Message<?> requestMessage) {
-			MessageValues messageValues;
-			if (this.extractEmbeddedHeaders) {
-				try {
-					messageValues = AbstractMessageChannelBinder.this.embeddedHeadersMessageConverter.extractHeaders(
+			MessageValues messageValues = (this.extractEmbeddedHeaders) ? extractEmbeddedHeaders(requestMessage) :
+					new MessageValues(requestMessage);
```
mbogoevici: Probably doing things in a different order helps - check if native deserialization happened (payload is not byte[]), then extract headers.

mbogoevici: Perhaps we can inline the extracted method back, too, since there are no changes.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -303,52 +309,71 @@ protected boolean shouldCopyRequestHeaders() {
...
+			this.supportsSerializationNatively = supportsSerializationNatively;
 		}
 
 		@Override
 		protected void handleMessageInternal(Message<?> message) throws Exception {
-			MessageValues transformed = serializePayloadIfNecessary(message);
-			byte[] payload;
+			MessageValues messageToSend = (this.supportsSerializationNatively) ? new MessageValues(message) :
+					serializePayloadIfNecessary(message);
+			this.delegate.handleMessage(embedHeadersIfApplicable(messageToSend));
```
mbogoevici: We shouldn't embed headers if encoding natively.

mbogoevici: Again, I think this can be simplified and the old logic can be restored as there should be no change to it - we just need to add a separate case where native encoding is being used - then we just pass the object to the binder.

spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
```diff
@@ -122,7 +122,7 @@ public void testSendAndReceive() throws Exception {
 		// Let the consumer actually bind to the producer before sending a msg
 		binderBindUnbindLatency();
 		moduleOutputChannel.send(message);
-		Message<?> inbound = receive(moduleInputChannel);
+		Message<?> inbound = receive(moduleInputChannel, 5);
```
ilayaperumalg: I had this locally to see if it helps avoiding those intermittent test errors we see on the binder tests. will revert this.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
```diff
@@ -50,6 +50,8 @@
 
 	private HeaderMode headerMode = HeaderMode.embeddedHeaders;
 
+	private boolean supportSerializationNatively = false;
```
ilayaperumalg: I like `useNativeEncoding`. Perhaps, `useBrokerNativeEncoding` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
```diff
@@ -125,5 +127,13 @@ public HeaderMode getHeaderMode() {
 	public void setHeaderMode(HeaderMode headerMode) {
 		this.headerMode = headerMode;
 	}
+
+	public boolean isSupportDeserializationNatively() {
```
ilayaperumalg: Given the native deserializer would have already kicked in before the ReceivingHandler at the consumer binding receives the message, there is no way of bypassing the codec at the consumer side I think.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
mbogoevici: It's not broker native.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
mbogoevici: > there is no way of bypassing the codec at the consumer side I think. 

I don't quite understand. If the native decoder/deserializer kicks in, the message payload won't be a `byte[]` (normally) - so the codec shouldn't handle it (it can't handle anything that's not a `byte[]`).

mbogoevici: I.e. we don't want to use the native decoding *and* the codec, that doesn't make sense.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -303,16 +316,25 @@ protected boolean shouldCopyRequestHeaders() {
...
 			this.delegate = delegate;
 			this.setBeanFactory(AbstractMessageChannelBinder.this.getBeanFactory());
 			this.embedHeaders = embedHeaders;
 			this.embeddedHeaders = headersToEmbed;
+			this.useNativeEncoding = useNativeEncoding;
 		}
 
 		@Override
 		protected void handleMessageInternal(Message<?> message) throws Exception {
+			Message<?> messageToSend = (this.useNativeEncoding) ? getMessageBuilderFactory().withPayload(message.getPayload()).build() :
```
mbogoevici: We shouldn't strip the headers if native encoding is being used. They may contain information useful for the sender (e.g. the partition id in partitioned scenarios, the key for Kafka, etc). if native encoding is being used, we should just use the incoming message as is.

ilayaperumalg: My understanding is that when we use `native` encoding, irrespective of whether we add the `headers` to the outbound messages, the consumer's native de-serializer would ignore those headers and the only headers we would is some `metadata` headers updated at the consumer side (via spring-integration-kafka). Also, I thought this is what we discussed [here] (https://github.com/spring-cloud/spring-cloud-stream/pull/692#discussion_r84076033). What am I missing?

mbogoevici:
The problem that I was highlighting was that message rebuilding without headers prevents some control information to be passed to the sending endpoint. Some headers are used for controlling the target partition for example, even if they are not embedded in the payload.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -329,8 +329,8 @@ private SendingHandler(MessageHandler delegate, boolean embedHeaders,
 
 		@Override
 		protected void handleMessageInternal(Message<?> message) throws Exception {
-			Message<?> messageToSend = (this.useNativeEncoding) ? getMessageBuilderFactory().withPayload(message.getPayload()).build() :
-					serializeAndEmbedHeadersIfApplicable(message);
+			Message<?> messageToSend = (this.useNativeEncoding) ? getMessageBuilderFactory().withPayload(message.getPayload())
+					.copyHeaders(message.getHeaders()).build() : serializeAndEmbedHeadersIfApplicable(message);
```
mbogoevici: We can just return the original message here and skip the redundant message re-creation. Will fix on merge.

mbogoevici:
Squashed and merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/e7e799b8d8e7704e8e8e4df02ba28e2819793976

Issue #697, opened 2016-10-26
Schema Server with Postgres


parunruaivo:
Me again :(

Well I was trying to use Postgres instead of MySQL but I'm getting

`org.postgresql.util.PSQLException: Large Objects may not be used in auto-commit mode.`

The table used is

``` sql
CREATE TABLE schema
(
  ID         SERIAL PRIMARY KEY,
  DEFINITION TEXT         NOT NULL,
  FORMAT     VARCHAR(255) NOT NULL,
  SUBJECT    VARCHAR(255) NOT NULL,
  VERSION    INTEGER      NOT NULL
);
```

the schema used was

``` json
{
  "type" : "record",
  "name" : "CustomerCreated",
  "namespace" : "com.payments.mobile.commons.domain.events.customer",
  "fields" : [ {
    "name" : "customerId",
    "type" : "long"
  }, {
    "name" : "firstName",
    "type" : "string"
  }, {
    "name" : "lastName",
    "type" : "string"
  }, {
    "name" : "emailAddress",
    "type" : "string"
  }, {
    "name" : "occurredOn",
    "type" : {
      "type" : "record",
      "name" : "Date",
      "namespace" : "java.util",
      "fields" : [ ]
    }
  } ]
}
```

Debugging the line

``` java
List<Schema> registeredEntities = this.repository.findBySubjectAndFormatOrderByVersion(
                schema.getSubject(), schema.getFormat());
```

and checking schema.getDefinition() it had the schema with a lot `\r\n`.

Thanks

mbogoevici:
Not sure if the formatting was the problem - I think it's the LOB mapping. That being said, we definitely shouldn't be in auto-commit mode I think - this hints at the operation being non-transactional.

mbogoevici:
(Thanks for posting these by the way - this is very valuable)

viniciusccarvalho:
Postgresql has a requirement on LOB being only used inside transactions. So even for read-only methods such as findbyXXX we need to add `@Transactional`. Thanks for bring this to attention @parunruaivo 

Reference: https://www.postgresql.org/docs/current/static/lo-interfaces.html

mbogoevici:
Fixed via https://github.com/spring-cloud/spring-cloud-stream/commit/18c894689c32e17769b929720b36390da8107ca6

PR #700, opened 2016-11-01
Fixes for GH-696 and GH-697

### LINKED ISSUE #696: Schema is a reserved keyword in MySQL 5.7
Using schema registry server with MySQL 5.7 throws 
`com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'schema`

viniciusccarvalho:
+ Renamed Table from Schema to SCHEMA_REPOSITORY to avoid potential reserved word clashes with some databases

+ Added `@EntityScan` annotation so managed entities can be scanned if the schema server is being used embedded inside another app and not as standalone

+ Postgresql GH-696 requires transactional semantics even for reading operations when using Lob columns

+ Removed dependency to h2 database. A database driver is now explicitly required when running the server

Resolves #696

### DIFF
--- a/spring-cloud-stream-schema-server/pom.xml
+++ b/spring-cloud-stream-schema-server/pom.xml
@@ -27,6 +27,7 @@
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
 			<version>1.4.192</version>
+			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.apache.avro</groupId>
--- a/spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/config/SchemaServerConfiguration.java
+++ b/spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/config/SchemaServerConfiguration.java
@@ -19,8 +19,10 @@ package org.springframework.cloud.stream.schema.server.config;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.springframework.boot.autoconfigure.domain.EntityScan;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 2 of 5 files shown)


spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/config/SchemaServerConfiguration.java
```diff
@@ -34,6 +36,7 @@
 @Configuration
 @EnableJpaRepositories(basePackageClasses = SchemaRepository.class)
 @EnableConfigurationProperties(SchemaServerProperties.class)
+@EntityScan(basePackageClasses = Schema.class)
```
ilayaperumalg: This can be removed as the fix is already available here: https://github.com/spring-cloud/spring-cloud-stream/commit/6c9976b77f5e54771516b62ed971382c3ffe36e5

viniciusccarvalho:
Done, there's a fix for GH-696 as well regarding Transactions on read-mode on postgres

spring-cloud-stream-schema-server/pom.xml
```diff
@@ -27,6 +27,7 @@
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
 			<version>1.4.192</version>
+			<scope>test</scope>
```
mbogoevici: Is this modification intentional? - i.e. do we want to require them to supply a database? I suppose that's okay.

mbogoevici: But not as as a backwards breaking change.

mbogoevici:
Squashed, rebased, merged via https://github.com/spring-cloud/spring-cloud-stream/commit/18c894689c32e17769b929720b36390da8107ca6

PR #703, opened 2016-11-02
Enable child application web context when set

### LINKED ISSUE #702: Applications that are part of an aggregate should be able to expose web endpoints
If the child applications of the aggregate application require web endpoints to be enabled (ex: aggregate source application being http source), those endpoints need to be enabled.

Here is the SO issue for more info: http://stackoverflow.com/questions/40295959/springcloudstream-httpsource-is-returning-404

ilayaperumalg:
- Make sure to disable the aggregate application's web context when any of the child application web context is enabled
   - This will allow the child applications to open up web endpoints (ex: http source)
 - Add tests

Resolves #702


mbogoevici:
As discussed with @ilayaperumalg , we don't want to go with individual web contexts per child app, but be able to forward the requests from the parent web context to the children.

mbogoevici:
Closing this PR until one based on the discussed changes is available.

PR #718, opened 2016-11-22
 Fix wrong property for the root path of the schema server

making:


spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/controllers/ServerController.java
```diff
@@ -44,7 +44,7 @@
  * @author Vinicius Carvalho
  */
 @RestController
-@RequestMapping(path = "${spring.cloud.stream.schema.server.prefix:}")
```
mbogoevici: Not very clear what is the intent here? Just changing the name of the property?

making: `spring.cloud.stream.schema.server.path` is mentioned in the [doc](
http://docs.spring.io/spring-cloud-stream/docs/Brooklyn.SR1/reference/htmlsingle/#_schema_registry_server) and defined in [`SchemaServerProperties`](https://github.com/making/spring-cloud-stream/blob/6d8cb58bdefa3ab49d562954b652e61a02f84aa5/spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/config/SchemaServerProperties.java#L33) but unavailable.

mbogoevici: Gotcha, thanks for catching this!

ilayaperumalg:
LGTM, merged.

PR #720, opened 2016-11-23
Remove logback config from spring-cloud-stream-binder-test

### LINKED ISSUE #704: Remove logback.xml in spring-cloud-stream-binder-test
Since we have used `spring-cloud-stream-binder-test` in our  test projects our logback-spring.xml don't work anymore. As soon as a logback.xml is in the classpath, the logback-spring.xml meachnism is no longer recognized. 
Is it possible to remove the logback.xml?

ilayaperumalg:
Resolves #704


mbogoevici:
There's a bit more to this then just removing the logback file I believe.  AFAIR  it exists so we can set logging levels during tests. I'm okay with removing it, but we should do something to set log levels.

ilayaperumalg:
I am under the impression that the subsequent projects which use the `spring-cloud-stream-binder-test` would specify `boot` supported `log-level` properties explicitly.

mbogoevici:
Well, that's additive to this setting I guess. Wouldn't just removing the file alter the default behaviour? Logging levels, logging formats, etc. - we still want to preserve some of that, I guess?

ilayaperumalg:
I pushed a change to set some logging defaults for the `spring-cloud-stream-binder-test` project. I noticed that the other entries in the existing `logback.xml` doesn't seem to be valid in the context of this project.

mbogoevici:
Merged w/https://github.com/spring-cloud/spring-cloud-stream/commit/5ea86e2f53f062b3868586f76e18d25c596fa005

PR #725, opened 2016-11-30
Improve application aggregation and testability

### LINKED ISSUE #723: Improve testability of aggregate applications
See http://stackoverflow.com/questions/40868331/unit-test-for-aggregated-spring-cloud-stream-application for reference.

Currently, it is possible to get ahold of the output channel of the aggregate, but not of the input channel, in case the source is a processor.

mbogoevici:
Fixes #723

- Add support for registering an `AggregateApplication` bean for accessing the
  components from underlying subcontexts
- Testing support and samples

### DIFF
--- a/spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/aggregate/AggregateTestWithMain.java
+++ b/spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/aggregate/AggregateTestWithMain.java
@@ -1,0 +1,61 @@
+package org.springframework.cloud.stream.test.aggregate;
+
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Test;
+
+import org.springframework.cloud.stream.aggregate.AggregateApplication;
+import org.springframework.cloud.stream.aggregate.AggregateApplicationBuilder;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.messaging.Processor;
+import org.springframework.cloud.stream.test.binder.MessageCollector;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.integration.annotation.Transformer;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class AggregateTestWithMain {
+
+	@Test
+	public void testAggregateApplication() throws InterruptedException {
+		// emulate a main method
+		ConfigurableApplicationContext context = new AggregateApplicationBuilder().from(UppercaseProcessor.class)
+				.namespace("upper").to(SuffixProcessor.class).namespace("suffix").run();
+
+		AggregateApplication aggregateAccessor = context.getBean(AggregateApplication.class);
+		MessageCollector messageCollector = context.getBean(MessageCollector.class);
+		Processor uppercaseProcessor = aggregateAccessor.getBinding(Processor.class, "upper");
+		Processor suffixProcessor = aggregateAccessor.getBinding(Processor.class, "suffix");
+		uppercaseProcessor.input().send(MessageBuilder.withPayload("Hello").build());
+		Message<?> receivedMessage = messageCollector.forChannel(suffixProcessor.output()).poll(1, TimeUnit.SECONDS);
+		assertThat(receivedMessage).isNotNull();
+		assertThat(receivedMessage.getPayload()).isEqualTo("HELLO WORLD!");
+		context.close();
+	}
+
+	@Configuration
+	@EnableBinding(Processor.class)
+	public static class UppercaseProcessor {
+
+		@Transformer(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT)
+		public String transform(String in) {
+			return in.toUpperCase();
+		}
+	}
+
+	@Configuration
+	@EnableBinding(Processor.class)
+	public static class SuffixProcessor {
+
+		@Transformer(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT)
+		public String transform(String in) {
+			return in + " WORLD!";
+		}
+	}
+}
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
@@ -1,98 +1,23 @@
-/*
- * Copyright 2015-2016 the original author or authors.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
 package org.springframework.cloud.stream.aggregate;
 
-import java.util.LinkedHashMap;
-import java.util.Map.Entry;
-
-import org.springframework.boot.Banner.Mode;
-import org.springframework.boot.builder.SpringApplicationBuilder;
-import org.springframework.context.ConfigurableApplicationContext;
-import org.springframework.integration.channel.DirectChannel;
-import org.springframework.messaging.SubscribableChannel;
-
 /**
- * Class that is responsible for embedding apps using shared channel registry.
+ * Handle to an aggregate application, providing access to the underlying
+ * components of the aggregate (e.g. bindable instances).
  *
  * @author Marius Bogoevici
- * @author Ilayaperumal Gopinathan
- * @author Venil Noronha
  */
-abstract class AggregateApplication {
+public interface AggregateApplication {
 
-	private static final String SPRING_CLOUD_STREAM_INTERNAL_PREFIX = "spring.cloud.stream.internal";
-
-	private static final String CHANNEL_NAMESPACE_PROPERTY_NAME =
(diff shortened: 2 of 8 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplication.java
```diff
@@ -1,98 +1,23 @@
...
 /**
- * Class that is responsible for embedding apps using shared channel registry.
+ * Handle to an aggregate application, providing access to the underlying
+ * components of the aggregate (e.g. bindable instances).
  *
  * @author Marius Bogoevici
- * @author Ilayaperumal Gopinathan
- * @author Venil Noronha
  */
-abstract class AggregateApplication {
```
mbogoevici: The original class was in fact renamed to `AggregateApplicationUtils`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -325,13 +368,38 @@ void embed() {
...
+				}
+				catch (Exception e) {
+					throw new IllegalStateException(
+							"Error while trying to register the aggregate bound interface '"
+									+ bindableProxyName + "' into namespace '" + this.getNamespace() + "'",
+							e);
+				}
+			}
+			// Register metrics if JMX enabled and exporter avalable
+			if (BeanFactoryUtils.beansOfTypeIncludingAncestors(AggregateApplicationBuilder.this.parentContext,
```
mbogoevici: Note: the code was just wrapped in a check, since previously it would fail if JMX was not enabled.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -35,21 +38,25 @@
...
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.core.env.PropertySources;
 import org.springframework.integration.monitor.IntegrationMBeanExporter;
 import org.springframework.util.StringUtils;
 
 /**
- * Application builder for {@link AggregateApplication}.
+ * Application builder for {@link AggregateApplicationUtils}.
```
markfisher: I think this change was unintentional.

mbogoevici: Yeah, it inadvertently picked up the renaming.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -35,21 +38,25 @@
...
  *
  * @author Dave Syer
  * @author Ilayaperumal Gopinathan
  * @author Marius Bogoevici
  * @author Venil Noronha
  */
-public class AggregateApplicationBuilder {
+@EnableBinding
```
mbogoevici: Not needed, can be removed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -154,16 +184,25 @@ public ConfigurableApplicationContext run(String[] parentArgsFromRun) {
...
 			appsToEmbed.put(appToEmbed, appConfigurer.namespace);
 			appConfigurers.put(appConfigurer, appConfigurer.namespace);
 		}
-		this.parentContext = AggregateApplication.createParentContext(this.parentSources.toArray(new Object[0]),
+		if (this.parentContext == null) {
+			this.parentContext = AggregateApplicationUtils.createParentContext(this.parentSources.toArray(new Object[0]),
 					this.parentArgs.toArray(new String[0]), selfContained(), this.webEnvironment, this.headless);
+		}
+		else {
+			if (BeanFactoryUtils.beansOfTypeIncludingAncestors(this.parentContext, SharedChannelRegistry.class)
```
ilayaperumalg: Is the reason to use `beansOfTypeIncludingAncestors` here is to get `SharedChannelRegistry` from other user provided sources' ancestors? 

I also wonder why do we need `ConditionalOnMissingBean` on `SharedChannelRegistry` at the first place (as we don't expose any contract over SharedChannelRegistry).

mbogoevici: This branch does not trigger the `ParentContext` configuration. The `parentContext` is the Spring context in which the `AggregateApplicationBuilder` is installed. I agree that we might want to skip the check if the bean already exists, but I would say it is more prudent not to (and do the check), even if it is an undocumented extension.

ilayaperumalg: At least for testing scenario where the `parentContext` is set using `ApplicationContextAware`, this is needed I guess.

ilayaperumalg: or, anywhere the `AggregateApplication` is used as a `bean` instead.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
ilayaperumalg: Why do we need `@EnableBinding` here?

spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/aggregate/AggregateTestWithMain.java
```diff
@@ -0,0 +1,61 @@
...
+		Processor uppercaseProcessor = aggregateAccessor.getBinding(Processor.class, "upper");
+		Processor suffixProcessor = aggregateAccessor.getBinding(Processor.class, "suffix");
+		uppercaseProcessor.input().send(MessageBuilder.withPayload("Hello").build());
+		Message<?> receivedMessage = messageCollector.forChannel(suffixProcessor.output()).poll(1, TimeUnit.SECONDS);
+		assertThat(receivedMessage).isNotNull();
+		assertThat(receivedMessage.getPayload()).isEqualTo("HELLO WORLD!");
+		context.close();
+	}
+
+	@Configuration
```
mbogoevici: They're in a different package and there's not much code to be shared, so I don't think that it would help a lot, but it instead it would make the test harder to read.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
mbogoevici: We can remove it, but if the builder is installed as a bean, this adds `@EnableBinding` as a configuration in the parent context. The alternative would be to add `@EnableBinding` to the main `@SpringBootApplication` itself as we do in the tests. I suppose that's a little cleaner.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
mbogoevici: In theory, we might use this hook to configure a `SharedChannelRegistry` as a bean at some point. I'd rather be more defensive and test.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
ilayaperumalg: The `ParentConfiguration` (which already has `@EnableBinding`) always gets invoked irrespective of whether it's a bean or main type.  Hence, I think we can safely remove this.

mbogoevici: > The ParentConfiguration (which already has @EnableBinding) always gets invoked irrespective of whether it's a bean or main type.

That is not correct. If the `parentContext` is set, then we don't create a new parent context.

ilayaperumalg: yeah, I missed that. It is just that we set the `parentSources` with the `ParentConfiguration` but that is not used at all in case of `bean` type obviously :-)

ilayaperumalg: Even in this case, do the `child` applications inside the aggregate application always have the `@EnableBinding`?  In that case, I guess that would be sufficient for the aggregate application.

I remember having this discussion before: what happens if we remove `@EnableBinding` from the `ParentConfiguration` completely? I don't see any issue with that. what do you think?

mbogoevici: I personally lean towards leaving it here, since we can have a more consistent application model (i.e. skip adding `@EnableBinding` explicitly in the aggregating application). Otherwise we'd need to provide some rules as to when to do it and when not.

mbogoevici: re: https://github.com/spring-cloud/spring-cloud-stream/pull/725#discussion_r90664756  - that would mean that each subcontext creates its own copy of a ChannelBindingService and so on - in this setup, they're shared from the parent context. I'd rather keep the latter.

ilayaperumalg: No, I think we are good here and this PR is ready for the merge.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 6c861135a3d4b4efb42d594af5df74c665598c8b

PR #727, opened 2016-12-01
Support flexible bound element types

### LINKED ISSUE #519: Generalize binding outside message channels

mbogoevici:
Fixes #519

Introduces some internal changes to the framework allowing the use
of other types than MessageChannel/SubscribableChannel as bindable
types (e.g. Flux, Observable, KStream, etc.)

- Modify BinderFactory to allow the retrieval and lookup of a
  binder not only by name, but also by BoundElement type
- Subsequent changes to ChannelBindingService and tests to account
  for the modified signature
- Introduce BoundElementFactory as the contract for creating bound
  elements


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -56,90 +57,105 @@
...
 
-	private final Map<String, Binding<MessageChannel>> producerBindings = new HashMap<>();
+	private final Map<String, Binding<?>> producerBindings = new HashMap<>();
 
-	private final Map<String, List<Binding<MessageChannel>>> consumerBindings = new HashMap<>();
+	private final Map<String, List<Binding<?>>> consumerBindings = new HashMap<>();
 
-	public ChannelBindingService(ChannelBindingServiceProperties channelBindingServiceProperties,
-			BinderFactory<MessageChannel> binderFactory) {
+	public ChannelBindingService(
```
ilayaperumalg: Should this class be renamed to something more general? may be just `BindingService`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -56,90 +57,105 @@
...
+			ChannelBindingServiceProperties channelBindingServiceProperties,
```
ilayaperumalg: Can we rename `ChannelBindingServiceProperties` as well to something like `BindingServiceProperties`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -56,90 +57,105 @@
...
 	}
 
 	@SuppressWarnings("unchecked")
-	public Binding<MessageChannel> bindProducer(MessageChannel outputChannel, String outputChannelName) {
-		String channelBindingTarget = this.channelBindingServiceProperties.getBindingDestination(outputChannelName);
-		Binder<MessageChannel, ?, ProducerProperties> binder =
-				(Binder<MessageChannel, ?, ProducerProperties>) getBinderForChannel(outputChannelName);
-		ProducerProperties producerProperties = this.channelBindingServiceProperties.getProducerProperties(outputChannelName);
+	public <T> Binding<T> bindProducer(T output, String outputName) {
+		String channelBindingTarget = this.channelBindingServiceProperties
```
ilayaperumalg: channelBindingTarget -> bindingTarget

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -56,90 +57,105 @@
...
-	public void unbindProducers(String outputChannelName) {
-		Binding<MessageChannel> binding = this.producerBindings.remove(outputChannelName);
+	public void unbindProducers(String outputName) {
+		Binding<?> binding = this.producerBindings.remove(outputName);
 		if (binding != null) {
 			binding.unbind();
 		}
 		else if (log.isWarnEnabled()) {
-			log.warn("Trying to unbind channel '" + outputChannelName + "', but no binding found.");
+			log.warn("Trying to unbind channel '" + outputName
```
ilayaperumalg: `channel` name reference can be removed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
markfisher: +1 to Ilaya's suggestion (as well as removing all mentions of "channel" from the JavaDoc), but let's take the opportunity to rename the "Service" part as well. I'm thinking... will post back if I have some ideas.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: I've considered that, but it's used by Spring Cloud Bus, so not in this version.

markfisher: might be good to go ahead with `@Deprecated` though?

mbogoevici: I'm not sure deprecated is the right term here, as there is no replacement for it. Clearly using  it like that was a bad idea, but there's not much we can do here.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: We can use a different name for the service. I've kept it for symmetry with the properties, but in the end that makes sense.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
```diff
@@ -56,90 +57,105 @@
...
+			Binding<T> binding = binder.bindConsumer(target,
+					channelBindingServiceProperties.getGroup(inputName), input,
+					consumerProperties);
 			bindings.add(binding);
 		}
-		this.consumerBindings.put(inputChannelName, bindings);
-		return bindings;
+		bindings = Collections.unmodifiableCollection(bindings);
+		this.consumerBindings.put(inputName, new ArrayList<Binding<?>>(bindings));
+		return Collections.unmodifiableCollection(bindings);
```
mbogoevici: Ok. Will fix.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: Ok, makes sense.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ChannelBindingService.java
mbogoevici: Actually, deprecation might be the right thing to do here as it indicates that this shouldn't be used directly.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BoundElementFactory.java
```diff
@@ -19,26 +19,35 @@
...
- * Defines methods to create/configure the {@link org.springframework.messaging.MessageChannel}s defined
- * in {@link org.springframework.cloud.stream.annotation.EnableBinding}.
+ * Defines methods to create/configure the bound elements defined by
+ * {@link org.springframework.cloud.stream.annotation.EnableBinding}.
+ *
+ * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
  */
-public interface BindableChannelFactory {
+public interface BoundElementFactory {
```
markfisher: I wonder if `BindingTarget` might be a better name for `BoundElement` throughout? That is how we refer to this same "element" elsewhere in the codebase already, right?.

mbogoevici: An earlier version used 'BindingTargetFactory' so it wouldn't be hard to change back. I liked 'BoundElement' for some reason, but I'm fine with 'BindingTarget'.

markfisher: if you prefer 'BoundElement' that's fine, but in that case we should probably make sure to update the other occurrences of `bindingTarget` for consistency (could be a subsequent PR).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -106,37 +110,59 @@ else if (!StringUtils.hasText(this.defaultBinder)) {
 					}
 				}
 				if (defaultCandidateConfigurations.size() == 1) {
-					this.defaultBinder = defaultCandidateConfigurations.iterator().next();
-					configurationName = this.defaultBinder;
+					this.defaultBindersPerBoundType.put(boundElementType.getName(),
+							defaultCandidateConfigurations.iterator().next());
```
mbogoevici: no, that's just the result of moving some code around and inlining. what you're suggesting makes more sense

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -106,37 +110,59 @@ else if (!StringUtils.hasText(this.defaultBinder)) {
...
+							Binder<Object, ?, ?> binderInstance = getBinderInstance(defaultCandidateConfiguration);
+							Class<?> binderType = GenericsUtils.getParameter(binderInstance.getClass(), Binder.class, 0);
+							if (binderType.isAssignableFrom(boundElementType)) {
+								candidatesForBindableType.add(defaultCandidateConfiguration);
+							}
+						}
+						if (candidatesForBindableType.size() == 1) {
+							this.defaultBindersPerBoundType.put(boundElementType.getName(),
+									candidatesForBindableType.iterator().next());
+							configurationName = this.defaultBindersPerBoundType.get(boundElementType.getName());
```
markfisher: same thing here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -106,37 +110,59 @@ else if (!StringUtils.hasText(this.defaultBinder)) {
...
+						} else if (candidatesForBindableType.size() > 1) {
+							throw new IllegalStateException(
+									"A default binder has been requested, but there is more than one binder available for '"
+											+ boundElementType.getName() + "' : "
+											+ StringUtils.collectionToCommaDelimitedString(defaultCandidateConfigurations)
```
markfisher: I think this one should only print out the `candidatesForBindableType` collection rather than the full `defaultCandidateConfigurations` collection.

mbogoevici: ok

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -57,66 +57,67 @@
...
 		}
-		//ignore
-		return messageChannel;
+		return null;
 	}
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-		Assert.notNull(BindableProxyFactory.this.channelFactory, "Channel Factory cannot be null");
+		Assert.notNull(BindableProxyFactory.this.boundElementFactories, "'boundElementFactories' cannot be null");
```
markfisher: don't need both of these lines... `Assert.notEmpty` will also check if `null`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -127,30 +128,33 @@ public void doWith(Method method) throws IllegalArgumentException {
...
 		});
 	}
 
-	private void validateChannelType(Class<?> channelType) {
-		Assert.isTrue(SubscribableChannel.class.equals(channelType) || MessageChannel.class.equals(channelType),
-				"A bound channel should be either a '" + MessageChannel.class.getName() + "', " +
-						" or a '" + SubscribableChannel.class.getName() + "'");
+	private BindingTargetFactory getBoundElementFactory(Class<?> boundElementType) {
+		for (BindingTargetFactory factory : boundElementFactories) {
+			if (factory.canCreate(boundElementType)) {
```
markfisher: do we enforce only one factory per-type? (and if not, do we intend to support more than one per type?... this would always return the first one that returns `true`)

mbogoevici: It's a reasonable assumption to have only one factory per type, tbh.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -177,65 +181,67 @@ public boolean isSingleton() {
 	}
 
 	@Override
-	public void bindInputs(ChannelBindingService channelBindingService) {
+	public void bindInputs(BindingService bindingService) {
 		if (log.isDebugEnabled()) {
 			log.debug(String.format("Binding inputs for %s:%s", this.channelNamespace, this.type));
 		}
-		for (Map.Entry<String, ChannelHolder> channelHolderEntry : this.inputHolders.entrySet()) {
+		for (Map.Entry<String, BoundElementHolder> channelHolderEntry : this.inputHolders.entrySet()) {
```
markfisher: should change the `channelHolderEntry` var name

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -127,30 +128,33 @@ public void doWith(Method method) throws IllegalArgumentException {
...
+			}
+		}
+		throw new IllegalStateException("No factory found for bound element type: " + boundElementType.getName());
 	}
 
-	private MessageChannel locateSharedChannel(String name) {
-		return this.sharedChannelRegistry != null ?
-				this.sharedChannelRegistry.get(getNamespacePrefixedChannelName(name)) : null;
+	private MessageChannel locateSharedBoundElement(String name, Class<?> boundElementType) {
+		return MessageChannel.class.isAssignableFrom(boundElementType) && this.sharedChannelRegistry != null
```
mbogoevici: In theory, yes, but it's beside the scope of this PR. I really didn't mean to support aggregation for other types.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -177,65 +181,67 @@ public boolean isSingleton() {
...
 
 	@Override
-	public void bindOutputs(ChannelBindingService channelBindingService) {
+	public void bindOutputs(BindingService bindingService) {
 		if (log.isDebugEnabled()) {
 			log.debug(String.format("Binding outputs for %s:%s", this.channelNamespace, this.type));
 		}
-		for (Map.Entry<String, ChannelHolder> channelHolderEntry : this.outputHolders.entrySet()) {
-			ChannelHolder channelHolder = channelHolderEntry.getValue();
+		for (Map.Entry<String, BoundElementHolder> channelHolderEntry : this.outputHolders.entrySet()) {
```
markfisher: same here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -177,65 +181,67 @@ public boolean isSingleton() {
...
 	}
 
 	@Override
-	public void unbindInputs(ChannelBindingService channelBindingService) {
+	public void unbindInputs(BindingService bindingService) {
 		if (log.isDebugEnabled()) {
 			log.debug(String.format("Unbinding inputs for %s:%s", this.channelNamespace, this.type));
 		}
-		for (Map.Entry<String, ChannelHolder> channelHolderEntry : this.inputHolders.entrySet()) {
+		for (Map.Entry<String, BoundElementHolder> channelHolderEntry : this.inputHolders.entrySet()) {
```
markfisher: ditto

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -177,65 +181,67 @@ public boolean isSingleton() {
...
 	}
 
 	@Override
-	public void unbindOutputs(ChannelBindingService channelBindingService) {
+	public void unbindOutputs(BindingService bindingService) {
 		if (log.isDebugEnabled()) {
 			log.debug(String.format("Unbinding outputs for %s:%s", this.channelNamespace, this.type));
 		}
-		for (Map.Entry<String, ChannelHolder> channelHolderEntry : this.outputHolders.entrySet()) {
+		for (Map.Entry<String, BoundElementHolder> channelHolderEntry : this.outputHolders.entrySet()) {
```
markfisher: ..and here

mbogoevici: Can we just have a separate PR for renaming things, as I've suggested previously, or is that a blocker. There are a few other places, far beyond the scope of this PR that need to be changed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -36,110 +37,125 @@
...
+			Binding<T> binding = binder.bindConsumer(target,
+					channelBindingServiceProperties.getGroup(inputName), input,
+					consumerProperties);
 			bindings.add(binding);
 		}
-		this.consumerBindings.put(inputChannelName, bindings);
-		return bindings;
+		bindings = Collections.unmodifiableCollection(bindings);
+		this.consumerBindings.put(inputName, new ArrayList<Binding<?>>(bindings));
+		return Collections.unmodifiableCollection(bindings);
```
markfisher: can just return `bindings` here since it's already unmodifiable?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -36,110 +37,125 @@
...
+		List<Binding<?>> bindings = this.consumerBindings.remove(inputName);
 		if (bindings != null && !CollectionUtils.isEmpty(bindings)) {
-			for (Binding<MessageChannel> binding : bindings) {
+			for (Binding<?> binding : bindings) {
 				binding.unbind();
 			}
 		}
 		else if (log.isWarnEnabled()) {
-			log.warn("Trying to unbind channel '" + inputChannelName + "', but no binding found.");
+			log.warn("Trying to unbind channel '" + inputName
```
markfisher: should replace "channel" with "target" in this message

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -36,110 +37,125 @@
...
-	public void unbindProducers(String outputChannelName) {
-		Binding<MessageChannel> binding = this.producerBindings.remove(outputChannelName);
+	public void unbindProducers(String outputName) {
+		Binding<?> binding = this.producerBindings.remove(outputName);
 		if (binding != null) {
 			binding.unbind();
 		}
 		else if (log.isWarnEnabled()) {
-			log.warn("Trying to unbind channel '" + outputChannelName + "', but no binding found.");
+			log.warn("Trying to unbind channel '" + outputName
```
markfisher: ditto

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingTargetFactory.java
```diff
@@ -19,26 +19,35 @@
...
+	 * @return bindable element
 	 */
-	SubscribableChannel createInputChannel(String name);
+	Object createInput(String name);
 
 	/**
-	 * Create an output {@link SubscribableChannel} that will be bound via
-	 * the message channel {@link org.springframework.cloud.stream.binder.Binder}.
+	 * Create an output {@link SubscribableChannel} that will be bound via the message
+	 * channel {@link org.springframework.cloud.stream.binder.Binder}.
```
markfisher: should not mention "message channel" here (looks like the corresponding input method's comment was genericized, but this one was not)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingTargetFactory.java
```diff
@@ -19,26 +19,35 @@
...
 	 * @param name name of the message channel
```
markfisher: should not mention channel in the `@param` or `@return` either

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BoundSubscribableChannelFactory.java
```diff
@@ -20,29 +20,30 @@
...
 /**
- * Class that {@link BindableProxyFactory} uses to create and configure message channels.
+ * An implementation of {@link BindingTargetFactory} for creating {@link SubscribableChannel}s.
  *
  * @author Marius Bogoevici
  * @author David Syer
  * @author Ilayaperumal Gopinathan
  */
-public class DefaultBindableChannelFactory implements BindableChannelFactory {
+public class BoundSubscribableChannelFactory extends AbstractBindingTargetFactory<SubscribableChannel> {
```
markfisher: maybe this class should be named `SubscribableChannelBindingTargetFactory`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/SingleChannelBindable.java
```diff
@@ -42,12 +42,12 @@ public SingleChannelBindable(String name, MessageChannel messageChannel) {
 	}
 
 	@Override
-	public void bindOutputs(ChannelBindingService adapter) {
+	public void bindOutputs(BindingService adapter) {
```
markfisher: might as well change this "adapter" param name (at some point)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/reflection/GenericsUtils.java
```diff
@@ -0,0 +1,44 @@
+package org.springframework.cloud.stream.reflection;
+
+import org.springframework.core.ResolvableType;
+
+/**
+ * Internal utilities for handling generics.
+ *
+ * @author Marius Bogoevici
+ */
+public class GenericsUtils {
```
markfisher: nit pick: this could be `abstract`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/reflection/GenericsUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+	public static Class<?> getParameter(Class<?> evaluatedClass, Class<?> interfaceClazz, int index) {
```
markfisher: should be `getParameterType`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
markfisher: sure thing... I was just tracking them as I encountered, but definitely not necessary to address all of those changes within the scope of this PR if it's easier to submit a subsequent PR that addresses just those naming changes... and I can review it! ;)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -57,7 +57,7 @@
 
 	private volatile ConfigurableApplicationContext context;
 
-	private Map<String, String> defaultBindersPerBoundType = new HashMap<>();
+	private Map<String, String> defaultBinderForBindingTargetType = new HashMap<>();
```
markfisher: actually this being singular doesn't sound like a collection... why not just "defaultBinders"? (with a comment that mentions the key of the map is the type)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/SingleBindingTargetBindable.java
```diff
@@ -21,38 +21,36 @@
 import java.util.HashSet;
 import java.util.Set;
 
-import org.springframework.messaging.MessageChannel;
-
 /**
- * A {@link Bindable} component that wraps a generic channel. Useful for binding channels outside the
- * {@link org.springframework.cloud.stream.annotation.Input} and {@link org.springframework.cloud.stream.annotation.Output}
- * annotated interfaces.
+ * A {@link Bindable} component that wraps a generic channel. Useful for binding channels
```
markfisher: still mentions "channels" ;)

mbogoevici:
Closing this and will issue a new PR with a comprehensive change set.

PR #728, opened 2016-12-01
Add support for defaults at binding level

### LINKED ISSUE #446: Support default properties for bindings
Suggested solution: add `@Value("${spring.cloud.stream.rabbit.binder.default.autoBindDlq}")` on binding properties and add `additional-spring-configuration-metadata.json` for completion.

mbogoevici:
Fixes #446

Supports properties prefixed with `spring.cloud.stream.default`
for specifying default values for large number of channels or
for dynamically generated channels.

Note: this does not cover binding specific properties,
as the property objects are created under binder's control
but the same technique can be applied there.

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -885,17 +885,21 @@ Default: false.
 Binding properties are supplied using the format `spring.cloud.stream.bindings.<channelName>.<property>=<value>`.
 The `<channelName>` represents the name of the channel being configured (e.g., `output` for a `Source`).
 
+To avoid repetition, Spring Cloud Stream supports setting values for all channels, in the format `spring.cloud.stream.default.<property>=<value>`.
+
 In what follows, we indicate where we have omitted the `spring.cloud.stream.bindings.<channelName>.` prefix and focus just on the property name, with the understanding that the prefix will be included at runtime.
 
 ==== Properties for Use of Spring Cloud Stream
 
-The following binding properties are available for both input and output bindings and
-must be prefixed with `spring.cloud.stream.bindings.<channelName>.`.
+The following binding properties are available for both input and output bindings and must be prefixed with `spring.cloud.stream.bindings.<channelName>.`, e.g. `spring.cloud.stream.bindings.input.destination=ticktock`.
+
+Default values can be overridden by using the prefix `spring.cloud.stream.default`, e.g. `spring.cloud.stream.default.contentType=application/json`.
 
 destination::
     The target destination of a channel on the bound middleware (e.g., the RabbitMQ exchange or Kafka topic).
     If the channel is bound as a consumer, it could be bound to multiple destinations and the destination names can be specified as comma separated String values.
     If not set, the channel name is used instead.
+    The default value of this property cannot be overridden.
 group::
     The consumer group of the channel.
 Applies only to inbound bindings.
@@ -915,7 +919,9 @@ Default: null (the default binder will be used, if one exists).
 
 ==== Consumer properties
 
-The following binding properties are available for input bindings only and must be prefixed with `spring.cloud.stream.bindings.<channelName>.consumer.`.
+The following binding properties are available for input bindings only and must be prefixed with `spring.cloud.stream.bindings.<channelName>.consumer.`, e.g. `spring.cloud.stream.bindings.input.consumer.concurrency=3`.
+
+Default values can be overridden by using the prefix `spring.cloud.stream.default.consumer`, e.g. `spring.cloud.stream.default.consumer.headerMode=raw`.
 
 concurrency::
   The concurrency of the inbound consumer.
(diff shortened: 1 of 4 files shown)


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -960,7 +966,9 @@ Default: `-1`.
 
 ==== Producer Properties
 
-The following binding properties are available for output bindings only and must be prefixed with `spring.cloud.stream.bindings.<channelName>.producer.`.
+The following binding properties are available for output bindings only and must be prefixed with `spring.cloud.stream.bindings.<channelName>.producer.`, e.g. `spring.cloud.stream.bindings.input.producer.partitionKeyExpression=payload.id`.
+
+Default values can be overridden by using the prefix `spring.cloud.stream.default.producer`, e.g. `spring.cloud.stream.default.producer.contentType=payload.id`.
```
ilayaperumalg: do you mean `partitionKeyExpression` instead of `contentType` here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
```diff
@@ -210,6 +181,13 @@ public ProducerProperties getProducerProperties(String outputChannelName) {
...
 		BindingProperties bindingProperties = new BindingProperties();
+		if (applicationContext != null) {
+			RelaxedDataBinder defaultsDataBinder = new RelaxedDataBinder(bindingProperties,
+					"spring.cloud.stream.default");
+			defaultsDataBinder.setDisallowedFields("destination");
+			defaultsDataBinder
+					.bind(new PropertySourcesPropertyValues(applicationContext.getEnvironment().getPropertySources()));
+		}
 		if (this.bindings.containsKey(channelName)) {
 			BeanUtils.copyProperties(this.bindings.get(channelName), bindingProperties);
```
ilayaperumalg: Doesn't this reset the values set by the dataBinder (with `spring.cloud.stream.default` prefix) above? While this is what we need to `override` the `spring.cloud.stream.default` values with the binding properties provided explicitly (via cmdline for instance), the `spring.cloud.stream.default` values will also get overridden with the application `default` values here.

For instance, if the application has `spring.cloud.stream.default.producer.partitionCount=5` that would never be set eventually because of BeanUtils.copyProperties at line:192.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/SourceBindingWithGlobalPropertiesTest.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest(classes = SourceBindingWithGlobalPropertiesTest.TestSource.class, properties = {
+		"spring.cloud.stream.default.contentType=application/json",
+		"spring.cloud.stream.default.producer.requiredGroups=someGroup" })
+public class SourceBindingWithGlobalPropertiesTest {
+
+	@SuppressWarnings("rawtypes")
+	@Autowired
+	private BinderFactory binderFactory;
```
ilayaperumalg: Unused instance variables can be removed.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
mbogoevici: Right.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -885,17 +885,21 @@ Default: false.
...
+
 In what follows, we indicate where we have omitted the `spring.cloud.stream.bindings.<channelName>.` prefix and focus just on the property name, with the understanding that the prefix will be included at runtime.
 
 ==== Properties for Use of Spring Cloud Stream
 
-The following binding properties are available for both input and output bindings and
-must be prefixed with `spring.cloud.stream.bindings.<channelName>.`.
+The following binding properties are available for both input and output bindings and must be prefixed with `spring.cloud.stream.bindings.<channelName>.`, e.g. `spring.cloud.stream.bindings.input.destination=ticktock`.
+
+Default values can be overridden by using the prefix `spring.cloud.stream.default`, e.g. `spring.cloud.stream.default.contentType=application/json`.
```
mbogoevici: Well, they all have default values, which are `empty`. We can use 'set' instead if you think that it's more accurate, but we have to document how to do it, don't you think?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingServiceProperties.java
mbogoevici: Hm, that's a pretty big oversight. Let me rework this.

ilayaperumalg:
The changes look good. Now the `default` properties wouldn't be overridden due to bean copy. One question though: It isn't clearer to me at what phase the `ChannelBindingServiceProperties` obtained via defaults (using the prefix `spring.cloud.stream.default`) gets overridden by the per binding properties.

mbogoevici:
They're overriden during the actual binding, which happens on the instance returned by the map.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as b3a0f2c2e1ab98581f3303ed5d6cc3864adcc1e7

yngwietiger:
So, if I'm understanding correctly, I can't do something like:

spring.cloud.stream.rabbit.binder.default.consumer.autoBindDlq=true
spring.cloud.stream.rabbit.binder.default.consumer.dlqTtl=10000
spring.cloud.stream.rabbit.binder.default.consumer.dlqTtlDeadLetterExchange=

?

I've tried different permutations of this, but nothing works.

Thanks.

PR #729, opened 2016-12-02
Do not check if the Lifecycle argument is null in DefaultBinding

### LINKED ISSUE #607: DefaultBinding with null endpoint fails after AbstractMessageChannelBinder.doBindProducer()
While implementing a SCS binder, dealing with MessageHandler in the producer side of it, if it does not implement the interface Lifecycle, `AbstractMessageChannelBinder` will send a null reference to `DefaultBinding`, which will cause an assertion exception, since the constructor verifies it.

```
public DefaultBinding(String name, String group, T target, Lifecycle endpoint) {
        Assert.notNull(target, "target must not be null");
        Assert.notNull(endpoint, "endpoint must not be null");
        this.name = name;
        this.group = group;
        this.target = target;
        this.endpoint = endpoint;
    }
```

The result is that non lifecycle implementing message handlers as the JmsSendingMessageHandler will not work out of the box.

mbogoevici:
Fixes #607

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
@@ -38,15 +38,22 @@ public class DefaultBinding<T> implements Binding<T> {
 
 	protected final T target;
 
-	protected final Lifecycle endpoint;
+	protected final Lifecycle runningComponent;
 
-	public DefaultBinding(String name, String group, T target, Lifecycle endpoint) {
+	/**
+	 * Creates a new instance.
+	 * @param name the name of the binding target
+	 * @param group the group (only for input targets)
+	 * @param target the binding target
+	 * @param runningComponent a {@link Lifecycle} object that is running while the binding is active
+	 *                         and will be stopped during unbinding
+	 */
+	public DefaultBinding(String name, String group, T target, Lifecycle runningComponent) {
 		Assert.notNull(target, "target must not be null");
-		Assert.notNull(endpoint, "endpoint must not be null");
 		this.name = name;
 		this.group = group;
 		this.target = target;
-		this.endpoint = endpoint;
+		this.runningComponent = runningComponent;
 	}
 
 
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -38,15 +38,22 @@
 
 	protected final T target;
 
-	protected final Lifecycle endpoint;
-
-	public DefaultBinding(String name, String group, T target, Lifecycle endpoint) {
+	protected final Lifecycle runningComponent;
+
+	/**
+	 * Creates a new instance.
```
ilayaperumalg: Can be more descriptive than just `Creates a new instance`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -38,15 +38,22 @@
...
+	 *                         and will be stopped during unbinding
+	 */
+	public DefaultBinding(String name, String group, T target, Lifecycle runningComponent) {
 		Assert.notNull(target, "target must not be null");
-		Assert.notNull(endpoint, "endpoint must not be null");
 		this.name = name;
 		this.group = group;
 		this.target = target;
-		this.endpoint = endpoint;
+		this.runningComponent = runningComponent;
```
ilayaperumalg: While `runningComponent` is a good name in general here, we could as well think of some other name as it is not yet running here. May be just `lifecycleComponent` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
mbogoevici: Do you have any suggestion in particular?

ilayaperumalg: I was thinking something simple description like `Creates an instance that contains information about the binding of the producer/consumer target`

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 8e5689c

PR #735, opened 2016-12-14
Refine the behaviour of the defaults map

### LINKED ISSUE #734: Defaults support breaks some older usage patterns of `BindingProperties`
The immutability of the environment-initialized Map as well as the fact that it overrides the `bindings` property when the context is set breaks the current behaviour in `AbstractBinderTest` and also possibly in Spring Cloud Bus which sets the binding properties directly.

This can be refined as we only need to start looking for defaults when the environment is available and we can also allow setting properties explicitly.

mbogoevici:
Fixes #734

- allow mutability
- use Environment directly for configuration

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
@@ -33,7 +33,10 @@ import org.springframework.cloud.stream.binder.ProducerProperties;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.EnvironmentAware;
 import org.springframework.core.convert.ConversionService;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.Environment;
 import org.springframework.integration.support.utils.IntegrationUtils;
 import org.springframework.util.Assert;
 
@@ -45,8 +48,7 @@ import org.springframework.util.Assert;
  */
 @ConfigurationProperties("spring.cloud.stream")
 @JsonInclude(Include.NON_DEFAULT)
-public class BindingServiceProperties
-		implements ApplicationContextAware, InitializingBean {
+public class BindingServiceProperties implements ApplicationContextAware, EnvironmentAware, InitializingBean {
 
 	private ConversionService conversionService;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/EnvironmentEntryInitializingTreeMap.java
```diff
@@ -52,18 +52,18 @@
...
 	 * 
-	 * @param applicationContext the application context that supplies the property values
+	 * @param environment the environment that supplies the default property values
 	 * @param entryClass the entry class
 	 * @param defaultsPrefix the prefix for initializing the properties
 	 * @param delegate the actual map that stores the values
 	 */
-	public EnvironmentEntryInitializingTreeMap(ConfigurableApplicationContext applicationContext, Class<T> entryClass,
+	public EnvironmentEntryInitializingTreeMap(ConfigurableEnvironment environment, Class<T> entryClass,
 			String defaultsPrefix, Map<String, T> delegate) {
```
ilayaperumalg: Not specific to this PR; but can we make the `defaultsPrefix` to be a constant (with the value `spring.cloud.stream.default`) in this class instead of passing from the invoking method?

mbogoevici: Well, if we made that a prefix then the class implementation would be effectively coupled to it. Suppose we'd want to reuse it for Rabbit, Kafka or other kinds of defaults?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
```diff
@@ -118,11 +120,17 @@ public void setDynamicDestinations(String[] dynamicDestinations) {
...
 		this.applicationContext = (ConfigurableApplicationContext) applicationContext;
-		// override the bindings store with the environment-initializing version if in a
-		// Spring context
-		this.bindings = new EnvironmentEntryInitializingTreeMap<>(this.applicationContext,
-				BindingProperties.class, "spring.cloud.stream.default",
-				new TreeMap<String, BindingProperties>(String.CASE_INSENSITIVE_ORDER));
+	}
+
+	@Override
+	public void setEnvironment(Environment environment) {
```
ilayaperumalg: I am trying to understand how does this different from application context aware invocation (previous implemenation) as both `setEnvironment` and `setApplicationContext` are invoked by `ApplicationContextAwareProcessor` bean post processor?

mbogoevici: This takes into account the possibility that `setApplicationContext()` may be called manually,  (outside a Spring context), e.g. the way it's currently done during test set up. We need the context and the environment for different reasons - if running within a Spring context, they would both be set, otherwise not. This allows not to override the map if there's no actual environment to read defaults from.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as c5b9a9f

PR #742, opened 2016-12-19
Skip `deserializePayloadIfNecessary` only if `originalContentType` is not set

### LINKED ISSUE #741: Allow for binders without embedded headers to process 'originalContentType'
As part of the changes that led to support for native serialization/deserialization we've made the assumption that any payload that is not a `byte[]` is already fully processed and so will skip `deserializePayloadIfNecessary`. However, if the binder is based on endpoints that do their own deserialization and populate headers (as for example RabbitMQ and the upcoming JMS), the current changes skip the part where `originalContentType` is processed. We should be more lenient and allow `deserializePayloadIfNecessary` to execute if we have an `originalContentType` header.

mbogoevici:
Fixes #741

The current code will skip `deserializePayloadIfNecessary` if the
payload is a `byte[]`. There are in fact two conditions that indicate
that a message does not require postprocessing in the binder: that
the payload is not a `byte[]` which indicates that the message has
already been deserialized, and that the `originalContentType` is not
present, or else it'd need to be moved to `contentType`. This fix
adds the secondary condition to the check.

Note: this requires backporting to the 1.1.x branch.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -278,7 +278,8 @@ public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties,
 		@Override
 		@SuppressWarnings("unchecked")
 		protected Object handleRequestMessage(Message<?> requestMessage) {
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Also backported to 1.1.x as e5319fa

PR #743, opened 2016-12-19
Fix `id` type on `SchemaRegistrationResponse`

### LINKED ISSUE #739: `SchemaRegistryClient`, `Schema` and `SchemaRegistrationResponse` have inconsistent types of `id`
Some use `Integer`, some use `long`.  While the value is not likely to exceed the scope of `Integer`, we need to bring some consistency into this.

mbogoevici:
Fixes #739

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/SchemaRegistrationResponse.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/SchemaRegistrationResponse.java
@@ -21,15 +21,15 @@ package org.springframework.cloud.stream.schema;
  */
 public class SchemaRegistrationResponse {
 
-	private long id;
+	private int id;
 
 	private SchemaReference schemaReference;
 
-	public long getId() {
+	public int getId() {
 		return this.id;
 	}
 
-	public void setId(long id) {
+	public void setId(int id) {
 		this.id = id;
 	}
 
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
A trivial question: while it is absolutely fine to have `Integer` or `int` for the `id` type. Doesn't make it look better if we use the primitive type across?

ilayaperumalg:
Also, for the merging of this PR, I guess this can go into `master` as well as `1.1.x`. Given the story #739 is actually `1.1.1.RELEASE` and this PR is issued from `master`, I think that's the plan.

mbogoevici:
Yes, we should merge this to 1.1.x as well.

When you referred to using the primitive type across I suppose you meant here: https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/SchemaRegistryClient.java#L48 ? I suppose we could do that.

ilayaperumalg:
Yeah, Is there any issue updating [the SchemaRegistryClient interface](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/SchemaRegistryClient.java#L48) and the [Schema](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/model/Schema.java#L33) ?

mbogoevici:
The PR's been updated for the `SchemaRegistryClient` interface. `Schema` id should stay nullable, as it is a `@GeneratedValue`

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as f6073ad and also cherry-picked and pushed to 1.1.x 85db822

PR #749, opened 2016-12-21
Add support for schema deletion

### LINKED ISSUE #716: Add support for deleting schemas via HTTP
For testing purposes, it is handy to be able to delete schemas from registry server via HTTP.

ilayaperumalg:
- Add REST endpoints for schema deletion
   - by subject, format, version
   - by subject
   - by id

 - Add tests

Resolves #716

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1280,6 +1280,7 @@ Spring Cloud Stream provides a schema registry server implementation.
 In order to use it, you can simply add the `spring-cloud-stream-schema-server` artifact to your project and use the `@EnableSchemaRegistryServer` annotation, adding the schema registry server REST controller to your application.
 This annotation is intended to be used with Spring Boot web applications, and the listening port of the server is controlled by the `server.port` setting.
 The `spring.cloud.stream.schema.server.path` setting can be used to control the root path of the schema server (especially when it is embedded in other applications).
+The `spring.cloud.stream.schema.server.alllowSchemaDeletion` boolean setting enables the deletion of schema. By default this is disabled.
 
 The schema registry server uses a relational database to store the schemas.
  By default, it uses an embedded database.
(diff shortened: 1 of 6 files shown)


spring-cloud-stream-schema-server/src/main/java/org/springframework/cloud/stream/schema/server/controllers/ServerController.java
```diff
@@ -126,6 +133,49 @@ public ServerController(SchemaRepository repository,
...
+			deleteSchema(schema);
+		}
+		else {
+			throw new SchemaDeletionNotAllowedException();
+		}
+	}
+
+	@RequestMapping(value = "/{subject}", method = RequestMethod.DELETE)
+	public void delete(@PathVariable("subject") String subject) {
+		if (this.schemaServerProperties.isAlllowSchemaDeletion()) {
```
mbogoevici: I think we need to throw an exception here as well.

mbogoevici: I am wondering though whether we shouldn't return a 405 response instead of just throwing an exception.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1280,6 +1280,7 @@ Spring Cloud Stream provides a schema registry server implementation.
 In order to use it, you can simply add the `spring-cloud-stream-schema-server` artifact to your project and use the `@EnableSchemaRegistryServer` annotation, adding the schema registry server REST controller to your application.
 This annotation is intended to be used with Spring Boot web applications, and the listening port of the server is controlled by the `server.port` setting.
 The `spring.cloud.stream.schema.server.path` setting can be used to control the root path of the schema server (especially when it is embedded in other applications).
+The `spring.cloud.stream.schema.server.alllowSchemaDeletion` boolean setting enables the deletion of schema. By default this is disabled.
```
mbogoevici: Need to fix the typoed property through the code and docs.

mbogoevici:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/5d1cef2715c251183db92b6e1fdab0608432b74c

PR #758, opened 2017-01-10
Fix duplicate StreamListener mapping from abstract methods

### LINKED ISSUE #683: Duplicate @StreamListener exception when Implementing an interface
using: 
- spring-boot-starter-parent 1.4.1.RELEASE;
- spring-cloud-dependencies Camden.RELEASE
- spring-cloud-stream-dependencies 1.0.3.RELEASE
- spring-cloud-starter-stream-rabbit

Given that I have the following interfaces:

```
public interface Iceberg {
    String INPUT = "input";

    @Input(Iceberg.INPUT)
    SubscribableChannel input();
}

public interface ISink {
    void sighted(String msg);
}
```

And the following implementation:

```
@EnableBinding(Iceberg.class)
public class Titanic implements ISink {
  @Override
  @StreamListener(Iceberg.INPUT)
  public void sighted(String msg) {
    System.out.println("Oh no! " + msg);
  }
```

Then I get `org.springframework.beans.factory.BeanInitializationException: Duplicate @StreamListener mapping for 'input' on Titanic#sighted[1 args] already existing for Titanic#sighted[1 args]`

But I should see on the console "Oh no! Iceberg sighted" when I send "Iceberg sighted".

ilayaperumalg:
- Add test
 - For StreamListener methods, check if the method modifier is not `bridge` to filter out abstract methods

Resolves #683

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerDuplicateMappingTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerDuplicateMappingTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.annotation.StreamListener;
 import org.springframework.cloud.stream.messaging.Processor;
+import org.springframework.cloud.stream.messaging.Sink;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.messaging.Message;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerDuplicateMappingTests.java
```diff
@@ -38,27 +39,73 @@
...
+	@EnableAutoConfiguration
+	public static class TestDuplicateMappingFromAbstractMethod implements ISink<TestBase> {
+
+		@Override
+		@StreamListener(Sink.INPUT)
+		public void testMethod(TestBase msg) {
+		}
+	}
+
+	public interface ISink<T extends Base> {
```
mbogoevici: We can possibly use a different name for the interface (not a huge fan of `I`-prefixed interfaces :) ).

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerDuplicateMappingTests.java
```diff
@@ -38,27 +39,73 @@
...
+	@EnableBinding(Sink.class)
 	@EnableAutoConfiguration
 	public static class TestDuplicateMapping {
 
-		@StreamListener(Processor.INPUT)
+		@StreamListener(Sink.INPUT)
 		public void receive(Message<String> fooMessage) {
 		}
 
 		@StreamListener(Processor.INPUT)
```
mbogoevici: We should use `Sink.INPUT` here as well, I guess.

mbogoevici:
LGTM in general - a couple of small comments that I can fix on merge.

mbogoevici:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/aa8c14dcd221ae1bf552dc282fe5c931e9ac53e3 and 1.1.x as https://github.com/spring-cloud/spring-cloud-stream/commit/9b655b2a05993e6d682b40dd70c2322b9ab46279

PR #760, opened 2017-01-12
Update doc for PartitionKeyExtractorClass

### LINKED ISSUE #631: PartitionKeyExtractorStrategy instances do not have access to application context
When selecting the partion of a kafka topic using a custom PartitionKeyExtractorStrategy, the instantiated objects are not initialized to use configuration values or beans from the application context. This would be very useful when selecting the partition based on a configurable value.

I have a provided a [demo](https://github.com/spring-cloud/spring-cloud-stream/files/434525/demo.zip). When starting it, the logging shows that the inserted `@Value` is null and not properly initialized.

If this is on purpose the documentation should provide more information what is initialized into the this object.

ilayaperumalg:
- update info on how to access application context

This resolves #631

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1519,22 +1519,9 @@ spring.cloud.stream.bindings.output.producer.partitionCount=5
 
 Once the message key is calculated, the partition selection process will determine the target partition as a value between `0` and `partitionCount - 1`.
 The default calculation, applicable in most scenarios, is based on the formula `key.hashCode() % partitionCount`.
-This can be customized on the binding, either by setting a SpEL expression to be evaluated against the key (via the `partitionSelectorExpression` property) or by setting a `org.springframework.cloud.stream.binder.PartitionSelectorStrategy` implementation (via the `partitionSelectorClass` property).
+This can be customized on the binding, either by setting a SpEL expression to be evaluated against the 'key' (via the `partitionSelectorExpression` property) or by setting a `org.springframework.cloud.stream.binder.PartitionSelectorStrategy` implementation (via the `partitionSelectorClass` property).
 
-Using the specific 'partitionSelectorExpression':
-
-----
-spring.cloud.stream.bindings.output.producer.partitionKeyExpression=payload.id
-spring.cloud.stream.bindings.output.producer.partitionSelectorExpression=payload.group
-----
-
-or using the 'partitionSelectorClass':
-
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1519,6 +1519,22 @@ This can be customized on the binding, either by setting a SpEL expression to be
 
 Additional properties can be configured for more advanced scenarios, as described in the following section.
 
+===== Accessing application context for PartitionKeyExtractorClass
+
+In cases where 'PartitionKeyExtractorClass' needs to access the underlying application context (to resolve application properties etc.,) the following can be done:
```
mbogoevici: Better to call it out as the 'Spring application context'.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1519,6 +1519,22 @@ This can be customized on the binding, either by setting a SpEL expression to be
...
+
+You can configure your strategy implementation as a @Bean in the application context with a bean name of the fully qualified class name, that bean will be fully initialized by Spring and used by the binder.
```
mbogoevici: We should explicitly mention `PartitionKeyExtractorClass ` instead of the more generic 'strategy'.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1519,6 +1519,22 @@ This can be customized on the binding, either by setting a SpEL expression to be
...
+If the binder doesn't find a bean with that name, it will create one (and add it to the context), but it won't get a full Spring initialization.
+
+----
+@Bean(name="com.example.MyKeyExtractor")
+public MyKeyExtractor extractor() {
+    return new MyKeyExtractor();
+}
+----
+
```
mbogoevici: Somewhere in this paragraph we need to explicitly mention setting the `partitionKeyExtractorClass` property on the binding. Also, since the same pattern applies to `partitionSelectorClass`.

mbogoevici:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/e93a03f792f8615982c2fa163c641fbe805d924f in master.

mbogoevici:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/f744857c32e89f85adb4da9bd80312101f96eada in the 1.1.x branch.

PR #762, opened 2017-01-13
Log deserialization exception messages

### LINKED ISSUE #497: Deserialization errors (e.g. w/Kryo) should be logged more explicitly
Currently, they are silently dropped and messages sent to the DLQ, but this doesn't help a lot especially on debugging.

ilayaperumalg:
- Log all the exception messages at the deserialization (especially at the codec's decode process)
 - Combine all the exception catches into Exception as the exception messages are almost same and redundant.

Resolves #497

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -78,10 +78,13 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 
 	private volatile Map<String, Class<?>> payloadTypeCache = new ConcurrentHashMap<>();
 
+	protected final Log logger = LogFactory.getLog(getClass());
+
 	/**
 	 * For binder implementations that support a prefix, apply the prefix to the name.
+	 *
 	 * @param prefix the prefix.
-	 * @param name the name.
+	 * @param name   the name.
 	 */
 	public static String applyPrefix(String prefix, String name) {
 		return prefix + name;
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -254,13 +260,11 @@ private Object deserializePayload(byte[] bytes, MimeType contentType) {
...
 				return this.codec.decode(bytes, targetType);
-			}
-			catch (ClassNotFoundException e) {
-				throw new SerializationFailedException("unable to deserialize [" + className + "]. Class not found.",
-						e); // NOSONAR
-			}
-			catch (IOException e) {
-				throw new SerializationFailedException("unable to deserialize [" + className + "]", e);
+			}// catch all exceptions that could occur during de-serialization
+			catch (Exception e) {
```
ilayaperumalg: Hope combining all the exceptions here makes sense as the exception messages appears redundant.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -78,10 +78,13 @@
 
 	private volatile Map<String, Class<?>> payloadTypeCache = new ConcurrentHashMap<>();
 
+	protected final Log logger = LogFactory.getLog(getClass());
+
```
ilayaperumalg: yeah :-)

sobychacko:
Cleaned up, squashed the commits - LGTM. Cherry picked into master branch also. Closing.

Issue #764, opened 2017-01-16
org.springframework.cloud.stream.binding.MessageConverterConfigurer$PartitioningInterceptor Performance


maximkir:
I've noticed that ~200 milliseconds are spend at `PartitioningInterceptor.preSend()` method. This happens due to `ExpressionUtils.createStandardEvaluationContext` call which happens on every invocation and is expensive.

![image](https://cloud.githubusercontent.com/assets/4746489/21986153/fefdb0d2-dc06-11e6-86de-90b907c3c94d.png)

I suggest to refactor the class to:

```
	private final class PartitioningInterceptor extends ChannelInterceptorAdapter {

		private final BindingProperties bindingProperties;
		private final PartitionHandler partitionHandler;

		private PartitioningInterceptor(BindingProperties bindingProperties) {
			this.bindingProperties = bindingProperties;
			partitionHandler = new PartitionHandler(
				 MessageConverterConfigurer.this.beanFactory,
				 ExpressionUtils.createStandardEvaluationContext(MessageConverterConfigurer.this.beanFactory),
				 null,
				 this.bindingProperties.getProducer());
		}

		@Override
		public Message<?> preSend(Message<?> message, MessageChannel channel) {
			int partition = this.partitionHandler.determinePartition(message);
			return new MutableMessageBuilderFactory().fromMessage(message)
					.setHeader(BinderHeaders.PARTITION_HEADER, partition).build();

		}
	}
```
The partition handler will be created once during interceptor creation and will save run time => smaller latency.

mbogoevici:
Thanks @maximkir !

If you don't mind changing your suggestion into a PR, we happily accept contributions!

maximkir:
No problem, Am I need a permissions to submit my changes?

ilayaperumalg:
No, You can submit a PR right away :-)  All you need is to sign the Contributor License Agreement which gives you notification when you submit a PR.

maximkir:
Submitted.

PR #766, opened 2017-01-17
Creating partition handler once during object instantiation instead of on every method invocation. Fixes gh-764

maximkir:
Fixes spring-cloud/spring-cloud-stream#764


artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/5faf1d812eadb217431e434ef75253c4cf0e7e64 and cherry-picked to `master`.

@maximkir ,

Thank you very much for the contribution!

Pay attention how I reuse a `messageBuilderFactory` as well.

Looking forward for more!

@mbogoevici , there is no issue with early `BeanFactory` access because this `configureMessageChannel()` is called from the `BindableProxyFactory.afterPropertiesSet()` via `getBindingTargetFactory(returnType).createInput(name)`. So, it is already safe to instantiate `EvaluationContext` from the `BeanFactory`.

We might revise caching beans in the `PartitionHandler` instead of `getBean()` on each call, but that is already different story...

PR #776, opened 2017-02-01
Access aggregate's child applications' request mapping/endpoint web endpoints

### LINKED ISSUE #702: Applications that are part of an aggregate should be able to expose web endpoints

ilayaperumalg:
- Use namespace to prefix the endpoints' paths
 - Lazy initialize EndpointHandlerMapping once child application context is updated with its actuator endpoints
 - Disable actuator endpoints for aggregator parent context

Resolves #702

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
Please note that this PR exposes the spring boot actuator endpoints of the child applications' and not the MVC request mappings of the child applications.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -386,15 +424,7 @@ void embed() {
 							e);
 				}
 			}
-			// Register metrics if JMX enabled and exporter avalable
-			if (BeanFactoryUtils.beansOfTypeIncludingAncestors(AggregateApplicationBuilder.this.parentContext,
```
ilayaperumalg: Given all the metrics are now exposed via `/<namespace>/metrics` endpoint, I believe this explicit registration of metrics can be removed.

ilayaperumalg:
The latest update to this PR takes care of mapping the `RequestMapping` from the child context to the aggregate parent.

ilayaperumalg:
Rebased and updated the PR with the changes needed to not enforcing the `EnableAutoConfiguration` on Aggregate ParentConfiguration.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationUtils.java
```diff
@@ -59,10 +65,16 @@ static String getDefaultNamespace(String appClassName, int index) {
...
+		// Child context needs to enable web MVC configuration and web enabled to obtain the MVC request mapping in the child applications
+		return new SpringApplicationBuilder(new Object[]{app, EmbeddedServletContainerAutoConfiguration.class,
+				EndpointWebMvcManagementContextConfiguration.class, WebMvcAutoConfiguration.EnableWebMvcConfiguration.class,
+				ServerPropertiesAutoConfiguration.class})
+				.web(true).main(app)
+				.bannerMode(Mode.OFF)
 				.properties("spring.jmx.default-domain=" + namespace)
 				.properties(
 						InternalPropertyNames.NAMESPACE_PROPERTY_NAME + "=" + namespace)
+				.properties("server.port=-1")
```
mbogoevici: Still not OK with doing this. If starting one web server per application is the only way we can get this done, then we should drop the idea altogether and not support embedding of web apps.

ilayaperumalg: In order to expose the web endpoints, the child applications need to be web `enabled` but here we make sure they don't expose the `server.port`. One possible way is to introduce a property to let the user decide whether to expose the web endpoints for the child application or not. That way, we will start web application context only for those child apps that have their web endpoints (request mappings + actuator endpoints) mapped to parent aggregate application.

mbogoevici:
Also it would be good to have some documentation on how this process works.

ilayaperumalg:
Rebased.

mbogoevici:
I've been reviewing this and the approach per se looks OK, the only caveat is that it couples the core even tighter to a dependency on the `web` module of Spring Boot. I think we can make all this stuff optional on whether web classes are on the classpath (otherwise it is not applicable anyway).

ilayaperumalg:
@mbogoevici Given we already have `spring-boot-starter-web` in core [classpath](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/pom.xml#L21), the web classes are already there. 

Also, We [enable](https://github.com/spring-cloud/spring-cloud-stream/pull/776/files#diff-e6a31d66d78de3178c1715226b5a4117R233) the `web` configuration only when the web is enabled for the aggregate application. The child application contexts' web is disabled as well.

mbogoevici:
@ilayaperumalg 
>Given we already have spring-boot-starter-web in core classpath, the web classes are already there.

That is a contentious point already and we might want to remove that dependency in the future. It is not effectively required at the moment. Currently it is possible to exclude it and still get a functional application (albeit without web actuator endpoints). With this change, it will be impossible to remove `spring-boot-starter-web` for an aggregate application (`AggregateApplicationBuilder` will probably not load).

ilayaperumalg:
@mbogoevici I agree, Didn't realize the strategy to remove `web` starter dependency. I will update this PR.

ilayaperumalg:
Given the complexity involved in maintaining the `Endpoint` handler mappings and the future efforts to maintain the compatibility with Spring Boot changes, we are closing this PR.

The work around to fix this issue would be to run the `source` application as a bean outside the aggregate application and have `outbound` channel of the `source` manually connect to the `inbound` channel of the aggregate application. This way, the `source` application acts as a `gateway` to receive the incoming request endpoints while the `binding` between the `source` and the aggregate application is wired inside the same application.

PR #781, opened 2017-02-02
Overload EmbeddedHeadersMessageConverter method

### LINKED ISSUE #780: EmbeddedHeadersMessageConverter extractHeader method can operate on byte[]
It can be a convenient option to have the `EmbeddedHeadersMessageConverter` to have an overloaded method to extractHeader using just the `payload` and `requestHeaders` instead of using the `Message` object especially in cases where the message payload is already retrieved.

ilayaperumalg:
- Overload extractHeaders method to accept bytes payload
  - This will help avoiding unnessary message creation during ad-hoc extractHeader invocation
  - add test

Resolves #780

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeadersMessageConverter.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeadersMessageConverter.java
@@ -139,6 +139,19 @@ public class EmbeddedHeadersMessageConverter {
 		}
 	}
 
+	/**
+	 * Return a message where headers, that were originally embedded into the payload,
+	 * have been promoted back to actual headers. The new payload is now the original
+	 * payload.
+	 *
+	 * @param payload            the message payload
+	 * @return the message with extracted headers
+	 * @throws Exception
+	 */
+	public MessageValues extractHeaders(byte[] payload) throws Exception {
+		return extractHeaders(payload, false, null);
+	}
+
 	private MessageValues oldExtractHeaders(ByteBuffer byteBuffer, byte[] bytes, int headerCount,
 			boolean copyRequestHeaders, MessageHeaders requestHeaders)
 			throws UnsupportedEncodingException {
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeadersMessageConverter.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeadersMessageConverter.java
@@ -103,18 +103,7 @@ public class EmbeddedHeadersMessageConverter {
 		return extractHeaders(message.getPayload(), copyRequestHeaders, message.getHeaders());
 	}
 
-	/**
-	 * Return a message where headers, that were originally embedded into the payload,
-	 * have been promoted back to actual headers. The new payload is now the original
-	 * payload.
-	 *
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeadersMessageConverter.java
```diff
@@ -93,38 +95,52 @@ public static String decodeExceptionMessage(Message<?> requestMessage) {
...
+	 * have been promoted back to actual headers. The new payload is now the original
+	 * payload.
+	 *
+	 * @param payload            the message payload
+	 * @param copyRequestHeaders boolean value to specify if the request headers should be copied
+	 * @param requestHeaders     the request headers to copy
+	 * @return the message with extracted headers
+	 * @throws Exception
+	 */
+	public MessageValues extractHeaders(byte[] payload, boolean copyRequestHeaders, MessageHeaders requestHeaders) throws Exception {
```
mbogoevici: Since we're overloading things anyway, why don't we have a simple `public MessageValues extractHeaders(byte[] payload)` method that just extracts headers - it doesn't seem like copying of the headers should be the responsibility of this method - that can be simply done by the invoker of the method, on the result.

mbogoevici: We can make this method private.

ilayaperumalg: may be `protected` as this method is still being used by the `SendingHandler` in the `AbstractMessageChannelBinder`.

mbogoevici: Probably `SendingHandler` should use one of the public methods.

ilayaperumalg: yeah, in that case, I guess we need to have both the methods as `public` as `extractHeaders(payload)` is used by the Kafka binder DLQ forwarding mechanism while the `SendingHandler` requires the other one.

mbogoevici: We only need two public methods:

```
public MessageValues extractHeaders(Message<byte[]> message, boolean copyRequestHeaders) throws Exception
```
and
```
public MessageValues extractHeaders(byte[] payload) throws Exception
```

This method can be internal - in the future, this will need to be refactored as we don't need to pass both the boolean argument and the requestHeaders. Passing an empty or null set of request headers should be enough to signal that there's nothing to copy.

ilayaperumalg: yes, sorry; I overlooked the method parameter when I responded initially. I will fix that method `extractHeaders(byte[] payload, boolean copyRequestHeaders, MessageHeaders requestHeaders)` to be private. fixing that

mbogoevici:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/e1171a38f737e4b801a50831de8cdf3d930ef10d and backported as https://github.com/spring-cloud/spring-cloud-stream/commit/cded4e0c87c83f3d9d7b292cfecc158c771d0af4

PR #783, opened 2017-02-02
"spring.cloud.stream.default" prefix is not used by BindingServiceProperties

### LINKED ISSUE #782: "spring.cloud.stream.default" prefix is not used by BindingServiceProperties
While using Brooklyn SR2 release, I am trying to set a default `contentType` by adding this to the properties file:
```properties
spring.cloud.stream.default.contentType=application/x-kryo
```

It doesn't work. 

After debugging, it appears that `EnvironmentEntryInitializingTreeMap` default values are only loaded by call to the `get` method. But `BindingServiceProperties` calls `get` only if `containsKey` return true:
```java
if (this.bindings.containsKey(bindingName)) {
			BeanUtils.copyProperties(this.bindings.get(bindingName), bindingProperties);
		}
```

Therefore, I think that `EnvironmentEntryInitializingTreeMap` should override `containsKey` by returning `get(key) != null`

I am going to submit a pull request with the fix.

reda-alaoui:
Fix #782

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/EnvironmentEntryInitializingTreeMap.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/EnvironmentEntryInitializingTreeMap.java
@@ -90,4 +90,8 @@ public class EnvironmentEntryInitializingTreeMap<T> extends AbstractMap<String, 
 		return delegate.entrySet();
 	}
 
+	@Override
+	public boolean containsKey(Object key) {
+		return get(key) != null;
+	}
 }


pivotal-issuemaster:
@reda-alaoui Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=783)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=783) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@reda-alaoui Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=783)!

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/EnvironmentEntryInitializingTreeMap.java
```diff
@@ -90,4 +90,8 @@ public T put(String key, T value) {
 		return delegate.entrySet();
 	}
 
+	@Override
+	public boolean containsKey(Object key) {
```
mbogoevici: This effectively returns true every time. I wonder if we shouldn't rather do a 'get' inside `BindingServiceProperties`.

reda-alaoui: Maybe this TreeMap will be used by another consumer in a near futur. In this case, the developer responsible for this new consumer will certainly not be aware of this treemap specific behaviour.

I think that this treemap should fully implement the map contract to avoid speculations on the consumers behaviours.

mbogoevici: As a note: as the javadoc clearly states, we do not intend to use this as a general purpose data structure, but as a workaround until https://github.com/spring-projects/spring-boot/issues/7986 is fixed (only in Spring Boot 2, unfortunately). 

That being said, your solution is valid too in the context. Thanks for the contribution!

PR #787, opened 2017-02-07
Introducing provisioning SPI

### LINKED ISSUE #745: Separate binding from queue provisioning and introduce provisioning SPI
The current binder implementations address both connecting to the messaging middleware (via binding targets) and ensuring that the queues/topics exist - also enforcing naming conventions and topology conventions - e.g. TopicExchange + queue-per-consumer group in the case of RabbitMQ, virtual topics in the case of ActiveMQ, etc. We call the latter concern provisioning and it can be separated from the mere act of setting up endpoints for sending/receiving messages. Such separation of concerns is desirable because it allows to:

- use a common provisioning strategy for Kafka message-driven and Reactive Kafka binders;
- make the provisioning extensions of JMS binders general-purpose (and separate from the JMS binding implementation - in practice, entirely separate from JMS);

sobychacko:
- Provide an API for provisioning middleware specific concerns independent of the binder
 - Make the corresponding changes in AbstractMessageChannelBinder

Fixes #745

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.stream.binder;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cloud.stream.provisioning.ConsumerDestination;
 import org.springframework.cloud.stream.provisioning.ProducerDestination;
+import org.springframework.cloud.stream.provisioning.ProvisioningException;
 import org.springframework.cloud.stream.provisioning.ProvisioningProvider;
 import org.springframework.context.Lifecycle;
 import org.springframework.expression.ExpressionParser;
@@ -113,6 +114,9 @@ public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties,
 		catch (Exception e) {
 			if (e instanceof BinderException) {
 				throw (BinderException) e;
+			}
+			else if (e instanceof ProvisioningException) {
+				throw (ProvisioningException) e;
 			}
 			else {
 				throw new BinderException("Exception thrown while building outbound endpoint", e);
@@ -221,6 +225,9 @@ public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties,
 			}
 			if (e instanceof BinderException) {
 				throw (BinderException) e;
+			}
+			else if (e instanceof ProvisioningException) {
+				throw (ProvisioningException) e;
 			}
 			else {
 				throw new BinderException("Exception thrown while starting consumer: ", e);
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -17,6 +17,9 @@
 package org.springframework.cloud.stream.binder;
 
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.cloud.stream.provisioning.ConsumerDestination;
+import org.springframework.cloud.stream.provisioning.ProducerDestination;
+import org.springframework.cloud.stream.provisioning.ProvisioningProvider;
 import org.springframework.context.Lifecycle;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/provisioning/ConsumerDestination.java
```diff
@@ -0,0 +1,37 @@
...
+ * @author Soby Chacko
+ */
+public interface ConsumerDestination {
+
+	/**
+	 * Provides the destination name.
+	 *
+	 * @return destination name
+	 */
+	String getConsumerDestinationName();
```
artembilan: According to this `ConsumerDestination` contract it just can be `getName()`.

How do you handle `@since` tags on classes?
Also we in SI uses only the current year in the `Copyright` for new classes

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/provisioning/ProducerDestination.java
```diff
@@ -0,0 +1,44 @@
...
+	 */
+	String getProducerDestinationName();
+
+	/**
+	 * Provides the destination name for a given partition
+	 *
+	 * @param partition the partition to find destination for
+	 * @return destination name for the given partition
+	 */
+	String getPartitionedProducerDestinationName(int partition);
```
artembilan: This contract smells for me somehow...

Maybe a JavaDocs would explain that there might be several destinations on the target broker because of partition logic?
Thanks

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/provisioning/ProvisioningProvider.java
```diff
@@ -0,0 +1,63 @@
...
+ * @param <P>  the producer properties type
+ *
+ * @author Soby Chacko
+ *
+ * @since 1.2
+ */
+public interface ProvisioningProvider<C extends ConsumerProperties, P extends ProducerProperties> {
+
+	/**
+	 * Creates target destinations for outbound channels. The implementation
```
artembilan: Typically the methods JavaDocs is in the imperative.
I guess mentioning of the `outbound channels` is overhead here.
Might be something more generic, like " outbound bind entities".

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -110,6 +115,9 @@ public AbstractMessageChannelBinder(boolean supportsHeadersNatively, String[] he
 			if (e instanceof BinderException) {
 				throw (BinderException) e;
 			}
+			else if (e instanceof ProvisioningException) {
+				throw (ProvisioningException) e;
```
artembilan: M-m-m. How this exception can be possible here if we have done provisioning already a bit before via ` this.provisioningProvider.provisionProducerDestination()`?
Maybe the idea was to include that call to the `try...catch` as well?
Thanks

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/provisioning/ProducerDestination.java
```diff
@@ -0,0 +1,57 @@
...
+	 * @return destination name
+	 */
+	String getName();
+
+	/**
+	 * Provides the destination name for a given partition.
+	 *
+	 * If the producer provision the destination with partitions, on certain middleware brokers
+	 * there may exist multiple destinations distinguishable by the partition. For example,
+	 * if the destination name is <b>xyz</b> and is is provisioned with <b>4</b> partitions, there may be
```
artembilan: I guess it is a typo: 
and **it** is provisioned with

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/provisioning/ProducerDestination.java
```diff
@@ -0,0 +1,57 @@
...
+	 * 4 different destinations on the broker such as - <b>xyz-0, xyz-1, xyz-2 and xyz-3</b>.
+	 * However, please keep in mind that, this behavior is completely dependent on the broker
```
artembilan: M-m-m. Not sure in this `please`.
The JavaDoc is fully official text and should just describe facts about the API.
In this case to say that `A behavior is completely...` is pretty enough

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -98,7 +103,7 @@ public AbstractMessageChannelBinder(boolean supportsHeadersNatively, String[] he
 			final P producerProperties) throws BinderException {
 		Assert.isInstanceOf(SubscribableChannel.class, outputChannel,
 				"Binding is supported only for SubscribableChannel instances");
-		PD producerDestination = createProducerDestinationIfNecessary(destination, producerProperties);
+		final ProducerDestination producerDestination = this.provisioningProvider.provisionProducerDestination(destination, producerProperties);
```
artembilan: Yeah. `this.provisioningProvider.provisionConsumerDestination` is really included in the `try...catch`.
See `doBindConsumer()`

sobychacko:
PR review comments are addressed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/provisioning/ProvisioningException.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Generic unchecked exception to wrap middleware or technology specific exceptions.
+ * Wrapped exceptions could be either checked or unchecked.
+ *
+ * See {@link NestedRuntimeException} for more usage details.
+ *
+ * @author Soby Chacko
+ */
+public class ProvisioningException extends NestedRuntimeException {
```
garyrussell: Needs `@SupressWarnings("serial")` - will fix on merge.

garyrussell:
Merged as 392b087af631e54d15940e79913c41e29a06e2d2

PR #791, opened 2017-02-09
Add converter for Object->String conversion

### LINKED ISSUE #774: MessageConverter fails to convert files to string
See: https://github.com/spring-cloud-stream-app-starters/file/issues/2

mbogoevici:
Support `text/plain` outbound content type for generic objects

Fix #774


artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/f65dcbe975e802b3838a5a07f33a9f3329802749

PR #800, opened 2017-02-13
Add dispatching capabilities to `@StreamListener`

### LINKED ISSUE #682: Support of Multiple @StreamListener methods on the same bound element

mbogoevici:
`@StreamListener` has support for a `condition` parameter,
that contains a SpEL expression that is evaluated before the
method is invoked.

Fix #682

Move StreamListenerMessageHandler as a top level class

Use dispatching and add test

Refactor dispatching mechanism

Throw error when conditions are used in declarative mode

Make StreamListenerAnnotationBeanPostProcessor overridable

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.config;
+
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import org.junit.Test;
+
+import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.annotation.StreamListener;
+import org.springframework.cloud.stream.binding.StreamListenerAnnotationBeanPostProcessor;
+import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.handler.annotation.Payload;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.cloud.stream.config.BindingServiceConfiguration.STREAM_LISTENER_ANNOTATION_BEAN_POST_PROCESSOR_NAME;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class StreamListenerAnnotationBeanPostProcessorOverrideTest {
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testOverrideStreamListenerAnnotationBeanPostProcessor() throws Exception {
+		ConfigurableApplicationContext context = SpringApplication.run(TestPojoWithAnnotatedArguments.class,
+				"--server.port=0");
+
+		TestPojoWithAnnotatedArguments testPojoWithAnnotatedArguments = context
+				.getBean(TestPojoWithAnnotatedArguments.class);
+		Sink sink = context.getBean(Sink.class);
+		String id = UUID.randomUUID().toString();
+		sink.input().send(MessageBuilder.withPayload("{\"foo\":\"barbar" + id + "\"}")
+				.setHeader("contentType", "application/json").setHeader("testHeader", "testValue")
+				.setHeader("type", "foo").build());
+		sink.input().send(MessageBuilder.withPayload("{\"bar\":\"foofoo" + id + "\"}")
+				.setHeader("contentType", "application/json").setHeader("testHeader", "testValue")
+				.setHeader("type", "bar").build());
+		assertThat(testPojoWithAnnotatedArguments.receivedFoo).hasSize(1);
+		assertThat(testPojoWithAnnotatedArguments.receivedFoo.get(0)).hasFieldOrPropertyWithValue("foo",
+				"barbar" + id);
+		context.close();
+	}
+
+	@EnableBinding(Sink.class)
+	@EnableAutoConfiguration
+	public static class TestPojoWithAnnotatedArguments {
+
+		List<StreamListenerTestUtils.FooPojo> receivedFoo = new ArrayList<>();
+
+		List<StreamListenerTestUtils.BarPojo> receivedBar = new ArrayList<>();
+
+		@StreamListener(value = Sink.INPUT, condition = "foo")
+		public void receive(@Payload StreamListenerTestUtils.FooPojo fooPojo) {
+			this.receivedFoo.add(fooPojo);
+		}
+
+		/**
+		 * Overrides the default {@link StreamListenerAnnotationBeanPostProcessor}.
+		 */
+		@Bean(name = STREAM_LISTENER_ANNOTATION_BEAN_POST_PROCESSOR_NAME)
+		public static BeanPostProcessor streamListenerAnnotationBeanPostProcessor() {
+			return new StreamListenerAnnotationBeanPostProcessor() {
+				@Override
+				protected StreamListener postProcessAnnotation(StreamListener originalAnnotation, Method annotatedMethod) {
+					Map<String,Object> attributes = new HashMap<>(AnnotationUtils.getAnnotationAttributes(originalAnnotation));
+					attributes.put("condition", "headers['type']=='" + originalAnnotation.condition() + "'");
+					return AnnotationUtils.synthesizeAnnotation(attributes, StreamListener.class, annotatedMethod);
+				}
+			};
+		}
+	}
+}
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
@@ -73,9 +73,7 @@ public class StreamListenerAnnotationBeanPostProcessorOverrideTest {
 	public static class TestPojoWithAnnotatedArguments {
 
 		List<StreamListenerTestUtils.FooPojo> receivedFoo = new ArrayList<>();
-
-		List<StreamListenerTestUtils.BarPojo> receivedBar = new ArrayList<>();
-
+		
 		@StreamListener(value = Sink.INPUT, condition = "foo")
 		public void receive(@Payload StreamListenerTestUtils.FooPojo fooPojo) {
 			this.receivedFoo.add(fooPojo);
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithConditionsTest.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithConditionsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithConditionsTest.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithConditionsTest.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.config;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import org.junit.Test;
+
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.annotation.StreamListener;
+import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.handler.annotation.Payload;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class StreamListenerWithConditionsTest {
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testAnnotatedArgumentsWithConditionalClass() throws Exception {
+		ConfigurableApplicationContext context = SpringApplication.run(TestPojoWithAnnotatedArguments.class,
+				"--server.port=0");
+
+		TestPojoWithAnnotatedArguments testPojoWithAnnotatedArguments = context
+				.getBean(TestPojoWithAnnotatedArguments.class);
+		Sink sink = context.getBean(Sink.class);
+		String id = UUID.randomUUID().toString();
+		sink.input().send(MessageBuilder.withPayload("{\"foo\":\"barbar" + id + "\"}")
+				.setHeader("contentType", "application/json").setHeader("testHeader", "testValue")
+				.setHeader("type", "foo").build());
+		sink.input().send(MessageBuilder.withPayload("{\"bar\":\"foofoo" + id + "\"}")
+				.setHeader("contentType", "application/json").setHeader("testHeader", "testValue")
+				.setHeader("type", "bar").build());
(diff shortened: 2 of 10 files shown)


mbogoevici:
Some post-merge issues. Closing for a moment.

artembilan:
Just in case:
```
[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java:67: Annotation 'Autowired' should be alone on line. [AnnotationLocation]

[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java:67: Annotation 'Lazy' should be alone on line. [AnnotationLocation]

[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java:70: Annotation 'Autowired' should be alone on line. [AnnotationLocation]

[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java:70: Annotation 'Lazy' should be alone on line. [AnnotationLocation]
```

mbogoevici:
Fixed some issues with the previous version.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
```diff
@@ -189,10 +184,16 @@ public ChannelBindingServiceProperties channelBindingServiceProperties(
...
+	@Bean(name = STREAM_LISTENER_ANNOTATION_BEAN_POST_PROCESSOR_NAME)
+	public static StreamListenerAnnotationBeanPostProcessor streamListenerAnnotationBeanPostProcessor() {
+		return new StreamListenerAnnotationBeanPostProcessor();
+	}
+
 	// IMPORTANT: Nested class to avoid instantiating all of the above early
 	@Configuration
 	protected static class PostProcessorConfiguration {
 
+		public static final String DEFAULT_STREAM_LISTENER_ANNOTATION_BEAN_POST_PROCESSOR_NAME = "streamListenerAnnotationBeanPostProcessor";
```
mbogoevici: Needs to be removed.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2016 the original author or authors.
```
ilayaperumalg: copyright year ;)

ilayaperumalg: Please fix this for all the changed files too.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -59,56 +64,56 @@
...
 		}
-		Map<String, StreamListenerResultAdapter> resultAdapterMap =
-				BeanFactoryUtils.beansOfTypeIncludingAncestors(this.applicationContext, StreamListenerResultAdapter.class);
+		Map<String, StreamListenerResultAdapter> resultAdapterMap = BeanFactoryUtils
+				.beansOfTypeIncludingAncestors(this.applicationContext, StreamListenerResultAdapter.class);
 		this.streamListenerResultAdapters.add(new MessageChannelStreamListenerResultAdapter());
 		for (StreamListenerResultAdapter resultAdapter : resultAdapterMap.values()) {
 			this.streamListenerResultAdapters.add(resultAdapter);
 		}
+		this.evaluationContext = new StandardEvaluationContext();
```
ilayaperumalg: Can we use `IntegrationContextUtils.getEvaluationContext(BeanFactory)` here instead?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerMethodUtils.java
```diff
@@ -61,7 +61,7 @@ protected static int outputAnnotationCount(Method method) {
 
 	protected static void validateStreamListenerMethod(Method method, int inputAnnotationCount,
 			int outputAnnotationCount, String methodAnnotatedInboundName, String methodAnnotatedOutboundName,
-			boolean isDeclarative) {
+			boolean isDeclarative, StreamListener streamListener) {
```
ilayaperumalg: Given most of the other attributes are also derived from `StreamListener` annotation, may be we can just pass the `condition` String value here instead of the annotation itself?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -286,22 +292,45 @@ protected void registerHandlerMethodOnListenedChannel(Method method, StreamListe
...
+				StreamListenerMessageHandler streamListenerMessageHandler = new StreamListenerMessageHandler(
+						invocableHandlerMethod);
+				streamListenerMessageHandler.setApplicationContext(this.applicationContext);
+				streamListenerMessageHandler.setBeanFactory(this.applicationContext.getBeanFactory());
+				if (StringUtils.hasText(mapping.getDefaultOutputChannel())) {
+					streamListenerMessageHandler.setOutputChannelName(mapping.getDefaultOutputChannel());
+				}
+				streamListenerMessageHandler.afterPropertiesSet();
+				if (StringUtils.hasText(mapping.getCondition())) {
+					Expression condition = new SpelExpressionParser().parseExpression(mapping.getCondition());
```
ilayaperumalg: Nitpicking: Can we create `SpelExpressionParser` outside the for loop?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
```diff
@@ -0,0 +1,126 @@
...
+		return matchingMethods;
+	}
+
+	static class ConditionalStreamListenerHandler implements MessageHandler {
+
+		private Expression condition;
+
+		private StreamListenerMessageHandler streamListenerMessageHandler;
+
+		ConditionalStreamListenerHandler(Expression condition, StreamListenerMessageHandler streamListenerMessageHandler) {
```
ilayaperumalg: Would it be better to make `ConditionalStreamListenerHandler` with `non-null` condition value and have an intermediate base class to hold the `StreamListenerMessageHandler`. That way, instead of checking for `conditiion == null` for the non conditional stream listener handlers, we can have `type` check on `ConditionalStreamListenerHandler`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
```diff
@@ -0,0 +1,126 @@
...
+	protected boolean shouldCopyRequestHeaders() {
+		return false;
+	}
+
+	@Override
+	protected Object handleRequestMessage(Message<?> requestMessage) {
+		Collection<ConditionalStreamListenerHandler> matchingHandlers = findMatchingHandlers(requestMessage);
+		if (matchingHandlers.size() == 0) {
+			if (logger.isWarnEnabled()) {
+				logger.warn("Cannot find a @StreamListener matching " + requestMessage);
```
ilayaperumalg: Is it ok to display the `requestMessage` in the log?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
```diff
@@ -0,0 +1,126 @@
...
+			return null;
+		}
+		else if (matchingHandlers.size() > 1) {
+			for (ConditionalStreamListenerHandler matchingHandler : matchingHandlers) {
+				if (!matchingHandler.isVoid()) {
+					throw new MessagingException(
+							"Multiple matching methods cannot return values for " + requestMessage);
+				}
+			}
+			for (ConditionalStreamListenerHandler matchingMethod : matchingHandlers) {
```
ilayaperumalg: Does this mean, a single message can be processed by multiple message handlers? For instance, if there is a handler method without any condition but matches and the other one that has a condition that matches as well, the requestMessage will be processed by both the matched handler methods. I guess that's how this is intended to work.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithConditionsTest.java
```diff
@@ -0,0 +1,120 @@
...
+	@EnableBinding(Sink.class)
+	@EnableAutoConfiguration
+	public static class TestPojoWithAnnotatedArguments {
+
+		List<StreamListenerTestUtils.FooPojo> receivedFoo = new ArrayList<>();
+
+		List<StreamListenerTestUtils.BarPojo> receivedBar = new ArrayList<>();
+
+		@StreamListener(value = Sink.INPUT, condition = "headers['type']=='foo'")
+		public void receive(@Payload StreamListenerTestUtils.FooPojo fooPojo) {
```
ilayaperumalg: Can we have some tests with the methods that return value?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/StreamListener.java
```diff
@@ -131,6 +132,18 @@
 	/**
 	 * The name of the binding target (e.g. channel) that the method subscribes to.
 	 */
+	@AliasFor("target")
 	String value() default "";
 
+	/**
+	 * The name of the binding target (e.g. channel) that the method subscribes to.
+	 * @return
```
ilayaperumalg: empty @ return

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
mbogoevici: I think that introducing a type check would make the code much more complex than it already is.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
garyrussell: Yes @ilayaperumalg - that's the intention.

@mbogoevici I suppose we could support a return value as long as only one such matching method - or is that too confusing?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
garyrussell: Yes; we must - in order to get registered SpEL functions, a bean resolver, etc.

garyrussell: It doesn't seem right to do all this work in a setter `setApplicationContext`.

Maybe `InitializingBean` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
garyrussell: We also need to resolve property placeholders in the expression before parsing. See `resolveExpression` in `RabbitListenerAnnotationBeanPostProcessor`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerMessageHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessagingException;
+import org.springframework.messaging.handler.invocation.InvocableHandlerMethod;
+
+/**
+ * @author Marius Bogoevici
+ */
+public class StreamListenerMessageHandler extends AbstractReplyProducingMessageHandler {
```
garyrussell: `@Since`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerMessageHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+	public boolean isVoid() {
+		return invocableHandlerMethod.isVoid();
+	}
+
+	@Override
+	protected Object handleRequestMessage(Message<?> requestMessage) {
+		try {
+			return this.invocableHandlerMethod.invoke(requestMessage);
+		} catch (Exception e) {
```
garyrussell: new line for catch (several others too).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
```diff
@@ -0,0 +1,126 @@
...
+			else {
+				Object value = conditionalStreamListenerHandlerMethod.getCondition().getValue(this.evaluationContext, message);
+				if (Boolean.class.isAssignableFrom(value.getClass())) {
+					if ((Boolean) value) {
+						matchingMethods.add(conditionalStreamListenerHandlerMethod);
+					}
+				}
+				else if (Boolean.valueOf(String.valueOf(value))) {
+					matchingMethods.add(conditionalStreamListenerHandlerMethod);
+				}
```
garyrussell: Shouldn't we throw an exception (or at least log with ERROR) if the expression doesn't evaluate to a `String` or `Boolean` and clearly document that, if it's a `String`, then the `Boolean.valueOf()` algorithm applies.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
artembilan: `SpelExpressionParser` is thread-safe and stateless. Can be just as `static final` property.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
artembilan: I think `Expression.getValue()` with return type variant to `Boolean.class` does the trick for us.
As far as the conversion service is in charge the in SpEL.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
artembilan: +1

dturanski:
LLIWMMR *



*Looks like it will meet my requirements

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/DispatchingStreamListenerMessageHandler.java
mbogoevici: @garyrussell It currently does that - if a single value-returning method is a match, then it proceeds. If multiple methods match and at least one has a return value, the dispatching fails. (we could relax that a bit and make it so that we allow at most once to return a value). 

@artembilan can you clarify what you mean a bit?

artembilan: Well, let's imagine the situation when today one our message passes only one condition and with return. We are happy to process the message and get a reply. Another day the same message passes several conditions including that one with return. So, we fail because we can't have return in multi-matching. That doesn't feel like way to go. If we tied with conditions we can't jump from algorithm to algorithm. Would be better to have a stable behavior independently of number of matches. 

Or... Let all matches to deal with reply on their own. In other words: if you have a return on method, that its processor responsibility to handler return properly. At least I see that we have `HandlerMethodReturnValueHandler` which can be use in the loop for each matched handler as well.

mbogoevici: > Another day the same message passes several conditions including that one with return.

How is that possible without changing the application? The conditions remain the same, so message handling should be deterministic. 

>  If we tied with conditions we can't jump from algorithm to algorithm. 

Again, I don't quite understand what that means. 

> Would be better to have a stable behavior independently of number of matches.

The problem with replies is side effects on error handling. If we have multiple matches and some fail and some succeed, and it's getting very unpredictable. Not that we can really prevent users from adding side effects to their handlers, but I'm not convinced we should do it ourselves.

artembilan: > The conditions remain the same, so message handling should be deterministic.

Really? Looks like they are SpEL and, therefore, we can define there access to external state, e.g. some static method invocation, or if we will rely on the `integrationEvaluationContext`, we will have access to the application context and will be able to call any bean.

artembilan: > Again, I don't quite understand what that means.

Having conditions as SpEL we really make the matching logic unpredictable. (See my previous comment).

And that would be very confusing to expect reply via `return`, but fail because we have matched one more condition and in multi-matching we can't have any `return`. 

So, my vote do not allow `return` at all independently of the number of matches.

mbogoevici: I thought we were discouraging 'programming in SpEL' as a principle. Sure, but all these computations would have to be deterministic. We can recommend that the evaluation log be stateless ... otherwise it is going to be a mess.

artembilan: OK! Let it be...
I don't know what to say yet, but it really will be confusing if we have condition for `Number` on two methods, and second one has condition for odds and we have return for the last one. Sending a series of numbers, we will be surprised to end up with the `Multiple matching methods cannot return values for`, because odds match two conditions.

garyrussell: >and at least one has a return value, the dispatching fails. (we could relax that a bit and make it so that we allow at most once to return a value)

Right - that's what I meant - relax to allow if only one has a return - but perhaps that might be confusing.

mbogoevici: chatted a bit with @artembilan about this. One other possibility is to nix conditions on value-returning methods altogether - it eliminates the potential of erratic behaviour as to the return values of the methods. This would turn this feature into a SpEL-matching event handler that would be suitable for Event Streaming/Event Sourcing scenario and would avoid the risk of rebuilding a router via `@StreamListener`. The use case @dturanski has brought up definitely doesn't require return values . @garyrussell , wdyt?

garyrussell: Works 4 me.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerWithConditionsTest.java
mbogoevici: Updated how we handle multiple return values - a return value is an error condition now.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerAnnotationBeanPostProcessorOverrideTest.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * @author Marius Bogoevici
+ */
+public class StreamListenerAnnotationBeanPostProcessorOverrideTest {
+
+	@Test
+	@SuppressWarnings("unchecked")
+	public void testOverrideStreamListenerAnnotationBeanPostProcessor() throws Exception {
+		ConfigurableApplicationContext context = SpringApplication.run(TestPojoWithAnnotatedArguments.class,
+				"--server.port=0");
```
mbogoevici: >Although isn't clear why @SpringBootTest doesn't fit here...

It's the standard way, but this is more convenient for keeping things together. Although we might want to clean this up a little bit.

Good observation - removing Tomcat would speed up tests significantly. Since this is already happening in several places, adding a separate issue to address that.

garyrussell:
LGTM; @artembilan ?

garyrussell:
Merged as 076f0ac1cb81f717cef70ad2a853b506560e4c5d

PR #804, opened 2017-02-14
Making ProvisioningProvider protected, javadocs

### LINKED ISSUE #803: Making provisioning provider protected in AbstractMessageChannelBinder
This would allow donwnstream binder impls to reference `provisioningProvider` without redefining it in their own classes.

sobychacko:
Fixes #803

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -74,7 +74,10 @@ public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties,
 	 */
 	private final String[] headersToEmbed;
 
-	private final ProvisioningProvider<C, P> provisioningProvider;
+	/**
+	 * {@link ProvisioningProvider} delegated by the downstream binder implementations.
+	 */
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -74,7 +74,10 @@
 	 */
 	private final String[] headersToEmbed;
 
-	private final ProvisioningProvider<C, P> provisioningProvider;
+	/**
+	 * {@link ProvisioningProvider} delegated by the downstream binder implementations.
+	 */
+	protected final ProvisioningProvider<C, P> provisioningProvider;
```
artembilan: This doesn't make me happy, because in the target `Binder` impl I have to cast to target `ProvisioningProvider` impl.
Adding one more generic argument - `PP extends ProvisioningProvider<C, P>` would make the trick.
But do we really need to worry about?

Just my 5 cents...

sobychacko: I am fine with adding that extra generic argument to avoid casting in implementations in subclasses. Will send another push.

garyrussell:
Merged as 48c53ef278889d0413e2aa82ed7f4fe365b98894

PR #807, opened 2017-02-15
Not enforcing EnableAutoConfiguration on aggregate parent

### LINKED ISSUE #737: AggregateApplicationBuilder forcing full auto-config
In `AggregateApplicationBuilder` there is no option to fully define what parent context will have as `ParentConfiguration` is always added. This config class has `@EnableAutoConfiguration` in it which makes it impossible to exclude auto-config classes which are known not to work in your app.

This is i.e. case with `gpfdist` which needs to disable `DataSourceAutoConfiguration` as otherwise we'll end up
```
2016-12-15 09:37:00.601 DEBUG 6016 --- [           main] o.s.b.d.LoggingFailureAnalysisReporter   : Application failed to start due to an exception

org.springframework.boot.autoconfigure.jdbc.DataSourceProperties$DataSourceBeanCreationException: Cannot determine embedded database driver class for database type NONE. If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).

```

We should not force feed `@EnableAutoConfiguration` to a user app.

ilayaperumalg:
- The aggregate parent configuration now imports the required auto configuration classes such as ChannelBindingAutoConfiguration and EndpointAutoConfiguration
while `@EnableBinding` is still at the parent configuration to be able to share the binding configuration among the child applications
 - Add `ConditionalOnMissingBean` in spring cloud stream auto configuration classes so that if `EnableAutoConfiguration` is enabled at child application classes
they don't get instantiated again
 - Conditionally import EmbeddedServlet auto-configuration when web environment is enabled for parent
 - Add test

Resolves #737

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
@@ -29,16 +29,19 @@ import java.util.Set;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.beans.factory.SmartInitializingSingleton;
+import org.springframework.boot.actuate.autoconfigure.EndpointAutoConfiguration;
 import org.springframework.boot.actuate.endpoint.MetricReaderPublicMetrics;
 import org.springframework.boot.actuate.endpoint.MetricsEndpoint;
-import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.ImportAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.web.EmbeddedServletContainerAutoConfiguration;
 import org.springframework.boot.bind.PropertySourcesPropertyValues;
 import org.springframework.boot.bind.RelaxedDataBinder;
 import org.springframework.boot.bind.RelaxedNames;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.binding.BindableProxyFactory;
+import org.springframework.cloud.stream.config.ChannelBindingAutoConfiguration;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ConfigurableApplicationContext;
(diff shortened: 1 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -455,7 +462,7 @@ public ConfigurableApplicationContext run() {
 
 	}
 
-	@EnableAutoConfiguration
+	@ImportAutoConfiguration({ChannelBindingAutoConfiguration.class, EndpointAutoConfiguration.class})
```
jvalkeal: I think this is making things only marginally better. I remember that my original nitpick was above where this configuration if force fed. I don't like `@ImportAutoConfiguration` any better than `@EnableAutoConfiguration` on a framework level as framework should never force any auto-config. 
```
	private void addParentSources(Object[] sources) {
		if (!this.parentSources.contains(ParentConfiguration.class)) {
			this.parentSources.add(ParentConfiguration.class);
		}
		this.parentSources.addAll(Arrays.asList(sources));
	}
```
Maybe we could add some hooks to replace this config if user so want/need.

ilayaperumalg: @jvalkeal The `@ImportAutoConfiguration` only imports the binding related auto configs which are internal to SCSt; and this is something user wouldn't want to get control over.

> Maybe we could add some hooks to replace this config if user so want/need

We already do that where we use external auto configuration (EmbeddedServlet auto config)

jvalkeal:
Ok, tried with real aggregated app and works better, merging.

jvalkeal:
Just leaving a note here, this now works as is even if postgre libs are on CP
```
@SpringBootApplication(exclude = DataSourceAutoConfiguration.class)
public class FilegpfdistApplication {

		AggregateApplicationBuilder builder = new AggregateApplicationBuilder(FilegpfdistApplication.class, args);
		builder
			.from(LoadGeneratorSourceConfiguration.class).args(args)
			.to(ThroughputSinkConfiguration.class).args(args)
			.run();
	}
}
```
Previously I had to run it with `--spring.autoconfigure.exclude=org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration` to disable that auto-config on child contexts as well.

I still need to exclude on main app as otherwise I get
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Cannot determine embedded database driver class for database type NONE

Action:

If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).

```

jvalkeal:
cherry picked per https://github.com/spring-cloud/spring-cloud-stream/commit/8792423abceb6e12ddbc851a116e3402348bcdb5

ilayaperumalg:
@jvalkeal thank you for the review/verifying and merging. 

In the case of `exclusion` of `DataSourceAutoConfiguration` at the `@SpringBootApplication`, isn't this something user has to exclude anyway because of the `EnableAutoConfiguration` that is brought by the `@SpringBootApplication`?

jvalkeal:
well yes but previously aggregated app forced that full auto-config, which this PR made better. Now I can actually exclude that on main class and not using boot's global property.

PR #811, opened 2017-02-16
Copy existing `BindingProperties` values when replacing bound map

### LINKED ISSUE #810: Regression when `BindingProperties` are manipulated programmatically
This has been observed after introducing support for defaults.

Currently we use a map that initializes with values based on existing defaults. When these properties are preset using this pattern (specific to Spring Cloud) https://github.com/spring-cloud/spring-cloud-bus/blob/master/spring-cloud-bus/src/main/java/org/springframework/cloud/bus/BusAutoConfiguration.java#L69 , the values set are overwritten when the bound map is attached in `org.springframework.cloud.stream.config.BindingServiceProperties#setEnvironment`. This has caused regressions across different Spring Cloud components, with issues like: https://github.com/spring-cloud/spring-cloud-bus/issues/55

The solution is to copy existing properties before replacing the map with the one that initializes missing fields with defaults.

mbogoevici:
Fix #810

(backport to 1.1.x)

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
@@ -127,9 +127,10 @@ public class BindingServiceProperties implements ApplicationContextAware, Enviro
 		if (environment instanceof ConfigurableEnvironment) {
 			// override the bindings store with the environment-initializing version if in
 			// a Spring context
+			Map<String, BindingProperties> delegate = new TreeMap<String, BindingProperties>(String.CASE_INSENSITIVE_ORDER);
+			delegate.putAll(this.bindings);
 			this.bindings = new EnvironmentEntryInitializingTreeMap<>((ConfigurableEnvironment) environment,
-					BindingProperties.class, "spring.cloud.stream.default",
-					new TreeMap<String, BindingProperties>(String.CASE_INSENSITIVE_ORDER));
+					BindingProperties.class, "spring.cloud.stream.default", delegate);
 		}
 	}
 


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
```diff
@@ -127,9 +127,10 @@ public void setEnvironment(Environment environment) {
 		if (environment instanceof ConfigurableEnvironment) {
 			// override the bindings store with the environment-initializing version if in
 			// a Spring context
+			Map<String, BindingProperties> delegate = new TreeMap<String, BindingProperties>(String.CASE_INSENSITIVE_ORDER);
```
ilayaperumalg: Isn't the `bindings` always be `null` at this phase of the lifecycle? Referring the original issue from Spring Cloud Bus [here](https://github.com/spring-cloud/spring-cloud-bus/blob/master/spring-cloud-bus/src/main/java/org/springframework/cloud/bus/BusAutoConfiguration.java#L69), the `bindings` would be set only after the beanPostProcessor EnvironmentAware interfaces are invoked. I am trying to understand what impact does this change make by updating the `this.bindings` at this level.

mbogoevici: `bindings` can't be null here since the field is set to a default value.  The referenced `init()` method executes before this because it's a post-construct  on an autoconfiguration class and will update the map before the environment is set. 

This change ensures that any changes made to the `bindings` field before we introduce the environment-based map is preserved (which was the original issue).

ilayaperumalg: It is observed that `EnvironmentAware` interfaces are invoked via bean post processors every time the `refreshContext` is invoked. In this case, the `bindings` would have already been set with some values.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as a64860abc8f1b6d7d1d1b3d5d8815c96be9f9c70

also cherry-picked and pushed into 1.1.x: 1443ae8c671ec20203f85c9e5e5e0d11e9372edf

PR #814, opened 2017-02-19
AbstractMessageChannelBinder can have a `null` headers list

### LINKED ISSUE #812: AbstractMessageChannelBinder may fail with null headersToEmbed
For example `JMSMessageChannelBinder` call super constructor

```
super(true, null, provisioningProvider)
```

no problem there but if in my custom binder I call it with

```
super(false, null, provisioningProvider)
```

then I will end up with error
```
Caused by: java.lang.NullPointerException
	at org.springframework.cloud.stream.binder.EmbeddedHeadersMessageConverter.embedHeaders(EmbeddedHeadersMessageConverter.java:59)
	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder$SendingHandler.serializeAndEmbedHeadersIfApplicable(AbstractMessageChannelBinder.java:346)
	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder$SendingHandler.handleMessageInternal(AbstractMessageChannelBinder.java:329)
	at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:133)
	... 25 more
```

Looks like this should be either handled manually or `AbstractMessageChannelBinder` should fail fast.

I got pass this by using
```
super(false, new String[0], provisioningProvider);
```

mbogoevici:
Fix #812


jvalkeal:
yup, looks ok as it's the workaround I used manually. merging.

PR #817, opened 2017-02-20
Fix non self-contained aggregate application binding

ilayaperumalg:
- Remove conditionalOnMissingBean check on the necessary beans that are required for exernal bindings of the child application
 - Add test to verify

Resolves #816

PR #823, opened 2017-02-21
Register references to child context instead of the bindable proxies

### LINKED ISSUE #819: Sources that @Autowire Source Cannot be Aggregated
>I receive an error message "Field channels in org.springframework.cloud.stream.app.tcp.source.TcpSourceCon‌​figuration required a single bean, but 2 were found: 1- Source 2- .Processor

See http://stackoverflow.com/questions/42346872/how-to-handle-message-using-spring-cloud-stream-app-starter-tcp/42347372#comment71855225_42347372

One solution might be to configure the output channel by name so it is bound to the actual channel at runtime.

mbogoevici:
Fix #819 

(Merge #817 first as the newly enabled tests will not pass otherwise)


sobychacko:
LGTM - merged to master, back-ported to 1.1.x.

PR #830, opened 2017-02-23
Doc for non self-contained aggregate application

### LINKED ISSUE #818: Binding service properties for aggregate child application are not used when binding
When binding the aggregate child application's bindable types to external inbound/outbound binder, the binding properties are not used.

The reason for this seems to be the inclusion of `conditionalOnMissingBean` on `BindingService` [here](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java#L98).

If this `conditional` check is necessary, we could as well check if using explicit `BindingServiceProperties` as a bean would be an option here.

ilayaperumalg:
Resolves #818

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -745,6 +745,26 @@ For instance, to override the default `fixedDelay` and `debug` properties of 'so
 java -jar target/MyAggregateApplication-0.0.1-SNAPSHOT.jar --source.fixedDelay=10000 --sink.debug=false
 
 ----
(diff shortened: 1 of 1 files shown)


jvalkeal:
Would you mind adding more info what `non self-contained` actually means. There is the sample PR for this which I didn't merge as it's missing readme for same reason. Maybe `non self-contained` is wrong wording as at least I don't connect with its meaning. It may be obvious to you but not to me.

ilayaperumalg:
@jvalkeal sure thing.

jvalkeal:
Thx, much better. Merging...

jvalkeal:
Squashed/merged per https://github.com/spring-cloud/spring-cloud-stream/commit/347cac75f2bd2d8a61dba2f8332f3502ddc374f6

Issue #752, opened 2017-01-04
Support caching of schema instances in `DefaultSchemaRegistryClient`


mbogoevici:

viniciusccarvalho:
This is being closed since #832 is supposed to fix and enhance the caching capabilities of the message converter

PR #831, opened 2017-02-24
Gh 752

viniciusccarvalho:
Adds caching capabilities to the schema registry client. Also pushing docs changes from previous PR

Resolves spring-cloud/spring-cloud-stream#752


spring-cloud-stream-reactive/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,144 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ilayaperumalg: I noticed this file keeps getting generated every time we run the build. May be we better add here. Is that the idea? /cc @mbogoevici

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/CachingRegistryClient.java
```diff
@@ -0,0 +1,62 @@
...
+	private CacheManager cacheManager;
+
+	public CachingRegistryClient(SchemaRegistryClient delegate) {
+		this.delegate = delegate;
+	}
+
+	@Override
+	public SchemaRegistrationResponse register(String subject, String format, String schema) {
+		SchemaRegistrationResponse response = delegate.register(subject,format,schema);
+		cacheManager.getCache("refCache").put(response.getSchemaReference(),schema);
```
ilayaperumalg: Would be better to make these `cache` names as constants.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientConfiguration.java
```diff
@@ -25,17 +29,28 @@
...
 @EnableConfigurationProperties(SchemaRegistryClientProperties.class)
+@EnableCaching
 public class SchemaRegistryClientConfiguration {
 
+	@Autowired
+	private SchemaRegistryClientProperties schemaRegistryClientProperties;
+
 	@Bean
+	@ConditionalOnMissingBean(SchemaRegistryClient.class)
 	public SchemaRegistryClient schemaRegistryClient(SchemaRegistryClientProperties schemaRegistryClientProperties) {
```
ilayaperumalg: Given this `SchemaRegistryClientProperties` is autowired and used inside this method, I believe we can remove the method argument.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientProperties.java
```diff
@@ -28,6 +29,8 @@
 
 	private boolean enabled = true;
 
+	private boolean caching = true;
```
ilayaperumalg: How is this property being used?

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientProperties.java
```diff
@@ -28,6 +29,8 @@
 
 	private boolean enabled = true;
```
ilayaperumalg: I am not sure I understand how is the property `enabled` used as well?

spring-cloud-stream-schema/src/test/java/org/springframework/cloud/schema/client/SchemaClientTests.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = SchemaClientTests.BootApplicationTest.class, webEnvironment = SpringBootTest.WebEnvironment.MOCK, properties = {"spring.cloud.stream.schemaRegistryClient.caching=true","debug=true"})
+@SpyBean(SchemaRegistryClient.class)
+public class SchemaClientTests {
+
```
ilayaperumalg: We can avoid extra spaces in the classes in general.

spring-cloud-stream-reactive/dependency-reduced-pom.xml
mbogoevici: We should add it to .gitignore  and not check it in.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientConfiguration.java
mbogoevici: `@ConditionalOnMissingBean` is unsafe for use outside an autoconfiguration in general. (We make an exception for it in `BindingServiceConfiguration` and that is exclusively for hierarchies).

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientProperties.java
viniciusccarvalho: I don't think it is, I can't really remember why we have it, I'll leave this one out, fixing all other comments on a new commit.

spring-cloud-stream-reactive/dependency-reduced-pom.xml
mbogoevici: We also need to remove this file from the commit.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientConfiguration.java
```diff
@@ -16,7 +16,11 @@
 
 package org.springframework.cloud.stream.schema.client.config;
 
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
```
mbogoevici: Unused import.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientProperties.java
```diff
@@ -43,4 +45,6 @@ public boolean isEnabled() {
 	public void setEnabled(boolean enabled) {
```
mbogoevici: We can remove this property as it is not used in code. The usage within `@ConditionalOnProperties` depends on the presence (or absence) of the property in the environment.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/CachingRegistryClient.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * @author Vinicius Carvalho
+ */
+public class CachingRegistryClient implements SchemaRegistryClient {
+
+	private SchemaRegistryClient delegate;
+
+	protected static final String ID_CACHE = "idCache";
+	protected static final String REF_CACHE = "refCache";
```
mbogoevici: I think the names of the caches should be more Stream-specific (easy to clash with someone's code). Maybe we can consider making them configurable in the future.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1439,6 +1451,13 @@ A client for the Spring Cloud Stream schema registry can be configured using the
   }
 ----
 
+[NOTE]
+====
+The default client caches all responses locally to avoid unnecessary network calls. It leverages Spring caching abstraction, so you can provide
+an external cache provider such as eh-cache or hazelcast if you want to have distributed cache control and invalidation. Schemas are append only entities
+but deletes may occur, make sure you configure your external cache provider to invalidate the cache from time to time to avoid synchronization issues.
+====
```
mbogoevici: Can we detail a little more how that works, as well as the names of the two cache regions that we use?

mbogoevici:
Closing in favour of https://github.com/spring-cloud/spring-cloud-stream/pull/835

PR #840, opened 2017-03-07
Add failing test

dturanski:
@mbogoevici  Not sure if this is an SCSt issue, a core Spring issue, or one of my brain farts. Would appreciate your assistance. Thanks


mbogoevici:
@dturanski We should have used `AnnotatedElementUtils` instead of `AnnotationUtils`. Will provide a PR with the test and the fix shortly.

PR #847, opened 2017-03-09
Exclude test support binder from binder type registry

### LINKED ISSUE #827: Disabling TestSupportBinderAutoConfiguration leads to exception related to mapping default binder
I disabled TestSupportBinderAutoConfiguration by adding

spring.autoconfigure.exclude= org.springframework.cloud.stream.test.binder.TestSupportBinderAutoConfiguration
in applicaiton.properties but it resulted in the error
"Failed to start bean 'outputBindingLifecycle'; nested exception is java.lang.IllegalStateException: A default binder has been requested, but there is more than one binder available: test,kafka, and no default binder has been set."

So even with disabling TestSupportBinderAutoConfiguration test binder is still getting initialized . Am I missing something ?

I am using 1.0.2.RELEASE of spring-cloud-stream-test-support

ilayaperumalg:
- Given the TestSupportBinder has its own BinderFactory that returns the TestSupportBinder, we need to exclude the BinderType entry of TestSupportBinder detected from classpath
  - This will make sure to exclude test support binder when TestSupportBinderAutoConfiguration is excluded when using other explicit binder in tests

Resolves #827


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -117,7 +117,11 @@ public BinderTypeRegistry binderTypeRegistry(ConfigurableApplicationContext conf
 				URL url = resources.nextElement();
 				UrlResource resource = new UrlResource(url);
 				for (BinderType binderType : parseBinderConfigurations(classLoader, resource)) {
-					binderTypes.put(binderType.getDefaultName(), binderType);
+					// Exclude TestSupportBinder when detected as the TestSupport binder is expected to use its own
+					// BinderFactory when used.
+					if (!binderType.getDefaultName().equalsIgnoreCase("test")) {
```
mbogoevici: We shouldn't rely on a hardcoded binder name here.

ilayaperumalg: What do you think would be a better option here? Given it is the name we have given for the `TestSupportBinder` and `spring-cloud-stream-test-support` is part of SCSt, I thought it is ok to use the name here.

mbogoevici: We may just need to add `spring.cloud.stream.binder.test.defaultCandidate=false` in the environment if the TestBinder is on the CP ?

ilayaperumalg: That was my initial approach. But, once we have the property declared, then it is considered `declaredBinders` and our current mechanism overrides the `defaultCandidate` to `true` when there are no more declared binders [here](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java#L91). In any case, we hard code the name `test` here in the property as well ;)

mbogoevici: >In any case, we hard code the name test here in the property as well ;)

There's a difference between using the binder name in the module in which it is declared and  using the hardcoded name in `core`.

ilayaperumalg: yeah, I agree if we use the property in `spring-cloud-stream-test-support`. But the issue I mentioned above still arises.

mbogoevici: Interesting, because `TestSupportBinderAutoConfiguration` would replace the `DefaultBinderFactory`. I would guess that someone would want to suppress the auto-configuration only if they had another binder they wanted to use. Otherwise, what would be the point?

ilayaperumalg: right, but the `BinderTypeRegistry` would always get all the binders from `classpath` even if `TestSupportBinderAutoConfiguration` is disabled and the `DefaultBinderFactory` is used.

mbogoevici: I guess this causes the bean creation to fail. So probably we should provide an empty `BinderTypeRegistry` as part of the autoconfiguration too?

ilayaperumalg: But in this case, the `TestSupportBinderAutoConfiguration` is already **disabled** and no more beans from this. The `BinderFactoryConfiguration` is the one that is active and creates the `BinderFactory`; and hence if at all we need to make changes, then that has to be in the `BinderTypeRegistry` created from `BinderFactoryConfiguration`

mbogoevici: Maybe we should think if other changes are required then. I still think we shouldn't use a hardcoded value `"test"`.

ilayaperumalg: ok

mbogoevici:
Squashed, rebased merged via https://github.com/spring-cloud/spring-cloud-stream/commit/69cbc7f05be60f51ef5d0c62613edca328e0e6bc

PR #848, opened 2017-03-09
Fix StreamListener method with no explicit declarative method params

### LINKED ISSUE #844: Regression when checking declarative methods when argument type is 'Object'
The logic for fixing https://github.com/spring-cloud/spring-cloud-stream/issues/664 is treating methods that take `Object` as argument as declarative, which is a regression. We should revert to previous behaviour (e.g. if `Object` is an argument, treat it as a message handling method).

ilayaperumalg:
- Make sure to check if the method parameters have annotation specified to conclude the method as `declarative` otherwise, it is `handler` method.
 - Add test

Resolves #844

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
@@ -71,7 +71,7 @@ import org.springframework.util.StringUtils;
  */
 public class StreamListenerAnnotationBeanPostProcessor
 		implements BeanPostProcessor, ApplicationContextAware, BeanFactoryAware, SmartInitializingSingleton,
-				InitializingBean {
+		InitializingBean {
 
 	private static final SpelExpressionParser SPEL_EXPRESSION_PARSER = new SpelExpressionParser();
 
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -208,25 +208,25 @@ private boolean checkDeclarativeMethod(Method method, String methodAnnotatedInbo
...
 			}
 		}
 		return false;
 	}
 
-	private boolean isDeclarativeMethodParameter(Object targetBean, MethodParameter methodParameter) {
+	private boolean isDeclarativeMethodParameter(Object targetBean, MethodParameter methodParameter, boolean hasParameterAnnoation) {
 		if (targetBean != null) {
-			if (methodParameter.getParameterType().isAssignableFrom(targetBean.getClass())) {
+			if (hasParameterAnnoation && methodParameter.getParameterType().isAssignableFrom(targetBean.getClass())) {
```
mbogoevici: I don't think this will work properly. Consider this use case:

```
public interface KStreamProcessor {
	@Input("input")
	KStream<?, ?> input();

	@Output("output")
	KStream<?, ?> output();
}
```
and
```
@StreamListener("input")
@SendTo("output")
public KStream<?, WordCount> process(KStream<?, String> input) {
// omitted
}
```

This change will prevent the parameter from being treated as declarative, won't it?

ilayaperumalg: I think we are good here as the next conditional check to verify if there are matching `StreamListenerParameterAdapters` is the one being used for the above case. or, in this case, Is the target bean (inbound/outbound) are assignable from `KStream` ?

mbogoevici: Yeah, in this case the target bean is a `KStream` and we may not have a matching `StreamListenerParameterAdapter`.

mbogoevici:
Squashed, rebased merged as https://github.com/spring-cloud/spring-cloud-stream/commit/8230a8b3485d07a93707f099c2b97b6227b29a5e

PR #851, opened 2017-03-09
Kafka GH-21: Don't Override User Partition

garyrussell:
See spring-cloud/spring-cloud-stream-binder-kafka#109

If the user app sets the `BinderHeader.PARTITION_HEADER`, don't override it in the binder.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
@@ -275,12 +275,16 @@ public class MessageConverterConfigurer implements MessageChannelConfigurer, Bea
 
 		@Override
 		public Message<?> preSend(Message<?> message, MessageChannel channel) {
-			int partition = this.partitionHandler.determinePartition(message);
-			return MessageConverterConfigurer.this.messageBuilderFactory
-					.fromMessage(message)
-					.setHeader(BinderHeaders.PARTITION_HEADER, partition)
-					.build();
-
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
LGTM, would be nice to have some tests for this case. will add it when merging.

ilayaperumalg:
@garyrussell @mbogoevici I believe this needs to be backported to 1.1.x as well?

ilayaperumalg:
Merged as 34acc2d190ee5798d8ebb3df4639969f75d84368 along with the tests f14d4d85c5cfd76da7643456891ab45277031ba3 into `master`.

Keeping this open until backport of the changes is confirmed.

garyrussell:
I need to rethink this a bit - with Rabbit, the header will be propagated from an input channel to an output channel, which is not what we desire. We just need a way for a user to override the binder partition selection (such as for replaying dead letters). 

Perhaps we need a new header `BinderHeaders.PRESELECT_PARTITION` which is used to populate `BinderHeaders.PARTITION_HEADER` (and is not propagated).

mbogoevici:
Should we at the very least log a warning? This is a generally undesirable use case: the user has configured partitioning on the output and for some reason (e.g. erroneous header propagation) the message contains a partition number - which renders the partitioning setting moot. I do not think that the user should be allowed to manually bypass the partitioning configuration, at least not silently. I also do think that the output partitioning configuration shiuld take precedence.

garyrussell:
See the referenced PR in the kafka binder - replaying dead-letter messages needs some mechanism to bypass binder partitioning.

mbogoevici:
This was merged in master already, I guess we should revert ?

garyrussell:
I issued a new PR

PR #852, opened 2017-03-10
Clarify content type conversion in doc

### LINKED ISSUE #797: Improve the content type and transformation section of the documentation
I have been struggling with serialization/deserialization for some days now, reading and rereading the documentation and debugging the source code.

Maybe it's just me but I would suggest that the documentation was cleared up a bit:

1. **Default serialization:** this [closed issue](https://github.com/spring-cloud/spring-cloud-stream/issues/156) discussed about the default serialization and the bad practice of having to share classes between the producer and consumer sides. The point is **that that bad practice is still the default option in the framework but we can't guess it from the docs.**. 

In my case, I send a `service1.Order` object:

`orderProcessor.output().send(MessageBuilder.withPayload(order).build());`

And the consuming service fails:

```
Caused by: org.springframework.core.serializer.support.SerializationFailedException: unable to deserialize [service1.Order]. Class not found.; nested exception is java.lang.ClassNotFoundException: service1.Order
	at org.springframework.cloud.stream.binder.AbstractBinder.deserializePayload(AbstractBinder.java:255) ~[spring-cloud-stream-1.1.0.RELEASE.jar:1.1.0.RELEASE]
```

This is my listener that expects service2.Order:

```
@StreamListener(OrderProcessor.INPUT)
public void handle(Order order){
```
I can see that the embedded headers indicate `contentType=application/x-java-object;type=service1.Order` so it's obvious why it's failing.

Specifying content-type: application/json in the configuration of the output channel solves this but, shouldn't it be indicated more clearly in the docs? I was kind of naive expecting a JSON conversion as the default one.

The documentation currently states: 

> Spring Cloud Stream attaches, by default, a contentType header to outbound messages

But IMO we can't directly guess that we will need by default to share classes between services.

2. **6.2 section:** I don't get what the columns really mean, my interpretation is:

    - **Source payload**: the object passed to the MessageBuilder.withPayload() function.
    - **Target payload**: the expected payload after the serialization, the one that will actually be transmitted (well, actually the source of the transmitted byte[]).
    - **content-type header vs content type**: the former a content type header passed while building the Message on the producer side. The latter the param that can be specified in the configuration property.

Am I right? 

Also, I guess that table only applies regarding the producing side, no the consuming side.

Finally I just want to point out that this is only a suggestion based on my Spring Cloud Stream newbie experience. The platform is great and I do appreciate the clarifications and feedback I have been getting from the team.

Regards

ilayaperumalg:
Resolves #797

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1257,9 +1257,11 @@ In addition, Spring Cloud Stream provides custom MIME types, notably, `applicati
 === MIME types and Java types
 
 The type conversions Spring Cloud Stream provides out of the box are summarized in the following table:
+Source Payload represents the 'payload' before conversion and Target Payload represents the 'payload' after conversion.
+The message type-conversion can occur either at the 'producer' side (outbound) or at the 'consumer' side (inbound).
 
 |===
-|Source Payload |Target Payload |content-type header | content-type | Comments
+|Source Payload |Target Payload |content-type header at source payload | content-type at target payload after conversion | Comments
 
 |POJO
 |JSON String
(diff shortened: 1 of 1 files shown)


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1317,14 +1319,19 @@ The type conversions Spring Cloud Stream provides out of the box are summarized
 
 |===
 
-[[NOTE]]
+[NOTE]
+====
 Conversion applies to payloads that require type conversion.
 For example, if a module produces an XML string with outputType=application/json, the payload will not be converted from XML to JSON.
-This is because the payload at the module's output channel is already a String so no conversion will be applied at runtime.
+This is because the payload at the module's outbound channel is already a String so no conversion will be applied at runtime.
```
mbogoevici: Not entirely clear why we needed to change 'input'->'inbound' and 'output'->'outbound' ? Both are valid choices, but we use `@Input` and `@Output`.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1257,9 +1257,11 @@ In addition, Spring Cloud Stream provides custom MIME types, notably, `applicati
 === MIME types and Java types
 
 The type conversions Spring Cloud Stream provides out of the box are summarized in the following table:
+Source Payload represents the 'payload' before conversion and Target Payload represents the 'payload' after conversion.
```
mbogoevici: We don't need apostrophes.

mbogoevici:
Closed via https://github.com/spring-cloud/spring-cloud-stream/commit/567bdb4b2cc2d888493175160ad3873646b24da9

PR #853, opened 2017-03-10
kGH-21: Add BinderHeaders.PARTITION_OVERRIDE

garyrussell:
See spring-cloud/spring-cloud-stream-binder-kafka#109
See #851

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderHeaders.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderHeaders.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderHeaders.java
```diff
@@ -40,6 +40,12 @@
 	public static final String PARTITION_HEADER = "partition";
 
 	/**
+	 * Indicates the target partition of an outbound message. Overrides
+	 * any partition selected by the binder.
+	 */
+	public static final String PARTITION_OVERRIDE = "partitionOverride";
```
artembilan: Not related to the subject, but why there is no `prefix`?
Imagine downstream application which has some logic around headers.
So, how to distinguish binder headers from all others?
Or just how to understand that the header in target message is related to binder not some other "hell"?

Thanks

garyrussell: Yes; I thought about that too but I fear it's too late now 😦

artembilan: Well, we always can do that and support both headers sets some time.
Like that was once in Spring Cloud Sleuth.

garyrussell:
Talked to Marius; we decided it's ok to prefix the partition headers but not the oCT - testing now...

mbogoevici:
@garyrussell Small updates to Javadoc - let me know if you agree. https://github.com/mbogoevici/spring-cloud-stream/commit/458ed08688a483d798f25bb3698e3bfeff73f42c

garyrussell:
LGTM; thanks.

mbogoevici:
Squashed and merged w/updates via https://github.com/spring-cloud/spring-cloud-stream/commit/bd1454bffff52d29f113f040076eb8fc656122cf

PR #854, opened 2017-03-10
Initial metrics implementation

### LINKED ISSUE #846: Add metrics module
Add the new metrics emitter module that will publish boot metrics to a channel

viniciusccarvalho:
Fix #846 

Initial metrics support, adds emitter to send metrics from boot's `/metrics` endpoint on a certain frequency

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1806,8 +1806,12 @@ Spring Cloud Stream provides a health indicator for binders.
 It is registered under the name of `binders` and can be enabled or disabled by setting the `management.health.binders.enabled` property.
 
 == Metrics Emitter
-Spring Cloud Stream provides a module called `spring-cloud-stream-metrics` that can be used to emit any available metric from boot's `/metrics` endpoint to a named channel.
-This module allow operators to collect metrics from stream applications without relying on polling their endpoints, which could be on a private network or behind an LB that prevents per instance access.
+Spring Cloud Stream provides a module called `spring-cloud-stream-metrics` that can be used to emit any available metric from https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html[Spring Boot metrics endpoint] to a named channel.
+This module allow operators to collect metrics from stream applications without relying on polling their endpoints.
+
+HTTP polling can be challenging on cloud environments since applications could be on a private network or behind a Load balancer that prevents per instance access.
+
+This module is included on the each of the out of box https://github.com/spring-cloud-stream-app-starters[application starters], if you want to emit metrics from your application, just import it on your pom.xml and follow the activation instructions.
 
 The module is activated when you set the destination name for its channel, `spring.cloud.stream.bindings.streamMetrics.destination=<DESTINATION_NAME>`.
 
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1805,6 +1805,46 @@ Once we have received the message, we can validate that the component functions 
 Spring Cloud Stream provides a health indicator for binders.
 It is registered under the name of `binders` and can be enabled or disabled by setting the `management.health.binders.enabled` property.
 
+== Metrics Emitter
+Spring Cloud Stream provides a module called `spring-cloud-stream-metrics` that can be used to emit any available metric from boot's `/metrics` endpoint to a named channel.
+This module allow operators to collect metrics from stream applications without relying on polling their endpoints, which could be on a private network or behind an LB that prevents per instance access.
+
+The module is activated when you set the destination name for its channel, `spring.cloud.stream.bindings.streamMetrics.destination=<DESTINATION_NAME>`.
+
+By default the module is configured to only send Spring Integration message channel metrics.
+
+Available properties for customization using the prefix `spring.cloud.stream.metrics`.
+
+delay-millis::
+  The period in which metrics will be posted to the channel
++
+Default: `5000`.
++
+prefix::
+  Prefix string to be prepended to the metrics name
++
+Default: ``
++
+includes::
+  An array of strings containing regex patterns of metrics that should be included.
++
+Default: `integration**`
++
+excludes::
+  An array of strings containing regex patterns of metrics that should be excluded.
++
+Default: ``
++
+properties::
+  Just like the `includes` option, it allows white listing application properties that will be added to the metrics payload
++
+Default: null.
++
+spring.cloud.stream.bindings.streamMetrics.contentType::
+  Content-Type of the message
++
+Default: `application/json`
+
 == Samples
 
 For Spring Cloud Stream samples, please refer to the https://github.com/spring-cloud/spring-cloud-stream-samples[spring-cloud-stream-samples] repository on GitHub.
--- a/spring-cloud-stream-metrics/pom.xml
+++ b/spring-cloud-stream-metrics/pom.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-cloud-stream-metrics</artifactId>
+    <description>Emitter module to publish boot metrics</description>
+    <parent>
+        <groupId>org.springframework.cloud</groupId>
+        <artifactId>spring-cloud-stream-parent</artifactId>
+        <version>1.2.0.BUILD-SNAPSHOT</version>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-stream</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-stream-test-support</artifactId>
+        </dependency>
+    </dependencies>
+    
+</project>
--- a/spring-cloud-stream-metrics/pom.xml
+++ b/spring-cloud-stream-metrics/pom.xml
@@ -27,7 +27,8 @@
         <dependency>
             <groupId>org.springframework.cloud</groupId>
             <artifactId>spring-cloud-stream-test-support</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
-    
+
 </project>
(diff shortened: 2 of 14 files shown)


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1805,6 +1805,46 @@ Once we have received the message, we can validate that the component functions
 Spring Cloud Stream provides a health indicator for binders.
 It is registered under the name of `binders` and can be enabled or disabled by setting the `management.health.binders.enabled` property.
 
+== Metrics Emitter
+Spring Cloud Stream provides a module called `spring-cloud-stream-metrics` that can be used to emit any available metric from boot's `/metrics` endpoint to a named channel.
```
mbogoevici: We should explicitly tell the user that they need to include the module in their application.

mbogoevici: Also, phrasing here - instead of 'boot's `/metrics` endpoint' we can say something more elaborate - maybe a reference to relevant documentation in Spring Boot ?

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1805,6 +1805,46 @@ Once we have received the message, we can validate that the component functions
...
+This module allow operators to collect metrics from stream applications without relying on polling their endpoints, which could be on a private network or behind an LB that prevents per instance access.
```
mbogoevici: LB->load balancer. Also security could be an issue.

spring-cloud-stream-metrics/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-stream-test-support</artifactId>
+        </dependency>
+    </dependencies>
+    
+</project>
```
mbogoevici: Missing EOL at the end of the file.

spring-cloud-stream-metrics/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+            <artifactId>spring-cloud-stream</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-stream-test-support</artifactId>
```
mbogoevici: Must be in test scope.

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/config/metrics/BinderMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.cloud.stream.metrics.Emitter;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@Configuration
+@ConditionalOnClass({Binder.class})
```
mbogoevici: May not be an array.

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/BootMetricJsonSerializer.java
```diff
@@ -0,0 +1,78 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+
+import org.springframework.boot.actuate.metrics.Metric;
+import org.springframework.boot.jackson.JsonComponent;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@JsonComponent
+public class BootMetricJsonSerializer {
```
mbogoevici: Should this be called `SpringBootMetricJsonSerializer` instead?

spring-cloud-stream-metrics/src/main/resources/config/application.properties
```diff
@@ -0,0 +1,2 @@
+spring.cloud.stream.bindings.streamMetrics.contentType=application/json
```
mbogoevici: We shouldn't provide an `application.properties` file, because it will be overridden by the application's `application.properties` file (only one file from the classpath is considered). Instead, we should use a file with a specific name and include it via `ConfigurationProperties` or use an `EnvironmentPostProcessor`.

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/BootMetricJsonSerializer.java
viniciusccarvalho: I thought for the intimate we would just call it boot ;)

markpollack:
I think that the published message should have a key-value pair representing the version of the message, for example `version=1`.   We don't have to get into full-on schema registry support, but as this will be a key message in systems going forward, we can expect it to evolve and adding a version number will help with consumer implementations as the ecosystem grows.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1805,6 +1805,50 @@ Once we have received the message, we can validate that the component functions
 Spring Cloud Stream provides a health indicator for binders.
 It is registered under the name of `binders` and can be enabled or disabled by setting the `management.health.binders.enabled` property.
 
+== Metrics Emitter
+Spring Cloud Stream provides a module called `spring-cloud-stream-metrics` that can be used to emit any available metric from https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html[Spring Boot metrics endpoint] to a named channel.
+This module allow operators to collect metrics from stream applications without relying on polling their endpoints.
+
+HTTP polling can be challenging on cloud environments since applications could be on a private network or behind a Load balancer that prevents per instance access.
+
+This module is included on the each of the out of box https://github.com/spring-cloud-stream-app-starters[application starters], if you want to emit metrics from your application, just import it on your pom.xml and follow the activation instructions.
```
markpollack: This is probably a premature statement ATM, being included in app starters.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1805,6 +1805,50 @@ Once we have received the message, we can validate that the component functions
...
+Default: ``
++
+includes::
+  An array of strings containing regex patterns of metrics that should be included.
++
+Default: `integration**`
++
+excludes::
+  An array of strings containing regex patterns of metrics that should be excluded.
++
```
markpollack: should we have a default value for excludes wrt to that _handler endpoint?  not sure what that was.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
mbogoevici: Maybe we can just remove the statement altogether. Stream is upstream of that.

viniciusccarvalho:
@markpollack Agree on versioning, we should follow same pattern as the schema server and hypermedia applications IMO. So `application/vnd.metrics.v1+json` should be the header. We don't pollute the body, and gives a hint for the schema server once we support json (on the roadmap, but long shot giving total lack of schema support by JSON spec)

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/config/metrics/StreamMetricsProperties.java
```diff
@@ -0,0 +1,111 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties(prefix = "spring.cloud.stream.metrics")
+public class StreamMetricsProperties {
+
+
+	private String prefix;
```
markpollack: `prefix` isn't used anymore, so can be removed

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/config/metrics/StreamMetricsProperties.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+package org.springframework.cloud.stream.config.metrics;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Vinicius Carvalho
```
markpollack: A one liner of javadoc would be good.

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/ApplicationMetrics.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.boot.actuate.metrics.Metric;
+
+/**
+ * @author Vinicius Carvalho
+ */
+public class ApplicationMetrics {
+
+	private String name;
+
+	private int index;
```
markpollack: I think it would help a bit to call this `instanceIndex`

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/config/metrics/StreamMetricsProperties.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+@ConfigurationProperties(prefix = "spring.cloud.stream.metrics")
+public class StreamMetricsProperties {
+
+
+	private String prefix;
+
+	@Value("${spring.application.name:application}")
+	private String applicationName;
```
markpollack: I think this needs to be the same as https://github.com/spring-projects/spring-boot/blob/v1.5.2.RELEASE/spring-boot/src/main/java/org/springframework/boot/context/ContextIdApplicationContextInitializer.java#L76  so that it can also take into account names that come from config server.

markpollack:
Can you sprinkle some javadoc at the class level?  doesn't have to be too crazy...

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
markpollack: yea, better to keep the flow such that we don't ref downstream components/projects.

spring-cloud-stream-metrics/src/test/java/org/springframework/cloud/stream/metrics/BinderMetricsEmitterTests.java
```diff
@@ -0,0 +1,182 @@
...
+				"--spring.cloud.stream.metrics.properties=java**");
+		Emitter emitterSource = applicationContext.getBean(Emitter.class);
+		MessageCollector collector = applicationContext.getBean(MessageCollector.class);
+		Message message = collector.forChannel(emitterSource.metrics()).poll(1000, TimeUnit.MILLISECONDS);
+		Assert.assertNotNull(message);
+		ObjectMapper mapper = applicationContext.getBean(ObjectMapper.class);
+		ApplicationMetrics applicationMetrics = mapper.readValue((String)message.getPayload(),ApplicationMetrics.class);
+		Assert.assertFalse(contains("mem", applicationMetrics.getMetrics()));
+		Assert.assertTrue(contains("integration.channel.errorChannel.errorRate.mean",applicationMetrics.getMetrics()));
+		Assert.assertFalse(CollectionUtils.isEmpty(applicationMetrics.getProperties()));
```
markpollack: should check for actual property value.

markpollack:
squashed, merged and rebased as e05d5e03bd08df0c22a189774db83a29d9e985a4   Closes #846

PR #860, opened 2017-03-14
Avoiding duplicate entries on properties map

### LINKED ISSUE #859: Metrics Emitter sending duplicate entries for properties
It seems that the emitter is sending relaxed properties more than once as show in this snippet:

`"spring.cloud.stream.bindings.streamMetrics.destination" : "foo",
    "spring.cloud.stream.bindings.stream-metrics.destination" : "foo",`

We should prevent that same properties are sent more than once

viniciusccarvalho:
Fixes #859


ericbottard:
Dumb question: why use relaxed names in the first place?

ericbottard:
> The fix to avoid duplicates was easy, we are now looking into sending all properties in a canonical dot format. So JAVA_HOME becomes java.home

Not sure what you mean by that, as the changed test seems to imply the contrary. Props are now sent as all caps. Or do you mean "we'll need another PR to send them as canonical dot format"

viniciusccarvalho:
Closed after internal discussion, new PR to follow with changes

PR #861, opened 2017-03-14
Allow non byte array payload with raw HeaderMode

### LINKED ISSUE #855: Allow String Payload with headerMode=raw
Convert to UTF-8.

sobychacko:
Fixes #855

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -24,7 +24,6 @@ import org.springframework.cloud.stream.provisioning.ProvisioningProvider;
 import org.springframework.context.Lifecycle;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
-import org.springframework.http.MediaType;
 import org.springframework.integration.channel.FixedSubscriberChannel;
 import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.endpoint.EventDrivenConsumer;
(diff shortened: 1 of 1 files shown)


garyrussell:
Is the intention here, to allow any payload type, because the user can always configure a kafka `Serializer` to actually do the conversion?

It's not clear how the kafka test works because I don't see it configuring a `StringSerializer` on the outbound side.

If we want to support any type we should probably include a test for, say, the `JsonSerializer` too.

mbogoevici:
@garyrussell Using arbitrary payloads is already supported by setting `useNativeSerialization` in the producer properties https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java#L130 which is tested in the Kafka Binder via https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/Kafka10BinderTests.java#L183
(Uses Avro serialization). 

We could add an additional test for JsonDeserializer though. 

Unfortunately, there is an issue with simply removing the block, which is that if `embeddedHeaders` is `raw` will send out a `null` payload - so we should use the contents of `transformed` instead.

sobychacko:
@mbogoevici if the header mode is raw, we still get the payload from the transformed message as here:https://github.com/sobychacko/spring-cloud-stream/blob/d8a3a7a45474646280af51fb0ed1acc239a56400/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java#L349

Is that what you meant?

mbogoevici:
@sobychacko Yeah, I thought we were missing a branch, but figured out that we don't afterwards.

mbogoevici:
@garyrussell The test works because it relies on sending the String using its UTF-8 encoding. Maybe we can test using a `StringDeserializer` on the other end rather than reading the raw bytes?

garyrussell:
I wrote that comment without actually looking at any code.

I recall seeing that we configure `ByteArraySerializer` by default so it's just a bit confusing to me how sending a String will work because that serializer only takes a `byte[]`. So I assumed I would see an override to the `StringSerializer` in the test.

But I see we unconditionally convert String to byte[] (UTF-8) in `serializePayloadIfNecessary`.

I presume you mean `useNativeEncoding` - I see we pass the message unchanged in that case.

Cool

garyrussell:
Too many options (but all necessary, I guess 😄 )

mbogoevici:
@garyrussell Thanks! Actually - we need to highlight this better in the docs. Some of our more recent questions weren't aware that `useNativeEncoding` is even an option.

garyrussell:
We should maybe consider deprecating and renaming to `useNativeSerialization` - `useNativeEncoding` is sooooo 0.8.x.x 😄

garyrussell:
I wonder if we should also infer `useNativeEncoding` if the serializer is set to anything other than a `ByteArraySerializer` - or at least log a WARN?

mbogoevici:
This is a core feature, not a Kafka one - so we're not necessarily beholden to the Kafka naming conventions. In theory any binder can use it to delegate to a native serialization/encoding mechanism.

mbogoevici:
IIRC think one of the reasons why we punted in the past on `useNativeSerialization` vs `useNativeEncoding` was that it sounded to much like Java serialization (which we thankfully never use :D).

PR #870, opened 2017-03-22
Avoid eager creation of bound target beans

### LINKED ISSUE #767: Method checks in `StreamListenerBeanPostProcessor` involve eager bean creation
In particular this bit works only if the bean exists:

https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java#L169

We may want to be a bit lenient there, for cases when beans are created by other BeanPostProcessors.

Note: this might affect `invokeSetupMethodOnListenedChannel` as well.

ilayaperumalg:
- At StreamListenerBeanPostProcessor, avoid eager creation of inbound/outbound target beans whenever possible
  - Use `applicationContext.getType(targetName)` instead of `applicationContext.getBean(targetName)` which checks the type of target beans from bean factory instead of creating the bean eagerly.
  - For declarative method invocation, `applicationContext.getBean` is needed as it requires the actual bean as its argument

Resolves #767

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerHandlerMethodTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerHandlerMethodTests.java
@@ -30,10 +30,14 @@ import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.annotation.Input;
 import org.springframework.cloud.stream.annotation.Output;
 import org.springframework.cloud.stream.annotation.StreamListener;
+import org.springframework.cloud.stream.binding.StreamListenerErrorMessages;
 import org.springframework.cloud.stream.messaging.Processor;
 import org.springframework.cloud.stream.messaging.Sink;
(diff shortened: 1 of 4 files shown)


spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerHandlerMethodTests.java
```diff
@@ -138,9 +139,9 @@ public void testMethodInvalidInboundName() throws Exception {
 			SpringApplication.run(TestMethodInvalidInboundName.class, "--server.port=0");
 			fail("Exception expected on using invalid inbound name");
 		}
-		catch (BeanCreationException e) {
```
mbogoevici: Why did this change?

ilayaperumalg:
Updated the PR with the test and some changes related to handling the exceptions.

mbogoevici:
Merged, rebased squashed via https://github.com/spring-cloud/spring-cloud-stream/commit/ba18c01602f74267aa4a060105f700645523f12b

PR #872, opened 2017-03-22
Leniently support treating input Strings as JSON when converting to POJO

### LINKED ISSUE #871: Regression on JSON conversion when content-type is `text/plain`
We previously were lenient about inbound JSON content with content-type 'text/plain' and allowed the converter system to convert it to a POJO - with the introduction of `ObjectStringMessageConverter`, it takes over all the conversion of inbound 'text/plain' content and converts it to a String (even for a different targetClass - this is incorrect - it should only honour a `String` targetClass)

mbogoevici:
Fixes #871

* Only support conversion based on `text/plain` header if target type is String;

(Note: requires backport to 1.1.x)


sobychacko:
Looks good. Merged and back-ported to 1.1.x.

PR #874, opened 2017-03-24
Make spring-boot actuator as optional dependency

### LINKED ISSUE #869: spring-boot-starter-actuator as optional dependency
Hi,

is it possible to define spring-boot-starter-actuator as optional dependency?

Boot as well as Spring Cloud Bus have actuator as an option which I think is quite nice. However, Stream defines it as mandatory. It can't be excluded as DefaultBinderFactory has a dependency on it. Can you make that usage ConditionalOnClass?

We are using stream in our project and I would prefer not for force actuator on our users.

Kai

ilayaperumalg:
- BinderFactoryConfiguration has a conditional configuration that creates `BinderFactory` with binder health indicators only if the spring-boot `actuator` is in classpath
 - Move Channels endpoint configuration into a separate conditional configuration class

Resolves #869

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderFactoryWithHealthIndicator.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderFactoryWithHealthIndicator.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.binder;
+
+import java.util.Map;
+
+import org.springframework.boot.actuate.health.AbstractHealthIndicator;
+import org.springframework.boot.actuate.health.CompositeHealthIndicator;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.boot.actuate.health.OrderedHealthAggregator;
+import org.springframework.context.ConfigurableApplicationContext;
+
+/**
+ * {@link BinderFactory} that extends {@link DefaultBinderFactory} and provides {@link HealthIndicator} support.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class BinderFactoryWithHealthIndicator extends DefaultBinderFactory {
+
+	private volatile CompositeHealthIndicator bindersHealthIndicator;
+
+	public BinderFactoryWithHealthIndicator(Map<String, BinderConfiguration> binderConfigurations) {
+		super(binderConfigurations);
+	}
+
+	public void setBindersHealthIndicator(CompositeHealthIndicator bindersHealthIndicator) {
+		this.bindersHealthIndicator = bindersHealthIndicator;
+	}
+
+	@Override
+	protected <T> Binder<T, ?, ?> getBinderInstance(String binderConfigurationName) {
+		Binder<T, ?, ?> binder = super.getBinderInstance(binderConfigurationName);
+		if (this.bindersHealthIndicator != null) {
+			OrderedHealthAggregator healthAggregator = new OrderedHealthAggregator();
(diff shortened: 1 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderFactoryWithHealthIndicator.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.boot.actuate.health.OrderedHealthAggregator;
+import org.springframework.context.ConfigurableApplicationContext;
+
+/**
+ * {@link BinderFactory} that extends {@link DefaultBinderFactory} and provides {@link HealthIndicator} support.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class BinderFactoryWithHealthIndicator extends DefaultBinderFactory {
```
mbogoevici: I am not sure that subclassing is the best possible strategy here. Feels like this is something we might to address using a listener or observer. (see below)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -51,21 +59,36 @@
  * @author Ilayaperumal Gopinathan
  */
 @Configuration
+@Import(BinderFactoryConfiguration.BindersHealthIndicatorConfiguration.class)
```
mbogoevici: Do we need to import it if it's nested?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -229,7 +212,7 @@ else if (!StringUtils.hasText(this.defaultBinder)) {
 	/**
 	 * Utility class for storing {@link Binder} instances, along with their associated contexts.
 	 */
-	private static final class BinderInstanceHolder {
+	protected static final class BinderInstanceHolder {
```
mbogoevici: This is only necessary if we're subclassing. Otherwise we can keep it private.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderFactoryWithHealthIndicator.java
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	public void setBindersHealthIndicator(CompositeHealthIndicator bindersHealthIndicator) {
+		this.bindersHealthIndicator = bindersHealthIndicator;
+	}
+
+	@Override
+	protected <T> Binder<T, ?, ?> getBinderInstance(String binderConfigurationName) {
+		Binder<T, ?, ?> binder = super.getBinderInstance(binderConfigurationName);
+		if (this.bindersHealthIndicator != null) {
```
mbogoevici: Can we move this into a 'Listener' type of class registered with the factory - maybe add other types of customizable behaviour in the future. Subclassing feels very invasive.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -155,4 +178,21 @@ public BinderTypeRegistry binderTypeRegistry(ConfigurableApplicationContext conf
 		}
 		return parsedBinderConfigurations;
 	}
+
+	@ConditionalOnClass(name = "org.springframework.boot.actuate.health.HealthIndicator")
```
mbogoevici: This shouldn't move from an autoconfiguration to an imported configuration. I also think that in this case we need to make this a separate autoconfiguration (`@AutoConfigureBefore(EndpointAutoConfiguration.class)`) so that the added health indicators are properly injected.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/ChannelBindingAutoConfiguration.java
```diff
@@ -41,34 +33,20 @@
  *
  * @author Dave Syer
  * @author Marius Bogoevici
+ * @author Ilayaperumal Gopinathan
  */
 @Configuration
 @ConditionalOnBean(BindingService.class)
 @EnableConfigurationProperties(DefaultPollerProperties.class)
-@AutoConfigureBefore(EndpointAutoConfiguration.class)
+@Import(ChannelsEndpointConfiguration.class)
```
mbogoevici: We need `ChannelsEndpointConfiguration` (rename to `ChannelsEndpointAutoConfiguration`) to `spring.factories` rather than importing it here (among other things, explicit import renders `AutoConfigureBefore` moot, I think).

mbogoevici:
Added some refactoring here: https://github.com/mbogoevici/spring-cloud-stream/commit/993059e92a73ecc0015be5abab7c308f62667268

ilayaperumalg:
@mbogoevici The refactoring changes look good to me. Also verified the changes using a SCSt app (with/without actuator dependency).

PR #880, opened 2017-03-27
Changing name of property

### LINKED ISSUE #879: Refactor MetricsProperties applicationName property
Currently the the properties of metrics module allow one to set the applicationName, that is unfortunate a misleading property label, and we should change it to `key`, key could be by default anything the user or deployer desires, it defaults to `spring.application.name` but it does not necessary needs to be an application name, it could be any value a user or a deployer decides to use.

viniciusccarvalho:
Fixes #879 

Better naming for ApplicationMetrics unique identifier


mbogoevici:
Rebased, updated commit message, merged as https://github.com/spring-cloud/spring-cloud-stream/commit/9fdd7f9fc5eb8fffa3a309faf90ac1c6c664221c

PR #885, opened 2017-03-29
StreamEmitter: Initial commits [Not for merging - Only for reviewing and gathering feedback]

### LINKED ISSUE #722: Add support for Reactive sources
See http://stackoverflow.com/questions/40863807/spring-cloud-stream-source-returning-a-reactive-flux for details.

As `@StreamListener` is mainly intended for `Processor` and `Sink`, we need a mechanism that supports reactive sources (or something more generic) as well.

sobychacko:
Add StreamEmitter annotation and its corresponding bean post processor
Add tests to verify the basic behavior and validate the annotation method
Refactor the error message abstract classes
Add a common utility class where both StreamEmitter and StreamListener could share utility methods

Resolves #722

### DIFF
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
@@ -51,6 +51,11 @@ public class ReactiveSupportAutoConfiguration {
 		return new FluxToMessageChannelResultAdapter();
 	}
 
+	@Bean
+	public StreamEmitterAnnotationBeanPostProcessor streamEmitterAnnotationBeanPostProcessor() {
+		return new StreamEmitterAnnotationBeanPostProcessor();
+	}
+
 	@Configuration
 	@ConditionalOnClass(name = "rx.Observable")
 	public static class RxJava1SupportConfiguration {
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
@@ -52,7 +52,7 @@ public class ReactiveSupportAutoConfiguration {
 	}
 
 	@Bean
-	public StreamEmitterAnnotationBeanPostProcessor streamEmitterAnnotationBeanPostProcessor() {
+	public static StreamEmitterAnnotationBeanPostProcessor streamEmitterAnnotationBeanPostProcessor() {
 		return new StreamEmitterAnnotationBeanPostProcessor();
 	}
 
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
@@ -41,30 +41,36 @@ import org.springframework.cloud.stream.annotation.Output;
  *
  * A StreamEmitter method that has a return type cannot take any method parameters.
  *
+ * <pre class="code">
  * &#064;StreamEmitter
  * &#064;Output(Source.OUTPUT)
  * public Flux<String> emit() {
  * 	return Flux.intervalMillis(1000)
  * 		.map(l -> "Hello World!!");
  * }
+ * </pre>
  *
  * The following examples show how a void return type can be used on a method with StreamEmitter and how the
  * method signatures could be used in a flexible manner.
  *
+ * <pre class="code">
  * &#064;StreamEmitter
  * public void emit(&#064;Output(Source.OUTPUT) FluxSender output) {
  * 	output.send(Flux.intervalMillis(1000)
  *		.map(l -> "Hello World!!"));
  * }
+ * </pre>
  *
+ * <pre class="code">
  * &#064;StreamEmitter
  * &#064;Output(Source.OUTPUT)
  * public void emit(FluxSender output) {
  * 	output.send(Flux.intervalMillis(1000)
  *		.map(l -> "Hello World!!"));
  * }
+ * </pre>
  *
- *
+ * <pre class="code">
  * &#064;StreamEmitter
  * public void emit(&#064;Output("OUTPUT1") FluxSender output1,
  *					&#064;Output("OUTPUT2") FluxSender output2,
@@ -76,6 +82,7 @@ import org.springframework.cloud.stream.annotation.Output;
  *	output3.send(Flux.intervalMillis(1000)
  *		.map(l -> "Hello World!!"));
  *	}
+ *</pre>
  *
  * @author Soby Chacko
  */
@@ -83,4 +90,6 @@ import org.springframework.cloud.stream.annotation.Output;
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 public @interface StreamEmitter {
+
+	String value() default "";
 }
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.reactive;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.annotation.Output;
+
+/**
+ * Method level annotation that marks a method to be an emitter to outputs declared via
+ * {@link EnableBinding} (e.g. channels).
+ *
+ * This annotation is intended to be used in a Spring Cloud Stream application that requires
+ * a source to write to one or more {@link Output}s using the reactive paradigm.
+ *
+ * No {@link Input}s are allowed on a method that is annotated with StreamEmitter.
+ *
+ * Depending on how the method is structured, there are some flexibility in how the {@link Output} may be used.
+ *
+ * Here are some supported usage patterns:
+ *
+ * A StreamEmitter method that has a return type cannot take any method parameters.
+ *
+ * &#064;StreamEmitter
+ * &#064;Output(Source.OUTPUT)
+ * public Flux<String> emit() {
+ * 	return Flux.intervalMillis(1000)
+ * 		.map(l -> "Hello World!!");
+ * }
+ *
+ * The following examples show how a void return type can be used on a method with StreamEmitter and how the
+ * method signatures could be used in a flexible manner.
+ *
+ * &#064;StreamEmitter
+ * public void emit(&#064;Output(Source.OUTPUT) FluxSender output) {
+ * 	output.send(Flux.intervalMillis(1000)
+ *		.map(l -> "Hello World!!"));
+ * }
+ *
+ * &#064;StreamEmitter
+ * &#064;Output(Source.OUTPUT)
+ * public void emit(FluxSender output) {
+ * 	output.send(Flux.intervalMillis(1000)
+ *		.map(l -> "Hello World!!"));
+ * }
+ *
+ *
+ * &#064;StreamEmitter
+ * public void emit(&#064;Output("OUTPUT1") FluxSender output1,
+ *					&#064;Output("OUTPUT2") FluxSender output2,
+ *					&#064;Output("OUTPUT3)" FluxSender output3) {
+ *	output1.send(Flux.intervalMillis(1000)
+ *		.map(l -> "Hello World!!"));
+ *	output2.send(Flux.intervalMillis(1000)
+ *		.map(l -> "Hello World!!"));
+ *	output3.send(Flux.intervalMillis(1000)
+ *		.map(l -> "Hello World!!"));
+ *	}
+ *
+ * @author Soby Chacko
+ */
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface StreamEmitter {
+}
--- a/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
+++ b/spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
@@ -20,6 +20,10 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import reactor.core.Disposable;
 
 import org.springframework.aop.support.AopUtils;
 import org.springframework.beans.BeansException;
@@ -40,6 +44,7 @@ import org.springframework.context.SmartLifecycle;
 import org.springframework.core.MethodParameter;
 import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.core.annotation.SynthesizingMethodParameter;
 import org.springframework.util.Assert;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
@@ -48,24 +53,33 @@ import org.springframework.util.StringUtils;
  * {@link BeanPostProcessor} that handles {@link StreamEmitter} annotations found on bean methods.
  *
  * @author Soby Chacko
+ * @since 1.2.2
  */
 public class StreamEmitterAnnotationBeanPostProcessor
 		implements BeanPostProcessor, InitializingBean, ApplicationContextAware, SmartLifecycle {
 
 	private final List<StreamListenerParameterAdapter<?, Object>> streamListenerParameterAdapters = new ArrayList<>();
 
-	private final List<StreamListenerResultAdapter<?, ?>> streamListenerResultAdapters = new ArrayList<>();
+	private final List<StreamListenerResultAdapter<?, ?, ?>> streamListenerResultAdapters = new ArrayList<>();
+
+	private final List<FluxSender> fluxSenders = new ArrayList<>();
+
+	private final List<Disposable> fluxDisposables = new ArrayList<>();
 
 	private ConfigurableApplicationContext applicationContext;
 
-	private Method targetMethod;
+	private final List<Method> targetMethods = new ArrayList<>();
 
 	private Object targetBean;
 
 	private volatile boolean running;
 
+	private Lock lock = new ReentrantLock();
+
 	@Override
 	public final void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+		Assert.isTrue(applicationContext instanceof ConfigurableApplicationContext,
+				"ConfigurableApplicationContext is required");
 		this.applicationContext = (ConfigurableApplicationContext) applicationContext;
 	}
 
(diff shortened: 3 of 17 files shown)


spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * {@link BeanPostProcessor} that handles {@link StreamEmitter} annotations found on bean methods.
+ *
+ * @author Soby Chacko
+ */
+public class StreamEmitterAnnotationBeanPostProcessor
+		implements BeanPostProcessor, InitializingBean, ApplicationContextAware, SmartLifecycle {
+
+	private final List<StreamListenerParameterAdapter<?, Object>> streamListenerParameterAdapters = new ArrayList<>();
```
mbogoevici: Why do we need these?

sobychacko: I believe we need this for void methods that takes the ` @Output` annotation with types like FluxSender.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
```diff
@@ -51,6 +51,11 @@ public FluxToMessageChannelResultAdapter fluxToMessageChannelResultAdapter() {
 		return new FluxToMessageChannelResultAdapter();
 	}
 
+	@Bean
+	public StreamEmitterAnnotationBeanPostProcessor streamEmitterAnnotationBeanPostProcessor() {
```
artembilan: `BeanPostProcessor` s must be `static` bean methods: http://docs.spring.io/spring/docs/current/spring-framework-reference/html/beans.html#beans-java-injecting-imported-beans:

> Also, be particularly careful with BeanPostProcessor and BeanFactoryPostProcessor definitions via @Bean. Those should usually be declared as static @Bean methods, not triggering the instantiation of their containing configuration class. Otherwise, @Autowired and @Value won’t work on the configuration class itself since it is being created as a bean instance too early.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * No {@link Input}s are allowed on a method that is annotated with StreamEmitter.
+ *
+ * Depending on how the method is structured, there are some flexibility in how the {@link Output} may be used.
+ *
+ * Here are some supported usage patterns:
+ *
+ * A StreamEmitter method that has a return type cannot take any method parameters.
+ *
+ * &#064;StreamEmitter
```
artembilan: Hm.
All these must be wrapped to code to be readable from the target JavaDocs resource.
For example:
```
* <pre class="code">
 * &#064;Configuration
 * &#064;EnableIntegration
 * &#064;ComponentScan(basePackageClasses = { MyConfiguration.class })
 * public class MyIntegrationConfiguration {
 * }
 * </pre>
```
which results to the: http://docs.spring.io/spring-integration/api/org/springframework/integration/config/EnableIntegration.html

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
```diff
@@ -0,0 +1,86 @@
...
+ *	output3.send(Flux.intervalMillis(1000)
+ *		.map(l -> "Hello World!!"));
+ *	}
+ *
+ * @author Soby Chacko
+ */
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface StreamEmitter {
```
artembilan: Well, this encourage me to make our (Spring Integration) `@Publisher` supports `Publisher<?>` if target  channel is `ReactiveChannel` 😄

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/ReactiveSupportAutoConfiguration.java
sobychacko: good point - will change that to `static`

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
sobychacko: +1

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+	@Override
+	public final void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+		this.applicationContext = (ConfigurableApplicationContext) applicationContext;
+	}
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public void afterPropertiesSet() throws Exception {
+		Map<String, StreamListenerParameterAdapter> parameterAdapterMap = BeanFactoryUtils
+				.beansOfTypeIncludingAncestors(this.applicationContext, StreamListenerParameterAdapter.class);
```
artembilan: This is absolutely bad to instantiate beans from the `BPP` initialization.
The same problem as don't make `BPP` bean method `static`.

I name it "Early BeanFactory access paradox". May work in your tests, but will be surprise for end-user in your complex configuration.

I see you use from `star()`, so that is fully fine to consider to implement `SmartInitializingSingleton` for this purpose to get support beans.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+			}
+			this.running = true;
+		}
+	}
+
+	@SuppressWarnings({"rawtypes", "unchecked"})
+	private void invokeSetupMethodOnToTargetChannel(Method method, Object bean, String outboundName) {
+		Object[] arguments = new Object[method.getParameterTypes().length];
+		for (int parameterIndex = 0; parameterIndex < arguments.length; parameterIndex++) {
+			MethodParameter methodParameter = MethodParameter.forMethodOrConstructor(method, parameterIndex);
```
artembilan:
```
* @deprecated as of 5.0, in favor of {@link #forExecutable}
 */
@Deprecated
public static MethodParameter forMethodOrConstructor(Object methodOrConstructor, int parameterIndex) {
```
Although I don't believe that you worry about ctors here:
```
new SynthesizingMethodParameter(method, parameterIndex)
```
is there for you.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+			Object targetBean = this.applicationContext.getBean(outboundName);
+			boolean streamListenerResultAdapterFound = false;
+			for (StreamListenerResultAdapter streamListenerResultAdapter : this.streamListenerResultAdapters) {
+				if (streamListenerResultAdapter.supports(result.getClass(), targetBean.getClass())) {
+					streamListenerResultAdapter.adapt(result, targetBean);
+					streamListenerResultAdapterFound = true;
+					break;
+				}
+			}
+			Assert.isTrue(streamListenerResultAdapterFound, StreamEmitterErrorMessages.CANNOT_CONVERT_RETURN_TYPE_TO_ANY_AVAILABLE_RESULT_ADAPTERS);
```
artembilan: `isTrue()` throws `IllegalArgumentException`, unlike `state()` with its `IllegalStateException`.
I don't know what is better for you here - just share my concern that the target exception may be reviewed.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+		stop();
+		if (callback != null) {
+			callback.run();
+		}
+	}
+
+	@Override
+	public void stop() {
+		if (running) {
+			this.running = false;
```
artembilan: Well, this fully looks like `SmartInitializingSingleton` is for you. You don't really support `lifecycle` in this component.
All you need is fully initialized container but before `refresh()`.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
mbogoevici: Gotcha, that's what I was missing :)

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
mbogoevici: Wouldn't moving the code to `StreamListenerParameterAdapter.afterSingletonsInstantiated` mean that these variables are set after the singletons (which may have `@StreamListener` and `@StreamEmitter` annotations) are postprocessed?

We don't expect `StreamListenerParameterAdapter` and `StreamListenerResultAdapter` to undergo postprocessing, we just want to have them registered with the context via autoconfiguration - and the implementations are also generally framework-provided not user-provided.

Asking this as it might impact the current implementation for `@StreamListener`.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
sobychacko: I think `state` might be more suitable here as we are checking on the return type.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
sobychacko: I am afraid we still need to implement `SmartLifeCycle` to control the ordering of this bean instantiation. We need to make sure that this bean only gets instantiated after the `OutputLifecycleBinding`. Do you think `SmartInitializingSingleton` can solve this issue?

mbogoevici: It is actually subscribing to the flux produced by the method during `start()`. I suppose we should unsubscribe during `stop()`?

sobychacko: makes sense. We are programmatically subscribing though so might have to resort to something like that to unsubscribe as well.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+			Class<?> parameterType = methodParameter.getParameterType();
+			Object targetReferenceValue = null;
+			if (methodParameter.hasParameterAnnotation(Output.class)) {
+				targetReferenceValue = AnnotationUtils.getValue(methodParameter.getParameterAnnotation(Output.class));
+			}
+			else if (arguments.length == 1 && StringUtils.hasText(outboundName)) {
+				targetReferenceValue = outboundName;
+			}
+			if (targetReferenceValue != null) {
+				Assert.isInstanceOf(String.class, targetReferenceValue, "Annotation value must be a String");
```
artembilan: Looks like this is redundant:
1. The `outboundName` is `String` by definition
2. The `value` attribute of the `@Output` is `String`, too

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+			throw new BeanInitializationException("Cannot setup StreamEmitter for " + method, e);
+		}
+
+		if (!Void.TYPE.equals(method.getReturnType())) {
+			if (!StringUtils.hasText(outboundName)) {
+				for (int parameterIndex = 0; parameterIndex < method.getParameterTypes().length; parameterIndex++) {
+					MethodParameter methodParameter = MethodParameter.forMethodOrConstructor(method,
+							parameterIndex);
+					if (methodParameter.hasParameterAnnotation(Output.class)) {
+						outboundName = methodParameter.getParameterAnnotation(Output.class).value();
```
artembilan: Nit-pick, but a bit above in this method you use:
```
 AnnotationUtils.getValue(methodParameter.getParameterAnnotation(Output.class))
```
This one more addition to that my "obvious string" concern.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+					StreamEmitterErrorMessages.RETURN_TYPE_NO_OUTBOUND_SPECIFIED);
+			Assert.isTrue(method.getParameterCount() == 0, StreamEmitterErrorMessages.RETURN_TYPE_METHOD_ARGUMENTS);
+		}
+		else {
+			if (!StringUtils.hasText(methodAnnotatedOutboundName)) {
+				int methodArgumentsLength = method.getParameterTypes().length;
+				for (int parameterIndex = 0; parameterIndex < methodArgumentsLength; parameterIndex++) {
+					MethodParameter methodParameter = MethodParameter.forMethodOrConstructor(method, parameterIndex);
+					if (methodParameter.hasParameterAnnotation(Output.class)) {
+						String outboundName = (String) AnnotationUtils
```
artembilan: And already here you do explicit cast...

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamEmitterValidationTests.java
```diff
@@ -0,0 +1,277 @@
...
+import static org.assertj.core.api.Assertions.fail;
+import static org.springframework.cloud.stream.binding.StreamAnnotationErrorMessages.ATLEAST_ONE_OUTPUT;
+import static org.springframework.cloud.stream.binding.StreamAnnotationErrorMessages.INVALID_OUTBOUND_NAME;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.CANNOT_CONVERT_RETURN_TYPE_TO_ANY_AVAILABLE_RESULT_ADAPTERS;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.INPUT_ANNOTATIONS_ARE_NOT_ALLOWED;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.INVALID_OUTPUT_METHOD_PARAMETERS;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.NO_OUTPUT_SPECIFIED;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.OUTPUT_ANNOTATION_MISSING_ON_METHOD_PARAMETERS_VOID_RETURN_TYPE;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.RETURN_TYPE_METHOD_ARGUMENTS;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.RETURN_TYPE_NO_OUTBOUND_SPECIFIED;
```
artembilan: Usually we don't allow `static` imports: Only this set:
```
org.custommonkey.xmlunit.XMLAssert.*,
org.assertj.core.api.Assertions.*,
org.hamcrest.CoreMatchers.*,
org.hamcrest.Matchers.*,
org.hamcrest.collection.IsCollectionWithSize.*,
org.junit.Assert.*,
org.junit.Assume.*,
org.mockito.Mockito.*,
org.mockito.BDDMockito.*,
org.mockito.AdditionalAnswers.*,
org.mockito.ArgumentMatchers.*,
org.mockito.hamcrest.MockitoHamcrest.*,
```

And some other test support utilities.
These constants don't look like related to tests...

What are your Checkstyle rules ?

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+public class StreamEmitterAnnotationBeanPostProcessor
+		implements BeanPostProcessor, InitializingBean, ApplicationContextAware, SmartLifecycle {
+
+	private final List<StreamListenerParameterAdapter<?, Object>> streamListenerParameterAdapters = new ArrayList<>();
+
+	private final List<StreamListenerResultAdapter<?, ?>> streamListenerResultAdapters = new ArrayList<>();
+
+	private ConfigurableApplicationContext applicationContext;
+
+	private Method targetMethod;
```
sobychacko: It doesn't make sense to have more than one `StreamEmitter` in an end user application.

mbogoevici: Actually we can have multiple @StreamEmitters - either for merging data on a given outbound channel or for sending data on multiple channels.

artembilan:
Not necessarily related, but pay attention that we can do this with Spring Integration Java DSL:
````java
@Bean
Publisher<Message<String>> serverTimeFlow() {
	return IntegrationFlows
			.from(() -> new GenericMessage<>(new Date()),
					e -> e.poller(p -> p.fixedDelay(1000)))
			.transform(Object::toString)
			.toReactivePublisher();
}
````

So, maybe that would make sense to have this `@StreamEmitter` on these kind of beans?...

File: spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java [hunk 14]
```diff
@@ -0,0 +1,244 @@
...
+
+	private Object targetBean;
+
+	private volatile boolean running;
+
+	@Override
+	public final void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+		this.applicationContext = (ConfigurableApplicationContext) applicationContext;
```
olegz: Not to be too picky, but this is unsafe cast. Consider adding assertion to ensure that AC is an instance of `ConfigurableApplicationContext`. It may be safe within our whole bootstrap but it is public class after all.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+	@Override
+	public void stop() {
+		if (running) {
+			this.running = false;
+		}
+	}
+
+	@Override
+	public boolean isRunning() {
+		return false;
```
olegz: Not sure I understand why are we not returning `this.running`

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,244 @@
...
+			if (targetMethod != null) {
+				Assert.isTrue(targetMethod.getAnnotation(Input.class) == null,
+						StreamEmitterErrorMessages.INPUT_ANNOTATIONS_ARE_NOT_ALLOWED);
+				String methodAnnotatedOutboundName = StreamAnnotationCommonMethodUtils.getOutboundBindingTargetName(targetMethod);
+				int outputAnnotationCount = StreamAnnotationCommonMethodUtils.outputAnnotationCount(targetMethod);
+				validateStreamEmitterMethod(targetMethod, outputAnnotationCount, methodAnnotatedOutboundName);
+				invokeSetupMethodOnToTargetChannel(targetMethod, targetBean, methodAnnotatedOutboundName);
+			}
+			this.running = true;
+		}
```
olegz: There is a bit of a thread issue here. Again, it may be safe within its usage in AC, but I'd still go with `AtomicBoolean.compareAndSet`.
Also, not sure about the context, but is it a valid condition (for this bean to be marked as running) if IF did not materialize? Basically my concern is the lack of 'else'. It may be ok, just wanted to point it out just in case

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamEmitterValidationTests.java
sobychacko: its the standard scst checkstyle rules.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
sobychacko: that was a mistake. will fix.

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamEmitterValidationTests.java
artembilan: Well, your checkstyle is like this:
```
<module name="AvoidStaticImport">
            <property name="excludes"
                      value="org.junit.Assert.*,org.mockito.Mockito.*,org.mockito.Matchers.*,org.hamcrest.Matchers.*,org.assertj.core.api.Assertions.*"/>
        </module>
```
and this in the suppressions:
```
<suppress files="[\\/]src[\\/]test[\\/]java[\\/]" checks="AvoidStaticImport" />
```
I see now. I disagree with this, but of course I can't impact to your project.

OTOH if that was so standard I'd not pay attention to this...

artembilan: And yes, pay attention how `AvoidStaticImport` rule is configured.
I'm fully not sure that you will use those test utilities in the production code.
So, having afterwards such a suppress rule makes the original one useless.

Would be better to revise that some day.

sobychacko:
Closing this PR in favor of https://github.com/spring-cloud/spring-cloud-stream/pull/974

PR #889, opened 2017-03-30
Internal refactoring - expose header embedding and serialization

### LINKED ISSUE #888: Internal refactoring - expose serialization/header marshalling logic
Currently some logic such as serialization and header embedding is contained in private classes or hard to use. We should make it easy for third-party implementors to reuse that logic.

mbogoevici:
Fix #888

- Rename `EmbeddedHeaderMessageConverter` to `EmbeddedHeaderUtils`;
- Add method to populate default header set;
- Make `serializePayloadIfNecessary` protected so it can be accessed by subclasses;

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeaderUtils.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeaderUtils.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.stream.binder;
 
 import java.io.UnsupportedEncodingException;
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/EmbeddedHeaderUtils.java
```diff
@@ -174,4 +178,19 @@ private MessageValues buildMessageValues(byte[] payload, Map<String, Object> hea
 		return messageValues;
 	}
 
+	public static String[] headersToEmbed(String[] configuredHeaders) {
```
ilayaperumalg: We can add some tests to verify this method.

ilayaperumalg:
LGTM, merging. will add tests for the newly added method.

ilayaperumalg:
Rebased and merged as 27184a4cfcda0bc5b9d7fd2b06378e73c7e027b6 along with the test test for EmbeddedHeaderUtils.headersToEmbed c812fa0374b5f703f6c0e1b1851d7b2b52b51eba

PR #891, opened 2017-03-30
Allow `StreamListenerParameterAdapter` to be applied more broadly

### LINKED ISSUE #890: Allow the application of `StreamListenerParameterAdapter` even for matching targets
Currently, the `StreamListenerParameterAdapter` is not applied if the type of the target binding matches the parameter. In certain cases it might be useful to allow to happen, for example for inserting type conversion logic between the declared target type and the parameter type. 

One immediate use case for that is `KStream` support.

mbogoevici:
Fix #890

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
@@ -271,16 +271,16 @@ public class StreamListenerAnnotationBeanPostProcessor
 			if (targetReferenceValue != null) {
 				Assert.isInstanceOf(String.class, targetReferenceValue, "Annotation value must be a String");
 				Object targetBean = this.applicationContext.getBean((String) targetReferenceValue);
-				if (parameterType.isAssignableFrom(targetBean.getClass())) {
-					arguments[parameterIndex] = targetBean;
+				// Iterate existing parameter adapters first
+				for (StreamListenerParameterAdapter<?, Object> streamListenerParameterAdapter : this.streamListenerParameterAdapters) {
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerAnnotationBeanPostProcessor.java
```diff
@@ -271,16 +271,16 @@ else if (arguments.length == 1 && StringUtils.hasText(inboundName)) {
...
+					}
 				}
-				else {
-					for (StreamListenerParameterAdapter<?, Object> streamListenerParameterAdapter : this.streamListenerParameterAdapters) {
-						if (streamListenerParameterAdapter.supports(targetBean.getClass(), methodParameter)) {
-							arguments[parameterIndex] = streamListenerParameterAdapter.adapt(targetBean,
-									methodParameter);
-							break;
-						}
+				if (arguments[parameterIndex] == null) {
```
ilayaperumalg: Can we combine both the `if` statements into one:
```
if (arguments[parameterIndex] == null && parameterType.isAssignableFrom(targetBean.getClass()))
```

mbogoevici: Yes, good catch.

ilayaperumalg:
Not from the scope of this PR, but it would be nice to include a test configuration that uses something similar to how `KStream` support is implemented. We can cover all the tests for that type of bindings as well.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as ab9cf91803ba83f6a101f6a93581e040a69459ff along with 6592c1a8cdbe0064b96f1907006db9fbeb415305

PR #892, opened 2017-03-30
Use Spring Boot metric exporter machinery

dsyer:
The naming conventions and interfaces from Spring Boot are
used to align more closely with features already available
there. Same feature, slightly different names for things.


viniciusccarvalho:
Squashed, checkstyle'd and rebased as 07d0328

PR #896, opened 2017-03-31
Metrics refactoring

mbogoevici:
- Change the metrics channel name to `applicationMetricsChannel`
- Update documentation
- Remove `instanceIndex` to reduce coupling to
  Spring Cloud Stream semantics and because it can be sent via
  the `properties` metadata set


spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/ApplicationMetrics.java
```diff
@@ -35,19 +35,15 @@
 	private final Date createdTime;
 
 	private String name;
-
-	private int instanceIndex;
```
ilayaperumalg: We already use similar `instanceIndex` semantics in `BindingServiceProperties`. But, I don't seem to find any relevant usages of that property as well.

PR #899, opened 2017-04-03
Convert `byte[]` via `text/plain` to String

### LINKED ISSUE #898: When outbound type is `text/plain`, convert arrays to String
The issue is that when `byte[]` payloads are set to a `contentType` of `text/plain`, the output is the result of the `toString()` operation.

mbogoevici:
Fix #898

When the payload is `byte[]` ensure that the
String is the String representation, not the
toString() conversion result.


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 9ea4ff3

Issue #902, opened 2017-04-07
Fix metrics doc section


markpollack:
The docs indicate you can set delay, includes and excludes under the key `spring.cloud.stream.metrics`, but those properties are really under the control of boot with the prefix `spring.metrics.export`.  The docs should discuss the include/exclude props a bit since I think they aren't that well described in the boot docs.  There should be a link to the boot docs also http://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metric-writers

markpollack:
Also the example payload isn't accurate, should remove `instanceIndex`.   Prob best to run a sample and do a cut-n-paste job.

mbogoevici:
List `spring.metric.exporter.triggers.application.*` as an finer-grained option.  (Also wrt fixing #903 ).

PR #905, opened 2017-04-07
Align the exporter and the trigger name

### LINKED ISSUE #903: Align the exporter name and trigger configuration for metrics
Currently, `ApplicationMetricsProperties` assumes a trigger configuration named `application`, but the exporter bean is registered under `aggregateMetricsProperties`. This means that properties such as includes/excludes are configured under `spring.metrics.export.triggers.application.includes` but `delay-millis` is under `spring.metrics.export.triggers.aggregateMetricsExporter.delay-millis`.

mbogoevici:
Fix #903

Use 'application' as the trigger name consistently.

Since 'application' would be a too generic name for
the Exporter bean, use a BeanPostProcessor to install
the exporter in the `MetricExporters` bean.

(Capture  the option as part of #902)

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1914,44 +1914,55 @@ Spring Cloud Stream provides a health indicator for binders.
 It is registered under the name of `binders` and can be enabled or disabled by setting the `management.health.binders.enabled` property.
 
 == Metrics Emitter
+
 Spring Cloud Stream provides a module called `spring-cloud-stream-metrics` that can be used to emit any available metric from https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html[Spring Boot metrics endpoint] to a named channel.
 This module allow operators to collect metrics from stream applications without relying on polling their endpoints.
 
 HTTP polling can be challenging on cloud environments since applications could be on a private network or behind a Load balancer that prevents per instance access.
 
-The module is activated when you set the destination name for its channel, `spring.cloud.stream.bindings.applicationMetricsChannel.destination=<DESTINATION_NAME>`.
-`applicationMetricsChannel` can be configured in a similar fashion to any other bound output channel.
-The default `contentType` setting of `applicationMetricsChannel` is `application/json`.
+The module is activated when you set the destination name for metrics binding, e.g. `spring.cloud.stream.bindings.applicationMetrics.destination=<DESTINATION_NAME>`.
+`applicationMetrics` can be configured in a similar fashion to any other producer binding.
+The default `contentType` setting of `applicationMetrics` is `application/json`.
 
-By default the module is configured to only send Spring Integration message channel metrics.
+The following properties can be used for customizing the emission of metrics:
 
-Available properties for customization using the prefix `spring.cloud.stream.metrics`.
-
-key::
+spring.cloud.stream.metrics.key::
   The name of the metric being emitted. Should be an unique value per application.
 +
 Default:: `${spring.application.name:${vcap.application.name:${spring.config.name:application}}}`
 +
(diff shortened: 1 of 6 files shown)


dsyer:
LGTM. Maybe Spring Boot could accommodate "application" as a trigger key natively in the future (avoiding the BPP).

viniciusccarvalho:
I believe this is a wrong approach for the problem. We have already agreed that each channel (boot and stream) would be independent of each other. They do after all, have different payloads.

Now, by reusing the `includes` property from boot, it means that they do have different payloads, but only one way to filter. So let's imagine the scenario where a user is using boot metrics, and it's used to receive one message per metric emission, and he subscribes to everything in the `http` subsystem. Along comes stream and start pushing `integration` metrics on it, and now, if the client of the metrics from stream does not care about http, it's receiving that data as well.

The best analogy I can make here is of a Table on a database, where you have two clients with different connections, but all the suddenly you only have one projection to rule both connections.

mbogoevici:
> Now, by reusing the includes property from boot, it means that they do have different payloads, but only one way to filter.

This is doing exactly the opposite: `spring.export.metrics.trigger.application` is the namespace for properties that apply exclusively to `ApplicationMetricsExporter`. The one thing that the PR does is to make sure that the exporter is registered under the name `application` (and not a name derived from its), which means that properties such as `delay-millis` also apply to it.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1914,48 +1914,58 @@ Spring Cloud Stream provides a health indicator for binders.
...
 Default:: `${spring.application.name:${vcap.application.name:${spring.config.name:application}}}`
 +
-delay-millis::
-  The period in which metrics will be posted to the channel
-+
-Default: `5000`.
-+
-prefix::
+spring.cloud.stream.metrics.prefix::
   Prefix string to be prepended to the metrics name
```
markpollack: I believe this should be metrics key, not name.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1914,48 +1914,58 @@ Spring Cloud Stream provides a health indicator for binders.
...
+spring.cloud.stream.metrics.properties::
+  Just like the `includes` option, it allows white listing application properties that will be added to the metrics payload
++
+Default: null.
++
+
+Spring Cloud Stream provides a metric exporter named `application` that can be configured via regular Spring Boot metrics configuration properties.
+A detailed overview can be found in the https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html#production-ready-metric-writers[Spring Boot reference documentation].
+
+Summarizing, the exporter can be configured either by using the global Spring Boot configuration settings for exporters, or by using exporter-specific properties.
```
markpollack: Don't need to say "Summarizing"

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1914,48 +1914,58 @@ Spring Cloud Stream provides a health indicator for binders.
...
+A subset of the available properties is listed below.
```
markpollack: I'd remove this sentence and just have the other two usage example say,  "for example, you can configure blah blah..."

markpollack: link to the TriggerProperties class in github for "available properties"

viniciusccarvalho:
My bad, I thought we were only allowing global setting for all exporters. Having our own configuration is what I was really worried about. Please forgive and forget what I said :D

viniciusccarvalho:
Squashed and Merged as d6fe841

PR #923, opened 2017-04-13
Fixes unsafe access to DateFormat

### LINKED ISSUE #922: Fix unsafe access to DateFormat object
`MetricJsonSerializer` allows unsafe access to it's internal DateFormat object. Although the metrics emitter is executed on a single thread, this could cause issues if the bean is shared with different threads

viniciusccarvalho:
Fixes #922

### DIFF
--- a/spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializer.java
+++ b/spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializer.java
@@ -40,7 +40,12 @@ import org.springframework.boot.jackson.JsonComponent;
 @JsonComponent
 public class MetricJsonSerializer {
 
-	final static DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+	final static ThreadLocal<DateFormat> df = new ThreadLocal<DateFormat>(){
+			@Override
+			protected DateFormat initialValue() {
+					return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+			}
+	};
 
 	public static class Serializer extends JsonSerializer<Metric<?>> {
 
@@ -50,7 +55,7 @@ public class MetricJsonSerializer {
 			json.writeStartObject();
 			json.writeStringField("name", metric.getName());
 			json.writeNumberField("value", metric.getValue().doubleValue());
-			json.writeStringField("timestamp", df.format(metric.getTimestamp()));
+			json.writeStringField("timestamp", df.get().format(metric.getTimestamp()));
 			json.writeEndObject();
 		}
 	}
@@ -66,7 +71,7 @@ public class MetricJsonSerializer {
 			Date timestamp = null;
 
 			try {
-				timestamp = df.parse(node.get("timestamp").asText());
+				timestamp = df.get().parse(node.get("timestamp").asText());
 			}
 			catch (ParseException e) {
 			}


spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializer.java
```diff
@@ -40,7 +40,12 @@
 @JsonComponent
 public class MetricJsonSerializer {
 
-	final static DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+	final static ThreadLocal<DateFormat> df = new ThreadLocal<DateFormat>(){
+			@Override
+			protected DateFormat initialValue() {
+					return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+			}
+	};
```
garyrussell: I am not a fan of leaving dangling `ThreadLocals` attached to arbitrary threads. In [Spring Integration we usually use a dynamic pool of formatters](https://github.com/spring-projects/spring-integration/blob/4.3.x/spring-integration-core/src/main/java/org/springframework/integration/transformer/SyslogToMapTransformer.java#L93).

That said, it depends on how long this will live; if we're going to Java8 soon, we can let this stand and move to `DateTimeFormatter` which is thread-safe. @mbogoevici ??

viniciusccarvalho: I was not aware of the formatters pool, I can update the code to use it, seems a much cleaner option.

garyrussell: It's just a simple queue that we check formatters out/in

    private static final BlockingQueue<SimpleDateFormat> dateFormats = new LinkedBlockingQueue<SimpleDateFormat>();

mbogoevici: @garyrussell @viniciusccarvalho I would prefer the formatters pool as well if we can have it (as opposed to the ThreadLocal). I don't believe we will go full Java 8 before Spring Cloud Stream 2.

viniciusccarvalho: I'll make the changes, we may need to change the formatter itself, on the collector we changed it to round values to 2 decimal places. After this milestone of the collector, I want to make sure it uses the code that lives on stream-metrics, instead of duplicating it, too much duplicate effort. It was good to allow us to walk at a different pace, but after M1 I want to revisit this design.

mbogoevici: We may want to set the format as a configurable property for instance. But we can't break backwards compatibility.

viniciusccarvalho: I rather not let it to be configurable as it's a bit of hassle downstream. We have set it as the default format for timestamp, and allowing users to change it could break downstream apps such as the collector. Timestamp representation in json can either be millis since epoch or ISO 8601 which is the format currently being used. Consumers of the payload can always apply required transformations if say they want to store in a numerical format in a database for instance.

mbogoevici: My comment was specifically related to `we may need to change the formatter itself, on the collector we changed it to round values to 2 decimal places`. While I admit I don't know exactly what was done in the collector, what I intended to say was that that if we consider changing the format, it should be done  only as a configuration option and preserving the current behaviour as default. Otherwise I'm quite happy not to configure anything.

viniciusccarvalho: Well changing the precision does not really change the format right? And this was a discussion we had internally were it seems that this is not changing precision for accuracy, 2 decimal places should be enough for this type of metrics. Downstream consumers would still read a double (or a float) it's just truncated.

mbogoevici: Ok, I assume that you are talking about the precision of `value`, not the precision of the `timestamp`. That should be fine.

viniciusccarvalho:
Closing this PR. Will follow with a fix merging timezone fiz and potentially supporting number format configuration.

PR #926, opened 2017-04-17
GH-913: Support for ErrorMessageSendingRecoverer

### LINKED ISSUE #913: Spike on ErrorMessagePublishingRecoveryCallback + error routing
Use something like ErrorMessagePublishingRecoveryCallback (currently in SI PR https://github.com/spring-projects/spring-integration/pull/2091) in the binder and configure routing to the error channel:
- by exception type - decide what errors go to the destination.

garyrussell:
Needs https://github.com/spring-projects/spring-integration/pull/2091 + polishing commits mentioned

Resolves #913

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -17,8 +17,6 @@
 package org.springframework.cloud.stream.binder;
 
 
-import java.util.concurrent.atomic.AtomicInteger;
-
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
@@ -31,9 +29,7 @@ import org.springframework.context.Lifecycle;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.integration.channel.FixedSubscriberChannel;
-import org.springframework.integration.channel.PublishSubscribeChannel;
 import org.springframework.integration.context.IntegrationContextUtils;
-import org.springframework.integration.context.Orderable;
 import org.springframework.integration.core.MessageProducer;
 import org.springframework.integration.endpoint.EventDrivenConsumer;
 import org.springframework.integration.handler.AbstractMessageHandler;
@@ -45,7 +41,6 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.messaging.MessageHeaders;
-import org.springframework.messaging.MessagingException;
 import org.springframework.messaging.SubscribableChannel;
 import org.springframework.util.Assert;
 import org.springframework.util.MimeType;
@@ -300,23 +295,29 @@ public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties,
 	/**
 	 * Build an errorChannelRecoverer that writes to a pub/sub channel for the destination.
 	 * @param destination the destination.
+	 * @param group the group.
 	 * @param consumerProperties the properties.
-	 * @param errorMessageStrategy
-	 * @return the recoverer.
+	 * @return the ErrorInfrastructure which is a holder for the error channel, the recoverer and the
+	 * message handler that is subscribed to the channel.
 	 */
 	protected final ErrorInfrastructure registerErrorInfrastructure(ConsumerDestination destination, String group,
 			ConsumerProperties consumerProperties) {
 		ErrorMessageStrategy errorMessageStrategy = getErrorMessageStrategy();
 		ConfigurableListableBeanFactory beanFactory = getApplicationContext().getBeanFactory();
 		String errorChannelName = errorsBaseName(destination, group, consumerProperties);
-		SubscribableChannel errorChannel;
+		LastSubscriberAwareChannel errorChannel;
 		if (getApplicationContext().containsBean(errorChannelName)) {
-			errorChannel = getApplicationContext().getBean(errorChannelName, SubscribableChannel.class);
+			Object errorChannelObject = getApplicationContext().getBean(errorChannelName);
+			if (!(errorChannelObject instanceof LastSubscriberAwareChannel)) {
+				throw new IllegalStateException(
+						"Error channel '" + errorChannelName + "' must be a LastSubscriberAwareChannel");
+			}
+			errorChannel = (LastSubscriberAwareChannel) errorChannelObject;
 		}
 		else {
-			errorChannel = new ErrorChannel();
+			errorChannel = new BinderErrorChannel();
 			beanFactory.registerSingleton(errorChannelName, errorChannel);
-			errorChannel = (SubscribableChannel) beanFactory.initializeBean(errorChannel, errorChannelName);
+			errorChannel = (LastSubscriberAwareChannel) beanFactory.initializeBean(errorChannel, errorChannelName);
 		}
 		ErrorMessageSendingRecoverer recoverer;
 		if (errorMessageStrategy == null) {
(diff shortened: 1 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -258,6 +277,55 @@ protected abstract MessageProducer createConsumerEndpoint(ConsumerDestination de
...
+		return errorsBaseName(destination, group, extendedConsumerProperties) + ".errors";
+	}
+
+	protected String getErrorRecovererName(ConsumerDestination destination, String group,
+			ExtendedConsumerProperties<?> extendedConsumerProperties) {
+		return errorsBaseName(destination, group, extendedConsumerProperties) + ".recoverer";
+	}
+
+	protected String errorsBaseName(ConsumerDestination destination, String group,
+			ExtendedConsumerProperties<?> extendedConsumerProperties) {
```
mbogoevici: Does the argument need to be `ExtendedConsumerProperties` or can it be `ConsumerProperties` ? Seems we are just using instanceIndex which is coming from the base class.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -258,6 +277,55 @@ protected abstract MessageProducer createConsumerEndpoint(ConsumerDestination de
...
+			recoverer = new ErrorMessageSendingRecoverer(errorChannel, errorMessageStrategy);
+		}
+		String recovererBeanName = getErrorRecovererName(destination, group, extendedConsumerProperties);
+		beanFactory.registerSingleton(recovererBeanName, recoverer);
+		beanFactory.initializeBean(recoverer, recovererBeanName);
+		//TODO unregister on destroy
+		return recoverer;
+	}
+
+	protected String getErrorChannelName(ConsumerDestination destination, String group,
```
mbogoevici: I suspect that the main use case would be to forward all error messages to `errorChannel` (I believe this is what we originally discussed) - maybe we should make that the default and allow users to configure it?

garyrussell:
@mbogoevici Reworked for some of the discussions on our call last week.

- refactored infrastructure to be mostly in the abstract class
- bridge each error channel to the global error channel
- added a default handler that throws the exception to the container if there are no user (or recovering) handler subsribed - it's configured to be the last handler so its always run after the user handlers
- properly unsubscribe/destroy the error infrastructure beans on unbind
- unfortunately, I couldn't avoid the binder impl having to call `registerErrorInfrastructure` because we need the final group name (after anonymous determination)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -258,6 +295,125 @@ protected abstract MessageProducer createConsumerEndpoint(ConsumerDestination de
...
+
+	/**
+	 * Return the default error message handler, which throws the error message payload to
+	 * the caller if there are no user handlers subscribed. The handler is ordered so it
+	 * runs after any user-defined handlers that are subscribed.
+	 * @param errorChannel the error channel.
+	 * @return the handler.
+	 */
+	protected MessageHandler getDefaultErrorMessageHandler(final SubscribableChannel errorChannel) {
+		class FinalRethrowingErrorMessageHandler implements MessageHandler, Orderable {
```
artembilan: Maybe would be better to have this as a separate `static` class with the `errorChannel` as ctor arg?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -258,6 +295,125 @@ protected abstract MessageProducer createConsumerEndpoint(ConsumerDestination de
...
+	 * @return the recoverer.
+	 */
+	protected final ErrorInfrastructure registerErrorInfrastructure(ConsumerDestination destination, String group,
+			ExtendedConsumerProperties<?> extendedConsumerProperties) {
+		ErrorMessageStrategy errorMessageStrategy = getErrorMessageStrategy();
+		ConfigurableListableBeanFactory beanFactory = getApplicationContext().getBeanFactory();
+		String errorChannelName = errorsBaseName(destination, group, extendedConsumerProperties);
+		SubscribableChannel errorChannel;
+		if (getApplicationContext().containsBean(errorChannelName)) {
+			errorChannel = getApplicationContext().getBean(errorChannelName, SubscribableChannel.class);
```
artembilan: Well, there is no guarantee that end-user always declare error channel as `PublishSubscribeChannel`.
Or is it a part of binding service as well here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -366,4 +522,62 @@ public boolean isRunning() {
 			return this.delegate instanceof Lifecycle && ((Lifecycle) this.delegate).isRunning();
 		}
 	}
+
+	private interface SubscriberAwareChannel {
```
artembilan: Looks like nobody implements this interface.
I guess the idea was to make that `ErrorChannel` below to follow this contract.

Although looking to that `FinalRethrowingErrorMessageHandler` it really doesn't matter if use has subscribers or not.
You claim there like `run last if the user has subscribers too`.

Nevertheless maybe that would be much better to **remove** that `FinalRethrowingErrorMessageHandler` from the `ErrorChannel` just in the moment use calls its `subscribe()`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -258,6 +295,125 @@ protected abstract MessageProducer createConsumerEndpoint(ConsumerDestination de
...
+		beanFactory.initializeBean(recoverer, recovererBeanName);
+		MessageHandler handler = getErrorMessageHandler(destination, group, extendedConsumerProperties);
+		if (handler == null) {
+			handler = getDefaultErrorMessageHandler(errorChannel);
+		}
+		String errorMessageHandlerName = getErrorMessageHandlerName(destination, group, extendedConsumerProperties);
+		beanFactory.registerSingleton(errorMessageHandlerName, handler);
+		beanFactory.initializeBean(handler, errorMessageHandlerName);
+		errorChannel.subscribe(handler);
+		return new ErrorInfrastructure(errorChannel, recoverer, handler);
```
artembilan: I'm not sure yet why I would need such a complex object in return, but  believe some JavaDocs on the matter will be helpful for any target `Binder` implementators.
Thanks

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -216,6 +226,33 @@ protected void afterUnbindProducer(ProducerDestination destination, P producerPr
 				@Override
 				protected void afterUnbind() {
 					AbstractMessageChannelBinder.this.afterUnbindConsumer(destination, this.group, properties);
+					// TODO more error checking here
+					String recoverer = getErrorRecovererName(destination, group,
+							(ExtendedConsumerProperties<?>) properties);
+					if (getApplicationContext().containsBean(recoverer)) {
+						((DefaultSingletonBeanRegistry) getApplicationContext().getBeanFactory())
+							.destroySingleton(recoverer);
```
mbogoevici: I'm not sure if that will work - given that the binding is not a bean (which has been a discussion point for quite a while).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -258,6 +295,125 @@ protected abstract MessageProducer createConsumerEndpoint(ConsumerDestination de
...
+		return new FinalRethrowingErrorMessageHandler();
+	}
+
+	/**
+	 * Binders can return an {@link ErrorMessageStrategy} for building error messages; binder
+	 * implementations typically might add extra headers to the error message.
+	 * @return the implementation - may be null.
+	 */
+	protected ErrorMessageStrategy getErrorMessageStrategy() {
+		return null;
```
garyrussell: That happens at the SI level - see this code in the binder...

		if (errorMessageStrategy == null) {
			recoverer = new ErrorMessageSendingRecoverer(errorChannel);
		}
		else {
			recoverer = new ErrorMessageSendingRecoverer(errorChannel, errorMessageStrategy);
		}

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -366,4 +522,62 @@ public boolean isRunning() {
...
+
+		private int subscribers() {
+			return this.subscribers.get();
+		}
+
+	}
+
+	protected static class ErrorInfrastructure {
+
+		private final SubscribableChannel errroChannel;
```
garyrussell: Because we have to subscribe a handler to it.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -283,6 +292,163 @@ protected abstract MessageProducer createConsumerEndpoint(ConsumerDestination de
...
+		return new FinalRethrowingErrorMessageHandler(errorChannel, defaultErrorChannelPresent);
+	}
+
+	/**
+	 * Binders can return an {@link ErrorMessageStrategy} for building error messages; binder
+	 * implementations typically might add extra headers to the error message.
+	 * @return the implementation - may be null.
+	 */
+	protected ErrorMessageStrategy getErrorMessageStrategy() {
+		return null;
```
viniciusccarvalho: Not the right place to start this, but I just wanted to raise a concern I have on the way we handle names. It feels a bit odd that we have so many places to handle names for consumer groups, partitions, errors. And sometimes binders may not even use them. I wonder if another approach, less invasive to the base class, such as a strategy pattern could be applied, so any time we add a new feature that require a custom name, we don't need to change the base classes of the project. Happy to discuss this on our next call

viniciusccarvalho:
Squashed and merged via 3c2c7da5085559165cd4c1bc4aef44434a39f990

Issue #927, opened 2017-04-19
Fix Timezone error on MetricsJsonSerializer


viniciusccarvalho:
The Json serializer is set to write dates using GMT as base, but the DateFormatter is not set for that, causing wrong values to be written as string representations.

olegz:
Hmm. . . Actually based on the preliminary testing it writes date using local time zone which could obviously have implication if such timezone is not conveyed  properly to a subscriber in the different timezone.
There is also seem to be a bug in the actual format String
`"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"`
This simply prints 'Z' at the end
```
{"name":"Hello","value":123.0,"timestamp":"2017-04-24T10:28:55.902Z"}
```
I believe that it should be `"yyyy-MM-dd'T'HH:mm:ss.SSS Z"` to effectively convey the time zone of the date
```
{"name":"Hello","value":123.0,"timestamp":"2017-04-24T10:30:03.574 -0400"}
```

Back to the original issue. . . another approach would be to *always* print date as GMT. But I do not think it's the right idea and could lead to many mistakes due to misinterpretation.

viniciusccarvalho:
Adding a timezone to the formatter fix the issue, I've tested it already locally.  The formatters can read the timezone with a `Z` at the end as well as the time difference to GMT. What needs to be done is adding a timezone to the formatter that will be able to write according to GMT - The intention is to _always_ write in GMT.

olegz:
Well, then we have two issues
1. The 'Z' in the current format is meaningless since it's quoted, so the quotes needs to be removed to get any meaning out of it.
2. If we add GMT timezone to the `DateFormat`, then what is the purpose of the 'Z' at the end? It will always be `0000`. Perhaps you meant `z` (lowercase). That with combination of `df.setTimeZone(TimeZone.getTimeZone("GMT"));` will accomplish what I believe you are looking for `{"name":"Hello","value":123.0,"timestamp":"2017-04-24T15:15:59.261 GMT"}`

viniciusccarvalho:
Unit tests suggest the opposite. The Z needs to be quoted to be parsed correctly. And there's several examples on stackoverflow of usage of 'Z' quoted. 
I have a unit tests to try it with 'Z' and then setting and reading from GMT. It's not meaningless because it can hint other parsers (outside of java land) that this time is actually at zulu time.

mbogoevici:
@olegz 
[...]

PR #929, opened 2017-04-24
GH-927 Fixed GMT formatting issue

olegz:
Fixed DateFormat definition to include 'ISO 8601 time zone' designator (see https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html)
Added constamt TimeZone to dateFormat
Added test

### DIFF
--- a/spring-cloud-stream-metrics/src/test/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializerTests.java
+++ b/spring-cloud-stream-metrics/src/test/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializerTests.java
@@ -16,10 +16,7 @@
 
 package org.springframework.cloud.stream.metrics.config;
 
-
-
 import java.io.StringWriter;
-import java.lang.reflect.Field;
 import java.util.Date;
 
 import com.fasterxml.jackson.core.JsonFactory;
@@ -30,7 +27,6 @@ import org.junit.Test;
 
 import org.springframework.boot.actuate.metrics.Metric;
 import org.springframework.cloud.stream.metrics.config.MetricJsonSerializer.Serializer;
-import org.springframework.util.ReflectionUtils;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@olegz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=929)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=929) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@olegz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=929)!

olegz:
@viniciusccarvalho actually X adds Z at the end (not the number) when DF is set to use GMT timezone. There is test showing this. Do you see something different?

olegz:
@viniciusccarvalho 
Just to clarify a bit. . . 
The quoted 'Z' was just that. In fact that was part of the issue where the date appeared to be in GMT but was actually not. With 'X' option you actually delegating to the Formatter to inject 'Z' (see code snippet below) if and only if the offset zone equals to 0 (which can only be IF your current timezone is GMT or you have written the date as GMT), thus verifying that your date was written in GMT format.
```
case PATTERN_ISO_ZONE:   // 'X'
            value = calendar.get(Calendar.ZONE_OFFSET)
                    + calendar.get(Calendar.DST_OFFSET);

            if (value == 0) {
                buffer.append('Z');
                break;
            }
```

spring-cloud-stream-metrics/src/test/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializerTests.java
```diff
@@ -0,0 +1,65 @@
...
+		Serializer ser = new Serializer();	
+		ser.serialize(metric, jsonGenerator, null);
+		jsonGenerator.flush();
+		
+		JSONObject json = new JSONObject(writer.toString());
+		String serializedTimestamp = json.getString("timestamp");
+		assertEquals("2017-04-24T18:56:37.188Z", serializedTimestamp);
+	}
+	
+	private void injectTimestamp(Date date, Metric<Number> metric) throws Exception {
```
viniciusccarvalho: Metric has a constructor that accepts a Date object, there's no need for overriding access rules of the the class. Just use `Metric<Number> metric = new Metric("Hello",123,date);`

olegz: Fair enough, didn't realize that at the time, will fix

viniciusccarvalho:
Squashed and merged via 8fc0cf1a0c1b8f8edcf542734048e59e42c74383

PR #934, opened 2017-04-27
Increase poll timeout for Metrics tests to 10 sec

artembilan:
Sometimes 1 second is really not enough to wait for the answer
from other thread, especially on the slow, hard-busy CI server


artembilan:
>  except for the checkstyle violation on import order

OK. I see SF has changed its rules again: https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style#import-statements. Sigh...
I wonder who and when fixes all those changes there 😉

mbogoevici:
Closed via https://github.com/spring-cloud/spring-cloud-stream/commit/ce772bc8a21ef110280ddb0031f04481bc650450 (after import reformatting).

PR #938, opened 2017-05-01
Fixes #922

### LINKED ISSUE #922: Fix unsafe access to DateFormat object

viniciusccarvalho:


spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializer.java
```diff
@@ -70,16 +73,28 @@ public void serialize(Metric<?> metric, JsonGenerator json,
...
 
 			return metric;
 		}
 
 	}
+
+	public static DateFormat defaultDateFormat(){
+		DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSX");
+		df.setTimeZone(TimeZone.getTimeZone("GMT"));
+		return df;
```
olegz: Not sure I understand this change since it will recreate the same DateFormat every time it is called instead of once per class instance. Any reason why?

olegz: Actually i'll take that back, just remembered why we did it in the first place (DF is not thread safe, thus can not be shared while in the middle of formatting)

olegz: The only other comment is does it really have to be public?
Otherwise +1

spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/config/MetricJsonSerializer.java
```diff
@@ -56,8 +54,13 @@ public void serialize(Metric<?> metric, JsonGenerator json,
 			json.writeStartObject();
 			json.writeStringField("name", metric.getName());
 			json.writeNumberField("value", metric.getValue().doubleValue());
+			DateFormat df = formattersQueue.poll();
+			if(df == null){
+				df = defaultDateFormat();
```
mbogoevici: We should have a `try`/`finally` block here as well I suppose ?

mbogoevici:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/516ba13402e035ef9ccb4507e56f25f9ebc0df20

Issue #897, opened 2017-04-02
AvroSchemaRegistryClientMessageConverter NPE


lwshuo:
Hi team,

I'm on 1.1.1 and still seeing https://github.com/spring-cloud/spring-cloud-stream/issues/694

I'm trying to use confluent schema registry. I'm using 
`@EnableSchemaRegistryClient` and I also have these beans registered:

 @Bean
  public MessageConverter userMessageConverter() {
    return new AvroSchemaMessageConverter(MimeType.valueOf("avro/bytes"));
  }

  @Bean
  ConfluentSchemaRegistryClient confluentSchemaRegistryClient(){
    return new ConfluentSchemaRegistryClient();
  }

  @Bean
  AvroSchemaRegistryClientMessageConverter avroSchemaRegistryClientMessageConverter(){
    return new AvroSchemaRegistryClientMessageConverter( confluentSchemaRegistryClient());
  }

Is there anything that I'm missing?

lwshuo:
Can someone point me to examples of how to properly use `ConfluentSchemaRegistryClient ` ?

olegz:
@lwshuo

NPE is definitely a bug on our end, and we're about to fix it. That said, the issue you are describing is due to miss-configuration on your end which results in NPE instead of a more informative message, so let me explain. . .
When using `@EnableSchemaRegistryClient` you are relying on auto-configuration which could and should work out of the box. The only variable (customization) you are trying to introduce is `ConfluentSchemaRegistryClient`, which you can simply do by declaring only one Bean (see below):
```java
@Bean
@Primary
ConfluentSchemaRegistryClient confluentSchemaRegistryClient() {
	return new ConfluentSchemaRegistryClient();
}
```
That is all you need to ensure that instead of `DefaultSchemaRegistryClient`. The `@Primary` _. . .Indicates that a bean should be given preference when multiple candidates. . ._
 But when you defining all those beans (up above) you are essentially attempting to override what's already provided via auto-config and when you do that you are missing some of the dependencies required by some of the beans and in this case it's a `CacheManager` that is missing from your instance of `AvroSchemaRegistryClientMessageConverter`.

Anyway, as I said, we do recognize that NPE is not a good exception to be seen by the client and will fix it, you also need to adjust your configuration.

Please let us know

lwshuo:
@olegz  I tried what you said and it works now. Thank you so much!

PR #939, opened 2017-05-01
GH-897 fixed the NPE issue with CacheManager

olegz:
- Added a new constructor which takes CacheManager as an argument
- Deprecated the old constructor and setter for CacheManager
- Added fallback to NoOpsCacheManager to the deprecatd construtor to avoid the possiblity of NPE.
- Added warning to notify user that caching is _effectively disabled_ if NoOpCacheManager is used.

### DIFF
--- a/spring-cloud-stream-schema/src/test/java/org/springframework/cloud/schema/avro/AvroSchemaRegistryClientMessageConverterTests.java
+++ b/spring-cloud-stream-schema/src/test/java/org/springframework/cloud/schema/avro/AvroSchemaRegistryClientMessageConverterTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-schema/src/test/java/org/springframework/cloud/schema/avro/AvroSchemaRegistryClientMessageConverterTests.java
```diff
@@ -23,24 +23,31 @@
...
 import org.springframework.cloud.stream.test.binder.MessageCollector;
 import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.support.MessageBuilder;
 
 import static org.assertj.core.api.Assertions.assertThat;
-
+import static org.junit.Assert.assertTrue;
```
mbogoevici: Just as a note - will fix on merge -I'd much prefer we use AssertJ. Did a sweep back in the day but seems like we have a few usages of AspectJ now again.

olegz: Fair enough. Will keep in mind in the future

mbogoevici:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/8e77993fad95231c50cf358d6f40f291a5ef012e

PR #942, opened 2017-05-03
Fix JSON contentType handling in payload serialization

### LINKED ISSUE #941: JSON contentType handling when serializing the payload
When setting the `contentType` message header, the original content type from the payload should be checked with MimeType's type/subtype/parameter. 

For instance, the original contentType with the value `application/json;charset=UTF-8` currently serialized with the contentType `text/plain` based on this logic [here](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java#L308)

ilayaperumalg:
- Check if the original contentType starts with `application/json` and this way any original contentType with parameters such as `charset` can also be considered correctly

Resolves #941

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -306,7 +306,7 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 				return MimeTypeUtils.APPLICATION_OCTET_STREAM;
 			}
 			if (payload instanceof String) {
-				return MimeTypeUtils.APPLICATION_JSON_VALUE.equals(originalContentType) ? MimeTypeUtils.APPLICATION_JSON
+				return (originalContentType.startsWith(MimeTypeUtils.APPLICATION_JSON_VALUE)) ? MimeTypeUtils.APPLICATION_JSON
 						: MimeTypeUtils.TEXT_PLAIN;
 			}
 			String className = payload.getClass().getName();


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -306,7 +306,7 @@ static MimeType mimeTypeFromObject(Object payload, String originalContentType) {
 				return MimeTypeUtils.APPLICATION_OCTET_STREAM;
 			}
 			if (payload instanceof String) {
-				return MimeTypeUtils.APPLICATION_JSON_VALUE.equals(originalContentType) ? MimeTypeUtils.APPLICATION_JSON
+				return (originalContentType.startsWith(MimeTypeUtils.APPLICATION_JSON_VALUE)) ? MimeTypeUtils.APPLICATION_JSON
```
artembilan: OK. My `originalContentType` is `application/json;charset=UTF-8`, but according the outcome of this operation I'll get only `application/json`.
Since the `param` maybe any, wouldn't it be better to return `originalContentType` instead of raw `MimeTypeUtils.APPLICATION_JSON`?

mbogoevici: Shouldn't we actually return the `originalContentType` here instead to preserve the encoding?

ilayaperumalg: yeah, that is better.

ilayaperumalg: Yes @artembilan, that is correct and better! :-)

mbogoevici: FWIW we might always send back the originalContentType if it's not empty. The original change was meant to prevent the binder to override `application/json` but it seems like it shouldn't actually override anything.

artembilan:
```
SptringUtils.hasText(originalContentType) ? MimeType.valueOf(originalContentType) : MimeTypeUtils.TEXT_PLAIN;
```
?

mbogoevici:
We also need a test that reproduces the issue.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
mbogoevici: Actually I take that back.

It would be better overall for this to be explained with a test case.

Even if `contentType` is set to `text/plain`, `originalContentType` remains `application/json;charset=UTF-8`. Since on receiving the message, the binder uses `originalContentType` to replace `contentType`, it will handle a message that effectively has the charset-bearing version as `contentType`. The `contentType` header just instructs a binder that understands headers (e.g. Rabbit) that the content of the message is textual. The special treatment of JSON allows non-Spring Cloud Stream applications to handle the content as JSON. Overall, it is better to preserve the content type, but it only affects non-SCSt consumers.

ilayaperumalg:
Closing this PR as the `contentType` setting here would just be for the binder's serialization/de-serialization and we may not the `parameter` included in there.

Issue #935, opened 2017-04-30
Runtime errors with Boot 2.x


jvalkeal:
Mostly for heads up but boot is starting to break a lot of things
```
Caused by: java.lang.NoClassDefFoundError: org/springframework/boot/bind/RelaxedDataBinder
	at org.springframework.cloud.stream.binding.BindingService.validate(BindingService.java:174)
	at org.springframework.cloud.stream.binding.BindingService.bindConsumer(BindingService.java:96)
	at org.springframework.cloud.stream.binding.BindableProxyFactory.bindInputs(BindableProxyFactory.java:221)
	at org.springframework.cloud.stream.binding.InputBindingLifecycle.start(InputBindingLifecycle.java:55)
```
`RelaxedDataBinder` is gone, see https://github.com/spring-projects/spring-boot/issues/9000

I think we need to create `2.x` so that we start to see what issues spring5/boot2 brings in.

mbogoevici:
@jvalkeal Yes, the plan is to get started on a 2.x branch in May.

davidwadden:
Is there a build snapshot maven repository tracking the `2.x` branch anywhere?  I don't see one at http://repo.spring.io/libs-snapshot/org/springframework/cloud/spring-cloud-stream/

mbogoevici:
Not yet, will have one next week once the binders are upgraded as well.

mbogoevici:
Closing this after merging @jvalkeal 's PR in the `2.0.x` branch and making respective branches for all binders.

@davidwadden To your question - the BOM for `Elmhurst` release train (Based on 2.0) is CI-built now and available here https://repo.spring.io/libs-snapshot-local/org/springframework/cloud/spring-cloud-stream-dependencies/Elmhurst.BUILD-SNAPSHOT/

PR #959, opened 2017-05-12
Upgrade build to boot 2.x

jvalkeal:
This PR is take on what it needed to pump up to boot 2.x. Not yet sure if everything actually works beyond unit tests.

- Pump up version to 2.0.0
- Some generic polish
- All changes around breakage with boot 2.x
- Some boot classes has been moved around
- You can't no longer have binding key ending with
  camelCase.
- New Binder now has illegal keys.
- Some changes to tests as we can directly do end-to-end
  testing with ENV_VAR_FORMAT as normal keys
- Spring data repo changes as now uses Optional
- Remove relaxed binder and its tests in favor of new Binder
- Some mockito api changes
- One Ingored test TextPlainToJsonConversionTest.testTextPlainToJsonConversionOnInput
- Relates to #935

For `testTextPlainToJsonConversionOnInput`, failure is
```
org.junit.ComparisonFailure: expected:<"transformed-[Bar]"> but was:<"transformed-[{"name":"Bar"}]">
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at org.springframework.cloud.stream.config.TextPlainToJsonConversionTest.testTextPlainToJsonConversionOnInput(TextPlainToJsonConversionTest.java:73)
```
which is something I couln't figure out. There seem to be 4th converter `ConfigurableCompositeMessageConverter` as I had to add check in `CustomMessageConverterTests` for it. Maybe something caused by spring int.

github.com/spring-projects/spring-boot/issues/9152 is fixed which fixed some tests.


spring-cloud-stream-metrics/src/main/java/org/springframework/cloud/stream/metrics/ApplicationMetricsProperties.java
```diff
@@ -99,6 +120,29 @@ public void setProperties(String[] properties) {
 	private Map<String, Object> exportProperties = new HashMap<>();
 
 	public Map<String, Object> getExportProperties() {
+		this.exportProperties.clear();
```
mbogoevici: Wouldn't this  cause the `exportProperties` array to be computed each time we send out a metrics message? That would be very inefficient, as the list doesn't change between refresh operations. That was the main reason why it's been done once via `onApplicationEvent` - based on the discussion that @jvalkeal and I had, that seems to have some race condition issue, but we should probably make sure that this is cached rather than computed every time.

jvalkeal: Yes, I was supposed to cache that but forget.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/config/SchemaRegistryClientProperties.java
```diff
@@ -22,7 +22,7 @@
  * @author Marius Bogoevici
  * @author Vinicius Carvalho
  */
-@ConfigurationProperties(prefix = "spring.cloud.stream.schemaRegistryClient")
+@ConfigurationProperties(prefix = "spring.cloud.stream.schema-registry-client")
```
mbogoevici: Actually, this is my mistake see https://github.com/spring-cloud/spring-cloud-stream/pull/959#discussion_r116241186 as well

mbogoevici:
The test failure happens because the converter using the `ConverterService` is invoked before the content type converters - if we remove the String constructor then everything works alright (but I think we should be conscious about this issue).

spring-cloud-stream-metrics/src/test/java/org/springframework/cloud/stream/metrics/ApplicationMetricsExporterTests.java
```diff
@@ -38,17 +38,18 @@
 
 /**
  * @author Vinicius Carvalho
+ * @author Janne Valkealahti
  */
 public class ApplicationMetricsExporterTests {
 
 	@BeforeClass
 	public static void setSystemProps() {
-		System.setProperty("SPRING_TEST_ENV_SYNTAX", "testing");
```
jvalkeal: yes, not supported anymore

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/aggregate/AggregateApplicationBuilder.java
```diff
@@ -223,40 +229,22 @@ public ConfigurableApplicationContext run(String... parentArgs) {
 		for (Map.Entry<AppConfigurer, String> appConfigurerEntry : appConfigurers.entrySet()) {
 
 			AppConfigurer appConfigurer = appConfigurerEntry.getKey();
+			if (appConfigurerEntry.getValue() == null) {
```
jvalkeal: Well old code would NPE if entry value is null so not sure what you're asking.

mbogoevici:
Squashed and merged into new branch https://github.com/spring-cloud/spring-cloud-stream/tree/2.0.x as https://github.com/spring-cloud/spring-cloud-stream/commit/3850d6c8c977c69a564476cf1db4cc1c969a471c

PR #974, opened 2017-05-22
Add support for Reactive sources through @StreamEmitter

### LINKED ISSUE #722: Add support for Reactive sources

sobychacko:
Fix #722 

Rebased against 1.3.0
Incorporated most of the comments from https://github.com/spring-cloud/spring-cloud-stream/pull/885


spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitter.java
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @author Soby Chacko
+ * @since 1.3.0
+ */
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface StreamEmitter {
+
+	String value() default "";
```
artembilan: Need JavaDoc - what is this about.
Plus I don't see sample for this case in the general JavaDoc on the class.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,273 @@
...
+
+	private final List<FluxSender> fluxSenders = new ArrayList<>();
+
+	private final List<Disposable> fluxDisposables = new ArrayList<>();
+
+	private ConfigurableApplicationContext applicationContext;
+
+	private final List<Method> targetMethods = new ArrayList<>();
+
+	private Object targetBean;
```
artembilan: We talked with you before: my application can have several `@StreamEmitter` s and in the different classes.
Just add a test-case when you have several `@Configuration`, each with its own `@StreamEmitter` method!

I suggest to have a `Map<Method, Object>` since you are going to call all those method and we have to do that on the proper object.

sobychacko: thought we can do that in a separate iteration. Wonder if thats a widely applied use case or not. Not sure if StreamListener supports that approach.

artembilan: Well, I can't insist, but I'd not let it go in SI. This is an obvious ([observed](https://github.com/spring-cloud/spring-cloud-stream/pull/885#discussion_r108976014)) bug and no, `@StreamListener` supports that:
```
mappedListenerMethods.add(streamListener.value(),
				new StreamListenerHandlerMethodMapping(bean, method, streamListener.condition(), defaultOutputChannel));
```

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/FluxSenderImpl.java
```diff
@@ -0,0 +1,68 @@
...
+class FluxSenderImpl implements FluxSender, Disposable {
+
+	private Log log = LogFactory.getLog(FluxSenderImpl.class);
+
+	private Subscription subscription;
+
+	private Consumer<Object> consumer;
+
+	FluxSenderImpl(Consumer<Object> consumer) {
+		this.consumer = consumer;
```
mbogoevici: Needs null check assertion.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/FluxSenderImpl.java
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	@Override
+	public Mono<Void> send(Flux<?> flux) {
+		MonoProcessor<Void> sendResult = MonoProcessor.create();
+		// add error handling and reconnect in the event of an error
+		flux
+				.doOnError(e -> this.log.error("Error during processing: ", e))
+				.retry()
+				.doOnSubscribe(subscription -> this.subscription = subscription)
```
mbogoevici: A bit minor, but looking at the original suggestion in https://github.com/spring-cloud/spring-cloud-stream/pull/974/files/bd6ba25ef04e8928d279cb26772f005bb523e180#r118102222, I wonder if it wouldn't be simpler to store the Disposable returned by `subscribe` and have `dispose` delegate to it directly. A bit of a nit, but it avoids the extra `doOnSubscribe` and is a bit more straightforward IMO - @artembilan  ?

artembilan: OK. Fine with me.
That is something new from the Reactor.
I see now that in `Flux`:
```
public final Disposable subscribe(Consumer<? super T> consumer) {
```
My mind has been blinded  by the `Publisher` API:
```
public void subscribe(Subscriber<? super T> s);
```
So, yes, indeed, let's go with the return from the `subscribe()`

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,273 @@
...
+		});
+		return bean;
+	}
+
+	@Override
+	public void start() {
+		try {
+			lock.lock();
+			if (!running) {
+				mappedStreamEmitterMethods.forEach((k, v) -> v.forEach(item -> {
```
artembilan: Hm. Seems for me `MultiValueMap` is an overhead here.
Why simple `Map<Method, Object>` isn't enough ?

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,273 @@
...
+			callback.run();
+		}
+	}
+
+	@Override
+	public void stop() {
+		try {
+			lock.lock();
+			for (FluxSender fluxSender : fluxSenders) {
+				if (fluxSender.getClass().isAssignableFrom(FluxSenderImpl.class)) {
```
artembilan: ??? `FluxSenderImpl` **is** `Disposable`, too, so can be added to the `fluxDisposables` as well.
OTOH it would be better to avoid tangle with that package protected `FluxSenderImpl` class.
It `implements Disposable`, so we can `instanceof` it against that interface.
Unclear though why you are using `.getClass().isAssignableFrom()` on the evaluated object...

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/FluxSenderImpl.java
sobychacko: will change it then.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
sobychacko: `MultiValueMap` here ensures that the map is keyed by the bean and it contains a list of all the `method`s that are annotated with `StreamEmitter`. I found it cleaner to use it that way. What are your objections for doing this way? 
With `Map<Method, Object>` we have `StreamEmitter` method as the key and then value as the bean that contains the method, right?

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
sobychacko: will clean it up.

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
artembilan: Well, I see the point of mapping by bean, but at the same time we don't get a gain of that mapping and it is really unimportant in our logic. What should be strong is a one-to-one relationship between method and the target object. 
But yeah, that might be only me to be so nit-pick... Never mind, I can live with MultiValueMap as well

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
mbogoevici: The other option would be to simply have `FluxSender` extend `Disposable` - that would have some logic as the outgoing data stream should be cancellable/disposable.

artembilan: Fine with me, too. 
And with that we may come back to the anonymous implementation in that ParameterAdapter.
And already without that ugly `...Impl` name. 

I've chosen `FluxSenderImpl` just as a quick sample about target implementation for the interface. The name really was just an arbitrary. You call name it with other proper name. If we will stick with that yet, of course...

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerResultAdapter.java
```diff
@@ -26,7 +26,7 @@
  * annotated method is operating in declarative mode.
  * @author Marius Bogoevici
  */
-public interface StreamListenerResultAdapter<R, B> {
+public interface StreamListenerResultAdapter<R, B, D> {
```
mbogoevici: Chatting with @sobychacko about this offline. It seems like we could skip the generic parameter and simply return a `Closeable` from `adapt`. This would give more flexibility to the core to act on the result (e.g. right now `StreamListenerResultAdapter` can't really do anything with the return value, because it doesn't know what it is. 

@artembilan - any thoughts?

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
mbogoevici: Another idea would be to continue what on the suggestion here: https://github.com/spring-cloud/spring-cloud-stream/pull/974/files#r119424616 and rely on the more generic `Closeable.close()` instead of `Disposable` This would make it more reusable at the core framework level.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerResultAdapter.java
artembilan: Sounds reasonable, but I don't know the whole context.

Not sure though why you are asking me since I'm not so good in this project at all...

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,279 @@
...
+
+		if (!Void.TYPE.equals(method.getReturnType())) {
+			if (targetBean == null) {
+				targetBean = this.applicationContext.getBean(outboundName);
+			}
+			boolean streamListenerResultAdapterFound = false;
+			for (StreamListenerResultAdapter streamListenerResultAdapter : this.streamListenerResultAdapters) {
+				if (streamListenerResultAdapter.supports(result.getClass(), targetBean.getClass())) {
+					Object adapted = streamListenerResultAdapter.adapt(result, targetBean);
+					if (adapted != null && Disposable.class.isAssignableFrom(adapted.getClass())) {
```
artembilan: According to the latest changes it looks like we don't have `Disposable` returns any more.
So, can be dropped altogether here as well.

Also I still don't understand why do you use `class.isAssignableFrom()` instead of `instanceof`.
You have already `Object adapted` in place. What's wrong? Thanks

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,279 @@
...
+			}
+			for (Disposable disposable : fluxDisposables) {
+				disposable.dispose();
+			}
+			if (running) {
+				this.running = false;
+			}
+		}
+		catch (IOException e) {
+			log.error("Error closing reactive source", e);
```
artembilan: Shouldn't we consider `try...catch` individually?
Why an error on one `FluxSender.close()` should prevent all others to be closed afterward?

spring-cloud-stream-reactive/src/main/java/org/springframework/cloud/stream/reactive/StreamEmitterAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,274 @@
...
+			lock.lock();
+			for (Closeable closeable : fluxDisposables) {
+				try {
+					closeable.close();
+				}
+				catch (IOException e) {
+					log.error("Error closing reactive source", e);
+				}
+			}
+			if (running) {
```
artembilan: Why it is here, after the hard loop of closes?
Why that loop isn't wrapped to this `if`?
Thanks

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamEmitterBasicTests.java
```diff
@@ -0,0 +1,280 @@
...
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.messaging.MessageChannel;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Soby Chacko
+ */
+public class StreamEmitterBasicTests {
```
artembilan: These tests are very slow:

image

Would be better to reconsider them, at least switch of Web facet for Boot.

spring-cloud-stream-reactive/src/test/java/org/springframework/cloud/stream/reactive/StreamEmitterValidationTests.java
```diff
@@ -0,0 +1,277 @@
...
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.INVALID_OUTPUT_METHOD_PARAMETERS;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.NO_OUTPUT_SPECIFIED;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.OUTPUT_ANNOTATION_MISSING_ON_METHOD_PARAMETERS_VOID_RETURN_TYPE;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.RETURN_TYPE_METHOD_ARGUMENTS;
+import static org.springframework.cloud.stream.reactive.StreamEmitterErrorMessages.RETURN_TYPE_NO_OUTBOUND_SPECIFIED;
+
+/**
+ * @author Soby Chacko
+ */
+public class StreamEmitterValidationTests {
```
artembilan: This is about validation. Why should it be Spring Boot in full harness?
Slow as well...

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerResultAdapter.java
```diff
@@ -41,6 +43,6 @@
 	 * @param streamListenerResult the result of invoking the method.
 	 * @param bindingTarget the binding target.
 	 */
-	void adapt(R streamListenerResult, B bindingTarget);
+	Closeable adapt(R streamListenerResult, B bindingTarget);
```
artembilan: This is breaking change in the minor version.
I understand that you have an argument like impls are an internal stuff and it is unusual to have something custom in the target app and so on.
But would be better to keep that in mind for similar changes in the future.
Only major version can afford such a squander.

sobychacko: @artembilan agree with the fact that this is a breaking change in 1.3. However, we discussed this with @mbogoevici before and came to the conclusions above - i.e. the `StreamListenerResultAdapter` implementations are internal exclusively at this point.

mbogoevici: Due to the fact that we have other breaking changes (e.g. dependency requirements or so) I think that it is safe to to do this, especially as this never was a publicly used API.

artembilan:
Please, find a polishing here: https://github.com/artembilan/spring-cloud-stream/commit/fca75f42d014ee3edfd0e8556d9305f36ab51a6d.

Great to have a collaboration with you, so we have caught that Java DSL bug 😄 
Would be glad to meet more often for similar pair programming!

artembilan:
The fix for SI Java DSL: https://github.com/spring-projects/spring-integration-java-dsl/commit/7ff8aa947f964542364358e77656a8fa64175fe2

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/c3246f32571b4e64065d7fd489ea70909d2be272

PR #975, opened 2017-05-24
Ignore unknown binder types if not used

### LINKED ISSUE #972: Ignore binder configurations referencing unknown binder types if they are not used
When multiple binder configurations are specified, we check whether the binder type is recognized (present on the classpath). It is, however possible to create global configurations that reference multiple binders in scenarios like the one described in http://docs.spring.io/spring-cloud-dataflow/docs/1.2.0.RELEASE/reference/htmlsingle/#dataflow-multiple-brokers. Spring Cloud Stream will aggressively reject these configurations, even if, for example, the binder 'kafka1' is never used in the application (due to `defaultBinder` being set to `rabbit1`). Instead of that, we should be lenient and only complain if the configuration is actually being used (e.g. if a binder instance is requested).

mbogoevici:
Fix #972

Do not throw an error when binder configurations are
processed, unless a binder configuration that references
an unknown binder type (i.e. not found on the classpath)
is actually used by an application.

This allows supporting scenarios like
http://docs.spring.io/spring-cloud-dataflow/docs/1.2.0.RELEASE/reference/htmlsingle/#dataflow-multiple-brokers
where a set of binder configurations is passed to a number
of applications that might or might not have a specific
binder type on the classpath. If a specific binder configuration
is not used by the application, it can be ignored.

Backport to 1.2.x branch.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
@@ -64,12 +64,6 @@ public class BinderFactoryConfiguration {
 	@Value("${" + SELF_CONTAINED_APP_PROPERTY_NAME + ":}")
 	private String selfContained;
 
-	@Autowired
-	private BinderTypeRegistry binderTypeRegistry;
-
-	@Autowired
-	private BindingServiceProperties bindingServiceProperties;
-
 	@Autowired(required = false)
 	private Collection<DefaultBinderFactory.Listener> binderFactoryListeners;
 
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -93,16 +87,19 @@
...
+			BindingServiceProperties bindingServiceProperties) {
+		DefaultBinderFactory binderFactory = new DefaultBinderFactory(
+				getBinderConfigurations(binderTypeRegistry, bindingServiceProperties), binderTypeRegistry);
 		binderFactory.setDefaultBinder(bindingServiceProperties.getDefaultBinder());
 		binderFactory.setListeners(binderFactoryListeners);
 		return binderFactory;
 	}
 
-	public Map<String, BinderConfiguration> getBinderConfigurations() {
+	public Map<String, BinderConfiguration> getBinderConfigurations(BinderTypeRegistry binderTypeRegistry,
```
sobychacko: Shall we make this method as `private`?

mbogoevici: Possibly.

sobychacko:
LGTM - polished & squashed the changes. 
Cherry picked to 1.2.x

PR #984, opened 2017-06-07
Option to suppress header propagation with SI message handlers

### LINKED ISSUE #943: Explicitly exclude `contentType` header propagation with handlers
Ensure that `spring.integration.readOnly.headers=contentType` is set for SCSt applications (prevents contentType headers of the input to be propagated to the output).

mbogoevici:
Add `SpringIntegrationProperties` class with
`spring.cloud.stream.integration` prefix for managing
Spring Integration properties.

Add `spring.cloud.stream.integration.messageHandlerNotPropagatedHeaders`
for controlling headers propagated by reply producing handlers.

Fix #943 

Merge in `master` and `1.2.x` - once Boot 1.5.4 is available.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
@@ -56,6 +56,7 @@ import org.springframework.expression.PropertyAccessor;
 import org.springframework.integration.channel.PublishSubscribeChannel;
 import org.springframework.integration.config.IntegrationEvaluationContextFactoryBean;
 import org.springframework.integration.context.IntegrationContextUtils;
+import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
 import org.springframework.integration.json.JsonPropertyAccessor;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.converter.MessageConverter;
@@ -76,7 +77,7 @@ import org.springframework.util.CollectionUtils;
  * @author Gary Russell
  */
 @Configuration
-@EnableConfigurationProperties(BindingServiceProperties.class)
+@EnableConfigurationProperties({BindingServiceProperties.class, SpringIntegrationProperties.class})
 public class BindingServiceConfiguration {
 
 	public static final String STREAM_LISTENER_ANNOTATION_BEAN_POST_PROCESSOR_NAME = "streamListenerAnnotationBeanPostProcessor";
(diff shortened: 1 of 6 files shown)


mbogoevici:
(also needs docs)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpringIntegrationProperties.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.messaging.MessageHeaders;
+
+/**
+ * Contains properties for Spring Integration settings.
```
mbogoevici: Needs author tag.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpringIntegrationProperties.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+@ConfigurationProperties("spring.cloud.stream.integration")
+public class SpringIntegrationProperties {
+
+	private String[] messageHandlerNotPropagatedHeaders = new String[] { MessageHeaders.CONTENT_TYPE };
```
artembilan: This probably should go to `IntegrationProperties` in Spring Boot 2.0: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java

Please, consider to add some JavaDocs for Boot meta-data any way: http://docs.spring.io/spring-boot/docs/1.5.3.RELEASE/reference/html/configuration-metadata.html#configuration-metadata-annotation-processor

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpringIntegrationProperties.java
artembilan: ... and `@since`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
```diff
@@ -245,6 +246,28 @@ public MessageChannel resolveDestination(String name) throws DestinationResoluti
...
 		}
+
+		@Bean
+		public static BeanPostProcessor messageHandlerHeaderPropagationBeanPostProcessor(
+				final SpringIntegrationProperties springIntegrationProperties) {
+			return new BeanPostProcessor() {
+				@Override
+				public Object postProcessBeforeInitialization(Object bean, String beanName) throws BeansException {
+					// TODO: Filter out beans created by SCSt (not currently necessary)
+					if (bean instanceof AbstractReplyProducingMessageHandler) {
```
artembilan: Not sure if that is critical right now or will be possible according your TODO, but the bean can be proxy coming to this BPP, therefore won't be `instanceof` for class. Only interfaces are possible here.
Or something like this:
```
Class<?> targetClass = bean.getClass();
if (AopUtils.isAopProxy(bean) {
targetClass = AopUtils.getTargetClass(this.handler);

}
boolean replyMessageHandler = AbstractReplyProducingMessageHandler.class.isAssignableFrom(targetClass);
```

artembilan:
@mbogoevici ,

Spring Boot `1.5.4` is here, but let me guess that doesn't mean that we are able to easily to switch to it to pull, for example, SI-4.3.10. I think we need some specific upgrade in Spring Cloud Build. No?

Thanks

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SpringIntegrationProperties.java
mbogoevici: We should do a better job for all properties I think :(

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
mbogoevici: Thanks - that makes sense.

mbogoevici:
@artembilan Once https://github.com/spring-cloud/spring-cloud-build/pull/55 the baseline version will be Spring Boot 1.5.4.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/51deb128f0250ed0212bd7ab5cc2ba723a4bb1e3 and cherry-picked to `1.2.x`

PR #994, opened 2017-06-20
GH-993: Conversion Interceptor Improvements

### LINKED ISSUE #993: MessageConverterConfigurer - Fall Back to no contentType Header if Converter Returns null
Currently, if the `contentType` header in the message causes the converter to return null, the conversion fails with "Null payload not allowed" or similar.

Add something like...

```java
if (this.provideHint) {
	converted = ((AbstractMessageConverter) this.messageConverter).fromMessage(message, this.klazz,
			this.mimeType);
	if (converted == null) {
		// remove content type header and try again
		// add a warn log indicating the header needs to be remove
	}
	if (converted == null) // throw exception
}
```

garyrussell:
Fixes #993

If the converter cannot convert due to a lingering content-type header, take a second
shot with the header removed.

If we still fail, throw a `MessageConversionException`.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
@@ -243,34 +243,59 @@ public class MessageConverterConfigurer implements MessageChannelConfigurer, Bea
 					if (this.provideHint) {
 						converted = ((AbstractMessageConverter) this.messageConverter).fromMessage(message, this.klazz,
 								this.mimeType);
+						if (converted == null && message.getHeaders().containsKey(MessageHeaders.CONTENT_TYPE)) {
+							converted = ((AbstractMessageConverter) this.messageConverter).fromMessage(
+									MessageConverterConfigurer.this.messageBuilderFactory.fromMessage(message)
+										.removeHeader(MessageHeaders.CONTENT_TYPE)
+										.build(), this.klazz, this.mimeType);
+						}
 					}
 					else {
 						converted = this.messageConverter.fromMessage(message, this.klazz);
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -243,34 +243,59 @@ else if (this.mimeType.getType().equals("text") || this.mimeType.getSubtype().eq
 					if (this.provideHint) {
 						converted = ((AbstractMessageConverter) this.messageConverter).fromMessage(message, this.klazz,
 								this.mimeType);
+						if (converted == null && message.getHeaders().containsKey(MessageHeaders.CONTENT_TYPE)) {
+							converted = ((AbstractMessageConverter) this.messageConverter).fromMessage(
+									MessageConverterConfigurer.this.messageBuilderFactory.fromMessage(message)
+										.removeHeader(MessageHeaders.CONTENT_TYPE)
+										.build(), this.klazz, this.mimeType);
```
artembilan: Not sure what is that `messageBuilderFactory` configured for, but that seems for me it would be much efficient to use `MutableMessageBuilder` here any way.

That's really unfortunate that we have to rebuild message in this interceptor several time before it reaches the target `MH`...

I SI 5.0 we use `ConfigurableCompositeMessageConverter` for the `DefaultMessageHandlerMethodFactory` to enrich target `InvocableHandlerMethod` with proper argument conversion, including `Content-Type` negotiation.

I understand that this discussion isn't related to the PR, but just wanted to express the pain in my hurt each time I see message rebuild overhead...

sobychacko: These changes look good to me.

sobychacko:
These changes look good to me.

PR #1002, opened 2017-06-27
Allow custom JSON types to be handled properly

### LINKED ISSUE #940: Allow json contentType to support versioning schemes
Currently the converter for json only supports `application/json` we should allow a more flexible definition to support vnd and version content-types such as `application/vnd[...]+json` and `application/vnd[...].version+json`

viniciusccarvalho:
Fixes #940 

Changed logic to instead of strict comparison of contentType using
application/json we would only search for subtype json.

This allow us to support vnd.vendor.type.version+json types

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -56,6 +56,7 @@ import org.springframework.util.StringUtils;
  * @author Ilayaperumal Gopinathan
  * @author Mark Fisher
  * @author Marius Bogoevici
+ * @author Vinicius Carvalho
  */
 public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends ProducerProperties>
 		implements ApplicationContextAware, InitializingBean, Binder<T, C, P> {
@@ -238,7 +239,7 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 	}
 
 	private Object deserializePayload(byte[] bytes, MimeType contentType) {
-		if ("text".equalsIgnoreCase(contentType.getType()) || MimeTypeUtils.APPLICATION_JSON.equals(contentType)) {
+		if ("text".equalsIgnoreCase(contentType.getType()) || contentType.getSubtype().toLowerCase().contains(MimeTypeUtils.APPLICATION_JSON.getSubtype())) {
 			try {
 				return new String(bytes, "UTF-8");
 			}
@@ -303,13 +304,13 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 		private static ConcurrentMap<String, MimeType> mimeTypesCache = new ConcurrentHashMap<>();
 
 		static MimeType mimeTypeFromObject(Object payload, String originalContentType) {
+
 			Assert.notNull(payload, "payload object cannot be null.");
 			if (payload instanceof byte[]) {
 				return MimeTypeUtils.APPLICATION_OCTET_STREAM;
 			}
 			if (payload instanceof String) {
-				return MimeTypeUtils.APPLICATION_JSON_VALUE.equals(originalContentType) ? MimeTypeUtils.APPLICATION_JSON
-						: MimeTypeUtils.TEXT_PLAIN;
+				return originalContentType.toLowerCase().contains(MimeTypeUtils.APPLICATION_JSON.getSubtype()) ? MimeType.valueOf(originalContentType) : MimeTypeUtils.TEXT_PLAIN;
 			}
 			String className = payload.getClass().getName();
 			MimeType mimeType = mimeTypesCache.get(className);
--- a/spring-cloud-stream-binder-test/src/test/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupportTests.java
+++ b/spring-cloud-stream-binder-test/src/test/java/org/springframework/cloud/stream/binder/MessageChannelBinderSupportTests.java
@@ -132,6 +132,20 @@ public class MessageChannelBinderSupportTests {
 	}
 
 	@Test
+	public void testContentTypePreservedForCustomJson() throws IOException {
+		Message<String> inbound = MessageBuilder.withPayload("{\"foo\":\"foo\"}")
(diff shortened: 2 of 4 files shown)


artembilan:
Just my five fingers: 
no tests - not merge.

Rephrasing a popular Ukrainian author book:

> "Can we merge in the evening and add tests in the morning?"
"We can, but tests first!"

https://en.wikipedia.org/wiki/The_Twelve_Chairs

viniciusccarvalho:
ugh, forgot that `MessageChannelBinderSupportTests` only had tests for `application/json`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -303,13 +304,13 @@ public RetryTemplate buildRetryTemplate(ConsumerProperties properties) {
...
 		static MimeType mimeTypeFromObject(Object payload, String originalContentType) {
+
 			Assert.notNull(payload, "payload object cannot be null.");
 			if (payload instanceof byte[]) {
 				return MimeTypeUtils.APPLICATION_OCTET_STREAM;
 			}
 			if (payload instanceof String) {
-				return MimeTypeUtils.APPLICATION_JSON_VALUE.equals(originalContentType) ? MimeTypeUtils.APPLICATION_JSON
-						: MimeTypeUtils.TEXT_PLAIN;
+				return originalContentType.toLowerCase().contains(MimeTypeUtils.APPLICATION_JSON.getSubtype()) ? MimeType.valueOf(originalContentType) : MimeTypeUtils.TEXT_PLAIN;
```
ilayaperumalg: Since the `contentType` value extracted here is only for the binder serialization and for the payload `String`, we may want to still set `application/json` as the `contentType` here instead of `MimeType.valueOf(originalContentType)`. This way, we can still preserve the `originalContentType` and use it after deserialization.

ilayaperumalg: In that case, I believe we may not need `CustomJackson2MessageConverter` at all :-)

viniciusccarvalho: @ilayaperumalg this defeats the purpose of this PR. The intention is to **not** set the contentType to `application/json` but instead respect the contentType set by the user. If we modify it, then the contentType that was `application/vnd.type.version+json` will become `application/json` and then we loose the contentType on the flow. 

Although the `originalContetType` could preserve on the first `hop` but in my opinion is **wrong** what we do: `contentType:text/plain`, `originalContentType:application/vnd.type.version+json`. 

**No**, this indicates a transformation, think about it, I'm saying the content type now is X where it used to be Y.
This PR is to make sure we honor the contentType set by the user in case of custom json types.

IMHO we should actually discuss our strategy for contentType for 2.0, I have a few pointers to share with the team

ilayaperumalg: @viniciusccarvalho Please check this set of changes https://github.com/ilayaperumalg/spring-cloud-stream/commit/a5b5bb2c4ca6188f4abfb5a15eff5bd3eef9e374 and let me know what is missing in terms of supporting the feauture this PR addresses. I am trying to understand why do we need to set the `contentType` while it is *only* used for `binder` serialization/de-serialization proces. 


>  If we modify it, then the contentType that was application/vnd.type.version+json will become application/json and then we loose the contentType on the flow.

As long as we keep the `originalContentType` I think we are still good retriveing the actual `contentType` because the `contentType` we set for serialization/de-serialization is only valid at the `binder` level.

> Although the originalContetType could preserve on the first hop but in my opinion is wrong what we do: contentType:text/plain, originalContentType:application/vnd.type.version+json.

here, I meant to use `contentType:application/json, originalContentType:application/vnd.type.version+json` during binder's serialization process.

viniciusccarvalho: @ilayaperumalg I don't understand why we would want to keep this behavior. It looks like a bad smell to me having this:
`ContentType: application/json` `OriginalContentType: application/vnd.type.version+json` I've never got my head around this.

If can have only `ContentType` why do I want to enforce originalContentType? And it's very awkward that we have only `ContentType` when it's JSON payload, and still we have `originalContentType` for anything else. 

I think there's a big gap to be improved in ContentType handling on SCSt. If we believe that this should wait for a bigger refactor, I'm ok, but I don't see any reason why forcing an extra header in this case.

viniciusccarvalho: You know have opened the pandora box my friend :)

My humble opinion is that we are mixing a little bit of content negotiation and serialization on our current model. And perhaps it's what causes a bit of confusion and disconnection on our discussion

I think we will benefit from a f2f call, I'll send an invite to all members of the team that are interested in chiming in.

I have an opinion that we should only really send byte[], not even care if payload is of type String. If contentType proves to be a String representation (text/plain, application/csv) we could convert to String.class, otherwise most converters such as jackson are ok in reading byte[].

I am pretty sure my view is limited and will be proven wrong, but I'm glad to get educated on this.

garyrussell: I tend to agree that if the contentType is **any** json type, it should be retained (and that will avoid carrying an OCT header at all, since that's only added if it's different to the CT).

One problem I see with not setting the CT to `text/plain` for a `String` payload is the consumer might have a `@StreamListener` with parameter `Message<?>` and be expecting `Message<String>` but he'll get `Message<byte[]>`. But, as @mbogoevici has pointed out, we could document that in 2.0, if they want `Message<String>`, then the parameter should be `Message<String>`, not `Message<>` (or raw `Message`).

ilayaperumalg: yeah, @viniciusccarvalho and I had a discussion around this yesterday. Avoiding OCT header is the `eventual` benefit here. We would also want to bring this discussion again for 2.0 where we could potentially think through other options of handling `contentType` .

viniciusccarvalho:
We are moving this to 2.0 release and probably a broader discussion on contentType handling

Issue #1014, opened 2017-07-11
Improve testing on Confluent Schema Registry


viniciusccarvalho:
We currently offer no testing support for `ConfluentSchemaRegistryClient`, testing the server is an integration test, since it depends on kafka and zookeeper, so we can't easily embedded it.

One alternative is to use `Spring-Cloud-Contracts` and create mocks for a given version of the server, and make sure our tests pass against that revision.

Related: #985 #965

PR #1015, opened 2017-07-12
Fixes Confluent Schema Registry registration process error

### LINKED ISSUE #985: Unable to register schema in ConfluentSchemaRegistry
With the current implementation the register does not work since if a schema is not present it returns an HTTP 404. I have seen that the implementation has changed and now the service which is called is POST /subjects/%s instead of /subjects/%s/versions. 

Which means the schema is never registered.

viniciusccarvalho:
Fixes #985 #1014 

- Fixes Registration error
- Added tests using `MockRestServiceServer` following Confluent API 3.2.2

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/ConfluentSchemaRegistryClient.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/ConfluentSchemaRegistryClient.java
@@ -60,6 +60,38 @@ public class ConfluentSchemaRegistryClient implements SchemaRegistryClient {
 	@Override
 	public SchemaRegistrationResponse register(String subject, String format, String schema) {
 		Assert.isTrue("avro".equals(format), "Only Avro is supported");
+		String path = String.format("/subjects/%s/versions", subject);
+		HttpHeaders headers = new HttpHeaders();
+		headers.put("Accept",
+				Arrays.asList("application/vnd.schemaregistry.v1+json", "application/vnd.schemaregistry+json",
+						"application/json"));
+		headers.add("Content-Type", "application/json");
+		Integer version = null;
+		Integer id = null;
+		try {
+			String payload = this.mapper.writeValueAsString(Collections.singletonMap("schema", schema));
+			HttpEntity<String> request = new HttpEntity<>(payload, headers);
+			ResponseEntity<Map> response = this.template.exchange(this.endpoint + path, HttpMethod.POST, request,
+					Map.class);
+			id = (Integer) response.getBody().get("id");
+			version = getSubjectVersion(subject,schema);
+		}
+		catch (Exception e) {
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/ConfluentSchemaRegistryClient.java
```diff
@@ -49,59 +49,111 @@
...
+	public SchemaRegistrationResponse register(String subject, String format,
+			String schema) {
 		Assert.isTrue("avro".equals(format), "Only Avro is supported");
-		String path = String.format("/subjects/%s", subject);
+		String path = String.format("/subjects/%s/versions", subject);
 		HttpHeaders headers = new HttpHeaders();
-		headers.put("Accept",
-				Arrays.asList("application/vnd.schemaregistry.v1+json", "application/vnd.schemaregistry+json",
-						"application/json"));
+		headers.put("Accept", Arrays.asList("application/vnd.schemaregistry.v1+json",
```
ilayaperumalg: We could possibly move this list as constant. Can fix while merging.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/ConfluentSchemaRegistryClient.java
```diff
@@ -49,59 +49,111 @@
...
+				.setSchemaReference(new SchemaReference(subject, version, "avro"));
 		return schemaRegistrationResponse;
 	}
 
+	/**
+	 * Confluent register API returns the id, but we need the version of a given schema
+	 * subject. After a successful registration we can inquire the server to get the
+	 * version of a schema
+	 * @param subject
+	 * @return
```
ilayaperumalg: missing return type.

spring-cloud-stream-schema/src/test/java/org/springframework/cloud/schema/avro/client/ConfluentSchemaRegistryClientTests.java
```diff
@@ -0,0 +1,160 @@
...
+ */
+public class ConfluentSchemaRegistryClientTests {
+
+	private RestTemplate restTemplate;
+	private MockRestServiceServer mockRestServiceServer;
+
+	@Before
+	public void setup(){
+		this.restTemplate = new RestTemplate();
+		this.mockRestServiceServer = MockRestServiceServer.createServer(restTemplate);
```
ilayaperumalg: This is nice! :-)

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 661f8494a3c5c646edbbf891110c98d11711fa9e along with some minor refactoring b10a014273b9d7414a67d82dc9937908bbd99eaf

PR #1016, opened 2017-07-13
Improve support for disabling TestSupportBinderAutoConfiguration

### LINKED ISSUE #573: Test binder must also honour normal binder resolution when not autoconfigured
The typical usage for the test binder is to supersede the normal binder resolution process when autoconfiguration is enabled. However, when the test binder is not used in autoconfiguration mode, and resolved from the classpath, the `DefaultBinderFactory` cannot create the `Binder` instance, since the context does not provide a definition (it relies on the replacement of the DefaultBinderFactory).

Solution: when not in autoconfigured mode, provide the ability for the test binder to create a regular binder as well.

mbogoevici:
Fixes #573

- Split TestSupportBinderAutoConfiguration into separate configs
  for the binder, binderfactory and message collector.
  This enables to address the testBinder as a regular binder when
  autoconfiguration is disabled, and to ensure that the message
  collector is available when the autoconfiguration is disabled.
- reorganize tests to use the default autoconfiguration options
- add documentation for disabling test binder autoconfiguration


sobychacko:
LGTM - merging.

PR #1018, opened 2017-07-13
GH-1010: @StreamListener: Fix Header Propagation

### LINKED ISSUE #1010: @StreamListener with POJOs should propagate message headers
[StreamListenerMessageHandler](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerMessageHandler.java#L38) explicitly doesn't copy headers. I believe it should.

> Mark Fisher: the basic idea is that IF you are operating at the level of Message, you have opted to take over full control... obviously not the case for POJOs

garyrussell:
Fixes #1010

Propagate headers by default.

Add a `copyHeaders` property to `@StreamListener` to allow suppression of header propagation.

__cherry-pick to 1.2.x__

### DIFF
--- a/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerHandlerMethodTests.java
+++ b/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerHandlerMethodTests.java
@@ -63,6 +63,7 @@ import static org.springframework.cloud.stream.binding.StreamListenerErrorMessag
 /**
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
+ * @author Gary Russell
  */
 public class StreamListenerHandlerMethodTests {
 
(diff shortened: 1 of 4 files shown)


artembilan:
> cherry-pick to 1.2.x

> String copyHeaders() default "true";

This is breaking change in the point release...
Unexpected behavior for those who haven't care about headers before.

Any particular reason do not make it `false` by default for `1.2.x` ?

Thanks

garyrussell:
Well, it's really a bug; I originally was just going to make it `true` unconditionally, then added this property so people, who need to, can revert to the previous behavior; but, yes, it would require a coding change to do so.

Thoughts @dsyer @mbogoevici @viniciusccarvalho ??

dsyer:
I'd be happy to have a property that switched on the "correct" behaviour, keeping the old buggy behaviour as the default for 1.2.x. Then we can change the default in 1.3?

artembilan:
>  Then we can change the default in 1.3?

Right, that's my idea to avoid tears on the breaking changes in the point release.

I can do that during merge, cherry-picking if time permit.
Otherwise I have shared my fears and you know already what to do. 😄 

At least we are on the same side of boat 🤝

mbogoevici:
I agree with @dsyer and @artembilan that we shouldn't change behaviour in 1.2.x.

mbogoevici:
Apart from the change around the defaults that we may want to include for 1.2.x, the implementation must honour header propagation settings (e.g. ignore `contentType`, which is very relevant for `@StreamListener`) - this doesn't happen automatically right now, as `StreamListenerMessageHandler`s are not installed as beans - added this commit https://github.com/mbogoevici/spring-cloud-stream/commit/154d0dde561a4056d12292f2f3b89eb08df00328 to align header propagation  - @garyrussell let me know if you're OK with the change and I can merge.

garyrussell:
Agreed; thanks.

mbogoevici:
Cherry-picking requires also backporting  https://github.com/spring-cloud/spring-cloud-stream/commit/51deb128f0250ed0212bd7ab5cc2ba723a4bb1e3 (as otherwise the `contentType` header gets overwritten if the `copyHeaders` is set to true).

mbogoevici:
Merged in master with indicated changes via: 205d457

Cherry-picked in 1.2 with changes to preserve the current default behaviour: e386d11

PR #1023, opened 2017-07-18
Add @ErrorListener to consumer error messages

### LINKED ISSUE #914: Introduce `@ErrorListener` for receiving errors sent to error channel
Note: this will be independent of what gets published on the error channel (i.e. #913).

ilayaperumalg:
- Add @ErrorListener to consumer error messages from global Spring Integration `errorChannel`
 - Add ErrorListener annotation bean post processor that maps the listener method to subscribe to the errorChannel
 - Reuse some existing util method from StreamListener annotation bean post processor
 - Add tests
 - Update docs

Resolves #914

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -433,6 +433,42 @@ at the broker by configuring a binding for the outbound target named `error`. Fo
 publish error messages to a broker destination named "myErrors", provide the following property:
 `spring.cloud.stream.bindings.error.destination=myErrors`
 
+===== Using @ErrorListener for handling error messages
+
+The error messages sent to the global Spring Integration `errorChannel` can be consumed via explicit methods annotated with `@ErrorListener`.
+Since the `errorChannel` is a pub-sub message channel, there can be multiple `@ErrorListener` methods consuming the error messages and process them based on the requirements.
+
+[NOTE]
+====
+
+Since the error messages received from the `errorChannel` is always of the type `ErrorMessage`, the `@ErrorListener` annotated method should have parameter argument type `ErrorMessage`.
+
+[source,java]
+----
+
+@EnableBinding(Sink.class)
+public class MySink {
+
+	@Autowired
+    VotingService votingService;
+
+    @StreamListener(Sink.INPUT)
+    public void handle(Vote vote) {
+      votingService.record(vote);
+    }
+
+	@ErrorListener
+	public void errorHandler1(ErrorMessage errorMessage) {
+	  // do something with the errorMessage
+	}
+
+	@ErrorListener
+	public void errorHandler1(ErrorMessage errorMessage) {
+	  // do something with the errorMessage
+	}
+----
+====
+
 ===== Using @StreamListener for Automatic Content Type Handling
 
 Complementary to its Spring Integration support, Spring Cloud Stream provides its own `@StreamListener` annotation, modeled after other Spring Messaging annotations (e.g. `@MessageMapping`, `@JmsListener`, `@RabbitListener`, etc.).
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/ErrorListener.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/ErrorListener.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.messaging.handler.annotation.MessageMapping;
+import org.springframework.messaging.support.ErrorMessage;
+
+/**
+ * Annotation that marks a method to consume {@link ErrorMessage}s from global Spring Integration errorChannel.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@MessageMapping
+@Documented
+public @interface ErrorListener {
+}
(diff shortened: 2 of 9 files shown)


artembilan:
Looks like this PR has conflicts with `master`.
Should be rebased and fixed.

Thanks

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -433,6 +433,42 @@ at the broker by configuring a binding for the outbound target named `error`. Fo
 publish error messages to a broker destination named "myErrors", provide the following property:
 `spring.cloud.stream.bindings.error.destination=myErrors`
 
+===== Using @ErrorListener for handling error messages
+
+The error messages sent to the global Spring Integration `errorChannel` can be consumed via explicit methods annotated with `@ErrorListener`.
```
artembilan: with the `@ErrorListener`

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -433,6 +433,42 @@ at the broker by configuring a binding for the outbound target named `error`. Fo
...
+Since the `errorChannel` is a pub-sub message channel, there can be multiple `@ErrorListener` methods consuming the error messages and process them based on the requirements.
```
artembilan: Since that was `consuming` it is my feeling that it should be `processing` 😄

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -433,6 +433,42 @@ at the broker by configuring a binding for the outbound target named `error`. Fo
...
+
+[NOTE]
+====
+
+Since the error messages received from the `errorChannel` is always of the type `ErrorMessage`, the `@ErrorListener` annotated method should have parameter argument type `ErrorMessage`.
```
artembilan: `are always` . Because that was `messages` - plural.

Why is it always `ErrorMessage`? Why can't I rely on the `Exception` `payload`? Or be more strict and expect `MessagingException` to get an access to the `failedMessage` without any casts.

Thanks

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -433,6 +433,42 @@ at the broker by configuring a binding for the outbound target named `error`. Fo
...
+
+	@ErrorListener
+	public void errorHandler1(ErrorMessage errorMessage) {
+	  // do something with the errorMessage
+	}
+
+	@ErrorListener
+	public void errorHandler1(ErrorMessage errorMessage) {
+	  // do something with the errorMessage
+	}
```
artembilan: We should use only space indents in the Docs.
You can see here from GH how it is broken with the mix with tabs.
Also you have missed closing `}` for the class.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/ErrorListener.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Annotation that marks a method to consume {@link ErrorMessage}s from global Spring Integration errorChannel.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Target({ ElementType.METHOD, ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@MessageMapping
+@Documented
+public @interface ErrorListener {
```
artembilan: I think this must be supplied with the `value` for any possible error channel. Even if with the `errorChannel` value by default.
What I mean is here in the `AbstractMessageChannelBinder`:
```
protected final ErrorInfrastructure registerErrorInfrastructure(ConsumerDestination destination, String group,
			C consumerProperties) {
	ErrorMessageStrategy errorMessageStrategy = getErrorMessageStrategy();
	ConfigurableListableBeanFactory beanFactory = getApplicationContext().getBeanFactory();
	String errorChannelName = errorsBaseName(destination, group, consumerProperties);
```

I'm pretty sure that end-user might want to be able to use this `@ErrorListener` for those channels as well.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/StreamListenerMethodUtils.java
```diff
@@ -145,4 +148,37 @@ protected static String getOutboundBindingTargetName(Method method) {
 		}
 		return null;
 	}
+
+	protected static Method checkProxy(Method methodArg, Object bean) {
+		Method method = methodArg;
+		if (AopUtils.isJdkDynamicProxy(bean)) {
+			try {
+				// Found a @StreamListener method on the target class for this JDK proxy
```
artembilan: This utility method is not relevant to the ` @StreamListener` any more.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/ErrorListenerAnnotationBeanPostProcessor.java
```diff
@@ -0,0 +1,97 @@
...
+			}
+		});
+		return bean;
+	}
+
+	protected final void registerHandlerMethodOnListenedChannel(Method method, ErrorListener errorListener,
+			Object bean) {
+		Method targetMethod = StreamListenerMethodUtils.checkProxy(method, bean);
+		final InvocableHandlerMethod invocableHandlerMethod = this.messageHandlerMethodFactory
+				.createInvocableHandlerMethod(bean, targetMethod);
```
artembilan: Well, since you rely on the `messageHandlerMethodFactory`, we really can deal just with the exception `payload` and even `@Header` and some other tricks. See `DefaultMessageHandlerMethodFactory.initArgumentResolvers()`.

artembilan:
>  I believe we should follow a different implementation strategy for this, namely to meta-annotate `@ErrorListener` with `@StreamListener`

Hm... Thank you, @mbogoevici , for heads up! Right now I even think that we don't need this `@ErrorListener` at all since the regular `@StreamListener` (or any other SI Messaging annotation) can simply do that for us. Only what we need is the name of the error channel and proper method arguments for `ErrorMessage` or `Exception`. 
This issue should just fall into the Documentation - "How to handle exception in the SCSt apps?".

I won't mind to see some arguments otherwise to still introduce this `@ErrorListener` .

Thanks

ilayaperumalg:
@mbogoevici @artembilan Great ideas! 

We could still have `@ErrorListener` as a meta annotation so that it would be a convenient option from the usability perspective. What do you think?

artembilan:
>  What do you think?

I think we don't need anything unless document how is the `@StreamListener` cool feature that it even can be used for error-handling:
```
@StreamListener("errorChannel")
public void handleError(MessagingException e) {
    ...
}
```
And that's all.

What ma I missing here, please?

viniciusccarvalho:
I actually like the meta-annotation idea. Indeed it does not bring much to the table, but it does help on reading the code IMHO. When I look at it, it jump to my eyes that it's an error handler method.

artembilan:
Spring Integration allows to create custom annotations with `@ServiceActivator` and his brothers, but we don't provide any out-of-the-box.

This `@ErrorListener` approach with meta-annotation smells like a noise without any ground.
I will be able to configure it for the `Source.OUTPUT` or any other non-error channel. With that the signature of the method really can be not an `Exception`. So, all that readability is eliminated by the side-effect of the meta-annotation.

That's why the `errorChannel`  in SI is treated as equal to any other channel and you can simply declared any handler to subscriber that channel. Or, OTOH, you can always use any other channel for sending `ErrorMessage`s.

That is IMHO. 
If you decide to go with this non-important addition, I will be silent.

viniciusccarvalho:
@ilayaperumalg We've discussed this earlier and we will move this and the GenericError handling to a 2.0 release. For now 1.3 will stick with just making sure we collect error on the appropriate message channels, users should then be responsible to subscribe to them in order to consume.

PR #1029, opened 2017-07-25
Adding custom interceptors during error channel binding

### LINKED ISSUE #1028: Error channel binding is not adding any custom interceptors.
When we set content-type on error channel destinations, it's not taking any effect and that's due to the fact that we are not adding any interceptors for error channel binding. 

```
spring.cloud.stream.bindings:
  error:
    destination: error-channel
    content-type: application/*+avro
```

sobychacko:
Adding unit test to verify message conversion for error channel

Fix #1028 

Needs cherry-picking to the 2.0 branch.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
@@ -54,12 +54,15 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.DependsOn;
 import org.springframework.expression.PropertyAccessor;
+import org.springframework.integration.channel.DirectChannel;
 import org.springframework.integration.channel.PublishSubscribeChannel;
 import org.springframework.integration.config.IntegrationEvaluationContextFactoryBean;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
+import org.springframework.integration.handler.BridgeHandler;
 import org.springframework.integration.json.JsonPropertyAccessor;
 import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.SubscribableChannel;
 import org.springframework.messaging.converter.MessageConverter;
 import org.springframework.messaging.core.DestinationResolutionException;
 import org.springframework.messaging.core.DestinationResolver;
@@ -83,7 +86,9 @@ public class BindingServiceConfiguration {
 
 	public static final String STREAM_LISTENER_ANNOTATION_BEAN_POST_PROCESSOR_NAME = "streamListenerAnnotationBeanPostProcessor";
 
-	private static final String ERROR_CHANNEL_NAME = "error";
+	public static final String ERROR_BRIDGE_CHANNEL = "errorBridgeChannel";
+
+	private static final String ERROR_NAME = "error";
 
 	@Autowired(required = false)
 	private ObjectMapper objectMapper;
@@ -164,12 +169,23 @@ public class BindingServiceConfiguration {
 	}
 
 	@Bean
-	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_CHANNEL_NAME + ".destination")
-	public SingleBindingTargetBindable<MessageChannel> errorChannelBindable(
-			@Qualifier(IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME) PublishSubscribeChannel errorChannel,
+	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_NAME + ".destination")
+	public MessageChannel errorBridgeChannel(
+			@Qualifier(IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME) PublishSubscribeChannel errorChannel) {
+		SubscribableChannel errorBridgeChannel = new DirectChannel();
+		BridgeHandler handler = new BridgeHandler();
+		handler.setOutputChannel(errorBridgeChannel);
+		errorChannel.subscribe(handler);
+		return errorBridgeChannel;
+	}
+
+	@Bean
+	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_NAME + ".destination")
+	public SingleBindingTargetBindable<MessageChannel> errorBridgeChannelBindable(
+			@Qualifier(ERROR_BRIDGE_CHANNEL) MessageChannel errorBridgeChannel,
 			CompositeMessageChannelConfigurer compositeMessageChannelConfigurer) {
-		compositeMessageChannelConfigurer.configureOutputChannel(errorChannel, ERROR_CHANNEL_NAME);
-		return new SingleBindingTargetBindable<MessageChannel>(ERROR_CHANNEL_NAME, errorChannel);
+		compositeMessageChannelConfigurer.configureOutputChannel(errorBridgeChannel, ERROR_NAME);
+		return new SingleBindingTargetBindable<>(ERROR_NAME, errorBridgeChannel);
 	}
 
 	@Bean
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
```diff
@@ -166,7 +166,9 @@ public BinderAwareChannelResolver binderAwareChannelResolver(BindingService bind
 	@Bean
 	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_CHANNEL_NAME + ".destination")
 	public SingleBindingTargetBindable<MessageChannel> errorChannelBindable(
-			@Qualifier(IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME) PublishSubscribeChannel errorChannel) {
+			@Qualifier(IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME) PublishSubscribeChannel errorChannel,
+			CompositeMessageChannelConfigurer compositeMessageChannelConfigurer) {
+		compositeMessageChannelConfigurer.configureOutputChannel(errorChannel, ERROR_CHANNEL_NAME);
```
mbogoevici: Instead of directly manipulating the error channel, it would be better to bridge it to a secondary channel and actually bind the latter. Applying the interceptors directly on the error channel means that all its recipients will start receiving marshaled objects, and so on.

artembilan: Well, what @sobychacko have done here is fully logical from the binding perspective:
```
@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_CHANNEL_NAME + ".destination")
```
what means that end-user expect any out-of-the-box conversion applied exactly for this channel.

But on the other hand I agree that it is very dangerous to apply it unconditionally because `errorChannel` is `PublishSubscribe` by default and has that famous `LoggingHandler` as one of the default subscriber.

I believe what @mbogoevici want to see here is like `BridgeHandler` as a subscriber to the `errorChannel` and a `new DirectChannel()` to bridge to and already only this new channel return for the target binding. And yes, apply any appropriate configurations, including conversion interceptor.

mbogoevici: @artembilan What @sobychacko did is in principle correct but you've captured pretty well the motivation (avoiding to change the default behaviour of `errorChannel`) and solution (bridge) that are implied by my comment.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
```diff
@@ -164,12 +169,23 @@ public BinderAwareChannelResolver binderAwareChannelResolver(BindingService bind
 	}
 
 	@Bean
-	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_CHANNEL_NAME + ".destination")
-	public SingleBindingTargetBindable<MessageChannel> errorChannelBindable(
-			@Qualifier(IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME) PublishSubscribeChannel errorChannel,
+	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_NAME + ".destination")
```
mbogoevici: Do we need to expose this channel as a bean? It feels like we can do this inside `errorBridgeChannelBindable`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
```diff
@@ -164,10 +169,23 @@ public BinderAwareChannelResolver binderAwareChannelResolver(BindingService bind
 	}
 
 	@Bean
-	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_CHANNEL_NAME + ".destination")
-	public SingleBindingTargetBindable<MessageChannel> errorChannelBindable(
+	@ConditionalOnProperty("spring.cloud.stream.bindings." + ERROR_KEY_NAME + ".destination")
+	public MessageChannel errorBridgeChannel(
 			@Qualifier(IntegrationContextUtils.ERROR_CHANNEL_BEAN_NAME) PublishSubscribeChannel errorChannel) {
```
artembilan: There is no guarantee that end-user doesn't override global `errorChannel` with its own implementation.
We may insist to restrict it into the `SubscribeChannel` to simplify our life, but that really should not be just `PublishSubscribeChannel`.

We may reconsider this in the version `2.0` when SI 5.0 is available with its `FluxMessageChannel` though

artembilan:
Looking into this task after @ilayaperumalg 's `@ErrorListener` #1023 , I don't understand why should we care about content type in the error channel.

We send there `ErrorMessage` which payload can be only `Throwable`. 
So, what are you going to convert there? 
Maximum what we can do is serialization (raw or Krio). Well, yes, there is always `toString()` 😄 .

So, what is the issue from the first place, please?

Thanks

sobychacko:
@abilan - this PR originated from some use case requirements. If someone has the following properties set and they want the error messages to go through the content type conversion, currently it's not possible. 
```
spring.cloud.stream.bindings:
  error:
    destination: error-topic
    content-type: application/*+avro
```
Doesn't @ilayaperumalg 's PR only sends `ErrorMessage` if the user decides to use `@ErrorListener`?

artembilan:
OK. Thanks @sobychacko !

But how that `application/*+avro` can convert the `Exception` payload, please?

@ilayaperumalg 's PR doesn't send any messages. It is about subscribing to the error channel and handle exception from there.

Having this problem with conversion one more time, I'd again raise my old concern that it must not be done from the `ChannelInterceptor`, but in the target particular handler. 
At least SI 5.0 does that exactly on the method level...
Or I don't understand something here again...

Thanks

mbogoevici:
@artembilan This is an outbound channel, so the user wants to serialize the outbound exception via Avro and send it to the target messaging system. This is completely different than #1023 where we want to handle the exception. The `ChannelInterceptor` is the standard way of doing that in 1.x - in 2.x that can be something different.

artembilan:
Thank you, @mbogoevici  for the pointer!

Well, is it really possible?
More over as @sobychacko is showing me right now we have this code to reproduce:
```
 return IntegrationFlows
                .from(input)
                .transform(transformer)
                .channel(errorChannel)
                .get();
```
They send the result of the transformation directly to the `errorChannel`!!!
Where is the exception here though?

I noticed that reading their comments:
```
 while on the error-person topic, the content type is application/x-java-object;type=org.apache.avro.generic.GenericData$Record.
```
But we really expect there exception...

If they are going to send normal payloads and support auto-conversion, that would be better to have regular binding, but not `spring.cloud.stream.bindings.error`...

That's my feeling.
Sorry if I'm ruing everything here because I'm so ignorant in the subject.

artembilan:
Well, on the other hand, after debating with my second alter ego, I would leave your solution (I merger it) because if that is really possible to send anything to the message channel, that is already target application responsibility to deal with the payload and if one of the expectation is really to have out-of-the-box conversion, let it be. But right, the way you do it here - via bridged channel, very close to the binder.

Only the problem how is it going to work if there is really going to be an `exception` in the payload.
Do you have a test-case on the matter?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
sobychacko: @mbogoevici it might be useful to have this exposed as a bean for testing purposes, as we can then subscribe to the bridged channel directly and make assertions.

sobychacko:
@artembilan I added tests for the `exception` payload in the error channel. If the content type is set for the error destination and if the `exception` can be converted (for ex - using the `json` content-type), then it should work. However, it is up to the user in that case, how to handle that, i.e. the user knows what kind of data is expected and if it is an exception how it should be handled etc.

artembilan:
Great!

I'll take a look in the evening or on the weekend.
And maybe even merge.

Have a good weekend any way!

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/22141a8dce93f32d10825ba9fb0cb4d5e29fcb95

PR #1031, opened 2017-07-25
Extracting payload serde code from AbstractBinder into a utility class

### LINKED ISSUE #1030: Payload serialization/deserialization code extraction
Extract code used for payload serde in `AbstractBinder` into a utility class so that other components that are not part of the `AbstractBinder` hierarchy can reuse them.

sobychacko:
Fix #1030

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -16,9 +16,6 @@
 
 package org.springframework.cloud.stream.binder;
 
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -66,8 +63,6 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 	private volatile Codec codec;
 
 	private volatile EvaluationContext evaluationContext;
-
-	private volatile Map<String, Class<?>> payloadTypeCache = new ConcurrentHashMap<>();
 
 	/**
 	 * For binder implementations that support a prefix, apply the prefix to the name.
@@ -165,13 +160,11 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 	}
 
 	protected final MessageValues deserializePayloadIfNecessary(Message<?> message) {
-		return MessageSerializationUtils.deserializePayload(new MessageValues(message), this.contentTypeResolver,
-				this.payloadTypeCache, this.codec);
+		return MessageSerializationUtils.deserializePayload(new MessageValues(message), this.contentTypeResolver, this.codec);
 	}
 
 	protected final MessageValues deserializePayloadIfNecessary(MessageValues messageValues) {
-		return MessageSerializationUtils.deserializePayload(messageValues, this.contentTypeResolver,
-				this.payloadTypeCache, this.codec);
+		return MessageSerializationUtils.deserializePayload(messageValues, this.contentTypeResolver, this.codec);
 	}
 
 	protected String buildPartitionRoutingExpression(String expressionRoot) {
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -16,12 +16,8 @@
 
 package org.springframework.cloud.stream.binder;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -32,20 +28,14 @@ import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.support.AbstractApplicationContext;
-import org.springframework.core.serializer.support.SerializationFailedException;
 import org.springframework.expression.EvaluationContext;
 import org.springframework.integration.codec.Codec;
 import org.springframework.integration.expression.ExpressionUtils;
 import org.springframework.messaging.Message;
-import org.springframework.messaging.MessageHeaders;
 import org.springframework.retry.backoff.ExponentialBackOffPolicy;
 import org.springframework.retry.policy.SimpleRetryPolicy;
 import org.springframework.retry.support.RetryTemplate;
 import org.springframework.util.Assert;
-import org.springframework.util.ClassUtils;
-import org.springframework.util.MimeType;
-import org.springframework.util.MimeTypeUtils;
-import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PayloadSerdeUtils.java
```diff
@@ -0,0 +1,230 @@
+/*
+ * Copyright 2016 the original author or authors.
```
mbogoevici: 2017

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PayloadSerdeUtils.java
```diff
@@ -0,0 +1,230 @@
...
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
+import org.springframework.util.ObjectUtils;
+
+/**
+ * Utility class for serializing and de-serializing the message payload.
+ *
+ * @author Soby Chacko
+ */
+public abstract class PayloadSerdeUtils {
```
mbogoevici: Can we call this something different? E.g. `MessageSerializationUtils` (we don't need to reference deserialization in the class name and it is more consistent to e.g. `org.springframework.util.SerializationUtils`).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PayloadSerdeUtils.java
```diff
@@ -0,0 +1,230 @@
...
+	protected static final Log LOGGER = LogFactory.getLog(PayloadSerdeUtils.class);
+
+	/**
+	 * Serialize the message payload unless it is a byte array.
+	 *
+	 * @param message the message with the payload to serialize
+	 * @param codec the codec used for serialization
+	 * @return the Message with teh serialized payload
+	 */
+	public static MessageValues serializePayloadIfNecessary(Message<?> message, Codec codec) {
```
mbogoevici: Can we drop the `ifNecessary` here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PayloadSerdeUtils.java
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+
+	/**
+	 * Handles representing any java class as a {@link MimeType}.
+	 *
+	 * @author David Turanski
+	 * @author Ilayaperumal Gopinathan
+	 */
+	public abstract static class JavaClassMimeTypeConversion {
```
mbogoevici: If the class needs to be publicly accessible, then its static methods need to be accessible as well. I would move this to top level to avoid `PayloadSerdeUtils.JavaClassMimeTypeConversion.mimeTypeFromObject()` and I would change its name to something else - `JavaClassMimeTypeUtils` or so.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageSerializationUtils.java
```diff
@@ -0,0 +1,171 @@
...
+				if (targetType == null) {
+					targetType = ClassUtils.forName(className, null);
+					payloadTypeCache.put(className, targetType);
+				}
+				return codec.decode(bytes, targetType);
+			} // catch all exceptions that could occur during de-serialization
+			catch (Exception e) {
+				String errorMessage = "Unable to deserialize [" + className + "] using the contentType [" + contentType
+						+ "] " + e.getMessage();
+				MessageSerializationUtils.LOGGER.error(errorMessage);
```
artembilan: I wouldn't log error if I throw it.
That is already this method user to decide what to do with the exception, but not utility

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageSerializationUtils.java
```diff
@@ -0,0 +1,171 @@
...
+			try {
+				if (originalPayload instanceof String) {
+					return ((String) originalPayload).getBytes("UTF-8");
+				}
+				codec.encode(originalPayload, bos);
+				return bos.toByteArray();
+			}
+			catch (IOException e) {
+				throw new SerializationFailedException(
+						"unable to serialize payload [" + originalPayload.getClass().getName() + "]", e);
```
artembilan: Would be better to mention in the JavaDoc what is the exception this method is going to throw and why

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/JavaClassMimeTypeUtils.java
```diff
@@ -0,0 +1,94 @@
...
+	 * Retrieve the class name from {@link MimeType}.
+	 *
+	 * @param mimeType {@link MimeType} to retrieve class name from
+	 * @return class name from mime type
+	 */
+	public static String classNameFromMimeType(MimeType mimeType) {
+		Assert.notNull(mimeType, "mimeType cannot be null.");
+		String className = mimeType.getParameter("type");
+		if (className == null) {
+			return null;
```
artembilan: Need to say in JavaDocs when it returns `null`. I was confused to think that I send to this method `text/plain` and expect some class from there. Maybe it is so smart that build a `String` class for me back?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -166,107 +157,21 @@ protected final String groupedName(String name, String group) {
...
-						"unable to serialize payload [" + originalPayload.getClass().getName() + "]", e);
-			}
-		}
+		return MessageSerializationUtils.serializePayload(originalPayload, this.codec);
 	}
 
 	protected final MessageValues deserializePayloadIfNecessary(Message<?> message) {
-		return deserializePayloadIfNecessary(new MessageValues(message));
+		return MessageSerializationUtils.deserializePayload(new MessageValues(message), this.contentTypeResolver,
+				this.payloadTypeCache, this.codec);
```
artembilan: This `payloadTypeCache` must go to the `MessageSerializationUtils` directly.
I see you have something similar in the `JavaClassMimeTypeUtils` for the `mimeTypesCache`. 
So, why don't be consistent?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/JavaClassMimeTypeUtils.java
```diff
@@ -74,7 +74,7 @@ public static MimeType mimeTypeFromObject(Object payload, String originalContent
 	 * Retrieve the class name from {@link MimeType}.
 	 *
 	 * @param mimeType {@link MimeType} to retrieve class name from
-	 * @return class name from mime type
+	 * @return class name from MimeType and null if the class name cannot be determined from the MimeType
```
artembilan: But you determine it from there only if there is `type` parameter.
Why don't mention that in JavaDocs?
This is public useful API, but a bit confusing without details.

Just IMHO though.

Feel free to leave as is.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageSerializationUtils.java
```diff
@@ -41,7 +39,7 @@
  */
 public abstract class MessageSerializationUtils {
 
-	protected static final Log LOGGER = LogFactory.getLog(MessageSerializationUtils.class);
+	private static Map<String, Class<?>> payloadTypeCache = new ConcurrentHashMap<>();
```
artembilan: `final` is missed.

mbogoevici:
Squashed and merged as https://github.com/spring-cloud/spring-cloud-stream/commit/6f1bc59ee232e72f22aa635fc2c4c7bbaf0e19b9

PR #1037, opened 2017-08-09
Use Checkstyle configs from src directly

artembilan:
Since we rely on the Checkstyle configs in the `spring-cloud-stream-tools`
module, we really need its artifact if we would like to use it as a
dependency for the `maven-checkstyle-plugin`.
But at the same time it is possible only if we run `package` phase.
During `compile/test` phase we don't have artifact yet and end up with
the error that no property set for Checkstyle.

* Use Checkstyle config files directly for the `maven-checkstyle-plugin`
properties.
This way we don't need any artifact to build in advance

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -159,21 +159,14 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


viniciusccarvalho:
Merged via 51e71610561e09a4006017df613851617a667b71

Issue #916, opened 2017-04-11
Catch exceptions thrown when publishing via `ProducerHandler`


mbogoevici:
- wherever the underlying messaging client supports asynchronous notifications - or support returns and confirms for Rabbit
- might be also invocable synchronously from the producer?

PR #1038, opened 2017-08-10
Generic error handling

### LINKED ISSUE #1036: Extract Error Handling strategy to be agnostic of binder implementation
Change the error handling strategy to be agnostic of binder implementation. Currently only supported via `MessageChannel` binders.

viniciusccarvalho:
This PR fixes #1036 

A great amount of changes happened here, I'll try to summarize some of them in order to make reviewing easier.

General idea was to pull any code for ErrorInfrastructure from `AbstractMessageChannelBinder` to become available on `AbstractBinder`. 

In order to achieve some of that, some core Interfaces needed a boost:

 - `Binding` -> A new lifecycle `bind()` method was added, so that `BindingService` could trigger any error handling configuration at the right time. The RabbitBinder had issues if we triggered the lifecycle of MessageProducers too early. A new `getTarget` method was also provided to allow implementations to fetch the target binding in case they needed to configure it.  As a side note, this is probably hinting that `Binding` should be a full blown Spring Bean, where we would allow the container to control it's lifecycle as well provide way for users to add interceptors and other stuff. Should be something to consider on 2.0 baseline

- Several `Binding` implementations. `DefaultBinding` was given a boost, and added a generic type `D` used by implementations to pass along `ConsumerDestination` or `ProducerDestination` both which are required by the error infrastructure.

- `ProvisioningProvider` - The `ConsumerDestination` and `ProducerDestination` now hold a reference to the relevant `ConsumerProperties` and `ProducerProperties`. The creation method for those types already had that reference and since the error configurer needed those, this was the least invasive way to provide : `defaultBinding.getDestination().getProperties()`

- Several test cases were updated, since we now rely on lifecycle `bind()` called via `BindingService` and those tests create binders directly, the invocation for this method was added there.

- `BinderErrorConfigurer` : a new interface that Binder implementations should provide, the `AbstractBinder` has hooks to delegate to the binder implementation of this interface. An abstract version of MessageChannel types is provided : `AbstractMessageChannelErrorConfigurer` much like the `AbstractMessageChannelBinder`

- `AbstractBinder` : Added a few lifecycle methods, and now requires an `BinderErrorConfigurer` type

- All Error types are now under `binder.error` package

The rabbit binder is complete, and I'm starting to work on the kafka binder now, I realize that this could trigger changes on the proposed changes on this PR, but given the size of this PR I'd rather have people providing feedback sooner rather than later.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -44,6 +44,7 @@ import org.springframework.util.StringUtils;
  * @author Mark Fisher
  * @author Marius Bogoevici
  * @author Soby Chacko
+ * @author Vinicius Carvalho
  */
 public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends ProducerProperties>
 		implements ApplicationContextAware, InitializingBean, Binder<T, C, P> {
@@ -63,6 +64,12 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 	private volatile Codec codec;
 
 	private volatile EvaluationContext evaluationContext;
+
+	protected final BinderErrorConfigurer errorConfigurer;
+
+	protected AbstractBinder(BinderErrorConfigurer errorConfigurer) {
+		this.errorConfigurer = errorConfigurer;
+	}
 
 	/**
 	 * For binder implementations that support a prefix, apply the prefix to the name.
@@ -128,7 +135,10 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 		if (StringUtils.isEmpty(group)) {
 			Assert.isTrue(!properties.isPartitioned(), "A consumer group is required for a partitioned subscription");
 		}
-		return doBindConsumer(name, group, target, properties);
+		Binding<T> binding = doBindConsumer(name, group, target, properties);
+		registerErrorInfrastructure(name,group,properties);
+		configureErrorInfrastructure(name,binding.getTarget());
+		return binding;
 	}
 
 	protected abstract Binding<T> doBindConsumer(String name, String group, T inputTarget, C properties);
@@ -169,6 +179,18 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 
 	protected String buildPartitionRoutingExpression(String expressionRoot) {
 		return "'" + expressionRoot + "-' + headers['" + BinderHeaders.PARTITION_HEADER + "']";
+	}
+
+	protected void registerErrorInfrastructure(String destination, String group, C consumerProperties){
+		this.errorConfigurer.register(destination,group,consumerProperties);
+	}
+
+	protected void destroyErrorInfrastructure(String destination, String group, C consumerProperties){
+		this.errorConfigurer.destroy(destination,group,consumerProperties);
+	}
+
+	protected void configureErrorInfrastructure(String destination, T target){
+		this.errorConfigurer.configure(destination,target);
 	}
 
 	/**
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -137,7 +137,7 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 		}
 		Binding<T> binding = doBindConsumer(name, group, target, properties);
 		registerErrorInfrastructure(name,group,properties);
-		configureErrorInfrastructure(name,binding.getTarget());
+		configureErrorInfrastructure(name,binding);
 		return binding;
 	}
 
@@ -189,15 +189,16 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 		this.errorConfigurer.destroy(destination,group,consumerProperties);
 	}
 
-	protected void configureErrorInfrastructure(String destination, T target){
-		this.errorConfigurer.configure(destination,target);
+	protected void configureErrorInfrastructure(String destination, Binding<T> binding){
+		this.errorConfigurer.configure(destination,binding);
 	}
 
 	/**
 	 * Create and configure a retry template.
-	 *
+	 * TODO: Perhaps we should deprecate this with BinderErrorConfigurer
 	 * @param properties The properties.
 	 * @return The retry template
+	 *
 	 */
 	public RetryTemplate buildRetryTemplate(ConsumerProperties properties) {
 		RetryTemplate template = new RetryTemplate();
@@ -211,4 +212,6 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 		template.setBackOffPolicy(backOffPolicy);
 		return template;
 	}
+
+
 }
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -22,6 +22,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.cloud.stream.error.BinderErrorConfigurer;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.support.AbstractApplicationContext;
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -137,7 +137,7 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 			Assert.isTrue(!properties.isPartitioned(), "A consumer group is required for a partitioned subscription");
 		}
 		Binding<T> binding = doBindConsumer(name, group, target, properties);
-		registerErrorInfrastructure(binding,group,properties);
+		registerErrorInfrastructure(binding, group, properties);
 		configureErrorInfrastructure(binding);
 		return binding;
 	}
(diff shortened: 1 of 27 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -128,7 +136,10 @@ protected void onInit() throws Exception {
 		if (StringUtils.isEmpty(group)) {
 			Assert.isTrue(!properties.isPartitioned(), "A consumer group is required for a partitioned subscription");
 		}
-		return doBindConsumer(name, group, target, properties);
+		Binding<T> binding = doBindConsumer(name, group, target, properties);
+		registerErrorInfrastructure(binding,group,properties);
```
viniciusccarvalho: Can you elaborate? our checkstyle rules did not fire for anything specific here.

artembilan: I know, that's why I am concerned. 
You see you don't have whitespaces after commas.
That looks a bit clumsy and hard to review, believe me...

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -37,4 +39,17 @@
...
+	 * lifecycle method on the represented target. Implementations must be idempotent.
+	 * This method is called after the concrete {@link Binder} implementation returns from
+	 * its create binding methods
+	 */
+	void bind();
+
+	/**
+	 * @return the binding target instance
+	 */
+	T getTarget();
```
ilayaperumalg: Could you elaborate the reasoning behind having this as part of the `Binding` contract itself? With this PR, I don't see `getTarget` used in a way (unless I miss something). Perhaps we need it in some other cases.

I am wondering if we can move this into `DefaultBinding` and not having this as part of the `Binding` interface?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -224,18 +222,14 @@ protected void afterUnbindProducer(ProducerDestination destination, P producerPr
 			bridge.setBeanName("bridge." + name);
 			consumerEndpoint = createConsumerEndpoint(destination, group, properties);
 			consumerEndpoint.setOutputChannel(bridge);
-			if (consumerEndpoint instanceof InitializingBean) {
```
ilayaperumalg: I'd like to understand the advantage of moving the lifecyle operations into `afterBind()`. I believe they, in fact, belong to `bind()` which already happens here.  

This makes me ask this question: Why do we need `bind()` while the actual binding is expected to happen during `doBindProducer` or `doBindConsumer`. I think we wanted to have `unbind()` to directly invoke the lifecycle `stop` operations and in the case of `bind()` we may not need an explicit method to do that as the binding operation is done  `doBindProducer` or `doBindConsumer`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -37,4 +39,17 @@
...
 	 */
 	void unbind();
+
+	/**
+	 * Binds the target component represented by this instance. Should be used to call any
+	 * lifecycle method on the represented target. Implementations must be idempotent.
+	 * This method is called after the concrete {@link Binder} implementation returns from
+	 * its create binding methods
+	 */
+	void bind();
```
ilayaperumalg: Please see my comment related to the introduction of this method [here](https://github.com/spring-cloud/spring-cloud-stream/pull/1038/files#r132620289)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -64,6 +66,12 @@
 
 	private volatile EvaluationContext evaluationContext;
 
+	protected final BinderErrorConfigurer errorConfigurer;
+
+	protected AbstractBinder(BinderErrorConfigurer errorConfigurer) {
+		this.errorConfigurer = errorConfigurer;
```
ilayaperumalg: Do you think we can probably make `BinderErrorConfigurer` into setter level injection rather than constructor level injection? 

May be that could help these changes less invasive in terms of creating the Binder instance and also, the implementations wouldn't have to worry about setting this to `null` explicitly.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/error/BinderErrorConfigurer.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.cloud.stream.binder.ConsumerProperties;
+
+/**
+ * Binder implementations should implement their Error Handling strategy using sub classes of this interface.
+ * The {@link AbstractBinder} uses lifecycle hooks to call each of the methods describe in this interface.
+ * @author Vinicius Carvalho
+ *
+ * @since 1.3
+ */
+public interface BinderErrorConfigurer<C extends ConsumerProperties, T> {
```
ilayaperumalg: We may have producer side error configurer once https://github.com/spring-cloud/spring-cloud-stream/issues/916 is addressed. May be we could create a common `BinderErrorConfigurer` that is not specific to `ConsumerProperties` ?

Also, to avoid running into issues like https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/58, we could potentially have both Producer/Consumer properties for the `ErrorConfigurer` that knows how to use Consumer/Producer properties based on the context.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
viniciusccarvalho: It's meant to be used by the implementations on the `configure` step of the `BinderErrorConfigurer`. MessageChannel binders are a bit odd in that sense since we need to configure the endpoint and not the target. But the reactive binder for instance you would need to configure the target

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
viniciusccarvalho: A couple of reasons:

First problem arose when trying to get to split register and configure from the ErrorConfigurer. Because we were starting the components lifecycle before the configure had a chance to act. This is the whole reason, and it brings us closer to a model where we could have binding as a full fledge bean, I believe that once we have that, we can make it a lifecycle component and use proper start/stop hooks for the underlying components.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
viniciusccarvalho: I don't think we should, since we want this to be mandatory. This is actually the whole reason of this PR, making sure all binders adhere to a ErrorConfigurer strategy

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/error/BinderErrorConfigurer.java
viniciusccarvalho: Good point, Kafka already proved to be a pain to create the implementation since we need the producerProperties as well as the binderconfiguration. It was solved at a much lower abstraction, but still painful (not as painful when compared on how to handle the anonymous consumer groups, that took my sleep for a couple of days)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
viniciusccarvalho: Gotcha, thanks for spotting this

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageProducerBinding.java
```diff
@@ -0,0 +1,71 @@
...
+	public void setMessageProducer(MessageProducer messageProducer) {
+		this.messageProducer = messageProducer;
+	}
+
+	@Override
+	/**
+	 * Initializes the {@link MessageProducer} instance associated with this instance.
+	 */
+	protected void afterBind() throws Exception{
+		if (messageProducer instanceof InitializingBean) {
```
olegz: Given that `messageProducer` is not required, should we null check before `instanceof`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/error/AbstractMessageChannelErrorConfigurer.java
```diff
@@ -0,0 +1,273 @@
...
+			beanFactory.registerSingleton(errorChannelName, errorChannel);
+			errorChannel = (LastSubscriberAwareChannel) beanFactory.initializeBean(errorChannel, errorChannelName);
+		}
+		ErrorMessageSendingRecoverer recoverer;
+		if (errorMessageStrategy == null) {
+			recoverer = new ErrorMessageSendingRecoverer(errorChannel);
+		}
+		else {
+			recoverer = new ErrorMessageSendingRecoverer(errorChannel, errorMessageStrategy);
+		}
```
olegz: IMHO this is really an SI issue. I would address it in ```ErrorMessageSendingRecoverer```. It already supports default error message strategy via one of its constructor
```
public ErrorMessageSendingRecoverer(MessageChannel channel) {
	this(channel, new DefaultErrorMessageStrategy());
}
```
. . . so consider modifying both constructors in ```ErrorMessageSendingRecoverer``` to something like this:
```
public ErrorMessageSendingRecoverer(MessageChannel channel) {
	this(channel, null);
}

public ErrorMessageSendingRecoverer(MessageChannel channel, ErrorMessageStrategy errorMessageStrategy) {
	setChannel(channel);
	setErrorMessageStrategy(errorMessageStrategy == null ? new DefaultErrorMessageStrategy() : errorMessageStrategy);
}
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/error/AbstractMessageChannelErrorConfigurer.java
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+
+	/**
+	 * Binders can return an {@link ErrorMessageStrategy} for building error messages; binder
+	 * implementations typically might add extra headers to the error message.
+	 * @return the implementation - may be null.
+	 */
+	protected ErrorMessageStrategy getErrorMessageStrategy() {
+		return null;
```
olegz: Following on the previous comment, could it even be simpler if we return an instance of the ```DefaultErrorMessageStrategy``` here? Individual Binders can still override, but we can avoid all that null checking.
That said, I still believe the 'ErrorMessageSendingRecoverer' constructors need to be addressed , but that would be a separate and unrelated issue

viniciusccarvalho:
Due the large refactoring involved into this PR we are moving it to 2.0 line. Closing it for now

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/error/AbstractMessageChannelErrorConfigurer.java
olegz: I know it's a closed PR, but for tracking purposes I've just submitted PR to SI which addresses it - https://github.com/spring-projects/spring-integration/pull/2214

PR #1039, opened 2017-08-15
GH-916: MC Binder Producer Error Infrastructure

### LINKED ISSUE #802: Document error handling options.
As Spring Cloud Stream user, I'd be great to have documentation that describes the error handling options available. This may include options at the binder/listener, spring-integration or spring-{Kafka,Rabbitmq} levels and may refer to DLQs, errorChannels, errorHandlers or any other options.

At the moment, it is difficult to understand the current options available specially for users who have been exposed only to Spring Cloud Stream. A typical scenario that can be demonstrated is the case of an exception shield/catch-all scenario for the sync and async scenarios.

### LINKED ISSUE #916: Catch exceptions thrown when publishing via `ProducerHandler`

garyrussell:
Resolves #916
Resolves #802 

- register a pubsub error channel.
- register and subscribe a bridge handler to bridge it to the global error channel.
- pass the error channel to the implementation so it can wire it into the outbound endpoint.
- destroy the infrastructure when unbinding.

Resolves spring-cloud/spring-cloud-stream#916

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -432,7 +432,23 @@ Spring Cloud Stream supports publishing error messages received by the Spring In
 error channel. Error messages sent to the `errorChannel` can be published to a specific destination
 at the broker by configuring a binding for the outbound target named `error`. For example, to
 publish error messages to a broker destination named "myErrors", provide the following property:
-`spring.cloud.stream.bindings.error.destination=myErrors`
+`spring.cloud.stream.bindings.error.destination=myErrors`.
+
+[[binder-error-channels]]
+===== Message Channel Binders and Error Channels
+
+Starting with _version 1.3_, some `MessageChannel` - based binders publish errors to a discrete error channel for each destination.
+In addition, these error channels are bridged to the global Spring Integration `errorChannel` mentioned above.
+You can therefore consume errors for specific destinations and/or for all destinations, using a standard Spring Integration flow (`IntegrationFlow`, `@ServiceActivator`, etc).
+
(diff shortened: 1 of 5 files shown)


garyrussell:
Hold off merging; I will add some tests.

sobychacko:
@garyrussell how do these changes correlate to when someone set something like `spring.cloud.stream.bindings.error.destination=error-dest`. In that case, I think we already create a bridge handler and subscribe to the error channel and send any errors from the error channel to the destination. Referring to this recent commit:  https://github.com/spring-cloud/spring-cloud-stream/commit/22141a8dce93f32d10825ba9fb0cb4d5e29fcb95.

It may not be related, but I am just curious.

garyrussell:
@sobychacko Interesting; I wasn't aware of that. They're (sort of) related, but not really.

Currently, the binders don't send errors to the Spring Integration global `errorChannel`.

It looks to me like that code binds the global `errorChannel` to a destination (enabling errors to be sent to a binder destination). This would only kick in if the user app publishes to the error channel somehow.

This PR registers an error channel for (async) errors for producer destinations. (And [this one](https://github.com/spring-cloud/spring-cloud-stream/pull/926), which is already merged, does the same for consumer destinations). On the consumer side, this publication is performed regardless. DLQ processing now consumes from that channel before sending to a DLQ. On the producer side, the error channel is disabled by default because this adds overhead (in the case of Rabbit, but not Kafka).

In both cases, those error channels are bridged to the global `errorChannel`.

This allows the user to consume producer and consumer either individually or collectively (with a normal SI flow).

Since the global `errorChannel` is a pubsub channel, these errors will also go to that binding as well (`spring.cloud.stream.bindings.error.destination=error-dest`), if present.

viniciusccarvalho:
Merged via 0eab8a5fa700e19c2a026722418d920955d38c7a

PR #1067, opened 2017-09-05
Add codecov support

### LINKED ISSUE #1043: Add codecov support

ilayaperumalg:
Resolves #1043

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,6 @@ cache:
 language: java
 jdk:
   - oraclejdk8
-env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -5,8 +5,6 @@ cache:
 language: java
 jdk:
   - oraclejdk8
-services:
-  - redis-server
```
artembilan: Why do we remove this one?
I can guess because all the metrics stuff has been moved to the separate project, but at least some comment in the commit would shed light for "Why?"
Thanks

ilayaperumalg: @artembilan Updated the commit with some additional changes and description. The `codecov` report is now available as well.

PR #1073, opened 2017-09-11
Fixes issue with originalContentTypeHeader

### LINKED ISSUE #1072: Fix missing contentType header on octet/stream payloads
We introduced a bug with the performance optimizations that allow byte messages to bypass SerDes process.

The deserialization process is also responsible to swap originalContentType and contentType on message arrival, this lead to binary converters to stop working properly as they would not see the correct contentType

viniciusccarvalho:
- Fixes #1072 
- Added a new ContentTypeResolver that searches for originalContentType
  as well as contentType headers
- This should fix issues with converters not being able to find the proper contentType information. The new Resolver could be used in 2.0 line as a fallback to read messages from 1.3 producers

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/avro/AbstractAvroMessageConverter.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/avro/AbstractAvroMessageConverter.java
@@ -20,6 +20,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


viniciusccarvalho:
@sobychacko yes I did not changed that part of the code to avoid ripple effects that are already very hard to track with contentType

Instead I've added a new ContentTypeResolver that checks for `originalContentType` and `contentType`.

For avro messages the messages produced are always of contentType : octet/stream and originalContentType: `avro...`

This new resolver will work if the originalContentType is present or not, decided to fix just the avro use case that seems to be broken at the moment, the other contentTypes are handled with messages already converter into String or Pojos via codec

Issue #1057, opened 2017-09-01
Provide a default interceptor for originalContenType


viniciusccarvalho:
With originalContentType being deprecated in 2.0, the new consumers will not be able to read messages from producers of 1.3, because the real contentType for string types such as `application/xml` will be set on the originalContentType header.

We should provide an interceptor that changes the headers so the downstream converters could read the message correctly.

This should be disabled by default and enabled at a binding level with something like `spring.cloud.stream.bindings.input.legacyContentHeaders=enabled`

Issue #1051, opened 2017-09-01
Remove Codec support


viniciusccarvalho:
- Remove the Kryo Codec
- Create a Kryo Message converter instead

Issue #1050, opened 2017-09-01
Remove serialization from AbstractBinder


viniciusccarvalho:
Remove the Serialization step on the AbstractBinder, serialization should now be handle by the MessageConverters.

As part of this story we need to:

- Remove serialization from AbstractBinder
- Make sure the converter interceptor is ALWAYS set
- Create a bypass mechanism for cases where the type is byte[] and we just ignore any conversion
- Modify the Jackson converter for the String case (target being a String, or to unquote json string)

Issue #992, opened 2017-06-19
Provide comprehensive tests for `contentType` conversion


mbogoevici:
As a developer, I'd like to add test coverage for all the support content-type combinations.

**Tasks:**
- [ ] Source: `POJO` and Target: `JSON String`
- [ ] Source: `Tuple` and Target: `JSON String`
- [ ] Source: `POJO` and Target: `String (toString())`
- [ ] Source: `POJO` and Target: `byte[] (java.io serialized)`
- [ ] Source: `JSON byte[] or String` and Target: `JSON`
- [ ] Source: `byte[] or String` and Target: `Serializable`
- [ ] Source: `JSON byte[] or String` and Target: `Tuple`
- [ ] Source: `byte[]` and Target: `String`
- [ ] Source: `String` and Target: `byte[]`

garyrussell:
@mbogoevici This is probably hijacking this issue, but I don't see one for the `null` payload conversion problem.

I wonder if fixing that issue might be as simple as...

```java
if (this.provideHint) {
	converted = ((AbstractMessageConverter) this.messageConverter).fromMessage(message, this.klazz,
			this.mimeType);
	if (converted == null) {
		// remove content type header and try again
		// add a warn log indicating the header needs to be remove
	}
	if (converted == null) // throw exception
}
```

mbogoevici:
@garyrussell Definitely a welcome suggestion, but it might be a separate issue - what I wanted to capture here before I completely forgot is that we don't have a reference test for the http://docs.spring.io/autorepo/docs/spring-cloud-stream-docs/Chelsea.SR2/reference/htmlsingle/#mime-types-and-java-types

(The tests should capture the issues around null conversion).

PR #1074, opened 2017-09-11
Content Type redesign

### LINKED ISSUE #1052: Make application/json the default contenType
Starting with 2.0 conentType will be required, however we can set a default as application/json if the user hasn't set one

viniciusccarvalho:
Fixes #1052, #1051, #1050, #992 

This is a pretty big PR so I'll try to help out describing the main changes here.

The biggest change was around the behavior of contentType handling, we are now making sure that the converters are the only place serdes happens.

Codec was dropped and replaced by a KryoMessageConverter (needs some attention, probably around performance tuning that was set on the codec, and not yet ported)

AbstractBinder no longer deals with Serialization

Stream now sets SI 5.0 `ConfigurableCompositeMessageConverter` to use the same converters as the framework, so that `@ServiceActivator` and `@StreamListener` will now behave the same on deserialization

Dropped support for `originalContentType` header, not longer used, kept the header name in AbstractBinderHeaders for #1057 

Rule of thumb for all tests and serialization 

- on output:

1.  If type is byte[] we bypass conversion and set contentType if not present to be the default for the channel
2. Any other type goes through conversion, regardless of user setting a contentType property as now a default one is always applied (application/json)

-on input:
1. If method target is `Message<?>`, `byte[]` or `Message<byte[]>` we just don't try any conversion, since the InvocableMethodHandler will match the type and not try conversion
2. Converters are used to translate from byte[] -> Class<?> of method signature of `@ServiceActivator` or `@StreamListener`

Custom converters now require a `@StreamConverter` qualifier, to avoid picking up other converters configured via boot

**All converters** are now in strict mode. They can only convert if a header that matches their configuration is met. This means, no more safe net, if a message arrives an no converter can read it, we throw a MessageConversionException instead of trying to deserialize using a Json converter.


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageSerializationUtils.java
```diff
@@ -45,122 +38,36 @@
 	 * Serialize the message payload unless it is a byte array.
 	 *
 	 * @param message the message with the payload to serialize
-	 * @param codec the codec used for serialization
 	 * @return the Message with teh serialized payload
 	 */
-	public static MessageValues serializePayload(Message<?> message, Codec codec) {
+	public static MessageValues serializePayload(Message<?> message) {
```
ilayaperumalg: I guess we no longer need this method or even simplify to return `MessageValues` at the higher level in the invocation stack.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -152,19 +147,17 @@ protected final String groupedName(String name, String group) {
 	}
 
 	protected final MessageValues serializePayloadIfNecessary(Message<?> message) {
```
ilayaperumalg: Given we don't do any serialization at the `AbstractBinder` , maybe we can get rid of this method completely and have simple method or a step to do the transformation from `Message` to `MessageValues` ?

spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
```diff
@@ -199,13 +199,13 @@ public void testSendAndReceiveNoOriginalContentType() throws Exception {
...
 		binderBindUnbindLatency();
 
-		Message<?> message = MessageBuilder.withPayload("foo").build();
+		Message<?> message = MessageBuilder.withPayload("foo").setHeader(MessageHeaders.CONTENT_TYPE,MimeTypeUtils.TEXT_PLAIN).build();
 		moduleOutputChannel.send(message);
 		Message<?> inbound = receive(moduleInputChannel);
 		assertThat(inbound).isNotNull();
-		assertThat(inbound.getPayload()).isEqualTo("foo");
+		assertThat(inbound.getPayload()).isEqualTo("foo".getBytes());
 		assertThat(inbound.getHeaders().get(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE)).isNull();
```
ilayaperumalg: I guess it is good to check if `originalContentType` is null here (at least to check if the support is dropped). But, I believe we can remove the BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE reference given that we are dropping support for it.

ilayaperumalg: Also,  I still see some references of `originalContentType` especially in `AbstractMessageChannelBinder`'s `serializeAndEmbedHeadersIfApplicable` method

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -104,36 +107,39 @@ public void configureOutputChannel(MessageChannel messageChannel, String channel
 	 * @param channel message channel to set the data-type and message converters
 	 * @param channelName the channel name
 	 */
-	private void configureMessageChannel(MessageChannel channel, String channelName, boolean input) {
+	private void configureMessageChannel(MessageChannel channel, String channelName,
```
ilayaperumalg: @viniciusccarvalho some specific formatter setting in your IDE makes these changes (re-aligning method arguments in separate lines) in several places.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -185,116 +193,79 @@ public int selectPartition(Object key, int partitionCount) {
...
-			}
+
 			this.messageConverter = MessageConverterConfigurer.this.compositeMessageConverterFactory
-					.getMessageConverterForType(this.mimeType);
+					.getMessageConverterForAllRegistered();
 			this.provideHint = this.messageConverter instanceof AbstractMessageConverter;
 		}
 
 		@Override
 		public Message<?> preSend(Message<?> message, MessageChannel channel) {
```
ilayaperumalg: We don't seem to be using `provideHint` during the conversion. Is there a specific reason for that?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
```diff
@@ -54,7 +54,7 @@
 
 	// Properties for both inbound/outbound
 
-	private String contentType;
+	private String contentType = "application/json";
```
ilayaperumalg: We can probably use constant from ``MimeTypeUtils`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CustomJackson2MappingMessageConverter.java
```diff
@@ -0,0 +1,61 @@
...
+		super();
+		setSerializedPayloadClass(byte[].class);
+		setStrictContentTypeMatch(true);
+	}
+
+	@Override
+	protected Object convertFromInternal(Message<?> message, Class<?> targetClass, @Nullable Object conversionHint) {
+		try{
+			return super.convertFromInternal(message, targetClass, conversionHint);
+		}catch (MessageConversionException me){
```
ilayaperumalg: checkstyle didn't catch this `catch` ;)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/KryoMessageConverter.java
```diff
@@ -0,0 +1,263 @@
...
+			return kryo;
+		};
+		this.pool = new KryoPool.Builder(factory).softReferences().build();
+		this.supportedMimeTypes = Collections.singletonList(MimeType.valueOf(KRYO_MIME_TYPE));
+	}
+
+	@Nullable
+	@Override
+	public Object fromMessage(Message<?> message, Class<?> targetClass, @Nullable Object conversionHint) {
+
```
ilayaperumalg: please avoid inserting empty lines inside the method.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
ilayaperumalg: In case of `MessageChannel` bound components, this means that `MessageConverterConfigurer`'s  `ContentTypeConvertingInterceptor` always gets added to the corresponding `MessageChannel`. I guess this is by design of having `contentType` available all the time at the Messsage and we go by `application/json` as the default.

In case of non- message channel bound components (KStream etc,), having `contentType=application/json` as the binding property may be confusing?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageSerializationUtils.java
viniciusccarvalho: Yes that's the goal, just not yet for this PR giving it's size. We need to move header serdes into its own interceptor too, then we can remove this method completely, and probably Receiving and Sending Handlers too

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
viniciusccarvalho: Please see previous comment

spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
viniciusccarvalho: I kept it for a reason, in a next PR (there's an issue for it) we will provide an interceptor that will allow users to read messages from legacy SCSt. This interceptor will need to check the presence of ORIGINAL_CONTENT_TYPE_HEADER, so there's a use for that still.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
viniciusccarvalho: Yes, to be honest I never understood why we needed in first place. On the way in, the conversion is actually handled by MethodInvokerHandlers, and they will provide the target Class<?> every time.

The whole klazz logic is now gone.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
viniciusccarvalho: Why? Shouldn't we provide consistency across all binders? I need to sync with Soby on the KStream binder, but I believe that it could use the same logic for message converters as it's not SI specific.

I agree this is not the best place to put the default, the proper way should probably be a an EnvironmentPostProcessor, but having the same default for all binders should something we should strive for, having consistency for our users.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/converter/CustomMappingJackson2MessageConverterTests.java
```diff
@@ -0,0 +1,59 @@
...
+		Object converted = converter.convertFromInternal(message, Map.class,null);
+		Assert.assertNotNull(((Map)converted).get("id"));
+	}
+
+	@Test
+	public void convertFromStructuredJsonIntoString() throws Exception {
+		String payload = "{\"id\":1}";
+		Message message = MessageBuilder.withPayload(payload.getBytes()).setHeader(MessageHeaders.CONTENT_TYPE,"application/json").build();
+		CustomJackson2MappingMessageConverter converter = new CustomJackson2MappingMessageConverter();
+		Object converted = converter.convertFromInternal(message, String.class,null);
```
sobychacko: Assertions missing on the test.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/converter/CustomMappingJackson2MessageConverterTests.java
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	@Test
+	public void convertFromJsonString() throws Exception {
+		ObjectMapper mapper = new ObjectMapper();
+		String payload = mapper.writeValueAsString("foo");
+		Message message = MessageBuilder.withPayload(payload.getBytes()).setHeader(MessageHeaders.CONTENT_TYPE,"application/json").build();
+		CustomJackson2MappingMessageConverter converter = new CustomJackson2MappingMessageConverter();
+		Object converted = converter.convertFromInternal(message, String.class,null);
```
sobychacko: Assertitons missing on the test.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -185,116 +193,79 @@ public int selectPartition(Object key, int partitionCount) {
...
+			// bypass conversion for ErrorMessges
+			if (message instanceof ErrorMessage) {
+				return message;
+			}
+			// bypass conversion for raw bytes
+			if (message.getPayload() instanceof byte[]) {
+				return MessageConverterConfigurer.this.messageBuilderFactory
```
artembilan: Well, we had a discussion that we are going to use here `MutableMessageBuilder` for better performance.
Are you going to do that in the separate PR as well?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/converter/KryoMessageConverterTests.java
```diff
@@ -0,0 +1,87 @@
...
+		Message message = MessageBuilder.withPayload(baos.toByteArray()).build();
+		Object result = kryoMessageConverter.fromMessage(message,String.class);
+		Assert.assertNull(result);
+	}
+
+	@Test(expected = MessageConversionException.class)
+	public void readWithWrongPayloadType() throws Exception{
+		KryoMessageConverter kryoMessageConverter = new KryoMessageConverter(null,true);
+		Message message = MessageBuilder.withPayload("foo").setHeader(MessageHeaders.CONTENT_TYPE,KryoMessageConverter.KRYO_MIME_TYPE+";type=java.lang.String").build();
+		Object result = kryoMessageConverter.fromMessage(message,String.class);
```
sobychacko: unused `result` variable. We could remove that I guess.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/converter/KryoMessageConverterTests.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	@Test(expected = MessageConversionException.class)
+	public void readWithWrongPayloadFormat() throws Exception{
+		KryoMessageConverter kryoMessageConverter = new KryoMessageConverter(null,true);
+		Message message = MessageBuilder.withPayload("foo").setHeader(MessageHeaders.CONTENT_TYPE,KryoMessageConverter.KRYO_MIME_TYPE+";type=java.lang.String").build();
+		Object result = kryoMessageConverter.fromMessage(message,String.class);
```
sobychacko: unused `result` variable. We could remove that I guess.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -185,116 +193,79 @@ public int selectPartition(Object key, int partitionCount) {
...
-							.build();
-				}
-				else {
-					sentMessage = message;
-				}
+			Object converted;
+
+			if (this.input) {
+				return MessageConverterConfigurer.this.messageBuilderFactory
+						.withPayload(message.getPayload())
```
artembilan: Maybe something has missed here?
I only see that this `if(this.input)` bock is fully similar to the previous for the `if (message.getPayload() instanceof byte[]) {`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CustomJackson2MappingMessageConverter.java
```diff
@@ -0,0 +1,61 @@
...
+			throw me;
+		}
+
+	}
+
+
+	@Nullable
+	@Override
+	protected Object convertToInternal(Object payload, @Nullable MessageHeaders headers, @Nullable Object conversionHint) {
+		return super.convertToInternal(payload, headers, conversionHint);
```
artembilan: Why do we need this method at all if it just delegates to the `super`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/CustomJackson2MappingMessageConverter.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Custom implementation of {@link org.springframework.messaging.converter.MappingJackson2MessageConverter} that handles special String cases.
+ * If the target of conversion is a String, it tries to read it from a quoted json string, to properly remove the quotes, if it fails, it then just
+ * returns the original string as it is just a raw json string needed for the target.
+ *
+ * @author Vinicius Carvalho
+ */
+
+public class CustomJackson2MappingMessageConverter extends MappingJackson2MessageConverter{
```
artembilan: This doesn't sound as a good name for the class looking into its purpose.
Maybe `StringFallbackJackson2MappingMessageConverter` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/converter/KryoMessageConverter.java
sobychacko: Also, there are unnecessary empty lines (more than 1) before the ctor.

artembilan: Well, it depends, @ilayaperumalg .

If your method declaration wrap its arguments to a new line, the good practice indeed to insert blank line after that for better readability. But not in this case, agreed.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
viniciusccarvalho: I thought about leaving that for a separate PR as well @artembilan since it's more of a performance optimization. To be fair I need to understand more how would it work, but I'm not forgetting about it

spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/PartitionCapableBinderTests.java
```diff
@@ -198,13 +200,14 @@ public void testPartitionedModuleSpEL() throws Exception {
 		catch (UnsupportedOperationException ignored) {
 		}
 
-		Message<Integer> message2 = MessageBuilder.withPayload(2)
+		Message<String> message2 = MessageBuilder.withPayload("2")
```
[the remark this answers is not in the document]
viniciusccarvalho: It can, true, but on the receiving I still need to check for a String, I just thought it would be a bit more consistent.
In fairness I'd just send a byte if I was writing the test today, this was just to make sure it goes green with the new handling of payloads.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
viniciusccarvalho: I'm trying to track this down, I use intellij with the eclipse formatter plugin using the spring guidelines. I really can't say what else could be happening. I usually run a code format to avoid missing a space after comma and get burned by Artem. Or handling better alignment of freaking builder methods. Sorry, not sure what else, Any ideas?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/StreamConverter.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Marker to tag {@link org.springframework.messaging.converter.MessageConverter} beans that will be added to the {@link org.springframework.cloud.stream.converter.CompositeMessageConverterFactory}
+ * @author Vinicius Carvalho
+ */
+@Target({ElementType.FIELD,ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Qualifier
+public @interface StreamConverter {
+}
```
sobychacko: Are we settling on this name? Without reading javadocs, it is not apparent to me what this annotation is used for. Any possibilities to reflect what the annotation is intended for from the name itself? `StreamConverter` could mean a few things in the context of stream processing. We may want to convey what it is from the name itself since it feels like a fundamental part of the core stream API.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/StreamListenerTestUtils.java
```diff
@@ -55,6 +55,14 @@ public String getFoo() {
 		public void setFoo(String foo) {
 			this.foo = foo;
 		}
+
+		@Override
+		public String toString() {
+			final StringBuffer sb = new StringBuffer("FooPojo{");
```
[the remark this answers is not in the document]
viniciusccarvalho: That's just the IDE using generate toString. I really did not care as it was just meant for a simple string comparison now that objects are sent as bytes, so I could compare FooPojo.toString().

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
viniciusccarvalho: Hum, good point, I really did not notice as I was stripping the method away as functionality was ported to the converters.

Rules are indeed equal:

On the outbound we check if the payload is bytes so we bypass conversion (performance optimization)

On inbound, well we don't care about inbound at all. Perhaps the right approach should not configure this interceptor for inbound, since conversion on inbound happens at the MethodHandler on both `@StreamListener` and `@ServiceActivator`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/StreamConverter.java
viniciusccarvalho: I'm open to suggestions, the name came from a chat with @artembilan. I have no strong opinions on it.

PedroAlvarado: Allow me to suggest @StreamMessageConverter. The word "message" helps disambiguate and narrow down its purpose. M

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1361,122 +1361,187 @@ To allow you to propagate information about the content type of produced message
 For middleware that does not directly support headers, Spring Cloud Stream provides its own mechanism of automatically wrapping outbound messages in an envelope of its own.
 For middleware that does support headers, Spring Cloud Stream applications may receive messages with a given content type from non-Spring Cloud Stream applications.
 
-Spring Cloud Stream can handle messages based on this information in two ways:
+The content type resolution process have been redesigned for Spring Cloud Stream 2.0.
```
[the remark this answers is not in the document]
viniciusccarvalho: I disagree, a good amount of our users would like to know what changed. There will be a migrating from 1.3 section, but I think the changes are substantial enough for us to put on a paragraph

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/MessageChannelConfigurerTests.java
```diff
@@ -78,9 +77,6 @@ public void testMessageConverterConfigurer() throws Exception {
 		MessageHandler messageHandler = new MessageHandler() {
 			@Override
 			public void handleMessage(Message<?> message) throws MessagingException {
-				assertThat(message.getPayload()).isInstanceOf(Tuple.class);
-				assertThat(((Tuple) message.getPayload()).getFieldNames().get(0)).isEqualTo("message");
-				assertThat(((Tuple) message.getPayload()).getValue(0)).isEqualTo("Hi");
```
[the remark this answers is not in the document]
viniciusccarvalho: This test is perhaps irrelevant now. We no longer try to convert messages based on payload, and since there's no more serializeIfNecessary step, when you subscribe to an input channel directly, you get the original message written to the channel.

I'll revisit this, as the right test now should be putting a payload of `byte[]` into the channel, because that's what the Binders will offer anyways.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
viniciusccarvalho: It's here: https://github.com/spring-cloud/spring-cloud-stream/issues/1058

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
artembilan: Yeah... I think the problem that Eclipse Formatter doesn't feet to all our requirements and I really don't like some of its code styles.
I'm pretty sure that this reformatting we have on this code line is one of that weird stuff...

We still have to work for the proper and convenient formatting for all of us, including Checkstyle.
Right, some code will suffer of reformatting coincidentally, just because it doesn't feet rules. Like this line of code.

BTW the "space after comma" is one of Checkstyle rule 😄 : http://checkstyle.sourceforge.net/config_whitespace.html#WhitespaceAfter

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java
artembilan: Why is it bad? Spring Boot has default values in many places of the `@ConfigurationProperties`, for example: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java#L256

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/StreamConverter.java
artembilan: This has been named after `@IntegrationConverter` - just a blast after quick brainstorm.

I like @PedroAlvarado 's `@StreamMessageConverter`!

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -525,6 +525,7 @@ protected String errorsBaseName(ProducerDestination destination) {
 		return destination.getName() + ".errors";
 	}
 
+	@Deprecated
 	private final class ReceivingHandler extends AbstractReplyProducingMessageHandler {
```
artembilan: I wonder if we can remove this internal classes altogether. 
We are in major version, so any braking changes like this is possible.
Or do you leave them for the legacy data support?

viniciusccarvalho: I think we may be able to remove it, haven't really thought much about it. I know for a fact that once we take care of Embedding headers as an interceptor too, there won't be much left in those classes.
I just tagged them to keep an eye out during the epic of 2.0 redesign

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -151,22 +146,22 @@ protected final String groupedName(String name, String group) {
 		return name + GROUP_INDEX_DELIMITER + (StringUtils.hasText(group) ? group : "default");
 	}
 
+	@Deprecated
 	protected final MessageValues serializePayloadIfNecessary(Message<?> message) {
```
garyrussell: Shouldn't we just add these deprecations in 1.3 and remove them in 2.0? Otherwise we can't remove them until 3.0.

garyrussell: But I guess that does not have to happen before the merge.

artembilan: Well, we can't deprecated in `1.3` since it is a first class citizen in the logic there.
I'm fine stick with deprecation until `3.0` as far as this method will be out of use any way.

viniciusccarvalho:
merged via 40ca71e9dfef228b460f446e5edde1edadd80f9c Thank you @sobychacko

PR #1090, opened 2017-10-04
Introducing new annotation parameters

### LINKED ISSUE #1027: Add contentType support for @Output annotation
As described by @markpollack, it would be interesting to have support for contentType selection on the annotation level, as well as the properties file.

```java
    @Output(contentType="application/json")
    MessageChannel hotDrinks();

    @Output
    MessageChannel coldDrinks();
```

@mbogoevici @garyrussell do you think we can add this for 1.3 GA? Doesn't seem to be a complex thing to implement

viniciusccarvalho:
- Fixes #1027
- Just adding the properties, will issue separate PR for internal change
  in logic using them.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/Input.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/Input.java
@@ -24,12 +24,14 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cloud.stream.messaging.Sink;
 
 /**
  * Indicates that an input binding target will be created by the framework.
  *
  * @author Dave Syer
  * @author Marius Bogoevici
+ * @author Vinicius Carvalho
  */
 
 @Qualifier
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/Input.java
```diff
@@ -42,4 +44,8 @@
 
 	String value() default "";
 
+	String destination() default Sink.INPUT;
```
artembilan: Looks like `2.0.x` isn't rebased to `master`: https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/Input.java#L44

And here is no JavaDocs.

There is also no functionality around these new attributes.
Are you going to address it lately?
Why then do not introduce them there as well?

This PR doesn't bring value to the Framework, but may make some noise from the community:

> I try to use new attributes in my `@Input` annotation but they seems don't work. Am I doing anything wrong?

Does it make sense to you?

(Sorry if I sound as nasty)

viniciusccarvalho: @artembilan Very important cue here: "Just adding the properties, will issue separate PR for internal change
in logic using them."

PR #1095, opened 2017-10-10
KAFKA GH-223: Handle unexpected content type

garyrussell:
See https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/223

When using native encoding; NPE when no class name derived from the content type.

Do not attempt deserialization if `JavaClassMimeTypeUtils.classNameFromMimeType(contentType)`
returns `null`.

Test case in the Kafka binder.

__cherry-pick to master__


artembilan:
@garyrussell ,

You know looks like nothing to fix in `master` on the matter. 
Or there is needed a fix in different place. Sorry I can't figure out where, since `JavaClassMimeTypeUtils` is out of use on the `master`...

garyrussell:
OK; let's cherry-pick the kafka test and we can verify it works ok.

Issue #1106, opened 2017-10-19
Address content-type compatibility between messages generated by different versions of SCSt apps


olegz:
Due to the recent content-type negotiation [improvements](https://github.com/spring-cloud/spring-cloud-stream/commit/171f034a8c2dd7c89123170ffcd9cb5012c3a2f9) we unintentionally introduced compatibility issue where pre-2.0 _consumers_ may not be able to properly negotiate the content type of the message generated by 2.0+ _producers_.
This is due to the fact that pre-2.0 _producers_ and _consumers_ were relying on the `original-content-type` header which was removed in 2.0.
Another [change](https://github.com/spring-cloud/spring-cloud-stream/commit/08504952e69ae8f286ab1a54c10208d15abdff88#diff-b1929bb9f8da0256744c4d55a7762fdbR2) that was introduced as part of this effort deals with registering additional flag (`spring.cloud.stream.bindings.input.legacyContentTypeHeaderEnabled`) and interceptor (_LegacyContentTypeHeaderInterceptor_) to allow 2.0+ _consumers_ to successfully process pre-2.0 messages. This issue is that it makes _consumer_ statically aware of the _producer_ and its version and could result in more issues if such _consumer_ receives messages from the _producers_ of different versions. 

Proposal:
1. 2.0 _producer_ should re-introduce the `original-content-type` header. While this header will be ignored by 2.0 _consumers_ it will be used by pre-2.0 _consumers_ as before. It should also generate one additional header such as *version* (e.g., `version=2.x`). This additional header will be used by 2.0+ _consumers_ to ignore the re-introduced `original-content-type` header (more details below).
2. 2.0 consumer should have a mechanism (may be through the already existing _LegacyContentTypeHeaderInterceptor_) to check for the `version` header before checking for the existence of  the `original-content-type` header. If 'version' header is pre-2.0 or is not present (which is what's going to be in the case of pre-2.0 producers) then we know it's a legacy message and content-type negotiation should be done using the `original-content-type` header, otherwise such header will be ignored. Basically the version header will play the same role as newly introduced `spring.cloud.stream.bindings.input.legacyContentTypeHeaderEnabled` flag, but it will come as part of the message instead of static configuration of the consumer. Further more, making it a `version` instead of a boolean will give us more flexibility in the future when it comes to similar issues especially if we decide to drop support for certain version of messages.

garyrussell:
[...]

PR #1107, opened 2017-10-20
Prep work for Kafka Issue #236

garyrussell:
See https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/236

- add `none` as a synonym for `HeaderMode.raw`
- add `headers` meaning "use native headers"
- change the default header mode - `null`
- remove `supportsHeadersNatively` - allow binders to chose header mode for each binding
  binders decide their own default when the property is null
- Rabbit will choose `headers` as default (but can be overridden to `none` for a binding)
  `embeddedHeaders` will not be allowed
- Kafka will choose `embeddedHeaders` as default for compatibility with 1.x peer apps
- remove deprecations on header embedding code

Resolves spring-cloud/spring-cloud-stream#1106

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1165,7 +1165,7 @@ Default: null (the default binder will be used, if one exists).
 
 The following binding properties are available for input bindings only and must be prefixed with `spring.cloud.stream.bindings.<channelName>.consumer.`, e.g. `spring.cloud.stream.bindings.input.consumer.concurrency=3`.
 
-Default values can be set by using the prefix `spring.cloud.stream.default.consumer`, e.g. `spring.cloud.stream.default.consumer.headerMode=raw`.
+Default values can be set by using the prefix `spring.cloud.stream.default.consumer`, e.g. `spring.cloud.stream.default.consumer.headerMode=none`.
 
 concurrency::
   The concurrency of the inbound consumer.
@@ -1176,11 +1176,13 @@ partitioned::
 +
 Default: `false`.
 headerMode::
-    When set to `raw`, disables header parsing on input.
+    When set to `none`, disables header parsing on input.
 Effective only for messaging middleware that does not support message headers natively and requires header embedding.
-Useful when inbound data is coming from outside Spring Cloud Stream applications.
+This option is useful when consuming data from non-Spring Cloud Stream applications when native headers are not supported.
+When set to `headers`, uses the middleware's native header mechanism.
+When set to `embeddedHeaders`, embeds headers into the message payload.
 +
-Default: `embeddedHeaders`.
+Default: depends on binder implementation.
 maxAttempts::
   If processing fails, the number of attempts to process the message (including the first).
   Set to 1 to disable retry.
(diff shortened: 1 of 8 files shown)


olegz:
Thanks @garyrussell. Let me go over it so I can understand what's happening and i'll merge it over the weekend

garyrussell:
@olegz DO NOT MERGE

I need to work through the kafka binder changes first - it's not going to be so easy since the kryo deserialization (if present) is in an interceptor now; the embedded headers have to be extracted after the kryo deserialization.

garyrussell:
Actually, my bad, Vinny took care of everything - the headers are embedded after the kryo serialization (outbound) and before the deserialize (inbound).

Still might be worth waiting a few days until I work through the compatibility stuff.

garyrussell:
This works fine with my PR...

```java
@SpringBootApplication
public class KafkaScstBinder2Application {

		SpringApplication.run(KafkaScstBinder2Application.class, args);
	}

	@Bean
	public ApplicationRunner runner(MessageChannel output) {
		return args -> output.send(new GenericMessage<>(new Foo()));
	}

	@StreamListener(Sink.INPUT)
	public void listen(Foo in) {
		System.out.println(in);
	}

	public static class Foo {

	}

}
```

```yaml
spring:
  kafka:
    consumer:
      auto-offset-reset: earliest
      enable-auto-commit: false
  cloud:
    stream:
      bindings:
        input:
          destination: testEmbed
          group: testembed
          consumer:
            header-mode: embeddedheaders
        output:
          destination: testEmbed
          content-type: application/x-java-object
          producer:
            header-mode: embeddedheaders

```

but I think we'll still need some hooks to support falling back on the consumer side (if no native headers, check to see if there are embedded headers).

olegz:
@garyrussell so based on my understanding from the above yaml, the header strategy is defined per-consumer. Which means if another producer sends a message with different header strategy the consumer will barf.

olegz:
@garyrussell In other words it's the same issue of consumers being statically aware of producers (albeit indirectly), which may be something we are okay with.. . ., but in the last attempt to make it more dynamic. . .
After digging into information around Kafka's message protocol I may have found something we can rely on to make it more dynamic.
From reading [this](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets) it appears that pre-0.11 Kafka messages will have the value of the _magic byte_ set to 1 and since 0.11 it is 2. That alone would allow us to determine the version of a message (at least from pre/post 0.11 perspective). Once that is done it may be as simple as looking at the Kafka message's _value bytes_ and determine if they contain embedded headers and I am assuming we would know how to do that if we wrote them.

garyrussell:
>Which means if another producer sends a message with different header strategy the consumer will barf.

See my comment on [#1106](https://github.com/spring-cloud/spring-cloud-stream/issues/1106#issuecomment-338308019)

My plan is to look for native headers first and, if not present, fall back to look for embedded headers (if the header mode is so configured).

>magic byte

Right, but the client doesn't expose the magic byte. I don't want to use reflection.

olegz:
Ok, so if I understand correctly you still need to configure some header strategy on Consumer and if that

garyrussell:
>Ok, so if I understand correctly you still need to configure some header strategy on Consumer and if that

Only so that we can avoid the overhead of attempting the fallback if there are no native headers and we know that no producers will ever send embedded headers.

garyrussell:
Polishing; support fallback

- if the consumer side is configured to support embedded headers,
  binders should look for native headers first and signal their presence
  using the `BinderHeaders.NATIVE_HEADERS_PRESENT` header;
  fall back to embedded headers if it appears the payload might contain
  them
- remove support for XD embedded headers (first byte < 0xff)
- add a quick check for embedded headers by examining the first byte for 0xff
- change the log from error to debug when decoding embedded headers fails

- see the kafka binder PR for a test case with a consumer that can handle all
  message flavors (native, embedded, and no headers).

olegz:
Merging

olegz:
Merged

Issue #1101, opened 2017-10-17
Make `MessageCollector` 1.x/2.x compatible


marcingrzejszczak:
without this type no proper serialization takes place and messages get `byte[]` body instead of the one basing on the `contentType`. Previously the default was json.

olegz:
@marcingrzejszczak  We had some internal discussions and I am actually rather reluctant to do anything with _ MessageCollector_ (at least at this point). My primary motivation is that with _<?>_, one should not have expectation of any type of automatic conversion. So what you get is what message looks like on the wire (which could itself be beneficial). Sure this could be improved by possibly creating some type of test utility to process message through the same type conversion  logic used when dealing with user code, but that would be different requirement all together.
Also, the _contentType_ and it's role in the conversion is being questioned as well. 

Thoughts?

marcingrzejszczak:
Adding @dsyer 

IMO this is a breaking change. Anybody who was using the `MessageCollector` at this point will get broken if the sent message was a String JSON. 

I mean I've managed to work this around so I'm no longer affected. I guess I'll need to change the sent object (I currently send a String whereas apparently, I should be sending sth else).

olegz:
@marcingrzejszczak I am coming to the same realization ;)
Anyway, it is fixed with this PR https://github.com/spring-cloud/spring-cloud-stream/pull/1115 once it is merged which is hopefully by tomorrow

sobychacko:
This issue is addressed with PR #1115

PR #1115, opened 2017-10-29
GH-1106 - addressed contentTyped compatibility issues

olegz:
This PR consists of 3 commits. One from @sobychacko 6c259be which addresses the _consumer_ side of things and two from me where the first commit addresses the _producer_ side of things related to content type compatibility fixes. More details in individual commit. The third commit simply restored tests that use MessageCollector to depend on proper payload type

Resolves spring-cloud/spring-cloud-stream#1106
Resolves spring-cloud/spring-cloud-stream#1101
Resolves spring-cloud/spring-cloud-stream#1110


cppwfs:
Tried this PR against the TaskProcessor tests and it looks like we are getting the JSON vs. the actual class as shown below:
```
Expected: Channel to receive a message whose payload is <TaskLaunchRequest{uri='maven://org.springframework.cloud.task.app:timestamp-task:jar:1.0.1.RELEASE', commandlineArguments=[], environmentProperties={payload=hello}, deploymentProperties={}}>
     but: received: "{\"uri\":\"maven://org.springframework.cloud.task.app:timestamp-task:jar:1.0.1.RELEASE\",\"commandlineArguments\":[],\"environmentProperties\":{\"payload\":\"hello\"},\"deploymentProperties\":{},\"applicationName\":\"Task-67c130ec-bfc2-4867-a588-ae80c1090077\"}"
```
Is this to be expected?

cppwfs:
After talking it over with Oleg.  The downstream apps will convert the JSON to the POJO.  So in this case I will update the test, since the Stream Test Framework does not convert the JSON to POJO.

olegz:
After discussing with the team and to clarify (in interim) what these fixes are(arn't) addressing (although this may still change, so I'll follow up):
1. The backward/forward compatibility between the messages has been restored for majority of the use cases, where `byte[]` representing the data on the wire represent either a raw/binary data (data that always existed as `byte[]`) or `text` type communicated through `contentType` header. 
2. There are still concerns for cases where message payload is represented as POJO especially after removing the dependency on Cryo. Those concerns are split between producer and consumer:
	2.1 On the producer side we seem to have a good handle on converting POJOs to `byte[]` via _JavaSerializationMessageConverter_. However we can only do it if `contentType` is set to `application/x-java-serialized-object` and POJO is _Serializable_. 
	2.2 On the consumer side we do not have the appropriate interceptor in the stack to do the oposite
At the very least we need to adress 2.2 to make it fully symetrical to 2.1. However we are still thinking of possibility of bringing Cryo back to ensure full compatibility between the messages from different versions of SCSt.

spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
```diff
@@ -121,36 +129,52 @@ public void testClean() throws Exception {
...
-		QueueChannel moduleInputChannel = new QueueChannel();
-		Binding<MessageChannel> producerBinding = binder.bindProducer("foo.0",
-				moduleOutputChannel, outputBindingProperties.getProducer());
-		Binding<MessageChannel> consumerBinding = binder.bindConsumer("foo.0",
-				"testSendAndReceive", moduleInputChannel, createConsumerProperties());
-		// Bypass conversion we are only testing sendReceive
-		Message<?> message = MessageBuilder.withPayload("foo".getBytes())
-				.setHeader(MessageHeaders.CONTENT_TYPE,
-						MimeTypeUtils.APPLICATION_OCTET_STREAM)
+		BindingProperties outputBindingProperties = createProducerBindingProperties(createProducerProperties());
```
garyrussell: I simply __hate__ reformatting being done as part of a big PR like this - it detracts from the meaningful changes and makes the review much harder - I __strongly__ believe such efforts (while worthwhile) __must__ be done as a separate PR.

olegz: I agree Gary and that was an over site on my part.

garyrussell: Thanks; turned out to be not too bad; just that this was the first file and I thought uh-oh 😄

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -294,16 +295,78 @@ private ContentTypeConvertingInterceptor(String contentType, boolean input) {
 		}
 	}
 
-	public final static class LegacyContentTypeHeaderInterceptor extends ChannelInterceptorAdapter {
+	public final static class InbondMessageConvertingInterceprtor extends ChannelInterceptorAdapter {
```
sobychacko: typo - Inbond -> Inbound

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -294,16 +295,78 @@ private ContentTypeConvertingInterceptor(String contentType, boolean input) {
...
+				Object payload = converter.fromMessage(message, targetClass);
+				contentType = targetClass == null ? MessageConverterUtils.X_JAVA_SERIALIZED_OBJECT : MessageConverterUtils.X_JAVA_OBJECT;
+				message = MessageBuilder.withPayload(payload)
+						.copyHeaders(message.getHeaders())
+						.setHeader(MessageHeaders.CONTENT_TYPE, contentType)
+						.removeHeader(BinderHeaders.BINDER_ORIGINAL_CONTENT_TYPE)
+						.build();
+			}
+			else {
+				contentType = contentType == null ? contentTypeResolver.resolve(message.getHeaders()) : contentType;
```
sobychacko: Probably the reassignment of `contentType` is not needed here. Only need the assignment if it is null.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -294,16 +295,78 @@ private ContentTypeConvertingInterceptor(String contentType, boolean input) {
...
+			}
+
+			if (converter != null){
+				if (contentType.equals(MessageConverterUtils.X_JAVA_OBJECT) && targetClass == null){
+					throw new IllegalStateException("Can not deserialize into message since 'contentType` has not "
+							+ "being encoded with the actual target type."
+							+ "Consider 'application/x-java-object; type=foo.bar.MyClass'");
+				}
+				Object payload = converter.fromMessage(message, targetClass);
+				contentType = targetClass == null ? MessageConverterUtils.X_JAVA_SERIALIZED_OBJECT : MessageConverterUtils.X_JAVA_OBJECT;
```
sobychacko: `targetClass` cannot be `null` at this point.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/MessageSerializationUtils.java
```diff
@@ -16,58 +16,48 @@
...
 		Object originalContentType = message.getHeaders().get(MessageHeaders.CONTENT_TYPE);
-		MessageValues messageValues = new MessageValues(message);
-		messageValues.setPayload(originalPayload);
-		messageValues.put(MessageHeaders.CONTENT_TYPE, originalContentType);
-		return messageValues;
-	}
+		// Pass content type as String since some transport adapters will exclude
+		// CONTENT_TYPE Header otherwise
+		String contentType = setOriginalContentType ? JavaClassMimeTypeUtils.mimeTypeFromObject(originalPayload,
+				ObjectUtils.nullSafeToString(originalContentType)).toString() : originalContentType.toString() ;
```
garyrussell: This throws an NPE when `originalContentType` is `null` and `setOriginalContentType` is `false`.

garyrussell: This works around it... https://github.com/garyrussell/spring-cloud-stream-binder-kafka/commit/94daa9f2a3139d6e23628578c5c281c7e8474d7c

sobychacko:
LGTM - polished, squashed and merged upstream.

PR #1125, opened 2017-11-08
Initial cleanup of AbstractMessageChannelBinder

olegz:
Removed extra handler for embeddedHeaders in favor of ChannelInterceptor to avoid additional dispatch
Deprecated HeaderMode

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -28,13 +28,11 @@ import org.springframework.cloud.stream.provisioning.ProvisioningProvider;
 import org.springframework.context.Lifecycle;
 import org.springframework.expression.ExpressionParser;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
-import org.springframework.integration.channel.FixedSubscriberChannel;
+import org.springframework.integration.channel.AbstractMessageChannel;
 import org.springframework.integration.channel.PublishSubscribeChannel;
 import org.springframework.integration.context.IntegrationContextUtils;
 import org.springframework.integration.core.MessageProducer;
-import org.springframework.integration.endpoint.EventDrivenConsumer;
 import org.springframework.integration.handler.AbstractMessageHandler;
-import org.springframework.integration.handler.AbstractReplyProducingMessageHandler;
 import org.springframework.integration.handler.BridgeHandler;
 import org.springframework.integration.handler.advice.ErrorMessageSendingRecoverer;
 import org.springframework.integration.support.ErrorMessageStrategy;
@@ -43,6 +41,7 @@ import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.SubscribableChannel;
+import org.springframework.messaging.support.ChannelInterceptorAdapter;
 import org.springframework.util.Assert;
 import org.springframework.util.MimeType;
 
@@ -60,6 +59,7 @@ import org.springframework.util.MimeType;
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
  * @author Soby Chacko
+ * @author Oleg Zhurakousky
  * @since 1.1
  */
 public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties, P extends ProducerProperties, PP extends ProvisioningProvider<C, P>>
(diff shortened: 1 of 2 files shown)


olegz:
@garyrussell could you please take a look at this. This is specifically concerning cleanup around embedded headers and since they are only used in Kafka i figured you'd be the best to take a look.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -220,34 +219,28 @@ protected void afterUnbindProducer(ProducerDestination destination, P producerPr
...
-			final boolean extractEmbeddedHeaders = HeaderMode.embeddedHeaders.equals(
-					properties.getHeaderMode());
-			ReceivingHandler rh = new ReceivingHandler(extractEmbeddedHeaders);
-			rh.setOutputChannel(inputChannel);
-			final FixedSubscriberChannel bridge = new FixedSubscriberChannel(rh);
-			bridge.setBeanName("bridge." + name);
+			ConsumerDestination destination = this.provisioningProvider.provisionConsumerDestination(name, group, properties);
+
+			if (HeaderMode.embeddedHeaders.equals(properties.getHeaderMode())) {
+				this.enhanceMessageChannel(inputChannel);
```
garyrussell: No `this.` on method calls.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -220,34 +219,28 @@ protected void afterUnbindProducer(ProducerDestination destination, P producerPr
...
 			return new DefaultBinding<MessageChannel>(name, group, inputChannel,
-					endpoint instanceof Lifecycle ? (Lifecycle) endpoint : null) {
+					consumerEndpoint instanceof Lifecycle ? (Lifecycle) consumerEndpoint : null) {
 
 				@Override
 				protected void afterUnbind() {
 					try {
-						if (endpoint instanceof DisposableBean) {
-							((DisposableBean) endpoint).destroy();
+						if (this.getEndpoint() instanceof DisposableBean) {
```
garyrussell: `this.` only on field accesses.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -276,6 +269,31 @@ else if (e instanceof ProvisioningException) {
...
+	@Deprecated
+	private void enhanceMessageChannel(MessageChannel inputChannel) {
+		((AbstractMessageChannel)inputChannel).addInterceptor(new ChannelInterceptorAdapter() {
+			@SuppressWarnings("unchecked")
+			@Override
+			public Message<?> preSend(Message<?> message, MessageChannel channel) {
+				if (message.getPayload() instanceof byte[]) {
+					if (!message.getHeaders().containsKey(BinderHeaders.NATIVE_HEADERS_PRESENT)
+							&& EmbeddedHeaderUtils.mayHaveEmbeddedHeaders((byte[]) message.getPayload())) {
+						MessageValues messageValues;
```
garyrussell: I am not sure why you deprecated `HeaderMode` (comment elsewhere in this review) but I think we only want to fall back to looking for embedded headers for binders that might support it. For example, the rabbit binder doesn't set the `NATIVE_HEADERS_PRESENT` - that is intended for binders that support embedded headers (e.g. Kafka for backwards compatibility).

artembilan: Should I implement now something myself in the Kinesis Binder to embed headers?
Thanks

garyrussell: @artembilan it should just happen automatically if the headerMode is set.

This comment is redundant - discussed with @olegz - this method is only called if the header mode is `embeddedHeaders` - I was thrown off by the deprecation.

garyrussell: But, @artembilan you might want to copy the embedded header test(s) from the kafka binder.

artembilan: Sure, @garyrussell ! Thank you for the pointer!

Although it's still not clear for me why we deprecate `HeaderMode` and this method still supporting `embeddedHeaders` mode. I might just miss some other hook on the matter...
Will investigate eventually.

artembilan: Ah, OK. @olegz has just addressed my concerns.
LGTM then 😄

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -276,6 +270,30 @@ else if (e instanceof ProvisioningException) {
...
+						try {
+							messageValues = EmbeddedHeaderUtils.extractHeaders((Message<byte[]>) message, true);
+						}
+						catch (Exception e) {
+							logger.debug(EmbeddedHeaderUtils.decodeExceptionMessage(message),e);
```
artembilan: Why `debug` not `error`?
We have just caught an `Exception` though

olegz: that's an old/existing code so while you may be right, it would needs to be fixed in another issue, another PR

garyrussell: I changed it from `error` to `debug` since we are falling back and we don't know for sure that the headers are embedded - we only attempt to fall back if the first byte is 0xff - it is possible that a real message without embedded headers passes that condition (but the extract will fail).

garyrussell: Before I added my fallback code, it was error because the binder expected embedded headers. This now allows the reception of messages from producers using any header mode (native, embedded or none).

artembilan: Thank you for explanation!

In SI I would request a comment for this kind of concerns. 😉 
Eventually we all may forget why it is debug on exception, so someone may change it unconditionally.

garyrussell:
Merged as 62c8513197793b9b15798fd0081599f92aabafb5 after fixing a couple more `this.[a-zA-Z]\(`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
garyrussell: Good point - https://github.com/spring-cloud/spring-cloud-stream/commit/b9884ae71a90bb65a5471d06ff1e716cf647c7e6

Issue #1127, opened 2017-11-08
Create implementation of a Binder backed by Spring Integration


olegz:
This will greatly improve testing since tests could be written to interact with the actual binder rather then _input/output_ channels of the actual application, thus validating everything that is happening with the _message_ before it gets to _input/output_ (conversion, serialization etc). 

We kind of already have a variation of it in some tests, but is scattered around, so hopefully this will consolidate it into a single and configurable Binder that could be used internally by the framework for testing as well as demos.

PR #1128, opened 2017-11-08
GH-1127 Added initial support for SI backed Binder [DO NOT MERGE]

olegz:
The idea is to introduce a standard Stub Binder that relies on the same core configuration as Kafka and Rabbit binders (using Spring Integration is just a natural choice for implementing such binder).
**This is primarily to introduce consistency around testing various stream components.** 
Basically, we already have variations of such binder scattered around many tests, resulting in some code duplication across `setupContext()` methods. This could eventually lead to such tests going out of sink with the core configuration if it were to change making those tests less valid.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -78,6 +78,10 @@ public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties,
 	 */
 	private final String[] headersToEmbed;
 
+	public AbstractMessageChannelBinder(PP provisioningProvider) {
+		this(new String[0], provisioningProvider);
+	}
+
 	public AbstractMessageChannelBinder(String[] headersToEmbed,
 			PP provisioningProvider) {
 		this.headersToEmbed = headersToEmbed == null ? new String[0] : headersToEmbed;
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/AbstractBindingLifecycle.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/AbstractBindingLifecycle.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.binding;
+
+import java.util.Map;
+
+import org.springframework.beans.BeansException;
+import org.springframework.context.SmartLifecycle;
+
+/**
+ * @author Oleg Zhurakousky
+ *
+ */
+abstract class AbstractBindingLifecycle implements SmartLifecycle {
+
+	private volatile boolean running;
+
+	private final BindingService bindingService;
+
+	private final Map<String, Bindable> bindables;
+
+	private final boolean input;
+
+	AbstractBindingLifecycle(BindingService bindingService, Map<String, Bindable> bindables, boolean input) {
+		this.bindingService = bindingService;
+		this.bindables = bindables;
+		this.input = input;
+	}
+
+	@Override
+	public void start() {
+		if (!running) {
+			for (Bindable bindable : bindables.values()) {
+				if (input) {
+					bindable.bindInputs(bindingService);
+				}
+				else {
+					bindable.bindOutputs(bindingService);
+				}
+			}
+			this.running = true;
+		}
+	}
+
+	@Override
+	public void stop() {
+		if (running) {
+			try {
+				for (Bindable bindable : bindables.values()) {
+					if (input) {
+						bindable.unbindInputs(bindingService);
+					}
+					else {
+						bindable.unbindOutputs(bindingService);
+					}
+				}
+			}
+			catch (BeansException e) {
+				throw new IllegalStateException("Cannot perform unbinding, no proper implementation found", e);
+			}
+			this.running = false;
+		}
+	}
+
+	@Override
+	public boolean isRunning() {
+		return running;
+	}
+
+	@Override
+	public boolean isAutoStartup() {
+		return true;
+	}
+
+	@Override
+	public void stop(Runnable callback) {
+		stop();
+		if (callback != null) {
+			callback.run();
+		}
+	}
+
+	/**
+	 * Return a low value so that this bean is started after receiving Lifecycle beans are
+	 * started. Beans that need to start before bindings will set a lower phase value.
+	 */
+	@Override
+	public int getPhase() {
+		return this.input ? Integer.MAX_VALUE - 1000 : Integer.MIN_VALUE + 1000;
+	}
+}
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/InputBindingLifecycle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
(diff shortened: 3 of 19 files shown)


artembilan:
Travis doesn't like how you are ignoring Checkstyle 😉 : 
```
[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java:19:15: Unused import - org.junit.Assert.assertEquals. [UnusedImports]

[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java:21:8: Unused import - java.nio.charset.StandardCharsets. [UnusedImports]

[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java:23:8: Unused import - org.slf4j.Logger. [UnusedImports]

[ERROR] /home/travis/build/spring-cloud/spring-cloud-stream/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java:24:8: Unused import - org.slf4j.LoggerFactory. [UnusedImports]
```

olegz:
@sobychacko I initially thought the same and had it in test-support, but then we can't make test-support a dependency of a core since it will create a cycle and majority of the tests that would benefit from it are in the core. So it's kind of a catch 22 situation. 
On the other hand, this isn't really an issue and semantically correct to what's been established within Spring in general - "ability to include a default implementation of some strategy" (e.g., default implementation of MessageStore, etc).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java
```diff
@@ -0,0 +1,130 @@
...
+class SpringIntegrationChannelBinder extends AbstractMessageChannelBinder<ConsumerProperties,
+	ProducerProperties, SpringIntegrationProvisioner> implements DisposableBean {
+
+	SpringIntegrationChannelBinder(SpringIntegrationProvisioner provisioningProvider) {
+		super(provisioningProvider);
+	}
+
+	@Override
+	public void destroy() throws Exception {
+		// noop
```
artembilan: Why do we need this hook if it is `noop` anyway?
Do you have anything else in mind for the future improvements?

sobychacko:
@olegz cyclic dependency angle makes sense. In that case, is it possible to keep them in one of the test packages? By including it in the core published artifact in `src/main/java`, I am afraid, (a) whether we are sending the signal that this is a binder used outside of testing/demo purposes (b) other classpath issues since 2 binder implementations are available in applications during runtime - one from the middleware implementation (kafka, rabbit etc.) and this one.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	@Override
+	protected MessageHandler createProducerMessageHandler(ProducerDestination destination,
+			ProducerProperties producerProperties, MessageChannel errorChannel) throws Exception {
+		MessageChannel siBinderOutputChannel = ((SpringIntegrationProducerDestination)destination).getOutputChannel();
+		MessageHandler messageHandler = new MessageHandler() {
+			@Override
+			public void handleMessage(Message<?> message) throws MessagingException {
+				siBinderOutputChannel.send(message);
```
artembilan: There is just a `BridgeHandler` for such a purpose.

olegz:
Including it in _test_ package won't work either, since maven build will fail due to test packages not being visible across modules and/or projects (tried that as well).
As to your second point; With regard to 'a)', not I am not afraid of that since as I mentioned you can find samples of it across the entire spring portfolio. As for 'b)', that is indeed a valid point, however as you have noticed this binder is not bootstrapped vid _META-INF/spring.binders_, so the only way for it to be in play is to be physically included in your configuration. And even if that were to happen, there is "@Conditional..." on the entire configuration.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java
```diff
@@ -0,0 +1,130 @@
...
+
+	@Override
+	protected MessageProducer createConsumerEndpoint(ConsumerDestination destination,
+			String group, ConsumerProperties properties) throws Exception {
+
+		DirectChannel siBinderInputChannel = ((SpringIntegrationConsumerDestination)destination).getInputChannel();
+		BridgeHandler handler = new BridgeHandler();
+		siBinderInputChannel.subscribe(handler);
+
+		MessageProducer producer = new MessageProducer() {
```
artembilan: Why just don't return that `BridgeHandler` instance?
It is really a `MessageProducer` as well

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java
olegz: we don't, just leftover from copy/paste. Good catch

sobychacko:
@olegz There is a way to make the test classes visible for testing purposes to other modules by publishing a test-jar from core which will only include the testing classes (no new artifacts published from the build, but its handled at build time). We have done that in the kafka binder I believe in 1.3. With that said, if the general consensus is to have this new concrete binder part of the core, I am good.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/si/SpringIntegrationChannelBinder.java
olegz: Not following you

artembilan:
```
BridgeHandler handler = new BridgeHandler();
handler.setOutputChannel(siBinderOutputChannel);
return handler;
```
(Sorry didn't tested 😄 )

sobychacko:
Another reason that comes to my mind for keeping it out of core `src/main/java` or to a new module altogether. By providing this binder in the core published artifact, there is nothing that prevents a user to use it by just providing a `META-INF/spring.binders` in the application. Our binder API is opinionated around the concepts of consumer groups, partitioning etc. Since it is difficult to enforce these concepts in an SI binder, we are sort of moving away from those opinions in a binder implementation provided by us through core `scst`. However, since this is a binder impl used for testing, I think it is okay to provide it as long as the users don't get the idea that this is a full blown binder provided by the core framework. Thats why I thought its best to keep it out of `src/main/java`.

olegz:
@sobychacko Actually in SI you can emulate and simulate any kind of scenarios (groups, partitions etc.) So, the intention is to evolve this binder to be configurable to simulate any and all scenarios we support.

markfisher:
I agree with Soby. We need to maintain a clear separation between "stream" binders and "integration" binders. This is also something we discussed in a recent team meeting, where the idea of an integration binder for handling auto-configuration of Channel Adapters is a good idea, but with the StreamBinder contract extending that for those brokers where true pubsub-with-consumer-groups behavior can be supported. Our RabbitMQ binder is a good example, due to the Exchange and Queue model (i.e. even though "consumer groups" are not explicitly supported as they are in Kafka, we can emulate). The same couldn't/shouldn't be applied for binders such as email, file, etc.

olegz:
@markfisher with regard to revisiting the binder API and see what needs to be done to clearly isolate the concepts behind Streaming binders and Message binders I am completely on board, but this specific PR and discussion is primarily around consolidating what we already have scattered around the code base. The only question is where should it live?

olegz:
@markfisher with regard to revisiting the binder API and see what needs to be done to clearly isolate the concepts behind Streaming binders and Message binders I am completely on board, but this specific PR and discussion is primarily around consolidating what we already have scattered around the code base. The only question is where should it live?

sobychacko:
@olegz a middle ground might be to pull out this new binder implementation to a new module (or use one of the existing test modules) and move all the scattered tests that depend on this to that module, thus avoiding the cycle with core. That way, we can isolate the SI binder and all the tests in a separate module and not mix the core framework module with testing both stream and messaging binder specific concepts. What do you think?

olegz:
@sobychacko yes, very possible. As you and I discussed the main idea is to bring consistency and correctness to our testing . At the moment this is just a preview. No need to rush it. We can wait till  M4 or even RC.
I am currently playing with some of the existing tests in the core to see how we can benefit. Once I do the same in other modules, it will be more clear as to where and how something like this should live.

olegz:
Closing in favor of a different PR which only provides implementation of such binder for testing.

PR #1131, opened 2017-11-08
KGH-12: Late binding for kafka etc

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/12

Support late binding for binders that don't inherently support it.

Schedule attempts at 30 second intervals (default, configurable).

`IllegalState` and `IllegalArgument` exceptions are fatal.

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -1123,6 +1123,12 @@ The typical usage of this property is to be nested in a customized environment <
 +
 Default: false.
 
+spring.cloud.stream.bindingRetryInterval::
+  The interval (seconds) between retrying binding creation when, for example, the binder doesn't support late binding and the broker is down (e.g. Apache Kafka).
+Set to zero to treat such conditions as fatal, preventing the application from starting.
++
+Default: 30
+
 [[binding-properties]]
 === Binding Properties
 
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.stream.binding;
 
+import java.sql.Date;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -36,6 +37,7 @@ import org.springframework.cloud.stream.binder.ExtendedProducerProperties;
 import org.springframework.cloud.stream.binder.ExtendedPropertiesBinder;
 import org.springframework.cloud.stream.binder.ProducerProperties;
 import org.springframework.cloud.stream.config.BindingServiceProperties;
+import org.springframework.scheduling.TaskScheduler;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.validation.DataBinder;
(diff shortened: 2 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -95,16 +106,56 @@ public BindingService(
...
+		if (exception instanceof IllegalStateException || exception instanceof IllegalArgumentException) {
+			throw exception;
+		}
+		this.log.error("Failed to create consumer binding; retrying in " +
+			this.bindingServiceProperties.getBindingRetryInterval() + " seconds", exception);
+		this.taskScheduler.schedule(() -> {
+			try {
+				late.setDelegate(binder.bindConsumer(target,
+					this.bindingServiceProperties.getGroup(inputName), input,
+					consumerProperties));
```
sobychacko: @garyrussell just curious, why are we calling `setDelegate` like this in the case of consumer, but setting it directly in the case of producer.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -188,4 +274,39 @@ private void validate(Object properties) {
...
+			if (this.unbound) {
+				delegate.unbind();
+			}
+			else {
+				this.delegate = delegate;
+			}
+		}
+
+		@Override
+		public synchronized void unbind() {
```
sobychacko: if some thread calls `unbind` on the `LateBinding` object (since the reference is already returned from consumer binding) and `setDelegate` is called through retries at around same time, isn't there a chance that `unbind` could potentially be called on the same binding delegate object twice?

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1123,6 +1123,12 @@ The typical usage of this property is to be nested in a customized environment <
...
 Default: false.
 
+spring.cloud.stream.bindingRetryInterval::
+  The interval (seconds) between retrying binding creation when, for example, the binder doesn't support late binding and the broker is down (e.g. Apache Kafka).
+Set to zero to treat such conditions as fatal, preventing the application from starting.
++
+Default: 30
+
 [[binding-properties]]
 === Binding Properties
```
sobychacko: Since we support late binding from the core framework level with this PR, do we need to streamline the behavior in the rabbit binder? (i.e, all binders reuse this strategy for late binding?)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
garyrussell: That's why the methods are synchronized...

- if the successful bind wins, we set the field and the unbind will unbind after it gets the monitor.
- if the unbind wins, we set the boolean and when the bind gets the monitor it will unbind instead of setting the field.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
garyrussell: Good catch; we __must__ use the setter to avoid the race.

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
garyrussell: Rabbit's mechanism is more sophisticated (it listens for new connections and will bind faster); I don't think we want to "dumb it down".

sobychacko:
Looks good! squashed and merged upstream.

Issue #1009, opened 2017-07-05
Spring cloud stream RabbitMQ routing messages dynamically


abhdeb:
I have implemented the example as shown here [Dynamic Destinations](http://docs.spring.io/spring-cloud-stream/docs/Elmhurst.BUILD-SNAPSHOT/reference/htmlsingle/#dynamicdestination) Spring Dynamic Destination

In the rabbitmq, it is creating an exchange dynamically, but there is no option to provide binding or routing key. My requirement is to send a message to this dynamically created exchange with a routing key. How would i need to implement this to setup the routing key?

```java
@Component public class DDProducerBean {

@Autowired
private BinderAwareChannelResolver poChannelResolver = null;

public void publish(DDSocketVO ddSocketVO) throws Exception {
    this.poChannelResolver.resolveDestination(ddSocketVO.getDestination()).send(MessageBuilder.withPayload(new ObjectMapper().
            setVisibility(PropertyAccessor.FIELD, Visibility.ANY).
            writeValueAsString(ddSocketVO)).build());
}
```

My requirement is to route messages dynamically at runtime based on the payload. So i need a way to set the routing key at run time. As a result i can't use producer.routingKeyExpression, which is done in config or properties or yml, and not dynamically at run time, so that won't serve the purpose. 

This implementation is specific for spring.cloud.stream.dynamicDestinations where the BinderAwareChannelResolver takes care of dynamically creating/binding the outbound channel for these dynamic destinations. But it doesn't talk about how to bind the outbound channel to a queue and how to specify a routing key expression. As these are dynamic channels/ producers, we can't use producer.routingKeyExpression as the producer is created dynamically.

garyrussell:
Producers are not responsible for binding queues; consumers are.

There is some support for a producer binding queues using the `requiredGroups` property, but that is for fixed destinations; there is no such support for dynamic destinations.

You can programmatically declare queues and bindings using the `AmqpAdmin` bean after you have resolved the destination and before sending messages.

abhdeb:
That sounds perfect, i had the same idea, thank you.

abhdeb:
In case of routingKeyExpression, producer is responsible for setting the expression-  for example 
System.setProperty("spring.cloud.stream.rabbit.bindings.websocketOutputChannel.producer.routingKeyExpression", "'adam.ir.list'"); Do you mean "Producers are not responsible for binding queues; consumers are" only incase of dynamic destination implementation?

[...]

PR #1132, opened 2017-11-10
GH-1009: Producer properties for dynamic bindings

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/1009

Provide a callback mechanism to set properties on dynamically created bindings.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
@@ -45,7 +45,7 @@ public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestina
 	private final DynamicDestinationsBindable dynamicDestinationsBindable;
 
 	@SuppressWarnings("rawtypes")
-	private final NewBindingCallback newBindingCallback;
+	private final NewDestinationBindingCallback newBindingCallback;
 
 	private ConfigurableListableBeanFactory beanFactory;
 
@@ -58,7 +58,7 @@ public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestina
 	@SuppressWarnings("rawtypes")
 	public BinderAwareChannelResolver(BindingService bindingService,
 			AbstractBindingTargetFactory<? extends MessageChannel> bindingTargetFactory,
-			DynamicDestinationsBindable dynamicDestinationsBindable, NewBindingCallback callback) {
+			DynamicDestinationsBindable dynamicDestinationsBindable, NewDestinationBindingCallback callback) {
 		this.dynamicDestinationsBindable = dynamicDestinationsBindable;
 		Assert.notNull(bindingService, "'bindingService' cannot be null");
 		Assert.notNull(bindingTargetFactory, "'bindingTargetFactory' cannot be null");
@@ -136,7 +136,7 @@ public class BinderAwareChannelResolver extends BeanFactoryMessageChannelDestina
 	 *
 	 */
 	@FunctionalInterface
-	public interface NewBindingCallback<T> {
+	public interface NewDestinationBindingCallback<T> {
 
 		/**
 		 * Configure the properties or channel before binding.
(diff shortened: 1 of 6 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -93,6 +106,16 @@ public MessageChannel resolveDestination(String channelName) {
 					channel = this.bindingTargetFactory.createOutput(channelName);
 					this.beanFactory.registerSingleton(channelName, channel);
 					channel = (MessageChannel) this.beanFactory.initializeBean(channel, channelName);
+					if (this.newBindingCallback != null) {
+						ProducerProperties producerProperties = this.bindingService.getBindingServiceProperties()
+								.getProducerProperties(channelName);
+						Object extendedProducerProperties =
+								this.bindingService.getExtendedProducerProperties(channel, channelName);
+						this.newBindingCallback.configure(channelName, channel, producerProperties,
```
ilayaperumalg: Maybe we could only call the `callback.configure` only when the `extendedProducerProperties` isn't null ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -103,4 +126,29 @@ public MessageChannel resolveDestination(String channelName) {
...
+	/**
+	 * Configure a new destination before it is bound.
+	 * @param <T> the extended properties type. If you need to support dynamic binding
+	 * with multiple binders, use {@link Object} and cast as needed.
+	 *
+	 * @since 2.0
+	 *
+	 */
+	@FunctionalInterface
+	public interface NewBindingCallback<T> {
```
ilayaperumalg: Perhaps the name for this callback could be `DynamicDestinationBindingCallback` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
garyrussell: Why would we not want to configure just the common properties if the binder does not support extended properties? Perhaps the callback javadocs should clarify that the extension might be null.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
garyrussell: Sounds good; I'll hold off changing in case there are other suggestions 😄

olegz: @garyrussell I agree with @ilayaperumalg on the name

olegz: Actually I'd call it  `NewDestinationBindingCallback`. There is nothing dynamic about this interface's definition. Implementations on the other hand could have a more specific names.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -128,6 +128,17 @@ public BindingService(
...
 
+	@SuppressWarnings("rawtypes")
+	public Object getExtendedProducerProperties(Object output, String outputName) {
+		Binder binder =  getBinder(outputName, output.getClass());
+		if (binder instanceof ExtendedPropertiesBinder) {
+			return ((ExtendedPropertiesBinder) binder).getExtendedProducerProperties(outputName);
+		}
+		else {
+			return null;
+		}
```
olegz: Should we just remove `else`?

garyrussell: I generally prefer this style but don't really mind either way; the bytecode is identical.

Feel free to change on merge if you feel strongly.

olegz: I am good

PR #1133, opened 2017-11-10
Native encoding/decoding changes for interceptors

### LINKED ISSUE #1109: Native Encoding/Decoding support for the inbound side
Add a flag `useNativeDecoding` for the consumer side.

Do not apply the content type interceptors if native encoding/decoding is being used.

sobychacko:
Fixes #1109

- Disable outbound content type interceptor when nativeEndoding is enabled on the producer
- Disable inbound content type interceptor when nativeDecoding is enabled on the consumer

- Add a new flag in ConsumerProperties for useNativeDecoding

- Apply ContentType interceptor before the InboundInterceptor to ensure
  that content type set on the channel is applied on the message if content type
  is missing on the Message.

- Add tests for verifying interceptors are not applied when native encoding/decoding is enabled
- Fix a kryo integration test
- Remove Inbound interceptor from MessageCollector when native encoding is enabled on the producer
- Polishing

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.annotation.JsonInclude;
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
  * @author Gary Russell
+ * @author Soby Chacko
  */
 @JsonInclude(JsonInclude.Include.NON_DEFAULT)
 public class ConsumerProperties {
@@ -47,6 +48,8 @@ public class ConsumerProperties {
 	private double backOffMultiplier = 2.0;
 
 	private HeaderMode headerMode;
+
+	private boolean useNativeDecoding = false;
 
 	@Min(value = 1, message = "Concurrency should be greater than zero.")
 	public int getConcurrency() {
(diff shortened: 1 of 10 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -125,13 +126,16 @@ private void configureMessageChannel(MessageChannel channel, String channelName,
 					getPartitionKeyExtractorStrategy(producerProperties),
 					getPartitionSelectorStrategy(producerProperties)));
 		}
-		if (input) {
-			messageChannel.addInterceptor(new InboundMessageConvertingInterceptor());
-		}
-		// TODO: Set all interceptors in the correct order for input/output channels
+		ConsumerProperties consumerProperties = bindingProperties.getConsumer();
 		if (StringUtils.hasText(contentType)) {
```
ilayaperumalg: Isn't this only for the case where the user explicitly sets the `contentType` in the binding properties for the channel name and also use the nativeDecoding for the decoding?

sobychacko: @ilayaperumalg it doesn't matter if `contentType` is set on the channel or not. If native encoding/decoding is enabled, we will skip any message conversion. If we don't do that, based on the recent content type redesign, we could end up converting the message in certain cases.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
```diff
@@ -48,6 +49,8 @@
 
 	private HeaderMode headerMode;
 
+	private boolean useNativeDecoding = false;
```
olegz: Could you remove the assignment to `false`. Redundant.

olegz:
@sobychacko I'll address that one comment and will merge

Issue #1097, opened 2017-10-12
Consider to introduce @StreamTest for SCSt-specific slices to test


artembilan:
See Spring Boot [Reference Manual](https://docs.spring.io/spring-boot/docs/1.5.7.RELEASE/reference/html/boot-features-testing.html#boot-features-testing-spring-boot-applications-testing-autoconfigured-tests) for more info. 

This one might be marked with something like:
```
public @interface AutoConfigureTestBinder {

	/**
	 * If {@link TestSupportBinderAutoConfiguration} should be appied.
         * Defaults to {@code true}
	 * @return if test binder support is enabled
	 */
	boolean enabled() default true;

}
```

Right now we can exclude that one only via:
```
@SpringBootTest(
		properties = "spring.autoconfigure.exclude=org.springframework.cloud.stream.test.binder.TestSupportBinderAutoConfiguration")
```
which is not so convenient and obvious.

olegz:
Now with https://github.com/spring-cloud/spring-cloud-stream/commit/e50801c82427fc354f67ad1c3cab3704d9ba9bd7 and new TestBinder being publicly available we can consider doing something like what you are proposing

PR #1139, opened 2017-11-14
GH-1127 Added initial implementation of the stub binder back by SI

olegz:
- Added initial implementation of SI-backed stub binder to facilitate more consistent testing
- Improved BinderAwareChannelResolverTests and ExtendedPropertiesBinderAwareChannelResolverTests to make use of it
- Added sample Application that demonstrates the binder usage

### DIFF
--- a/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationBinderConfiguration.java
+++ b/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationBinderConfiguration.java
@@ -18,7 +18,7 @@ package org.springframework.cloud.stream.binder.integration;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -70,10 +70,8 @@ public class SpringIntegrationBinderConfiguration<T> {
 
 	@Bean
 	public BinderTypeRegistry binderTypeRegistry() {
-		BinderType bt = new BinderType(NAME, new Class[] {SpringIntegrationBinderConfiguration.class});
-		Map<String, BinderType> binderTypes = new HashMap<>();
-		binderTypes.put(NAME, bt);
-		BinderTypeRegistry btr = new DefaultBinderTypeRegistry(binderTypes);
+		BinderType binderType = new BinderType(NAME, new Class[] {SpringIntegrationBinderConfiguration.class});
+		BinderTypeRegistry btr = new DefaultBinderTypeRegistry(Collections.singletonMap(NAME, binderType));
 		return btr;
 	}
 
--- a/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationBinderConfiguration.java
+++ b/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationBinderConfiguration.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.binder.integration;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.binder.Binder;
+import org.springframework.cloud.stream.binder.BinderType;
+import org.springframework.cloud.stream.binder.BinderTypeRegistry;
+import org.springframework.cloud.stream.binder.ConsumerProperties;
+import org.springframework.cloud.stream.binder.DefaultBinderTypeRegistry;
+import org.springframework.cloud.stream.binder.ProducerProperties;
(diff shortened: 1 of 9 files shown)


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationProvisioner.java
```diff
@@ -0,0 +1,132 @@
...
+	private final Map<String, SubscribableChannel> provisionedDestinations = new HashMap<>();
+
+	@Autowired
+	private SourceDestination source;
+
+	@Autowired
+	private TargetDestination target;
+
+	/**
+	 * Will provision consumer destination as SI {@link PublishSubscribeChannel} thus
```
artembilan: s/consumer/producer

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationBinderConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+		Map<String, Object> annotationAttributes = AnnotationUtils.getAnnotationAttributes(annotation);
+		configClasses.addAll(Arrays.asList((Class<?>[])annotationAttributes.get("value")));
+		return configClasses.toArray(new Class<?>[] {});
+	}
+
+	@Bean
+	public BinderTypeRegistry binderTypeRegistry() {
+		BinderType bt = new BinderType(NAME, new Class[] {SpringIntegrationBinderConfiguration.class});
+		Map<String, BinderType> binderTypes = new HashMap<>();
+		binderTypes.put(NAME, bt);
```
artembilan: `Collections.singletonMap()` instead?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationChannelBinder.java
```diff
@@ -0,0 +1,274 @@
...
+ * Implementation of {@link Binder} backed by Spring Integration framework.
+ * It is useful for localized demos and testing.
+ * <p>
+ * This binder extends from the same base class ({@link AbstractMessageChannelBinder}) as
+ * other binders (i.e., Rabbit, Kafka etc). Interaction with this binder is done via a pair of named {@link MessageChannel}s.
+ * <br>
+ * The names of the channels are:
+ * <ul>
+ * <li>{@link SpringIntegrationBinderConfiguration#BINDER_INPUT}</li>
+ * <li>{@link SpringIntegrationBinderConfiguration#BINDER_OUTPUT}</li>
```
artembilan: There are no these constants any more.
Red JavaDocs is only here:
image

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationChannelBinder.java
```diff
@@ -0,0 +1,274 @@
...
+ * </ul>
+ * Simply autowire them in your your application and send/receive messages.
+ * </p>
+ * You must also add {@link SpringIntegrationBinderConfiguration} to your configuration.
+ * Below is the example using Spring Boot test.
+ * <pre class="code">
+ *
+ * &#064;RunWith(SpringJUnit4ClassRunner.class)
+ * &#064;SpringBootTest(classes = {SpringIntegrationBinderConfiguration.class, TestWithSIBinder.MyProcessor.class})
```
artembilan: Well, this reminds me my idea about `@StreamTest` slice: https://github.com/spring-cloud/spring-cloud-stream/issues/1097

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/TargetDestination.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ */
+public class TargetDestination extends AbstractDestination {
+
+	private Queue<Message<?>> messages;
+
+	/**
+	 * Allows to receive {@link Message}s that have been received by a {@link TargetDestination}.
+	 */
+	public Message<?> receive() {
```
artembilan: I think we have to make as `BlockingQueue` and use its `poll(timeout)` to block the testing thread for the message from the async internals. There is no guarantee that all tests are going to be direct and sync.

WDYT?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationChannelBinder.java
```diff
@@ -0,0 +1,274 @@
...
+				// Empty
+			}
+		}
+	}
+
+	/**
+	 * Implementation of ErrorMessageStrategy used by this binder.
+	 * Modeled after the one in Rabbit binder although doesn't do much, so a placeholder for now .
+	 */
+	private static class IntegrationHeaderErrorMessageStrategy implements ErrorMessageStrategy {
```
artembilan: I somehow think the `DefaultErrorMessageStrategy` is enough.
Although I guess you might have something in mind for the future.
Since it is semantically is the same, I won't mind to keep it here just for consistency with Rabbit Binder for example.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationChannelBinder.java
olegz: Probably. In fact I was looking at that issue today

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationChannelBinder.java
olegz: Ohh, no. That is exactly the feedback I needed as I was more concerned about modeling it after existing Rabbit binder, then the individual components like this one.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SpringIntegrationChannelBinder.java
olegz: In the future we'll probably wrap this under something like `@StreamTest` and/or `@Enable. . .`

olegz:
@artembilan just pushed suggested changes.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/21a2a84d86f63c34845bf13368e34a808a92eedd

PR #1161, opened 2017-12-21
Fix AbstractBinderTests to populate properties

artembilan:
Some tests in `AbstractBinderTests` are rely on the partition header,
but that one can be populated only by the `PartitionInterceptor`.
To allow it we need to provide appropriate producer properties, like
`partitionKeyExpression`.

* Populate test-based `ProducerProperties` to the
`createBindableChannel()` where `PartitionInterceptor` is populated
* Fix `checkstyle.xml` static imports rule for Mockito-2.x as the
current base line

### DIFF
--- a/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
+++ b/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
@@ -73,6 +73,7 @@ import static org.junit.Assert.assertTrue;
  * @author Marius Bogoevici
  * @author Oleg Zhurakousky
  * @author Jacob Severson
+ * @author Artem Bilan
  */
 @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java
```diff
@@ -521,12 +528,17 @@ public void testSendPojoReceivePojoKryoWithStreamListener() throws Exception {
...
-				new BindingProperties());
-		DirectChannel moduleInputChannel = createBindableChannel("input",
-				new BindingProperties());
+
+		PP producerProperties = createProducerProperties();
+
+		BindingProperties producerBindingProperties = createProducerBindingProperties(producerProperties);
+
+		DirectChannel moduleOutputChannel = createBindableChannel("output", producerBindingProperties);
+		DirectChannel moduleInputChannel = createBindableChannel("input", producerBindingProperties);
```
garyrussell: This looks wrong; input channels use `consumerBindingProperties`.

(multiple places).

artembilan:
OK! How about now?
Thanks

garyrussell:
Merged as 7d2e3a38150bb509359a64ec4ececaa5f734cc16

Issue #1170, opened 2018-01-03
The messageHandlerHeaderPropagationBeanPostProcessor causes early BeanFactory initialization


artembilan:
The `messageHandlerHeaderPropagationBeanPostProcessor` bean depends on the `SpringIntegrationProperties` bean:
```
@Bean
public static BeanPostProcessor messageHandlerHeaderPropagationBeanPostProcessor(
		final SpringIntegrationProperties springIntegrationProperties) {
```
This causes all the infrastructure to be activated for the `@ConfigurationProperties` processing. And that is too early since we might not have all the required beans loaded into the application context.

Reconsider solution without autowiring into the `BeanPostProcessor` bean definition.

PR #1171, opened 2018-01-03
GH-1170: Fix early BeanFactory initialization

artembilan:
Fixes: spring-cloud/spring-cloud-stream#1170

The `messageHandlerHeaderPropagationBeanPostProcessor` causes early
`BeanFactory` initialization via its `SpringIntegrationProperties`
dependency.
This triggers an initialization all the `@ConfigurationProperties`
infrastructure in Spring Boot, when we might not have all the necessary
beans, e.g. `Converter` for some configuration property

* Eliminate the "early `BeanFactory` initialization syndrome" through
the `SmartInitializingSingleton`

**Cherry-pick to 1.3.x**

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2017 the original author or authors.
+ * Copyright 2015-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
```diff
@@ -245,51 +245,75 @@ public Object postProcessBeforeInitialization(Object bean, String beanName) thro
...
+				// TODO: Filter out beans created by SCSt (not currently necessary)
+				Class<?> beanClass = ClassUtils.getUserClass(bean);
+				if (AbstractReplyProducingMessageHandler.class.isAssignableFrom(beanClass)) {
+					this.producingMessageHandlers.add((AbstractReplyProducingMessageHandler) bean);
 				}
+				return bean;
+			}
+
+			@Override
+			public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
```
olegz: Unless we're back-porting it we can remove it. Will do it during the merge

Issue #1174, opened 2018-01-08
Add support for Pollable Consumers


sabbyanandan:
As a continuation of spring-projects/spring-integration#2310, as a developer, I'd like to add support for `DeferredAcknowledgmentMessageSource` in the core framework, so we could support pollable-consumer use-cases.

An application of this would be for scenarios that require rate-limiting. For instance, a batch-job/task can be triggered by upstream events from the broker; however, for situations that need rate limiting the launch of tasks (esp. relevant for resource-constrained environments), this ability allows the consumer to react only when they are ready to process it. This avoids launching the tasks on a recurring basis when there aren't enough resources to do the processing - it could prevent overloading the system.

Relates to spring-cloud/spring-cloud-stream-binder-rabbit#121
Relates to spring-cloud/spring-cloud-stream-binder-kafka#277

PR #1176, opened 2018-01-10
GH-1174: Add Pollable Consumer

garyrussell:
Resolves: https://github.com/spring-cloud/spring-cloud-stream/issues/1174

Introduce `PollableMessageSource` for polled consumers.

The `@Input` annotation can now be applied to interface methods returning
`PollableMessageSource` and binders that can support polled consumers
will wire up an implementation `DefaultPollableMessageSource` which can
then be `poll()`ed.

The `poll()` method takes a `MessageHandler` callback to handle any message
returned by the poll; the method returns true if a message was found and
handled, false otherwise.

By default, acknowledging the message is deferred until the message handler
returns but that auto-acknowledging can be disabled if the user wishes to
defer the acknowledgment until later. Refer to the Spring Integration
documentation for more information (internally, a `MessageSourcePollingTemplate`
is used, which performs the ack/nack when the handler exits).

Usage:

```java
public interface PolledConsumer extends Processor {

	@Input("pollableSource")
	PollableMessageSource pollableSource();

}
```

and

```java
@Bean
public ApplicationRunner runner(PollableMessageSource pollableSource) {
	return args -> pollableSource.poll(message -> {
		System.out.println("Polled payload: " + message.getPayload());
	});
}
```


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -0,0 +1,117 @@
...
+import org.springframework.integration.endpoint.MessageSourcePollingTemplate;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.support.ChannelInterceptor;
+
+/**
+ * The default implementation of a {@link PollableMessageSource}.
+ *
+ * @author Gary Russell
+ * @since 5.0
```
olegz: 2.0 perhaps?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -0,0 +1,117 @@
...
+
+			@Override
+			public Object invoke(MethodInvocation invocation) throws Throwable {
+				Object result = invocation.proceed();
+				if (result instanceof Message) {
+					Message<?> received = (Message<?>) result;
+					for (ChannelInterceptor interceptor : this.interceptors) {
+						received = interceptor.preSend(received, null);
+						if (received == null) {
+							break;
```
olegz: Can we just simply `return received;` here instead of `break`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -152,6 +160,47 @@ public BindingService(
 		});
 	}
 
+	@SuppressWarnings({ "rawtypes", "unchecked" })
+	public <T> Binding<T> doBindPollableConsumer(T input, String inputName, Binder<T, ConsumerProperties, ?> binder,
```
olegz: Do we really need to spell it out as `doBindPollableConsumer`? The code up above `if (input instanceof PollableSource) {` would be much simpler if it was simply `doBindConsumer`. Since it's `PollableSource` it is rather obvious what type of consumer it will be, right?

garyrussell: We already have a `doBindConsumer` for the message-driven consumer; all this generic stuff makes it hard to collapse them into one method.

garyrussell:
I suggest holding off from further review/merge, until I finish the Rabbit implementation - finding a few quirks.

garyrussell:
Rabbit impl is now complete and awaiting review/merge of this before PRing.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -0,0 +1,253 @@
...
+					return null;
+				}, this.recoveryCallback);
+			}
+			AckUtils.autoAck(ackCallback);
+			return true;
+		}
+		catch (Exception e) {
+			AckUtils.autoNack(ackCallback);
+			throw (MessageHandlingException) e;
+		}
```
olegz: Would it be better to do something like this instead of ```catch(. . )```:
```
finally {
	if (!ackCallback.isAcknowledged()) {
		AckUtils.autoNack(ackCallback);
	}
. . .
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	private void doHandleMessage(MessageHandler handler, Message<?> message) {
+		try {
+			handler.handleMessage(message);
+		}
+		catch (Exception e) {
+			throw new MessageHandlingException(message, e);
+		}
+	}
```
olegz: Do we actually need `try/catch` here? It already throws ME by definition and I am assuming the `handler` in this case is always AMH, so it is handled there. 
Not sure, hence the question

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -0,0 +1,253 @@
...
+	public void setSource(MessageSource<?> source) {
+		ProxyFactory pf = new ProxyFactory(source);
+		class ReceiveAdvice implements MethodInterceptor {
+
+			private final List<ChannelInterceptor> interceptors = new ArrayList<>();
+
+			@Override
+			public Object invoke(MethodInvocation invocation) throws Throwable {
+				Object result = invocation.proceed();
+				if (result instanceof Message) {
```
olegz: Thinking simplification of logic. . . Is it actually possible here for the result to be not-Message?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -136,11 +140,54 @@ public void destroy() throws Exception {
...
+				if (psType != null) {
+					return GenericsUtils.getParameterType(binderInstance.getClass(), intf, 0)
+							.isAssignableFrom(psType);
+				}
+			}
+		}
+		return false;
+	}
+
+	private Class<?> findPollableSourceType(Class<?>[] targetInterfaces) {
```
olegz: Wondering if we should consider adding this to `org.springframework.cloud.stream.reflection.GenericsUtils`?
```
public static Class<?> findType(Class<?>[] targetInterfaces, Class<?> type) {. . }
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -136,11 +140,54 @@ public void destroy() throws Exception {
 			configurationName = name;
 		}
 		Binder<T, ?, ?> binderInstance = getBinderInstance(configurationName);
-		Assert.state(GenericsUtils.getParameterType(binderInstance.getClass(), Binder.class, 0).isAssignableFrom(bindingTargetType),
+		Assert.state(
+				(binderInstance instanceof PollableConsumerBinder
+						&& checkCompatiblePollableBinder(binderInstance, bindingTargetType))
+					|| GenericsUtils.getParameterType(binderInstance.getClass(), Binder.class, 0)
+							.isAssignableFrom(bindingTargetType),
```
olegz: Ouch. .  may be a private method with a descriptive name? Seems like a lot going on here

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PollableConsumerBinder.java
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * Configure a binding for a pollable message source.
+	 * @param name the binding name.
+	 * @param group the consumer group.
+	 * @param inboundBindTarget the binding target.
+	 * @param consumerProperties the consumer properties.
+	 * @return the binding.
+	 */
+	default Binding<PollableSource<H>> bindPollableConsumer(String name, String group,
```
olegz: I guess I need to see how it is used in the actual binder implementation, but at least for now I am wondering if it would be better NOT to provide a default implementation and instead force user to either implement the operation otherwise it is not `PollableConsumerBinder`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -286,6 +289,63 @@ else if (e instanceof ProvisioningException) {
 		}
 	}
 
+
+	@Override
+	public Binding<PollableSource<MessageHandler>> bindPollableConsumer(String name, String group,
```
olegz: should it be `doBind...` just to maintain symmetry with other methods?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SampleStreamApp.java
```diff
@@ -37,10 +41,11 @@
...
  *
  * @author Oleg Zhurakousky
+ * @author Gary Russell
  *
  */
 @SpringBootApplication
-@EnableBinding(Processor.class)
+@EnableBinding(SampleStreamApp.PolledConsumer.class)
 @Import(SpringIntegrationBinderConfiguration.class)
 public class SampleStreamApp {
```
olegz: Did you intent to modify this sample app?
When running it I get 
```
Caused by: java.lang.NullPointerException: null
	at org.springframework.cloud.stream.binder.DefaultPollableMessageSource.poll(DefaultPollableMessageSource.java:220) ~[classes/:na]
	at org.springframework.cloud.stream.binder.DefaultPollableMessageSource.poll(DefaultPollableMessageSource.java:1) ~[classes/:na]
	at org.springframework.cloud.stream.binder.integration.SampleStreamApp.lambda$0(SampleStreamApp.java:65) [test-classes/:na]
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:780) ~[spring-boot-2.0.0.M7.jar:2.0.0.M7]
	... 4 common frames omitted
```

garyrussell: Oops - I didn't do a final test; will take a look.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PollableConsumerBinder.java
garyrussell: Yes; this is a holdover from when I had this as a subinterface of Binder; the AMCB has a default that protects other binders (JMS, Kinesis).

I can create the PR for Rabbit if you like, but the PR won't pass travis until this is merged.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
garyrussell: Not in our code but we can't control what a user implementation might do. I tend to prefer a test over a possible CCE.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
garyrussell: Need to add the `failedMessage` to the MHE - otherwise the `ErrorMessage` won't have a reference and we can't publish it to a DLQ with diagnostic headers.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
garyrussell: No; this is the interface method; the others are `do...` because they are invoked by the super class. Only binders that implement `PollableConsumerBinder` have this method (which does not include the superclass).

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SampleStreamApp.java
garyrussell: Hmmm... runs fine for me...
```
Polled payload: polled data
Handling payload: Hello
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PollableConsumerBinder.java
olegz: That's fine, let's merge this one first

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/integration/SampleStreamApp.java
olegz: wait, I may have been making some modifications during the review (to better understand) and this could be totally on me. So disregard it for now

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
garyrussell: Decided to revert this; have to do a null check as well which is already done inside the util and

    finally {
        AckUtils.autoNack(ackCallback);
    }

just looks wrong.

Moved back to the catch.

PR #1182, opened 2018-01-16
Fixed support for 'partitionKeyExtractor' and 'partitionSelector' to …

olegz:
…be Spring configured

- Added new producer properties 'partitionKeyExtractorName' and 'partitionSelectorName'
- Deprecated 'partitionKeyExtractorClass' and 'partitionSelectorClass' properties
- Removed InitilaizingBean from MessageConverterConfigurer
- Updated documentation
- Added additional tests
- Removed PartitionedProducerTest  as redundant
- Fixed unrelated to this effort BinderPropertiesTests due to recent Boot changes


garyrussell:
Commit comment headlines must be <= 50 characters.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -165,50 +180,65 @@ private boolean isNativeEncodingNotSet(ProducerProperties producerProperties, Co
...
+			partitionKeyExtractor = instantiate(producerProperties.getPartitionKeyExtractorClass(), PartitionKeyExtractorStrategy.class);
 		}
-		return null;
+		else if (StringUtils.hasText(producerProperties.getPartitionKeyExtractorName())) {
+			partitionKeyExtractor = this.partitionKeyExtractors.get(producerProperties.getPartitionKeyExtractorName());
+			Assert.notNull(partitionKeyExtractor, "PartitionKeyExtractorStrategy bean with the name '" + producerProperties.getPartitionKeyExtractorName() 
+				+ "' can not be found. Has it been configured (e.g., @Bean)?");
+		}
+		else {
+			Assert.isTrue(CollectionUtils.isEmpty(this.partitionKeyExtractors) || this.partitionKeyExtractors.size() == 1,
```
garyrussell: Could simplify this to `...size() <= 1`.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -83,25 +88,35 @@
 	private final BindingServiceProperties bindingServiceProperties;
 
 	private ConfigurableListableBeanFactory beanFactory;
+	
+	private final Map<String, PartitionKeyExtractorStrategy> partitionKeyExtractors;
```
garyrussell: Need to configure STS to trim trailing blanks - I use the anyEdit plugin.

`git log -p --check` ...

screen shot 2018-01-16 at 2 09 00 pm

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -83,25 +88,35 @@
...
 	}
 
 	@Override
 	public void setBeanFactory(BeanFactory beanFactory) throws BeansException {
 		this.beanFactory = (ConfigurableListableBeanFactory) beanFactory;
 	}
 
-	@Override
-	public void afterPropertiesSet() throws Exception {
-		Assert.notNull(this.beanFactory, "Bean factory cannot be empty");
```
garyrussell: I think we still need this, otherwise we'll get this log warning...

    logger.warn("Creating EvaluationContext with no beanFactory", new RuntimeException("No beanFactory"));

from `ExpressionUtils`.

artembilan:
Is it going to be as a breaking change for downstream binder implementations?
Any instructions how to utilize this fix properly?
Thanks

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -165,50 +180,65 @@ private boolean isNativeEncodingNotSet(ProducerProperties producerProperties, Co
...
+			partitionSelector = instantiate(producerProperties.getPartitionSelectorClass(), PartitionSelectorStrategy.class);
+		}
+		else if (StringUtils.hasText(producerProperties.getPartitionSelectorName())) {
+			partitionSelector = this.partitionSelectors.get(producerProperties.getPartitionSelectorName());
+			Assert.notNull(partitionSelector, "PartitionSelectorStrategy bean with the name '" + producerProperties.getPartitionSelectorName() 
+				+ "' can not be found. Has it been configured (e.g., @Bean)?");
 		}
-		return new DefaultPartitionSelector();
+		else {
+			Assert.isTrue(CollectionUtils.isEmpty(this.partitionSelectors) || this.partitionSelectors.size() == 1,
```
garyrussell: ditto ( <= 1)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -83,25 +88,35 @@
...
+	@SuppressWarnings("unchecked")
+	public MessageConverterConfigurer(BindingServiceProperties bindingServiceProperties,
+			CompositeMessageConverterFactory compositeMessageConverterFactory,
+			Map<String, PartitionKeyExtractorStrategy> partitionKeyExtractors,
+			Map<String, PartitionSelectorStrategy> partitionSelectors) {
 		Assert.notNull(compositeMessageConverterFactory,
 				"The message converter factory cannot be null");
 		this.bindingServiceProperties = bindingServiceProperties;
 		this.compositeMessageConverterFactory = compositeMessageConverterFactory;
+		this.partitionKeyExtractors = partitionKeyExtractors == null ? Collections.EMPTY_MAP : partitionKeyExtractors;
```
garyrussell: Remove the `@SuppressWarnings` and use the type-safe variant - `Collections.emptyMap()`.

olegz:
@artembilan No breaking change. The two properties are are deprecated but will continue to function until some time in the future.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -165,50 +180,65 @@ private boolean isNativeEncodingNotSet(ProducerProperties producerProperties, Co
...
+	private <T> T instantiate(Class<?> implClass, Class<T> type) {
+		try {
+			return (T) implClass.newInstance();
 		}
-		else {
-			synchronized (this) {
-				T bean;
-				Class<?> clazz;
-				try {
-					clazz = ClassUtils.forName(className, this.beanFactory.getBeanClassLoader());
```
garyrussell: Why switch to `newInstance()` instead of `ClassUtils`?

olegz: If you look a the code and how it was before they already had `Class` (see ProducerProperties) which they were passing as `class.getName` and then creating a class again. So all that it really needed is new instance on the already existing class

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
olegz: Hmm, but this was just validating that it is there and in our case due to BFAware it always is since its a bean. Right?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
garyrussell: Doh - yes, of course.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
garyrussell: True, but only as long as nobody instantiates it outside of Spring for some reason.

olegz: And I think I am ok with that since outside of Spring this method may not even execute (unless someone calls it explicitly)

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/MessageConverterConfigurer.java
```diff
@@ -165,50 +178,65 @@ private boolean isNativeEncodingNotSet(ProducerProperties producerProperties, Co
...
+			partitionKeyExtractor = instantiate(producerProperties.getPartitionKeyExtractorClass(), PartitionKeyExtractorStrategy.class);
+		}
+		else if (StringUtils.hasText(producerProperties.getPartitionKeyExtractorName())) {
+			partitionKeyExtractor = this.partitionKeyExtractors.get(producerProperties.getPartitionKeyExtractorName());
+			Assert.notNull(partitionKeyExtractor, "PartitionKeyExtractorStrategy bean with the name '" + producerProperties.getPartitionKeyExtractorName()
+				+ "' can not be found. Has it been configured (e.g., @Bean)?");
 		}
-		return null;
+		else {
+			Assert.isTrue(CollectionUtils.isEmpty(this.partitionKeyExtractors) || this.partitionKeyExtractors.size() <= 1,
```
garyrussell: I think you misunderstood - we don't need the isEmpty check, it's handled by <= 1 - I can fix on merge if you agree.

olegz: oops, yeah, please! Thank you

garyrussell:
Merged as 00748985d622e84a25b23d16585d9d35e1be3816 with minor polish.

Issue #1186, opened 2018-01-17
Document Polled Consumer


garyrussell:
1) Add a sample to SCSt Samples repo
2) Include code snippet in the docs and point to the samples repo for the full example

PR #1190, opened 2018-01-23
GH-1186: Polled Consumer Documentation

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/1186

### DIFF
--- a/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
+++ b/spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
@@ -149,6 +149,18 @@ image::SCSt-groups.png[width=300,scaledwidth="50%"]
 All groups which subscribe to a given destination receive a copy of published data, but only one member of each group receives a given message from that destination.
 By default, when a group is not specified, Spring Cloud Stream assigns the application to an anonymous and independent single-member consumer group that is in a publish-subscribe relationship with all other consumer groups.
 
+[[consumer-types]]
+=== Consumer Types
+
+Two types of consumer are supported:
+
+- Message-driven (sometimes referred to as Asynchronous)
+- Polled (sometimes referred to as Synchronous)
+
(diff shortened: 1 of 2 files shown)


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -552,6 +584,80 @@ public static class TestPojoWithAnnotatedArguments {
...
+            }
+        }
+    };
+}
+----
+
+The `PollableMessageSource.poll()` method takes a `MessageHandler` argument (often a lambda expression as shown here).
+It returns `true` if the message was received and successfully processed.
+
+As with message-driven consumers, if the `MessageHandler` throws an exception, messages are published to error channels as discussed in <<binder-error-channels>>.
```
olegz: Gary, just trying to correlate this sentence to the above code and thinking it may be a bit confusing. The code above is already written to handle exception. May be remove` try/catch`? (I know you need it to `Thread.sleep()`, but for this example you can probably use `LockSupport` or something)

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -552,6 +584,80 @@ public static class TestPojoWithAnnotatedArguments {
...
+
+[source,java]
+----
+public interface PolledConsumer {
+
+    @Input
+    PollableMessageSource dest1In();
+
+    @Output
+    MessageChannel dest2Out();
```
olegz: For the sake of documentation clarity can in be simply `destIn` and `destOut`?

spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
garyrussell: maybe 

    // handle failure (throw an exception to reject the message)

olegz: I'm ok with that

PR #1197, opened 2018-01-26
Pollable Consumer Payload Conversion

garyrussell:
- support payload conversion for polled consumers

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
@@ -27,6 +27,7 @@ import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.aop.support.NameMatchMethodPointcutAdvisor;
 import org.springframework.context.Lifecycle;
 import org.springframework.core.AttributeAccessor;
+import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.core.MessagingTemplate;
 import org.springframework.integration.support.AckUtils;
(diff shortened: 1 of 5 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -74,7 +77,9 @@
 
 	private BiConsumer<AttributeAccessor, Message<?>> attributesProvider;
 
-	private volatile boolean running;
+	private SmartMessageConverter messageConverter;
+
+	private boolean running;
```
artembilan: Well, the pattern is still to have it with a`volatile`. Even if you `synchronized` there is still possibility to read from CPU cache, but not jump to the main memory for the actual value. That's exactly why `volatile` has been introduced.
Although with the fixes against Meltdown and Spectre we may not need it anymore because we don't do speculative execution 😄 . Or maybe opposite: since we have done speculative execution before we could read data from other threads. Therefore it is more dangerous do not use `volatile`. Need to think yet...

Just IMHO, but we have convention in Spring Integration do not use `volatile` only for properties which mostly be changed only once during configuration. The multi-thread access must be supplied with `volatile` anyway.
And that's exactly how it isn't good to rely on the `ArrayList.isEmpty()` because its `size` property isn't not `volatile`.

olegz:
@garyrussell I'll try to look at it later today. . . What would be interesting is to see how it gives with https://github.com/spring-cloud/spring-cloud-stream/pull/1198. In any event, my goal is to ensure the consistency - all content-type conversion must go through CompositeMessageConverter (and converters within it) as the event driven consumers

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
garyrussell: No; volatile is not needed when using `synchronized` - synchronization flushes the cache.

garyrussell:
Right - it should get the same treatment as 1198 - currently we get the double conversion here too.

PR #1207, opened 2018-02-02
Schema Server small fixes

### LINKED ISSUE #1202: Inconsistent use of this.prefix in AvroSchemaRegistryClientMessageConverter.resolveSchemaForWriting
Currently the prefix is hard coded in `AvroSchemaRegistryClientMessageConverter.resolveSchemaForWriting` to `vnd`:
 
```java
		if (headers instanceof MutableMessageHeaders) {
			headers.put(MessageHeaders.CONTENT_TYPE,
					"application/vnd." + schemaReference.getSubject() + ".v"
							+ schemaReference.getVersion() + "+avro");
}
```
I suspect the `prefix` field should be used consistently: 
```java
		if (headers instanceof MutableMessageHeaders) {
			headers.put(MessageHeaders.CONTENT_TYPE,
					"application/" + this.prefix + "." + schemaReference.getSubject() 
						+ ".v" + schemaReference.getVersion() + "+avro");
		}
```

### LINKED ISSUE #1206: Include more jdbc drivers on the schema registry server pom
Currently we only ship with H2, but we could copy what the dataflow and skipper servers do today, by shipping MariaDB/PostgreSQL/SQLServer drivers as well which would allow users to just switch them via properties not requiring to rebuild the uber jar and include the drivers themselves.

viniciusccarvalho:
Fixes #1206
- Added extra JDBC drivers (mysql, postgresql, sqlserver, hsqldb)
- Included boot plugin to create runnable jar for the server
- Added ddl update to true for the default application.yml

### DIFF
--- a/spring-cloud-stream-schema-server/pom.xml
+++ b/spring-cloud-stream-schema-server/pom.xml
@@ -27,7 +27,29 @@
 		<dependency>
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
-			<version>1.4.192</version>
+		</dependency>
+		<dependency>
+			<groupId>org.hsqldb</groupId>
+			<artifactId>hsqldb</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.mariadb.jdbc</groupId>
+			<artifactId>mariadb-java-client</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.postgresql</groupId>
+			<artifactId>postgresql</artifactId>
+		</dependency>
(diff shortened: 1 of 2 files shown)


olegz:
Vinicius, after thinking some more about it, I don't believe this is something we want to support at the moment unless there is a push from the community with a tangible argument. My biggest concern is  potential dependency conflicts that may be transitively pulled by each driver.
What I would consider is adding something in the docs or project WIKI ,so in case questions arise on how to bring additional drivers as well as which drivers we support we can point users to such doc.

PR #1216, opened 2018-02-13
1215 CachingCachingRegistryClient stores data by wrong keys resulting in fetch operations always missing.

jmax01:
1215  CachingCachingRegistryClient stores data by wrong keys resulting in fetch operations always missing.

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/CachingRegistryClient.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/CachingRegistryClient.java
@@ -47,8 +47,8 @@ public class CachingRegistryClient implements SchemaRegistryClient {
 	@Override
 	public SchemaRegistrationResponse register(String subject, String format, String schema) {
 		SchemaRegistrationResponse response = delegate.register(subject, format, schema);
-		cacheManager.getCache(ID_CACHE).put(response.getSchemaReference(), schema);
-		cacheManager.getCache(REF_CACHE).put(response.getId(), schema);
+		cacheManager.getCache(ID_CACHE).put(response.getId(), schema);
+		cacheManager.getCache(REF_CACHE).put(response.getSchemaReference(), schema);
 		return response;
 	}
 


spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/CachingRegistryClient.java
```diff
@@ -47,8 +47,8 @@ public CachingRegistryClient(SchemaRegistryClient delegate) {
 	@Override
 	public SchemaRegistrationResponse register(String subject, String format, String schema) {
 		SchemaRegistrationResponse response = delegate.register(subject, format, schema);
-		cacheManager.getCache(ID_CACHE).put(response.getSchemaReference(), schema);
-		cacheManager.getCache(REF_CACHE).put(response.getId(), schema);
+		cacheManager.getCache(ID_CACHE).put(response.getId(), schema);
+		cacheManager.getCache(REF_CACHE).put(response.getSchemaReference(), schema);
```
olegz: ouch! Good catch!

jmax01: Can you guys back port this?

olegz:
@jmax01 I'll start reviewing, polishing and merging it, but wanted to make one general comment. As you have probably noticed especially with this issue in particular, this could have been easily caught if there was a test for it, but I gather there wasn't. I'll write one up as part of polishing unless you're willing to contribute one, then I'll wait. LMK. Cheers

jmax01:
I would love to but am hyper busy at the moment. I will try if I can get a few.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/CachingRegistryClient.java
olegz: Sure, we can back port this, but I now wonder, why do we need two caches? We use the same delegate, cache is just a K/V pair. Should it really just be `schemaCache`?
```
cacheManager.getCache("org.springframework.cloud.stream.schema.client.schemaCache").put(response.getId(), schema);
cacheManager.getCache("org.springframework.cloud.stream.schema.client.schemaCache").put(response.getSchemaReference(), schema);
```
@viniciusccarvalho ^^

viniciusccarvalho: We decided that since they are two different entities we would put them on different caches as it's good practice with caches usually. One holds ints->schema, the other object->schema

PR #1220, opened 2018-02-13
1217 AvroSchemaRegistryClientMessageConverter registers schema formats inconsistently.

jmax01:
1217 AvroSchemaRegistryClientMessageConverter registers schema formats inconsistently.

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/ParsedSchema.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/ParsedSchema.java
@@ -42,10 +42,14 @@ public class ParsedSchema {
 	private SchemaRegistrationResponse registration;
 
 	public ParsedSchema(Schema schema) {
-		this.schema = schema;
-		this.representation = schema.toString();
+		this(schema,false);
 	}
 
+	public ParsedSchema(Schema schema, boolean prettyPrintrepresentation) {
+		this.schema = schema;
+		this.representation = schema.toString(prettyPrintrepresentation);
+	}
+	
 	public Schema getSchema() {
 		return schema;
 	}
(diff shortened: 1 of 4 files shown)


jmax01:
ugh missed checkstyle issue

olegz:
I am not so sure about this one yet 
1. Why do we care about the visual format? If I understand correctly it doesn't affect functionality of registration, retrieval or applying the schema. If so why, this is purely a question of visual representation, correct?
2. If we were to go with "prettyPrint", why should we than do it always to make it "consistent"?

jmax01:
What is stored in the DB?
If I am not incorrect, the representation is stored in the DB without modification.
As a user I would want them stored in a consistent fashion.

viniciusccarvalho:
What is the inconsistency? Do you want data to be stored in pretty print? Is that the case? Inconsistency to me is if data is malformed. Don't even start on the whole human readable discusion with me...

olegz:
@jmax01 Jus to clarify what @viniciusccarvalho said, presentation of the schema is out of scope  of schema  _application, storage and retrieval_ . Let's assume there is some UI to display schemas, then that responsibility would fall into such UI and as far as I know there are plenty of browser plugins and other tools to support that. 
So, I'll reject this one.
In any event thank you for your continuous contributions to the framework!

jmax01:
Here is the inconsistency: 
`AvroSchemaRegistryClientMessageConverter.afterPropertiesSet`:
``` java
    //with print pretty	
    this.schemaRegistryClient.register(toSubject(schema), AVRO_FORMAT,
							schema.toString(true));
```
`AvroSchemaRegistryClientMessageConverter.resolveSchemaForWriting`:

``` java
  //The ParsedSchema.representation is not print pretty
  SchemaRegistrationResponse response = this.schemaRegistryClient.register(
  toSubject(schema), AVRO_FORMAT, parsedSchema.getRepresentation());
```
If you aren't going to allow configuration at least make it consistent, is that too much to ask?
I would recommend not storing them pretty printed to minimize overhead (generating, parsing, wire, storage, etc).

olegz:
@jmax01 the above is indeed a valid point with regard to inconsistency and the one I missed.
But what I propose is to actually call a no arg `toString`.
Feel free to submit another PR for that if you agree.

Once again, thank you for helping out with this

jmax01:
Sure will do.

olegz:
@jmax01 and just to end it on the fun note I liked the `If I am not incorrect` ;) took me a while to process it ;)

PR #1227, opened 2018-02-14
GH-1219 Fixed channel instrumentation in BinderAwareChannelResolver

### LINKED ISSUE #1219: Can't instrument channels created via BinderAwareChannelResolver
when resolving a message channel dynamically there seems to be no option to hook in a channel interceptor. In case of Sleuth it's called `TraceChannelInterceptor`. Due to missing channel interceptor a message is not instrumented and trace is not continued.

Affected Cloud BOM versions: `Edgware`, `Finchley`
Sample project: https://github.com/marcingrzejszczak/trace-id-issue-reproducer/tree/edgware_2
Related issue: https://github.com/spring-cloud/spring-cloud-sleuth/issues/821

olegz:
- added support for instrumenting dynamic channels with global interceptors
- added test

Resolves #1219

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2017 the original author or authors.
+ * Copyright 2013-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,11 +16,16 @@
 
 package org.springframework.cloud.stream.binding;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.cloud.stream.binder.Binding;
 import org.springframework.cloud.stream.binder.ProducerProperties;
 import org.springframework.cloud.stream.config.BindingServiceProperties;
+import org.springframework.integration.channel.ChannelInterceptorAware;
+import org.springframework.integration.config.GlobalChannelInterceptorProcessor;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.core.BeanFactoryMessageChannelDestinationResolver;
 import org.springframework.messaging.core.DestinationResolutionException;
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BinderAwareChannelResolver.java
```diff
@@ -151,5 +158,18 @@ void configure(String channelName, MessageChannel channel, ProducerProperties pr
...
+	/**
+	 * Will add any globally available interceptors to the MessageChannel
+	 * 
+	 * See: https://github.com/spring-cloud/spring-cloud-stream/issues/1219
+	 */
+	private void instrumentChannel(MessageChannel messageChannel) {
+		Collection<GlobalChannelInterceptorWrapper> interceptors =
+				this.beanFactory.getBeansOfType(GlobalChannelInterceptorWrapper.class).values();
+		for (GlobalChannelInterceptorWrapper globalChannelInterceptorWrapper : interceptors) {
+			((AbstractMessageChannel)messageChannel).addInterceptor(globalChannelInterceptorWrapper.getChannelInterceptor());
```
artembilan: You do this unconditionally. Any `@GlobalChannelInterceptor` contributes `patterns` to the `GlobalChannelInterceptorWrapper`. So, if we go this way you have to use its `getPatterns()` over the `PatternMatchUtils.smartMatch()` to be sure that we definitely need it apply such an interceptor to the channel.

The `(AbstractMessageChannel)` cast is not good as well since it is a class and in case of Proxy the cast are not going to work. You need to cast to the `ChannelInterceptorAware` (if that is possible at all)...

See `GlobalChannelInterceptorProcessor` for more information.

Moreover I'm still not sure why you didn't go the way to inject all the end-user `MessageChannelConfigurer`s, which are going to be called from the `SubscribableChannelBindingTargetFactory` for any channel created by the SCSt.

Also pay attention how you call `this.beanFactory.getBeansOfType()` for each channel we create here.

Does it make sense to you?

olegz: The reason why it is not in `SubscribableChannelBindingTargetFactory` is because that is where we add our interceptors, the Tracing and others are added via `GlobalChannelInterceptorProcessor` so if I add this code there I need to make sure that the interceptors added via `GlobalChannelInterceptorProcessor` aren't added again. In other words it is only the concern for dynamic channels. 
The rest makes sense, so I'll polish.
Basically with everything that has been discussed here about `TraceChannelInterceptor` created a bit of a confusion . . . the solution turned out to be much simpler -  he interceptor was simply never added to a channel.
Now I can see all the sleuth traces

artembilan:
@olegz ,

Looking at this one more time I think that it would be better to implement (today) https://jira.spring.io/browse/INT-4402 now and back-port to SI-4.3.x. The problem when it will be implemented there (and it must), you will have to remove this fix. 

My thoughts for this fix here were around `MessageChannelConfigurer`s injections into the:
```
@Bean
public CompositeMessageChannelConfigurer compositeMessageChannelConfigurer(
```
The point is like end-user may decide to do some other channel customization, not only interceptors configuration. Right now there is no any other way unless `BPP`, although we all already know how dangerous it may be.

But since you go another way, I believe that we don't need it at all since SI fix will cover this as well.

WDYT?

garyrussell:
See https://github.com/spring-projects/spring-integration/pull/2363

Issue #1062, opened 2017-09-01
New Boot 2.0 actuator to manage bindings


viniciusccarvalho:
Create a new Boot 2.0 actuator endpoint with  `@WriteOperation` to allow control of Binding lifecycle such as start/stop, play/pause/seek (when applicable)

PR #1239, opened 2018-02-21
GH-1062 Added initial support for actuator endpoint for managing input bindings

olegz:
This is initial crack at actuator endpoints for managing binding lifecycle. Currently only for Input bindings.
To test:
Stop: `curl  -H "Content-Type: application/json" -X POST http://localhost:8080/actuator/bindings/stop/inOne`
Start: `curl  -H "Content-Type: application/json" -X POST http://localhost:8080/actuator/bindings/start/inOne`
. . . where `inOne` is the channel name
The `http://localhost:8080/actuator/bindings/` url  (GET method) will simply display available bindings
To enable actuator with new boot supply this property `--management.endpoints.web.exposure.include=*`

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.stream.binder;
 
+import org.springframework.context.Lifecycle;
+
 /**
  * Represents a binding between an input or output and an adapter endpoint that connects
  * via a Binder. The binding could be for a consumer or a producer. A consumer binding
@@ -26,9 +28,46 @@ package org.springframework.cloud.stream.binder;
  * @author Mark Fisher
  * @author Gary Russell
  * @author Marius Bogoevici
+ * @author Oleg Zhurakousky
+ * 
  * @see org.springframework.cloud.stream.annotation.EnableBinding
  */
-public interface Binding<T> {
+public interface Binding<T> extends Lifecycle {
+	
+	/**
+	 * Stops the target component represented by this instance.
+	 * NOTE: At the time the instance is created the component is already started. 
+	 * This operation is typically used by actuator to re-bind/re-start.
+	 * 
+	 * @see BindingsEndpoint
+	 */
+	default void start() {}
+	
+	/**
+	 * Starts the target component represented by this instance.
+	 * NOTE: At the time the instance is created the component is already started. 
+	 * This operation is typically used by actuator to re-bind/re-start.
+	 * 
+	 * @see BindingsEndpoint
+	 */
+	default void stop() {}
+	
+	/**
+	 * Returns 'true' if the target component represented by this instance is running.
+	 */
+	default boolean isRunning() {
+		return false;
+	}
+	
+	/**
+	 * Returns the name of this binding  (i.e., channel name)
+	 * 
+	 * @return binding name
+	 */
+	default String getName() {
+		return null;
+	}
+		
 
 	/**
 	 * Unbinds the target component represented by this instance and stops any active
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2016 the original author or authors.
+ * Copyright 2013-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,10 +16,14 @@
 
 package org.springframework.cloud.stream.binder;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.context.Lifecycle;
 import org.springframework.integration.support.context.NamedComponent;
 import org.springframework.util.Assert;
 import org.springframework.util.ObjectUtils;
+import org.springframework.util.StringUtils;
 
 /**
  * Default implementation for a {@link Binding}.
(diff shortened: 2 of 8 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.stream.binder;
 
+import org.springframework.cloud.stream.endpoint.BindingsEndpoint;
```
artembilan: We don't use imports when class is used only from the JavaDocs.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -26,9 +28,30 @@
...
  * @author Gary Russell
  * @author Marius Bogoevici
+ * @author Oleg Zhurakousky
+ * 
  * @see org.springframework.cloud.stream.annotation.EnableBinding
  */
 public interface Binding<T> {
+	
+	/**
+	 * Binds the target component represented by this instance.
```
artembilan: The JavaDocs must be in an an imperative style: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting. Also no blank lines in the method JavaDocs.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -64,9 +71,21 @@ public String getName() {
...
 	}
+	
+	@Override
+	public final synchronized void bind() {
+		if (this.lifecycle != null && !this.lifecycle.isRunning()) {
+			if (StringUtils.hasText(this.group)) {
+				this.lifecycle.start();
+			}
+			else {
+				logger.warn("Can not re-bind an anonymous binding");
```
artembilan: Won't this happen at start up time? Or do you mean that `group` is not empty there yet, but it is cleaned afterwards ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/Bindable.java
```diff
@@ -31,8 +34,23 @@
...
 	default void bindInputs(BindingService adapter) {}
+	
+	/**
+	 * Binds all the inputs associated with this instance.
+	 */
+	
+	/**
+	 * Binds all the inputs associated with this instance.
+	 * @param adapter instance of {@link BindingService}
+	 * @return collection of {@link Binding}s
```
artembilan: We usually add `@since ` to all the new public API.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
```diff
@@ -218,9 +230,10 @@ public void bindInputs(BindingService bindingService) {
 				if (log.isDebugEnabled()) {
 					log.debug(String.format("Binding %s:%s:%s", this.namespace, this.type, inputTargetName));
 				}
-				bindingService.bindConsumer(boundTargetHolder.getBoundTarget(), inputTargetName);
+				bindings.addAll(bindingService.bindConsumer(boundTargetHolder.getBoundTarget(), inputTargetName));
 			}
 		}
+		return bindings;
```
artembilan: Shound't we worry here about an *immutability* of the collection to return?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/BindingsEndpoint.java
```diff
@@ -0,0 +1,105 @@
...
+	@Endpoint(id = "bindings/stop/{name}")
+	public class StopEndpoint {	
+		@WriteOperation
+		public boolean stop(String name) {
+			List<Binding<Object>> inputBindings = BindingsEndpoint.this.gatherInputBindings();
+			
+			for (Binding<Object> binding : inputBindings) {
+				String bindingName = binding.getName();
+				if (name.equals(bindingName)) {
+					binding.unbind();
```
artembilan: Maybe we should `return true` exactly here (not sure if we can have more than one binding with that name) and `false` in the end of method to indicate that there is nothing to stop?

Also how about to consider to implement this logic via Java `Stream` ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
olegz: see the comment on the interface. It's all a bit awkward but at the moment I have to deal with what I have and in this case it's only to call IF and only IF you called `unbind` before, otherwise its idempotent

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindableProxyFactory.java
olegz: may be, I'll look how it is used or can be used

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/BindingsEndpoint.java
olegz: sure, once the approach is settled on

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/Bindable.java
```diff
@@ -31,8 +34,23 @@
...
+	 */
+	default Collection<Binding<Object>> bindInputBindings(BindingService adapter) {
+		return null;
```
sobychacko: may be an empty collection rather than returning a null.

garyrussell:
Completely unbinding on `stop()` will destroy all the error channel scaffolding set up in the `AbstractMessageChannelBinder`, and other stuff (see `AMCB.afterUnbindConsumer()` - called from `DefaultBinding.afterUnbind()`.

I don't think it has to be this complicated; can't we simply stop/start the Lifecyle (if present) in the `Binding`.

Why not just `Binding extends Lifecycle`, with the lifecycle methods delegating to the `Lifecycle` therein?

olegz:
@garyrussell good point and indeed at some point I had something similar to that, but given the existing `unbind` method i fell victim on calling it.
But yes I agree that simply delegating to the internal Lifecycle would do and its a simple change

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -64,12 +72,33 @@ public String getName() {
...
+		return this.lifecycle != null && this.lifecycle.isRunning();
+	}
+	
 	@Override
-	public final void unbind() {
-		if (this.lifecycle != null) {
+	public final synchronized void start() {
+		if (!this.isRunning()) {
+			if (StringUtils.hasText(this.group)) {
+				this.lifecycle.start();
```
garyrussell: NPE if `lifecycle == null`.

PR #1244, opened 2018-02-23
Make Actuator optional

dsyer:


olegz:
After discussing it off line with Dave we're going to target this for GA. The PR seems to solve the issue and we just need to test it to ensure it doesn't break anything esle

Issue #1055, opened 2017-09-01
New interfaces for binding control


viniciusccarvalho:
Provide a couple of new interfaces that binders may implement:

```java
public interface Seekable {
   boolean seek(Long offset);
}
```

```java
public interface Resumable
  boolean pause();
  boolean play();
}
```

Those two interfaces will be exposed via Actuator endpoints for users to play/pause/seek consumers

Implementations may reuse the lifecycle methods start/stop, but it may not be the case.

Start/Stop still plays a role for producers if setting `autoStartEnabled`

markfisher:
`play()` should be `resume()`, and I'd argue that the interface should be `Pausable` since you first have to pause before you can resume ;)

olegz:
Moved it to RC2, given the release timings issue of https://github.com/spring-projects/spring-integration-kafka/pull/199/
Basically we can't release RC1 depending on snapshot of spring-integration-kafka and the above issue, just missed SI 5.0.2 by a day ;)

sobychacko:
Closing via https://github.com/spring-cloud/spring-cloud-stream/commit/dff7416a7ded0d37152ca11f741d727feb8ca742.

yharsh:
How to disable a channel to get configure during application boot up using something like `autoStartEnabled`.

PR #1264, opened 2018-02-27
GH-1055 Added support for Pausable

olegz:
Added support for Pausable lifecycle to Binding and its implementation
Added additional actuator endpoints

Resolves spring-cloud/spring-cloud-stream#1055

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.stream.binder;
 
-import org.springframework.context.Lifecycle;
 import org.springframework.integration.endpoint.Pausable;
 
 /**
@@ -33,7 +32,7 @@ import org.springframework.integration.endpoint.Pausable;
  * 
  * @see org.springframework.cloud.stream.annotation.EnableBinding
  */
-public interface Binding<T> extends Lifecycle, Pausable {
+public interface Binding<T> extends Pausable {
 	
 	/**
 	 * Stops the target component represented by this instance.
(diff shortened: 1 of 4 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
```diff
@@ -32,7 +33,7 @@
  * 
  * @see org.springframework.cloud.stream.annotation.EnableBinding
  */
-public interface Binding<T> extends Lifecycle {
+public interface Binding<T> extends Lifecycle, Pausable {
```
garyrussell: Pausable extends Lifecycle

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
```diff
@@ -95,6 +96,20 @@ public final synchronized void stop() {
 			this.lifecycle.stop();
 		}
 	}
+	
+	@Override
+	public final synchronized void pause() {
+		if (this.lifecycle instanceof Pausable) {
+			((Pausable)this.lifecycle).pause();
```
garyrussell: space after `)` (checkstyle will complain).

garyrussell: I wonder if the actuator should return an error if it's not a pausable binding?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/Binding.java
olegz: nice!

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinding.java
olegz: it didn't but i'll fix

olegz: may be a WARN for now. If we return something now we'll set the expectations and I am not ready for it yet

sobychacko:
Tested this PR against Kafka binder. LGTM. Merging.

Issue #1267, opened 2018-02-28
Add additional lifecycle hooks related to MeterRegistry


olegz:
This is to support https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/320

PR #1268, opened 2018-02-28
GH-1267 Created BindingCreatedEvent

olegz:
- updated AbstractChannelBinder to fire this event whenever bindings are created

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -25,12 +25,15 @@ import org.apache.commons.logging.Log;
 
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -824,5 +844,12 @@ ErrorInfrastructure getErrorInfrastructure() {
 		}
 
 	}
+	
+	@SuppressWarnings("serial")
+	public class BindingCreatedEvent extends ApplicationEvent {
```
artembilan: Great! But can we make this as a top-level class?
And may be extract a super class like `StreamEvent`?
Similar to what we have in SI with the `IntegrationEvent`

olegz: Top level yes, good point
As far as adding `super`. . . not ready for it yet, will revisit for 2.1

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/ec0966875cf20687c6e1e53159e53eefb5e71b08

Issue #1273, opened 2018-03-04
Multiple-systems configuration always inherit default environment - Finchley.M8/Spring Boot 2.0


wildloop:
[My configuration](https://github.com/wildloop/streams-demo/blob/sb2.0/src/main/resources/application-2vh.yml) based on example from documentation ["Connecting to Multiple Systems"](https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/#multiple-systems):

```json
spring:
  rabbitmq:
    username: demo
    password: demo
    virtualHost: demo
  cloud:
    stream:
      binders:
        demoin:
          type: rabbit
          defaultCandidate: false
          inheritEnvironment: false
          environment:
            spring:
              rabbitmq:
                username: demo
                password: demo
                virtualHost: demo-a
        demoout:
          type: rabbit
          defaultCandidate: false
          inheritEnvironment: false
          environment:
            spring:
              rabbitmq:
                username: demo
                password: demo
                virtualHost: demo-b
      bindings:
        input:
          binder: demoin
          destination: in-demo-a-dest
          group: in-demo-a-group
        output:
          binder: demoout
          destination: out-demo-b-dest
          group: out-demo-b-group
```

This config should cause two connections as binders:
- demoin: virtual-host demo-a
- demoout: virtual-host demo-b

but we get (log):
- o.s.a.r.c.CachingConnectionFactory: Created new connection: rabbitConnectionFactory#2121d1f9:0/SimpleConnection@4a481728 [delegate=amqp://demo@127.0.0.1:5672/**demo**, localPort= 51692]
- o.s.a.r.c.CachingConnectionFactory: Created new connection: rabbitConnectionFactory#76bf1bb8:0/SimpleConnection@1b482cbf [delegate=amqp://demo@127.0.0.1:5672/**demo**, localPort= 51694]

and RabbitMQ Admin UI screen:

![rabbit-vh](https://user-images.githubusercontent.com/2461648/36950871-bcb3e468-1ffc-11e8-9b56-f4274f73cc6c.png)

[This configuration works](https://github.com/wildloop/streams-demo/blob/master/src/main/resources/application-2vh.yml) in the Spring Cloud Edgware.SR2 and Spring Boot 1.5.9.RELEASE

olegz:
@wildloop thank you for pointing this out. It appears to be related to the following issues https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/326
[...]

PR #1276, opened 2018-03-05
GH-1273 Added map flattening for binder configuration properties

olegz:
polishing


garyrussell:
A bit cleaner, perhaps?

```java
private void flatten(String propertyName, Object value, Map<String, String> flattenedProperties) {
	if (value instanceof Map) {
		((Map<Object, Object>) value).forEach((k, v) ->
			flatten((propertyName != null ? propertyName + "." : "") + k, v, flattenedProperties));
	}
	else {
		flattenedProperties.put(propertyName, value.toString());
	}
}
```

and `flatten(null, binderConfiguration.getProperties(), binderProperties);`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -165,11 +180,14 @@ public void destroy() throws Exception {
 			Assert.state(binderConfiguration != null, "Unknown binder configuration: " + configurationName);
 			BinderType binderType = this.binderTypeRegistry.get(binderConfiguration.getBinderType());
 			Assert.notNull(binderType, "Binder type " + binderConfiguration.getBinderType() + " is not defined");
-			Map<Object, Object> binderProperties = binderConfiguration.getProperties();
+			
+			Map<String, String> binderProperties = new HashMap<>();
+			this.flatten("", binderConfiguration.getProperties(), binderProperties);
```
garyrussell: Usual convention is to put called methods below callees. (and drop the `this.` on method calls).

PR #1282, opened 2018-03-06
RB-GH-134: Event Publisher: Add protected getter

garyrussell:
See https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/134

Also use the `...Aware` interface instead of `@Autowired`.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -25,7 +25,6 @@ import org.apache.commons.logging.Log;
 
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.support.DefaultSingletonBeanRegistry;
 import org.springframework.cloud.stream.provisioning.ConsumerDestination;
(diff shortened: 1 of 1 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -93,7 +94,6 @@
 	 */
 	private final String[] headersToEmbed;
 
-	@Autowired
```
olegz: I am almost considering a full scan for 2.1 to do exactly that. We should not be using in internally. Ever!

garyrussell: Yeah, exactly - that's why I was a bit shocked to see you added it and @artembilan merged it 😄 

https://github.com/spring-cloud/spring-cloud-stream/commit/ec0966875cf20687c6e1e53159e53eefb5e71b08#diff-737803e2a91ac21a17baf06ff7b4cbacR96

olegz: Yeah, I'll admit that one's on me. I was trying the publisher route and didn't want to change a lot of things and forgot to clean up.

artembilan: What is the problem here though?
I have just tested Kafka Binder for catching events from here and I have this:
```
event = {BindingCreatedEvent@8580} "org.springframework.cloud.stream.binder.BindingCreatedEvent[source= Binding [name=input, target=input, lifecycle=null]]"
this.applicationEventPublisher = {AnnotationConfigApplicationContext@8581} "org.springframework.context.annotation.AnnotationConfigApplicationContext@3df04fa1: startup date [Wed Mar 07 10:49:41 EST 2018]; parent: org.springframework.context.annotation.AnnotationConfigApplicationContext@2ffe243f"
```

I understand that it might not be good to only rely on the annotation configuration in the Framework code, but in most cases (Spring Boot ?) this `@Autowired` works well.

What am I missing in this discussion? Non of your comments enlighten me what is the problem.

Thanks

Issue #1286, opened 2018-03-07
Rework bindings endpoint to avoid using verbs in URIs


wilkinsona:
It's generally considered bad practice to use verbs in a URI of a RESTful API (even for very loose definitions of RESTful):

 - https://apigee.com/about/blog/technology/restful-api-design-nouns-are-good-verbs-are-bad
 - https://blog.philipphauer.de/restful-api-design-best-practices/#use-nouns-instead-of-verbs-for-resources

The current bindings endpoint exposes four different resources, each with a verb in its URI, for controlling the state of a binding:

- `http://<host>:<port>/actuator/bindings/start/{name}`
- `http://<host>:<port>/actuator/bindings/stop/{name}`
- `http://<host>:<port>/actuator/bindings/pause/{name}`
- `http://<host>:<port>/actuator/bindings/resume/{name}`

Each resource only supports `POST` requests.

The approach that's generally recommend for this sort of thing is to provide a single `state` resource instead:

- `http://<host>:<port>/actuator/bindings/state/{name}`

A `GET` request can be used to query the resource's current state and a `POST` request can be used to update the state. The payload of the `POST` request is used to indicate the desired state of the resource. For example, a stopped binding could be started by making a `POST` request with the following body:

```
{
    "state": "STARTED"
}
```

olegz:
Given that this is new functionality that has never been released publicly we can certainly change it and what Andy is saying makes perfect sense.

sobychacko:
Fixed via https://github.com/spring-cloud/spring-cloud-stream/commit/e939e09e95f51cffa16354721ccae6765e11cb25

PR #1288, opened 2018-03-07
GH-1286 Reworked 'bindings' endpoint

olegz:


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/BindingsEndpoint.java
```diff
@@ -62,25 +53,41 @@ public BindingsEndpoint(List<InputBindingLifecycle> inputBindingLifecycles) {
...
-	public StopEndpoint getStopEndpoint() {
-		return new StopEndpoint();
-	}
-	
-	public StartEndpoint getStartEndpoint() {
-		return new StartEndpoint();
+	@WriteOperation
+	public boolean changeState(@Selector String name, State state) {
+		Assert.notNull(name, "'name' must not be null");
+		Assert.notNull(name, "'state' must not be null");
```
wilkinsona: If these asserts ever do anything, there's a bug in Spring Boot. You should only get null if an argument is annotated with `@Nullable`. Your endpoint won't even be called if a required parameter is missing. For example, a `POST` that's missing a `state` will automatically be rejected with a 400 response.

artembilan: Looks like the the second `Assert` must be against `state` param, but `name` again :smile:

olegz: yeah, I am removing both based on Andy's explanation

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/BindingsEndpoint.java
```diff
@@ -62,25 +53,41 @@ public BindingsEndpoint(List<InputBindingLifecycle> inputBindingLifecycles) {
...
-	}
-	
-	public StopEndpoint getStopEndpoint() {
-		return new StopEndpoint();
-	}
-	
-	public StartEndpoint getStartEndpoint() {
-		return new StartEndpoint();
+	@WriteOperation
+	public boolean changeState(@Selector String name, State state) {
```
wilkinsona: Returning a `boolean` strikes me as a little bit odd, particularly as it'll be `true` unless there's a bug and `state` is `null`.

Issue #1292, opened 2018-03-09
Unnecessary allocation of a ByteBuffer in AbstractAvroMessageConverter.


jmax01:
Unnecessary allocation of a `ByteBuffer` in `AbstractAvroMessageConverter`.

PR forthcoming.

```java

    @Override
    protected Object convertFromInternal(Message<?> message, Class<?> targetClass, Object conversionHint) {
        Object result = null;
        try {
            byte[] payload = (byte[]) message.getPayload();

            //What purpose does this serve?
            ByteBuffer buf = ByteBuffer.wrap(payload);

            MimeType mimeType = getContentTypeResolver().resolve(message.getHeaders());
            if (mimeType == null) {
                if (conversionHint instanceof MimeType) {
                    mimeType = (MimeType) conversionHint;
                }
                else {
                    return null;
                }
            }

            //What purpose does this serve?
            buf.get(payload);

            Schema writerSchema = resolveWriterSchemaForDeserialization(mimeType);
            Schema readerSchema = resolveReaderSchemaForDeserialization(targetClass);
            DatumReader<Object> reader = getDatumReader((Class<Object>) targetClass, readerSchema, writerSchema);
            Decoder decoder = DecoderFactory.get().binaryDecoder(payload, null);
            result = reader.read(null, decoder);
        }
        catch (IOException e) {
            throw new MessageConversionException(message, "Failed to read payload", e);
        }
        return result;
```

garyrussell:
>Unnecessary allocation of a ByteBuffer

It's actually (much) worse than that; the `get()` causes an unnecessary copy of each `byte` in the buffer to itself.

`ByteBuffer` (at least the Sun/Oracle implementation) doesn't have the smarts to short-circuit the copy if the wrapped and target buffers are identical.

This should be back-ported to 1.3.x.

jmax01:
Yes please backport.

PR #1293, opened 2018-03-09
GH-1292 Remove Unnecessary ByteBuffer

jmax01:


olegz:
@jmax01 in the future regardless of how trivial the change please add your name to the `@author` tag?  Also, do you mind if I add it during the merge?

Issue #1294, opened 2018-03-09
NPE from AvroSchemaRegistryClientMessageConverter.resolveWriterSchemaForDeserialization when mimeType does not parse to a SchemaReference such as "application/*+avro".


jmax01:
Should return `null`.

I would suggest that if `this.readerSchema` is not set and the `Schema` is not resolvable from the `mimeType` and the `targetClass` is an instance of `SpecificRecordBase`, `SpecificExceptionBase`, or `SpecificFixed`, the base classes of generated Avro Java classes, an attempt should be made to reflectively invoke the those classes' static `getClassSchema()` method. Caching the returned `Schema` might advisable.

I do not yet know what to do when those attempts fail, do we return `null` or throw?

```java 
 @Override
    protected Schema resolveWriterSchemaForDeserialization(MimeType mimeType) {
        if (this.readerSchema == null) {
            Schema schema = null;
            ParsedSchema parsedSchema = null;
            SchemaReference schemaReference = extractSchemaReference(mimeType);
            if (schemaReference != null) {
                parsedSchema = cacheManager.getCache(REFERENCE_CACHE_NAME)
                        .get(schemaReference, ParsedSchema.class);
                if (parsedSchema == null) {
                    String schemaContent = this.schemaRegistryClient
                            .fetch(schemaReference);
                    schema = new Schema.Parser().parse(schemaContent);
                    parsedSchema = new ParsedSchema(schema);
                    cacheManager.getCache(REFERENCE_CACHE_NAME)
                            .putIfAbsent(schemaReference, parsedSchema);
                }

            }
            //Throws NPE if mimeType is "application/*+avro"
            return parsedSchema.getSchema();
        }
        else {
            return this.readerSchema;
        }
    }
```

jmax01:
PR almost ready, but I need to know what to do in case both reader and writer schemas are null.

jmax01:
Also in `AvroSchemaMessageConverter` should `resolveReaderSchemaForDeserialization` leverage the `toClassSchema`  feature?

```java
	@Override
	protected Schema resolveReaderSchemaForDeserialization(Class<?> targetClass) {
		return this.schema;
	}
```

```java
	@Override
	protected Schema resolveReaderSchemaForDeserialization(Class<?> targetClass) {
		return this.schema != null ? this.schema : toClassSchemaInternal(targetClass);
	}
```

jmax01:
Also I think if we can't a mime type we don't have to stop.

[...]

PR #1295, opened 2018-03-09
GH-1294 fixed NPE with wildcarded MimeTypes

jmax01:
[Do not merge] has caching for schemas discovered via reflection of the targetClass.
Questions are marked TODO 1294

Resolves spring-cloud/spring-cloud-stream#1294

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/avro/AbstractAvroMessageConverter.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/avro/AbstractAvroMessageConverter.java
@@ -22,6 +22,7 @@ import java.nio.ByteBuffer;
 import java.util.Collection;
 import java.util.Collections;
 
+import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericDatumReader;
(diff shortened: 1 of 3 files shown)


olegz:
@jmax01 added few comments. The bottom line is that given that we're very close to a RELEASE, I am reluctant to address any enhancements right now, so just bugs. We can always add things in the future. Also as I've mentioned one can always provide a custom converter if the behavior of the provided one is not sufficient.

sabbyanandan:
Hi, @jmax01. Any thoughts on @olegz's comment?

jmax01:
Sorry have been really busy, I will try to take a peek tomorrow.

olegz:
Thanks @jmax01 
Let us know. Given that GA is scheduled for March 29th we'd like your two current PRs to make it. So if you need some help let us know as well

jmax01:
Once I get my DLQ with raw headers support working I will be able to take a look.

olegz:
@jmax01 I am closing this one in favor of https://github.com/spring-cloud/spring-cloud-stream/pull/1340
See the commit comment.
Basically the NPE is addressed following the defensive programming practices. Also we do support cases where both resolution methods return null using the `ReflectDatumReader` over the provided type.
As commit comment states, we recognize that it may still not be enough, but given that this is a `MessageConverter` user can always provide a custom one, yet over time we can evolve this one.

PR #1325, opened 2018-03-23
GH-1324 Restored support for the old format of metric Message

### LINKED ISSUE #1324: Reintroduce old metric format to the new metric support
With the new metric support which is based on micrometer the format of the published message is different. We can make it the same as it was to preserve backward compatibility.

olegz:
- brought back the Metric class from the old boot
- modified Metric class slightly to adjust for the type of info available thru micrometer
- added 'meterFilter' property for metric filtering

Resolves #1324


sobychacko:
LGTM - merging.

PR #1353, opened 2018-04-06
Channel data types support

Viacheslav-Petriaiev:


pivotal-issuemaster:
@Viacheslav-Petriaiev Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=1353)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-stream&pullRequestId=1353) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

olegz:
@Viacheslav-Petriaiev First of all, thank you for your contribution! I hope it will continue ;)

Question (and pointer for the future issues): Can you please describe why do you think we need to have this feature? 
Yes indeed, SI does have a concept of datatype channels to primarily support filtering/routing Messages with different types of payload. This is the problem we do not have in spring-cloud-stream since every _incoming_ and _outgoing_ message has payload of type `byte[]` (wire-format, see [Content Type Negotiation ](https://docs.spring.io/spring-cloud-stream/docs/Fishtown.BUILD-SNAPSHOT/reference/htmlsingle/#content-type-management) section). It can NOT be any other type.;

Viacheslav-Petriaiev:
Hello, Oleg
I was waiting for this question)))

In my project, I use spring-cloud-stream with spring-integration
- spring-cloud-stream - as a very convenient channels configuration mechanism, transport layer between the microservices
- spring-integration java dsl - as business process framework inside a microservice
`IntegrationFlows.from("channelConfiguredWithSpringCloudStream")` - and I know what datatype exactly I expect from this channel

So, I need an opportunity to configure this standard spring-integration channel parameter to use this spring-integration conversion mechanism and have intuitive visual config(Channel channelConfiguredWithSpringCloudStream expected SomeDTO)
```
spring:
  cloud:
    stream:
      bindings:
        channelConfiguredWithSpringCloudStream:
          dataTypes: com.xxx.dto.SomeDTO
```
Thanks

olegz:
@Viacheslav-Petriaiev Thank you for the info. SI channels are the internal mechanisms by which we enable bridge between the handler and the external destination. In theory we don't have to do that. . .

That said, what I was trying to emphasize in the previous comment is that there is no way the above could benefit you or anyone else since you can never have payload other then `byte[]` on the _bound channel_ (channel that is bound to the external destination). 
For even more clarity; I see that you have provided few test cases, but those only operate at the level of internal classes and my point is there is no way the arriving message can have a payload other then `byte[]`. So we may need to tighten up some of our internal classes, but that is a whole other issue.

olegz:
I think you're missing the difference between the _internal channels- (if you have some SI-based internal flow) and _bound channels_ - channels that serve as bridge between the in/out adapters and user provided handlers. These _bound channels_ *only* receive or output `byte[]` because those channels are the last or the first stop between the application and outside world so the payload will always be in raw form which is `byte[]`. Once again, if you go through [this section](https://docs.spring.io/spring-cloud-stream/docs/Fishtown.BUILD-SNAPSHOT/reference/htmlsingle/#content-type-management) of the user guide it will become clear. 

Basically ask yourself a question; How would you ever send  `FooDTO` over Kafka or Rabbit or any other messaging system. The answer would almost certainly be some serialization mechanism to/from `byte[]`. And that is the what bound channels are. Yes the  type conversion will happen between these channels and handler methods, but that is whole other discussion. 
Hope that helps

Viacheslav-Petriaiev:
Thanks for clarifications
If I've understood correctly
```
1) spring-cloud-stream flow
message broker -> bound channel(with byte[] payload) -> message converters mechanism -> method annotated with @StreamListener(with @Payload FooDTO)
```
```
2) my flow with spring-integration
message broker -> bound channel(with byte[] payload) -> some conversion mechanism -> internal channel(with FooDTO payload)
```

Maybe you can give me some advice, what is the better way to implement the second flow correctly. Conversion between bound channel and internal channel

Thanks a lot

olegz:
No. There is only `1)`.
`message broker -> bound channel(with byte[] payload) -> message converters mechanism -> method annotated with @StreamListener(with @Payload FooDTO) OR method annotated with @ServiceActivator(with @Payload FooDTO) OR any other valid SI handler annotation including a bridge handler to some other channel defined in your IntegrationFlows ......`

All I am saying is that the very first channel that I refer to as _bound channel_ has absolutely no way (in real life) of receiving message with payload that is anything other then `byte[]`. So essentially the code you provided in your PR would never execute other then in your test.

artembilan:
There are a bunch of out-of-the-box transformers in Spring Integration to be used in between the bound channel with `byte[]` and such an internal channel: https://docs.spring.io/spring-integration/docs/5.0.4.RELEASE/reference/html/messaging-transformation-chapter.html#_common_transformers

On the other hand you really can configure an internal channel for some data types:
```
IntegrationFlows.from("channelConfiguredWithSpringCloudStream")
                     .channel(c -> c.direct().datatype(FooDTO.class))
```
And the conversion from `byte[]` to `FooDTO` will happens via built-in `DefaultDatatypeChannelMessageConverter`: https://docs.spring.io/spring-integration/docs/5.0.4.RELEASE/reference/html/messaging-channels-section.html#channel-datatype-channel

Viacheslav-Petriaiev:
Yes, Artem. I used exactly this technique with channel datatype but in the bound channel insted of the internal channel. So this transformation worked not only in tests with changes in my PR))))
Conversion was executed here
https://github.com/spring-projects/spring-integration/blob/master/spring-integration-core/src/main/java/org/springframework/integration/channel/AbstractMessageChannel.java#L427

Ok, I understood that the better place for conversion is between bound and internal channels instead of the bound channel itself

olegz:
@Viacheslav-Petriaiev I don't think we're there yet. 
The `better place for conversion is between bound and internal channels instead of the bound channel itself` is simply not true.

The bound channel is just channel as any other channel. Conversion mechanism used on either channel is identical since as of today it is done using Spring Integration.
The datatype channels in Spring Integration are designed to be _". . . a simpler way of accomplishing the same thing is to apply the Datatype Channel pattern.__ where "same" is _". . .process a particular type of payload. . ."_. In SI it is real requirement because one may send payload of different types as as the doc explains routers, filters and/or datatype channels can be used to handle the requirement of ensuring that only certain data types can hit this or that channel and that implies that the payload is already in that type or Message Converters are used. 

In Spring Cloud Stream we *always* know what the incoming type is. It's a constant - `byte[]`. It can never be anything else. What we also know is that the end goal is to map the incoming Message to the input argument of the handler method which is done using Spring Integration provided argument resolvers and is explained in the link I provided earlier. 

So given that we know the _input type_ and the _required type_(input argument of the handler method),  I hope you can see that there is simply no added value to have data type channel (as bound channel), since it simply introduces another mechanism for Message Conversion at the channel level, which presents no tangible value in the context of Spring Cloud Stream since the *only* type we care about is the input argument type of the handler method.

olegz:
@Viacheslav-Petriaiev I am hoping that we're all on the same page by now and you understand why we should not proceed with this feature. So we're going to reject/close this PR, but once again Thank You for the effort, persistence and contribution and we hope to see more contributions from you!

Issue #1354, opened 2018-04-09
Add SCSt version to message header 


sabbyanandan:
As a MetricsCollector developer, I'd like to receive a identifier representing SCSt version in the message header, so I can use it to identify 2.0 messages for downstream processing.

The goal for Metrics Collector via: spring-cloud/spring-cloud-dataflow-metrics-collector#55 is to add support for both 1.x and 2.x messages single-handedly and to produce "streaming pipeline" representation of metrics for SCDF consistently.

**Acceptance:**
- SCSt version included in the message-header
- The release version is included regardless of the binder implementation in use
- Test harness validates for version inclusion in the header

PR #1356, opened 2018-04-11
GH-1354 Added version header to Metric Message

olegz:


olegz:
Had a quick discussion with @tzolov and given that SCDF has the higher level concept of a Stream (implemented by SCSt) will change the header name to `STREAM_CLOUD_STREAM_VERSION` to be more explicit and clear as to what _version_ is represented by this header.

Issue #1272, opened 2018-03-03
Consider removing parent/child AC relationship between the core and the actual binder implementation


olegz:
Aside from issues described https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/326  and https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/318 (just a few of them), it will help to better address lifecycle management of binders.

PR #1361, opened 2018-04-18
GH-1272 Removal of parent/child context

olegz:
Initial removal of parent/child context for Binders. This commit contains bare minimum of what's required while still supporting 'the old way'.
Basically implementor of a binder must do the following things:
- name your configuration file as the usual boot 'spring.factories` (no more spring.binders)
- add name attribute to the `@Bean` which declares the actual binders (e.g., `@Bean('rabbit')`)

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2017 the original author or authors.
+ * Copyright 2015-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.beans.factory.BeanCreationException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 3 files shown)


garyrussell:
I am not sure I am following the objective here.

How will we support different boot properties for each binder (via the `environment` property: https://docs.spring.io/spring-cloud-stream/docs/Elmhurst.RELEASE/reference/htmlsingle/index.html#multiple-systems) ?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -150,16 +156,19 @@ public BinderTypeRegistry binderTypeRegistry(ConfigurableApplicationContext conf
 		try {
 			Enumeration<URL> resources = classLoader.getResources("META-INF/spring.binders");
 			if (!Boolean.valueOf(this.selfContained) && (resources == null || !resources.hasMoreElements())) {
-				throw new BeanCreationException("Cannot create binder factory, no `META-INF/spring.binders` " +
-						"resources found on the classpath");
+				this.logger.warn("Failed to locate 'META-INF/spring.binders' resources on the classpath."
+						+ " Assuming standard boot 'META-INF/spring.factories' configuration is used");
```
sobychacko: When we fall back to the `spring.binders`check, that means that we haven't found any binders yet. In that check, if we still don't find any binder, don't we want to throw an exception?  This change only logs the warning and let the application comes up without a binder implementation present. Is that intended?

olegz: Yes, the issue is that the above code executes way to early to create `BinderTypeRegistry` bean (so AC initialization), hence the warn message.

So basically we are letting it through with the WARN and then in `DefaultBinderFactory.getBinder(..)` is where we try to find binder in the current AC and if not found we fall back on `doGetBinder(..)` which is the old method (unchanged) where we go and create child AC.

But that makes me think if we should add that same check right before `DefaultBinderFactory.doGetBinder(..)`? It would be easy, but there are already two assertions in that method so it will fail properly. 
WDYT?

garyrussell: Do we even need this warning at all? It's noise. Maybe DEBUG at most is all that's needed since we'll detect the error later. "no beans and no spring.binders".

sobychacko: @olegz I wasn't thinking about the earlier execution phase you mentioned. I am fine either way then. As @garyrussell mentioned, we might want to change the logging to `DEBUG` to reduce potential noise.

olegz: @sobychacko if that's the only change then please change it to DEBUG during merge.

sobychacko:
Polished and squashed.
Merged upstream.

PR #1366, opened 2018-04-24
Native multiple input binding changes

### LINKED ISSUE #1365: Introduce ability to do native broker multiple input bindings.
If the underlying broker supports native multiple input bindings on a single input, then the framework should delegate that to the binder rather than individually bind each destination from the list. 

See this issue for more context: https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/361

and this PR: https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/362

sobychacko:
If the binder implementation enables native multiple input binding, skip individual
consumer binding per binding target (for example - from a comma separated targets
provided as input destination property).

By default, native multiple input binding is disabled at the core framework level
and the binder implementation explicitly needs to override that to enable it.

Fixes #1365

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
@@ -124,15 +124,16 @@ public class ConsumerProperties {
 	private boolean useNativeDecoding;
 
 	/**
-	 * When set to true, the underlying binder natively deals with multiple destinations on the same input.
+	 * When set to true, the underlying binder will natively multiplex destinations on the same input binding.
 	 * For example, in the case of a comma separated multiple destinations, the core framework will skip binding
 	 * them individually if this is set to true, but delegate that responsibility to the binder.
 	 *
-	 * By default this property is set to `false` and the end users are not expected to configure this
-	 * property directly in the application. The individual binder implementations that need to support multiple
-	 * input bindings natively can enable this property by calling the setter method to do so.
+	 * By default this property is set to `false` and the binder will individually bind each destinations in case
+	 * of a comma separated multi destination list. The individual binder implementations that need to support multiple
+	 * input bindings natively (multiplex) can enable this property. Under normal circumstances, the end users are
+	 * not expected to enable or disable this property directly.
 	 */
-	private boolean nativeMultipleBinding;
+	private boolean multiplex;
 
 	@Min(value = 1, message = "Concurrency should be greater than zero.")
 	public int getConcurrency() {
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ConsumerProperties.java
```diff
@@ -209,4 +220,12 @@ public boolean isUseNativeDecoding() {
...
 		this.useNativeDecoding = useNativeDecoding;
 	}
+
+	public boolean isNativeMultipleBinding() {
+		return nativeMultipleBinding;
+	}
+
+	public void setNativeMultipleBinding(boolean nativeMultipleBinding) {
+		this.nativeMultipleBinding = nativeMultipleBinding;
+	}
```
olegz: @sobychacko and I just had a quick chat about this. Currently the `native-multiple-binding` will be a public property visible through content assist and essentially exposed to the end user. Yet the intention for this property are strictly internal. 
So we agreed that it would probably be wise to remove both _setter/getter_ and in KafkaStreams binder (where it is now required) simply use `DirectFieldAccessor` to set the value and in `BindingService` to access the value.

Not ideal, but will allows us a simple and quick solution for the time being until we get a chance to refactor the code to ensure that we can introduce internal vs external properties properly.

garyrussell: >Yet the intention for this property are strictly internal.

I disagree - with the rabbit/kafka binders, the user should be able to choose whether to use a container per destination (as currently) or multiple destinations per container.

While I am not suggesting we can do this, but with rabbit/kafka we could even support

    destination=[foo,bar],[baz,qux]

(2 containers with 2 destinations each).

olegz: Gary, I can't say I am disagreeing with you, but unless we support this functionality already I do not want to make it public and base on my discussion with Soby, KafkaStreams is the only binder where it always has to be `true`.

garyrussell: >but unless we support this functionality already

I thought that was the whole point, i.e. we __do__ intend to support it in the other binders in 2.1 and this is the preparation for that.

olegz: This specific property and its name basically tells the Binder to "stop being clever" and pass the value to the underlying system exactly as provided. Nothing more.
What you're talking about is functionality for the binder to get "more clever" and I think these are two completely orthogonal concerns - _delegate to the target system_ OR _do something "clever" and here is another property that will help you to define what "clever" means_.

Issue #1385, opened 2018-05-29
Exception while registering error infrastructure, if multiple destinations use same group 


vishnukr1992:
The spring cloud stream failed to create consumer binding while multiple destinations trying to use the same group.
spring boot version: 2.0.1.RELEASE
spring-cloud.version: Finchley.M9

pom.xml is

      <project xmlns="http://maven.apache.org/POM/4.0.0" 
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven- 
        4.0.0.xsd">
	     <modelVersion>4.0.0</modelVersion>
	      <groupId>org.nets</groupId>
	      <artifactId>nets-alert-engine</artifactId>
	      <version>0.0.1-SNAPSHOT</version>
	      <name>nets-alert-engine</name>

	  <parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>2.0.1.RELEASE</version>
		<relativePath /> 
	  </parent>

	  <properties>
		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
		<java.version>8</java.version>
		<spring-cloud.version>Finchley.M9</spring-cloud.version>
	  </properties>

	<dependencies>
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-config</artifactId>
		</dependency>
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-netflix-eureka-client</artifactId>
		</dependency>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-actuator</artifactId>
		</dependency>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-web</artifactId>
		</dependency>
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-stream-rabbit</artifactId>
		</dependency>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-integration</artifactId>
		</dependency>
		<dependency>
			<groupId>org.projectlombok</groupId>
[...]

PR #1388, opened 2018-06-12
GH-1385: Add qualifier for error infrastructure

garyrussell:
Fixes https://github.com/spring-cloud/spring-cloud-stream/issues/1385

Avoid bean name collision when using multiple bindings to the destination/group.

Add a new property `qualifier`.


olegz:
After discussing it with @garyrussell we're not going with the above solution to avoid introducing yet another property to solve a problem that can be solved with the existing properties.
Basically do use different group to eliminate the conflict.
What we'll do also is provide better error handling with a more descriptive error message suggesting to the user about the potential issue and resolution.

PR #1406, opened 2018-07-03
GH-1405 added RetryTemplate qualifier

### LINKED ISSUE #1405: Add @Qualifier to AbstractBinder's RetryTemplate
Our DLQ ITs were failing mysteriously until we discovered that our Kafka DLQ topics were only being written to _after_ our `RetryTemplate`-executed assertion logic had timed out.

Removing our `RetryTemplate` from the context allowed the tests to [pass](https://gitter.im/spring-cloud/spring-cloud-stream?at=5b365ba4b086c90e8d0266c1).

Picking up an arbitrary `RetryTemplate` as the default for Bindings seems risky.

Since no mutator is available the current workaround is to subclass [AbstractBinder](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java) implementations and override `buildRetryTemplate`.

The culprits:

```java

      @Autowired(required=false) // this would need to be refactored into constructor in the future
       private RetryTemplate retryTemplate;

	/**
	 * Create and configure a default retry template unless one
	 * has already been provided via @Bean by an application.
	 *
	 * @param properties The properties.
	 * @return The retry template
	 */
	protected RetryTemplate buildRetryTemplate(ConsumerProperties properties) {
		RetryTemplate rt = this.retryTemplate;
		if (rt == null) {
			rt = new RetryTemplate();
			SimpleRetryPolicy retryPolicy = new SimpleRetryPolicy();
			retryPolicy.setMaxAttempts(properties.getMaxAttempts());
			ExponentialBackOffPolicy backOffPolicy = new ExponentialBackOffPolicy();
			backOffPolicy.setInitialInterval(properties.getBackOffInitialInterval());
			backOffPolicy.setMultiplier(properties.getBackOffMultiplier());
			backOffPolicy.setMaxInterval(properties.getBackOffMaxInterval());
			rt.setRetryPolicy(retryPolicy);
			rt.setBackOffPolicy(backOffPolicy);
		}
		return rt;
}
```

olegz:
Added @StreamRetryTemplate qualifier annotation
Resolves #1405

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -66,7 +66,7 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 
 	@Autowired(required=false) // this would need to be refactored into constructor in the future
 	@StreamRetryTemplate
-	private RetryTemplate retryTemplate;
+	private RetryTemplate consumerBindingRetryTemplate;
 
 	/**
 	 * For binder implementations that support a prefix, apply the prefix to the name.
@@ -178,7 +178,7 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 	 * @return The retry template
 	 */
 	protected RetryTemplate buildRetryTemplate(ConsumerProperties properties) {
-		RetryTemplate rt = this.retryTemplate;
+		RetryTemplate rt = this.consumerBindingRetryTemplate;
 		if (rt == null) {
 			rt = new RetryTemplate();
 			SimpleRetryPolicy retryPolicy = new SimpleRetryPolicy();
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -23,6 +23,7 @@ import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.cloud.stream.annotation.StreamRetryTemplate;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.support.AbstractApplicationContext;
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -64,6 +65,7 @@
 	private volatile EvaluationContext evaluationContext;
 
 	@Autowired(required=false) // this would need to be refactored into constructor in the future
+	@StreamRetryTemplate
 	private RetryTemplate retryTemplate;
```
garyrussell: This technique correctly fails to load with the following:

```java

	@Bean
	@StreamRetryTemplate
	public RetryTemplate retryTemplate1() {
		return new RetryTemplate();
	}

	@Bean
	@StreamRetryTemplate
	public RetryTemplate retryTemplate2() {
		return new RetryTemplate();
	}

```
```
Field retryTemplate in org.springframework.cloud.stream.binder.AbstractBinder required a single bean, but 2 were found:
	- retryTemplate1: defined by method 'retryTemplate1' in org.springframework.cloud.stream.binder.test.SampleStreamApp
	- retryTemplate2: defined by method 'retryTemplate2' in org.springframework.cloud.stream.binder.test.SampleStreamApp
```

However, with this...

```java
	@Bean
	@StreamRetryTemplate
	public RetryTemplate retryTemplate() {
		return new RetryTemplate();
	}

	@Bean
	@StreamRetryTemplate
	public RetryTemplate retryTemplate2() {
		return new RetryTemplate();
	}

```
The context loads ok (the first template is injected - by name).

Maybe we should rename the field to something like `consumerBindingRetryTemplate` ?

Then we'll get

```
Field consumerBindingRetryTemplate in org.springframework.cloud.stream.binder.AbstractBinder required a single bean, but 2 were found:
	- retryTemplate: defined by method 'retryTemplate' in org.springframework.cloud.stream.binder.test.SampleStreamApp
	- retryTemplate2: defined by method 'retryTemplate2' in org.springframework.cloud.stream.binder.test.SampleStreamApp
```

Or give it some obscure name so a user is unlikely to configure a bean with that name.

olegz: Hmm. . . let me think a bit

olegz:
@garyrussell went with your suggestion and renamed the field. Also, in the event there is an existing one coming from the library, `@Primary` could be used to break the tie.

marbon87:
This should be mentioned in the changelog, because it is a breaking change.
I'm glat that i have a test that verifies that my custom RetryTemplate is used...

olegz:
@marbon87 good point. I just raised #1524, so we'll take care of it

PR #1413, opened 2018-07-09
GH-1412 Ensured SpelExpressionConverterConfiguration present in multi…

### LINKED ISSUE #1412: Ensure SpelExpressionConverterConfiguration is equally configured during single and/or multi-binder configurations
See the original issue originated in Kafka binder [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/408)

Currently for multi-binder cases we do NOT set parent context to a binder context, thus making these type of configurations invisible to the binder's context. While the issue has been identified specifically with `SpelExpressionConverterConfiguration`, there may be other similar once.

olegz:
…-binder

Ensured that SpelExpressionConverterConfiguration is always present in the AC regardless of single or multi-binder application configuration
Resolves #1412


sobychacko:
Looks good - polished and squashed the commits. Merged upstream.

PR #1423, opened 2018-07-24
Properties propagation in custom environment

### LINKED ISSUE #1420: Provide ability to propagate properties to individual binders in a multi-binder scenario

sobychacko:
Currently when the binder configuration provide properties in a custom environment,
Spring Cloud Stream does not include any beans from the outer context.
This change will ensure that in such cases, the binder context will have access
to the outer context as a non-parent bean.

Resolves #1420

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
@@ -55,8 +55,6 @@ import org.springframework.util.StringUtils;
  */
 public class DefaultBinderFactory implements BinderFactory, DisposableBean, ApplicationContextAware {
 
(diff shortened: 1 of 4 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -50,9 +51,12 @@
  * @author Ilayaperumal Gopinathan
  * @author Gary Russell
  * @author Oleg Zhurakousky
+ * @author Soby Chacko
  */
 public class DefaultBinderFactory implements BinderFactory, DisposableBean, ApplicationContextAware {
 
+	private static ConfigurableApplicationContext OUTER_CONTEXT;
```
artembilan: Doesn't look like this property is in use.
And that's is good because `static` shared property is evil for Spring 😄

PR #1428, opened 2018-08-04
Upgrade Spring Cloud Build to 2.1.0

### LINKED ISSUE #1426: Support Spring Boot 2.1
Spring Boot 2.1.0.M1 is based on Spring Framework 5.1.0.RC1 which supports Java 11. To use the long-term Java 11 it would be nice if Spring Cloud Stream would be based on Spring Boot 2.1.

sobychacko:
This upgrades to Spring Boot 2.1.0
Fix tests where the beans need to get overridden
Minor cleaning up
Polishing

Resolves #1426


spencergibb:
will this require stream clients to set the property `spring.main.allow-bean-definition-overriding=true`?

sobychacko:
@spencergibb Yes, I think so. There is nothing that we did to add that property by default in the framework.

olegz:
Also, we don't have much beans to override, so we're probably least affected by this.

spencergibb:
I guess I mean if, as a client of stream, if I have to add `spring.main.allow-bean-definition-overriding=true` then there is something wrong

olegz:
@spencergibb no as a client you shouldn't have to, so we may need to address in on the case by case basis since we don't have much of it in SCSt.

PR #1434, opened 2018-08-10
Boot 2.1 updates

### LINKED ISSUE #1429: Stream clients are required to set 'spring.main.allow-bean-definition-overriding=true'
See PR #1428 

https://github.com/spring-cloud/spring-cloud-stream/issues/1426#issuecomment-411147908

### LINKED ISSUE #1430: Remove ConditionalOnMissingBean from BinderTypeRegistry bean definition.
As discussed [here](https://github.com/spring-cloud/spring-cloud-stream/issues/1429#issuecomment-411403792) the `Conditional` is only to support testing and there are absolutely no expectation for user to ever override `BinderTypeRegistry`.

sobychacko:
Resolves #1429 
Resolves #1430


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
```diff
@@ -37,13 +36,14 @@
...
  * @author David Turanski
+ * @author Soby Chacko
  */
 @Target({ ElementType.TYPE, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 @Inherited
 @Configuration
-@Import({ BindingServiceConfiguration.class, BindingBeansRegistrar.class, BinderFactoryConfiguration.class})
+@Import({ BindingBeansRegistrar.class, BinderFactoryConfiguration.class})
```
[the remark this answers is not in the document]
ilayaperumalg: Unless I miss anything specific, the `BindingServiceConfiguration` was intended to be associated with `EnableBinding` only and all the beans of `BindingServiceConfiguration` would be of no value without `EnableBinding`.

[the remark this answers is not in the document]
sobychacko: Hi, @ilayaperumalg with the boot 2.1 changes that don't allow bean name overriding kind of started us to do it in this direction. There were `ConditionalOnMissingBean` annotations present on beans in `BindingServiceConfigurartion` and this was not an auto-configurable class. Since, it is not advisable to add `ConditionalOnMissing..` annotations on non auto-configurable classes, it was not guaranteed that it would work. In fact, we ran into that scenario and that's the reason. Downstream binders, sample apps etc. ran without issues. Seems to me that `BindingServiceConfiguration` is such a fundamental configuration class and could be benefited from auto-configuration. What do you think?

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
```diff
@@ -41,25 +42,28 @@
  * @author Oleg Zhurakousky
  */
 @RunWith(SpringJUnit4ClassRunner.class)
+@DirtiesContext
 public class AggregateApplicationTests {
 
-	@Before
-	public void before() {
-		System.setProperty("server.port", "0");
```
garyrussell: Don't we still need this?

spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/aggregate/main/AggregateWithMainTest.java
```diff
@@ -43,12 +42,6 @@
  */
 public class AggregateWithMainTest {
 
-	@Before
-	public void before() {
-		System.setProperty("server.port", "0");
```
garyrussell: Ditto

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/aggregation/AggregationTest.java
```diff
@@ -57,17 +56,12 @@
...
+ * @author Soby Chacko
  */
 public class AggregationTest {
 
 	private ConfigurableApplicationContext aggregatedApplicationContext;
 
-	@Before
-	public void before() {
-		System.setProperty("spring.main.allow-bean-definition-overriding", "true");
-		System.setProperty("server.port", "0");
```
garyrussell: again

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
sobychacko: @garyrussell I think this property was probably a superfluous one. No tests in the class failed after removing it. All the individual test contexts are getting cleaned up properly and don't think the individual tests will conflict. It could have been a carryover from previous times while it was failing for other reasons.

sobychacko:
@garyrussell I can coordinate with whoever merging the PR to have more detailed commit messages.

spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/aggregate/main/AggregateWithMainTest.java
sobychacko: same reason as above. Tests passed.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/aggregation/AggregationTest.java
sobychacko: same reason as above. Tests passed.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
garyrussell: Well, the tests will pass if it starts a web container and nothing else is listening on port 8080.

The reason we set this to 0 is so the operating system picks an unused port.

It can safely be removed as long as you are sure the test doesn't run a web container.

olegz: Given that [web and actuator are actually optional](https://docs.spring.io/spring-cloud-stream/docs/Fishtown.BUILD-SNAPSHOT/reference/htmlsingle/#spring-cloud-stream-preface-actuator-web-dependencies) (one has to opt in), I say remove it.

sobychacko:
@garyrussell Added a few more commit messages in the first commit in the PR.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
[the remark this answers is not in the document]
ilayaperumalg: ok, Thanks for the clarification @sobychacko. With these changes, the `BinderTypeRegistry` is always going to be `DefaultBinderTypeRegistry` since we have now removed the `@ConditionalOnMissingBean(BinderTypeRegistry.class)` for the `BinderTypeRegistry` bean. Are we ok with that?

I thought both the `BindingServiceConfiguration` and the `BinderFactoryConfiguration` go hand in hand and the `BinderFactoryConfiguration` can also be moved into `BindingServiceConfiguration` as part of the autoconfiguration.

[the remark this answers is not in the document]
artembilan: @ilayaperumalg ,

One more time: as long as your `@Configuration` is not included into the `spring.factories` `EnableAutoConfiguration` property, all those `@Conditional...` don't work. And that's how we got all those problems with the new `BeanDefinitionOverrideException`.

The point is that this `BinderTypeRegistry` bean is really appear in our app only when we have that `@EnableBinding`, and as early as possible. And that's how that our new `BindingServiceConfiguration` (I believe we need to rename it into an `...Auto...`) depends on the `@ConditionalOnBean(value = BinderTypeRegistry.class, search = SearchStrategy.CURRENT)`. It's not going to appear in our context if we don't have `@EnableBinding`.

I don't think that `BinderTypeRegistry` is something intended for overriding. If you would like to have some other impl, you would need a custom `@EnableBinding` as well.

The bean override didn't work before properly and I'm not sure it is going to be so critical that we don't allow to override a `BinderTypeRegistry` since now...

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
artembilan: I believe:
```
<dependency>
	<artifactId>spring-boot-starter-actuator</artifactId>
	<optional>true</optional>
</dependency>
```
In the `spring-cloud-stream` POM does the trick for us and we really have a WEB in our test case. So, I would say remove, but change it into the explicit web environment exclusion (`web(false)` ?).

If we add something into the system properties, we  need to clean up it afterwards, otherwise the next test in the chain is going to reuse them. Just because it still the same JVM...
And that's why would be better to have a `@DirtiesContext` on all the `@SpringBootTest` to clean up all the application context related to this class.

artembilan: But, yeah... That `@DirtiesContext` could be as a separate story, indeed.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
```diff
@@ -41,25 +42,28 @@
  * @author Oleg Zhurakousky
  */
 @RunWith(SpringJUnit4ClassRunner.class)
+@DirtiesContext
```
artembilan: We don't need both of these annotations here since we don't kick off an application context by the Spring Test Framework. 

This class has only one test-case which takes care about an application context and its closure in the end.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
artembilan: BTW. This was always redundant in this test-case - we use there for application context a `web(false)` option. 🤷‍♂️

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
```diff
@@ -41,25 +42,28 @@
...
 				.poll(5, TimeUnit.SECONDS));
 		Assert.assertThat(received, notNullValue());
 		Assert.assertTrue(received.getPayload().endsWith("processed"));
+
+		context.close();
+	}
+
+	@Configuration
+	@EnableAutoConfiguration
+	static class FooConfig{
```
artembilan: I believe this class should have a much proper name. And we need to have in the commit message a point "why?" - the `TestSupportBinderAutoConfiguration` is really an auto-configuration so we need to rely on the Boot in this case to pick up everything properly from the classpath and resources.

spring-cloud-stream-schema-server/src/test/java/org/springframework/cloud/stream/schema/server/SchemaRegistryServerAvroTests.java
```diff
@@ -44,8 +44,7 @@
  * @author Ilayaperumal Gopinathan
  */
 @RunWith(SpringRunner.class)
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
-		properties = {"--spring.main.allow-bean-definition-overriding=true"})
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
```
artembilan: Do we really need WEB in this test case?
Just to be sure that we need such an overhead, especially when we use `@DirtiesContext(classMode = AFTER_EACH_TEST_METHOD)`...

spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/example/ExampleTest.java
```diff
@@ -39,8 +39,9 @@
  * correctly.
  */
 @RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest(classes = ExampleTest.MyProcessor.class, properties = { "server.port=-1", "--spring.cloud.stream.bindings.input.contentType=text/plain", "--spring.cloud.stream.bindings.output.contentType=text/plain",
-		"--spring.main.allow-bean-definition-overriding=true"})
+@SpringBootTest(classes = ExampleTest.MyProcessor.class, properties = { "server.port=-1",
```
artembilan: The `server.port=-1` has to be replaced with the `webEnvironment = SpringBootTest.WebEnvironment.NONE`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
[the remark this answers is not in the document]
artembilan: The Copyright has to be update.
BTW, for all the affected classes...

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceConfiguration.java
```diff
@@ -74,16 +84,76 @@
...
+									BindingServiceProperties bindingServiceProperties) {
+		DefaultBinderFactory binderFactory = new DefaultBinderFactory(
+				getBinderConfigurations(binderTypeRegistry, bindingServiceProperties), binderTypeRegistry);
+		binderFactory.setDefaultBinder(bindingServiceProperties.getDefaultBinder());
+		binderFactory.setListeners(binderFactoryListeners);
+		return binderFactory;
+	}
+
+	private static Map<String, BinderConfiguration> getBinderConfigurations(BinderTypeRegistry binderTypeRegistry,
+																			BindingServiceProperties bindingServiceProperties) {
```
artembilan: Something is wrong with indents.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/aggregation/AggregationTest.java
```diff
@@ -82,7 +76,8 @@ public void closeContext() {
 	@Test
 	public void aggregation() {
 		aggregatedApplicationContext = new AggregateApplicationBuilder(
-				MockBinderRegistryConfiguration.class, "--server.port=0", "--debug=true")
+				FooConfig.class, "--server.port=0", "--debug=true",
```
artembilan: We have a `.web(false)`. What is the point of out of use `--server.port=0`?
And I believe we don't need `--debug=true`. That's what we used to determine the reason of failure in this test.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/aggregation/AggregationTest.java
```diff
@@ -443,4 +443,10 @@ else if (factory.getObjectType() == Processor.class) {
 	public static class DummyConfig {
 
 	}
+
+	@Configuration
+	@EnableAutoConfiguration
+	public static class FooConfig {
```
artembilan: Suspicious name for the configuration again. Even if it is test, you can't be sure that people who are going to maintain them after you would determine easily what this is about...

More over when you make yourself to think twice about names even in tests, you are going to be much better in naming on the production code level 😉 

Thanks

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/InputOutputBindingOrderTest.java
```diff
@@ -46,15 +42,11 @@
...
-	@Before
-	public void before() {
-		System.setProperty("spring.main.allow-bean-definition-overriding", "true");
-	}
-
 	@SuppressWarnings({"rawtypes", "unchecked"})
 	@Test
 	public void testInputOutputBindingOrder() {
-		ConfigurableApplicationContext applicationContext = SpringApplication.run(TestSource.class, "--server.port=-1");
+		ConfigurableApplicationContext applicationContext = SpringApplication.run(TestSource.class, "--server.port=-1",
```
artembilan: The wrong web exclude property again.
I mean there is no guarantee that such a `server.port=-1` turn off web environment in your application context.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/PollableConsumerTests.java
```diff
@@ -63,7 +64,6 @@
 
 	@Before
 	public void before() {
-		//System.setProperty("spring.main.allow-bean-definition-overriding", "true");
```
artembilan: I believe this must be removed.
Well, the whole change to this class can be reverted.
Although I would leave an empty line after package 😉

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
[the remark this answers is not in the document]
olegz: @ilayaperumalg yes, we're ok with "`BinderTypeRegistry` always being a  `DefaultBinderTypeRegistry`". Soby and I have discussed it quite a bit and we don't envision it as an extension point any longer.

spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/config/aggregate/AggregateApplicationTests.java
olegz: True, the name is mea meaningless, but given that we're already "soft-deprecated" aggregate applications (by simply removing documentation for it), we can live with it until we remove it.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/annotation/EnableBinding.java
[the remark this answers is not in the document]
ilayaperumalg: @sobychacko @artembilan @olegz Thank you for the explanation! I get the intention for this change now.

One suggestion about this:

> The point is that this BinderTypeRegistry bean is really appear in our app only when we have that @EnableBinding, and as early as possible. And that's how that our new BindingServiceConfiguration (I believe we need to rename it into an ...Auto...) depends on the @ConditionalOnBean(value = BinderTypeRegistry.class, search = SearchStrategy.CURRENT). It's not going to appear in our context if we don't have @EnableBinding.

Maybe it is worth having a different bean or a marker interface to check on initializing the `BindingServiceConfiguration` as auto-configuration instead of using `ConditionOnBean` with `BinderTypeRegistry`.  I believe the bean `BinderTypeRegistry` is just the inner bean (the holder of the binder types) for `BinderFactoryConfiguration` and can't be relied upon initializing the whole `BindingServiceConfiguration` itself.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -32,33 +31,45 @@
...
  * @author Marius Bogoevici
  * @author Ilayaperumal Gopinathan
  * @author Oleg Zhurakousky
+ * @author Soby Chacko
  */
 @Configuration
 @Role(BeanDefinition.ROLE_INFRASTRUCTURE)
+@EnableConfigurationProperties({ BindingServiceProperties.class })
+@Import({ContentTypeConfiguration.class})
 public class BinderFactoryConfiguration {
```
ilayaperumalg: This can no longer be named as `BinderFactoryConfiguration` since the factory configuration is now moved.

olegz: Well, it's a public class, so we should deprecated with comment stating exactly what you said. Given that it's a configuration we can probably rename it in 2.2, - essentially giving a small warning window, but for now we stuck with it

artembilan: Right. Let's deprecate and make it `extends BinderFactoryAutoConfiguration` (or whatever you think is better name for this) with moving all the context over there.
However keep in mind that this one is already used in the SCT:
```
@AutoConfigureBefore(BindingServiceConfiguration.class)
@AutoConfigureAfter(SimpleTaskConfiguration.class)
public class TaskEventAutoConfiguration {
```

spring-cloud-stream-schema-server/src/test/java/org/springframework/cloud/stream/schema/server/SchemaRegistryServerAvroTests.java
```diff
@@ -45,7 +45,7 @@
  */
 @RunWith(SpringRunner.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
-		properties = {"--spring.main.allow-bean-definition-overriding=true"})
+				properties = "spring.main.allow-bean-definition-overriding=true")
```
artembilan: Is there anything what really makes us to keep this `allow-bean-definition-overriding` here?

I believe the point of this PR to get rid of it...

Thanks

spring-cloud-stream-schema/src/test/java/org/springframework/cloud/schema/avro/AvroMessageConverterSerializationTests.java
```diff
@@ -67,9 +67,8 @@
 
 	@Before
 	public void setup() {
-		System.setProperty("--spring.main.allow-bean-definition-overriding", "true");
 		schemaRegistryServerContext = SpringApplication
-				.run(SchemaRegistryServerApplication.class);
+				.run(SchemaRegistryServerApplication.class, "--spring.main.allow-bean-definition-overriding=true");
```
artembilan: DITTO

spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/example/ExampleTest.java
```diff
@@ -39,8 +39,10 @@
  * correctly.
  */
 @RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest(classes = ExampleTest.MyProcessor.class, properties = { "server.port=-1", "--spring.cloud.stream.bindings.input.contentType=text/plain", "--spring.cloud.stream.bindings.output.contentType=text/plain",
-		"--spring.main.allow-bean-definition-overriding=true"})
+@SpringBootTest(classes = ExampleTest.MyProcessor.class, webEnvironment = SpringBootTest.WebEnvironment.NONE,
+		properties = {
+		"--spring.cloud.stream.bindings.input.contentType=text/plain",
+		"--spring.cloud.stream.bindings.output.contentType=text/plain"})
```
artembilan: I think we need to remove these `--`. That's not command line arg, but key/value props.
Otherwise let's just remove them as a noise since it works even with them, wrongly provided.

Or am I missing anything and it really works like a charm even in this installment?

Thanks

cppwfs:
@olegz We ran across an issue on the task side with this PR when we need to use SCSt for event messaging.  Task is now required to include additional Reactive dependencies to support `ReactiveWebApplicationContext` when we include SCSt 2.1.0 even though task does not use it.   It is coming from the relationship change when SCF used to depend on SCSt but now that relationship is flipped where SCSt now depends on SCF (which brings this dependency).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
artembilan: Oh! Sorry, we talked here about different class.

So, yes, this one has to be renamed as well.
But as Oleg said: that's already a matter of separate issue/PR.

@ilayaperumalg ,  please, proceed and raise such an issue if you feel strong.

Thanks for your patience!

spring-cloud-stream-test-support/src/test/java/org/springframework/cloud/stream/test/example/ExampleTest.java
sobychacko: ya, tests passed but will look.

spring-cloud-stream-schema/src/test/java/org/springframework/cloud/schema/avro/AvroMessageConverterSerializationTests.java
sobychacko: @artembilan There is an isolated case with schema registry that I had to override the bean internally in the unit tests. For the real application, it worked. Since it worked for the sample apps, I think in order to reduce the scope of this PR, we should track this separately before M2. I will create an issue.

Issue #1441, opened 2018-08-15
Add support for wiring Functions and Consumers to binding channels of the current binders


olegz:
Part of #1436

PR #1444, opened 2018-08-16
GH-1440 Added initial support for Supplier as Source

### LINKED ISSUE #1440: Add support for bootstrapping green-field Source development based on Suppliers provided by users
Part of #1436

olegz:
Added initial support and configuration to create Source apps by simply providing a Supplier bean while also benefitting from composition provided by SCF
Refactored/renamed some of the  new classes/methods

Resolves #1440


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.cloud.stream.converter.CompositeMessageConverterFactory;
+import org.springframework.integration.dsl.IntegrationFlowBuilder;
+import org.springframework.integration.dsl.IntegrationFlows;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ *
```
artembilan: Need some Javadocs.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
```diff
@@ -0,0 +1,138 @@
...
+	 * @param functionInspector
+	 * @param messageConverterFactory
+	 * @param functionProperties
+	 */
+	public IntegrationFlowFunctionSupport(FunctionCatalog functionCatalog, FunctionInspector functionInspector,
+			CompositeMessageConverterFactory messageConverterFactory, FunctionProperties functionProperties) {
+
+		Assert.notNull(functionCatalog, "'functionCatalog' must not be null");
+		Assert.notNull(functionInspector, "'functionInspector' must not be null");
+		Assert.notNull(messageConverterFactory, "'messageConverterFactory' must not be null");
```
artembilan: I believe the `Assert.notNull()` for `functionProperties` should be here as well.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
```diff
@@ -0,0 +1,138 @@
...
+		this.functionCatalog = functionCatalog;
+		this.functionInspector = functionInspector;
+		this.messageConverterFactory = messageConverterFactory;
+		this.functionProperties = functionProperties;
+	}
+
+	/**
+	 * Creates an instance of the {@link IntegrationFlowBuilder} from a {@link Supplier} bean available in the context.
```
artembilan: Our Spring style is *imperative* on methods: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting. Also no blank lines in the method Javadocs.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
```diff
@@ -0,0 +1,138 @@
...
+	 * The name of the bean must be provided via `spring.cloud.stream.function.name` property.
+	 *
+	 * @return instance of {@link IntegrationFlowBuilder}
```
artembilan: I believe we can also add a `@throws IllegalStateException` with an appropriate description.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
```diff
@@ -0,0 +1,138 @@
...
+	 */
+	public <O> IntegrationFlowBuilder integrationFlowFromNamedSupplier() {
+		if (StringUtils.hasText(this.functionProperties.getName())) {
+			Supplier<?> supplier = functionCatalog.lookup(Supplier.class, this.functionProperties.getName());
+			if (supplier instanceof FluxSupplier) {
+				supplier = ((FluxSupplier<?>)supplier).getTarget();
+			}
+			return this.integrationFlowFromProvidedSupplier(supplier);
+		}
+		throw new IllegalStateException("Supplier name is not provided");
```
artembilan: I think this message should include an appropriate configuration property to lead people to the correct direction.
That always was a bonus from Spring: not only say what is wrong, but also point to the right direction to fix.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
```diff
@@ -0,0 +1,138 @@
...
+	 * @param outputChannel channel where the output of a function will be sent
+	 * @return true if {@link Function} was located and added and false if it wasn't.
+	 */
+	public <I,O> boolean andThenFunction(IntegrationFlowBuilder flowBuilder, MessageChannel outputChannel) {
+		if (StringUtils.hasText(this.functionProperties.getName())) {
+			FunctionInvoker<I,O> functionInvoker =
+					new FunctionInvoker<>(this.functionProperties.getName(), this.functionCatalog, this.functionInspector,
+							this.messageConverterFactory);
+
+			subscribeToInput(functionInvoker, flowBuilder.toReactivePublisher(), message -> outputChannel.send(message));
```
artembilan: Nit-pick: the method reference can be used instead: `outputChannel::send`.
At least this is what my IDEA suggests me. No?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -0,0 +1,226 @@
...
+ * @author Oleg Zhurakousky
+ *
+ * @since 2.1
+ *
+ */
+public class SourceToFunctionsSupportTests {
+
+	@Test
+	public void testFunctionIsAppliedToExistingMessageSource() {
+		ApplicationContext context =
```
artembilan: The `ApplicationContext` has to be closed in the end of test.
You can wrap its instantiation to the `try-with-resource` also. And have all the assertions in its body.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -0,0 +1,226 @@
...
+				.isEqualTo("44".getBytes(StandardCharsets.UTF_8));
+		assertThat(target.receive(10000).getPayload())
+				.isEqualTo("66".getBytes(StandardCharsets.UTF_8));
+		//etc
+	}
+
+	@EnableAutoConfiguration
+	@Import(ProvidedMessageSourceConfiguration.class)
+	public static class SupplierConfiguration {
+		AtomicInteger counter = new AtomicInteger();
```
artembilan: Need some code reformatting: its hard to read it in this non-wrapped with blank lines style.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -0,0 +1,226 @@
...
+		@Autowired
+		private FunctionProperties functionProperties;
+
+		@Bean
+		public IntegrationFlow messageSourceFlow(IntegrationFlowFunctionSupport functionSupport) {
+			Assert.hasText(functionProperties.getName(), "Supplier name must be provided");
+
+			IntegrationFlowBuilder flowBuilder = functionSupport
+					.integrationFlowFromNamedSupplier()
+					.split()
```
artembilan: I thought you were going to have this `.split()` as out-of-the-box feature.
Or do you leave a `Flux` processing to the end-user?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -0,0 +1,226 @@
...
+
+		@Autowired
+		private FunctionProperties functionProperties;
+
+		@Bean
+		public IntegrationFlow messageSourceFlow(IntegrationFlowFunctionSupport functionSupport) {
+			Assert.hasText(functionProperties.getName(), "Supplier name must be provided");
+
+			IntegrationFlowBuilder flowBuilder = functionSupport
+					.integrationFlowFromNamedSupplier()
```
artembilan: Well, I feel a bit uncomfortable with this API: or should request a supplier name from end-user.
Or just make it `Supplier`-based and end-users to inject resolved upfront name-based suppliers.
At least that is how we deal with auto-configuration.

On the other hand the logic in the `IntegrationFlowFunctionSupport` should not be aware of auto-configurations, e.g. the same `FunctionProperties`. 

Just thoughts from my feeling...

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -0,0 +1,226 @@
...
+
+		@Bean
+		public IntegrationFlow messageSourceFlow(IntegrationFlowFunctionSupport functionSupport) {
+			Supplier<Message<String>> messageSource = () -> MessageBuilder.withPayload("hello function")
+					.setHeader(MessageHeaders.CONTENT_TYPE, MimeTypeUtils.TEXT_PLAIN).build();
+
+			IntegrationFlowBuilder flowBuilder = functionSupport.integrationFlowFromProvidedSupplier(messageSource);
+
+			if (!functionSupport.andThenFunction(flowBuilder, source.output())) {
+				flowBuilder = flowBuilder.channel(source.output());
```
artembilan: I don't like this combination of operators, but this is not your fault. This is a leak of proper reactive support in SI...
Hope we will fix it eventually. For example seems for me I suggested in some our discussion something like `fluxTransform(Fucntion<Flux<?>, Flux<?>>)` for the current Java DSL. What you would need to do here is just a function resolution from auto-configuration. 

Maybe it's really a time to play with that there?..

olegz: Yes, consider opening appropriate INT issues, so once addressed we can simplify the code here.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
olegz: Ok, let me try to clarify. . .
Every bean (as `@Bean`) of type `Supplier`, `Function` and/or `Consumer` will be loaded to `FunctionCatalog` by SCF. 
What we're doing here is simply specifying which one to use via `spring.cloud.stream.function.name` property.
So, it's not really comparable with auto-configuration concepts.

Does this clarify?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
olegz: Agreed, but this thing is still evolving so i'll wait till it's clear. As you can see the class was renamed already form the last merge. 
But I did do [this](https://github.com/spring-cloud/spring-cloud-stream/issues/1441) to ensure it gets done

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
olegz: yes

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
artembilan: What you mean sounds like a `defaultFunction` (or name) in the `IntegrationFlowFunctionSupport`, since you indeed goes to the catalog for lookup. That's why I don't see reason to be tied with the implicit name from some hard-coded property. Would be really better to configure this `IntegrationFlowFunctionSupport` bean appropriately leaving a door for end-users to customize.

Does it make sense?

olegz: Unless I am missing something, I am not sure I follow the _". . .configure this IntegrationFlowFunctionSupport bean appropriately. . ."_

Just to clarify a bit:
Doing it via property ensures that I can
* change the actual function that I want to use without rebuilding the app
* the name can include composition (i.e., `foo|bar|baz`) which essentially extends on the previous point
* the long term goal is to also consider making this property optional if there is only one applicable function in the catalog

artembilan: Well, I don't mind about that configuration property. Only what is my concern that it is not an `IntegrationFlowFunctionSupport` responsibility. 
you need to resolve a property to the name in the `FunctionConfiguration` and allow end-user to override it any possible way without any dependency on the `FunctionProperties`. 
That is what my concern about mixing auto-configuration goodies in this support component...

Well, feel free to reject my opinion and leave it as is.
Just let me know what is the final decision do not come back to this subject anymore! 😄

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/IntegrationFlowFunctionSupport.java
```diff
@@ -0,0 +1,141 @@
...
+		if (StringUtils.hasText(this.functionProperties.getName())) {
+			Supplier<?> supplier = functionCatalog.lookup(Supplier.class, this.functionProperties.getName());
+			if (supplier instanceof FluxSupplier) {
+				supplier = ((FluxSupplier<?>)supplier).getTarget();
+			}
+			return this.integrationFlowFromProvidedSupplier(supplier).split();
+		}
+		throw new IllegalStateException("Supplier by the name '"
+				+ functionProperties.getName() + "' specified using 'spring.cloud.stream.function.name' property"
+						+ " can not be located.");
```
artembilan: This is wrong message. It doesn't reflect the real state.
I'm changing it during merge to this:
> "A Supplier is not specified in the 'spring.cloud.stream.function.name' property."

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -0,0 +1,229 @@
...
+			//etc
+		}
+	}
+
+	@Test
+	public void testMessageSourceIsCreatedFromProvidedSupplierComposedWithMultipleFunctions() {
+		try(ConfigurableApplicationContext context =
+				new SpringApplicationBuilder(
+						TestChannelBinderConfiguration.getCompleteConfiguration(SupplierConfiguration.class))
+						.web(WebApplicationType.NONE)
```
artembilan: Typo: need single `m` in the beginning of `multiplyByTwo`.

Fixing during merge...

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-stream/commit/6b1d584c6c317be26a7ce951c1a008e14c98fd1c

PR #1453, opened 2018-08-22
Remove ConditionalOnProperty

### LINKED ISSUE #1452: Remove ConditionalOnProperty in FunctionConfiguration
`@ConditionalOnProperty("spring.cloud.stream.function.name")` requires scst app starter developers to: 

```
@Autowired(required=false)
IntegrationFlowFunctionSupport function support
...
if (support != null)
``` 

Since we don't know a priori if it's enabled or not.

dturanski:
Fixes #1452


dturanski:
@ilayaperumalg since you're looking at a shared configuration for app starters, you may have  another way to address the issue that doesn't enable FunctionConfiguration unnecessarily.

dturanski:
Reverted

PR #1462, opened 2018-08-31
Function support at the binder level

ilayaperumalg:
- Allow Outbound channel to be configured with the function support at the MessageChannel binder level
 TBD: Configure inbound channel
 - Refactor the IntegrationFlowFunctionSupport methods
 - Update tests

### DIFF
--- a/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
+++ b/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
@@ -42,6 +42,7 @@ import org.springframework.context.annotation.Import;
 import org.springframework.integration.channel.QueueChannel;
 import org.springframework.integration.dsl.IntegrationFlow;
 import org.springframework.integration.dsl.IntegrationFlowBuilder;
+import org.springframework.integration.dsl.IntegrationFlows;
 import org.springframework.integration.support.MessageBuilder;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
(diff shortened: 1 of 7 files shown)


artembilan:
According Travis report, there is a test failure:
```
testLateBindingConsumer(org.springframework.cloud.stream.binding.BindingServiceTests)  Time elapsed: 121.08 sec  <<< FAILURE!
java.lang.AssertionError: 
Expecting:
 <Mock for Binding, hashCode: 1353052385>
and actual:
 <null>
to refer to the same object
	at org.springframework.cloud.stream.binding.BindingServiceTests.testLateBindingConsumer(BindingServiceTests.java:480)
```

Looks like sporadic, but as experience shows it always a result of some race condition...

ilayaperumalg:
yeah, My local tests ran fine as well.

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -70,7 +71,7 @@
 	@Test
 	public void testFunctionIsAppliedToExistingMessageSource() {
 		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(
-			TestChannelBinderConfiguration.getCompleteConfiguration(FunctionsConfiguration.class)).web(
+			TestChannelBinderConfiguration.getCompleteConfiguration(FunctionsConfiguration.class, TestChannelBinderConfiguration.class)).web(
```
olegz: Why do you need to add `TestChannelBinderConfiguration` to `TestChannelBinderConfiguration`?

ilayaperumalg: This was an oversight.

dturanski:
This PR causes a Spring lifecycle issue when testing with Sftp source, a fairly complex scenario, but the fix requires a different configuration approach. I am closing this PR and replacing with a new one that works around this problem by configuring `IntegrationFlowFunctionSupport` in `AbstractBinder.afterPropertiesSet`

Issue #1459, opened 2018-08-30
Pollable Source Needs a Requeue Feature


garyrussell:
With the task launcher, if there are no slots for an extended time, we need to throw an exception that will re-seek the current record so it will be returned on the next poll. This is to avoid Kafka rebalancing the topic.

PR #1467, opened 2018-09-06
GH-1459: Pollable Consumer and Requeue

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/1459

Add a mechanism to cause a requeue of the current message.


spring-cloud-stream-core-docs/src/main/asciidoc/spring-cloud-stream-overview.adoc
```diff
@@ -1506,6 +1523,12 @@ When set to a negative value, it defaults to `spring.cloud.stream.instanceCount`
 See "`<<spring-cloud-stream-overview-instance-index-instance-count>>`" for more information.
 +
 Default: `-1`.
+polled.recoverable::
+When set to true (default) exceptions thrown by the polled consumer callback will result in an ErrorMessage sent to the consumer's error channel (and the global Spring Integration errorChannel).
+Set to `false` if, say, you want to requeue messages by throwing `RequeueCurrentMessageException` s.
+See <<polled-errors>>.
```
olegz: After reading your explanation I now wonder why do we need this property at all? Given that user is in full control of message retrieval and processing should we just send to _error channel_ ***unless*** user throws `RequeueCurrentMessageException`, at which point we will requeue?

PR #1477, opened 2018-09-17
Support extended default properties

### LINKED ISSUE #1360: Default spring.cloud.stream.rabbit.* properties to appy to multiple consumers?
Is there a way to specify default RABBIT properties to apply to multiple consumers?

e.g. I currently have some configuration like this:

```
spring.cloud.stream.rabbit.bindings.testChannel.consumer.bindingRoutingKey=TEST
spring.cloud.stream.rabbit.bindings.testChannel.consumer.auto-bind-dlq=true
spring.cloud.stream.rabbit.bindings.testChannel.consumer.dlq-ttl=10000
spring.cloud.stream.rabbit.bindings.testChannel.consumer.dlq-dead-letter-exchange=

spring.cloud.stream.rabbit.bindings.testChannel2.consumer.bindingRoutingKey=TEST2
spring.cloud.stream.rabbit.bindings.testChannel2.consumer.auto-bind-dlq=true
spring.cloud.stream.rabbit.bindings.testChannel2.consumer.dlq-ttl=10000
spring.cloud.stream.rabbit.bindings.testChannel2.consumer.dlq-dead-letter-exchange=
```

But it would be cleaner if I could just specify defaults for the DLQ properties, like:

```
spring.cloud.stream.rabbit.bindings.default.consumer.auto-bind-dlq=true
spring.cloud.stream.rabbit.bindings.default.consumer.dlq-ttl=10000
spring.cloud.stream.rabbit.bindings.default.consumer.dlq-dead-letter-exchange=

spring.cloud.stream.rabbit.bindings.testChannel.consumer.bindingRoutingKey=TEST
spring.cloud.stream.rabbit.bindings.testChannel2.consumer.bindingRoutingKey=TEST2
```
And it becomes even more obvious when I have like 5 or more queueus.

I thought I saw a ticket once (can't find it now), but not sure if there was ever any resolution.

BTW, I'm on version 1.2.1.RELEASE of spring-cloud-starter-stream-rabbit.

Is this possible? Is it in the works? Is there a better way to do this for lots of queues? I'm tempted to develop my own utility to build all queues/bindings/etc and avoid having a ton of configuration. 

Thanks.

sobychacko:
Currenlty, we only support default properties for core producer/consumer (spring.cloud.stream.default.producer|consumer...).
These changes add support for configuring default properties for extended producer and consumer properties.

For example, if the binder type is foo, then this allows the applications to configure default properties across
multiple producer or consumer bindings in the form of spring.cloud.stream.foo.default.producer|consumer.property.

The default prefixes for the extended properties are dictated by the respective binder implementations.

Resolves #1360

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ExtendedBindingProperties.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ExtendedBindingProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,10 +22,38 @@ package org.springframework.cloud.stream.binder;
  *
  * @author Marius Bogoevici
  * @author Mark Fisher
+ * @author Soby Chacko
  */
 public interface ExtendedBindingProperties<C, P> {
 
 	C getExtendedConsumerProperties(String channelName);
 
 	P getExtendedProducerProperties(String channelName);
+
+	/**
+	 * Extended binding properties can define a default prefix to place all the extended
+	 * common producer and consumer properties. For example, if the binder type is foo
+	 * it is convenient to specify common extended properties for the producer or consumer
+	 * across multiple bindings in the form of `spring.cloud.stream.foo.default.producer.x=y`
+	 * or `spring.cloud.stream.foo.default.consumer.x=y`.
+	 *
+	 * The binding process will use this defaults prefix to resolve any common extended
+	 * producer and consumer properties.
+	 *
+	 * @return default prefix for extended properties
+	 * @since 2.1.0
+	 */
+	String getDefaultsPrefix();
+
+	/**
+	 *
+	 * Extended properties class against which default extended producer and consumer properties
+	 * are resolved. It is expected that this class has two properties - one called producer
+	 * and another called consumer that contains the extended properties for producer and
+	 * consumer respectively.
+	 *
+	 * @return extended properties class that contains extended producer/consumer properties
+	 * @since 2.1.0
+	 */
+	Class<?> getExtendedPropertiesEntryClass();
 }
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.stream.binding;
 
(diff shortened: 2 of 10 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -228,6 +251,30 @@ public BindingService(
 		return binding;
 	}
 
+	private void handleExtendedDefaultProperties(ExtendedPropertiesBinder binder, MergableProperties extendedProperties, String filedName) {
```
garyrussell: `fieldName` ? (But see comment about `Class<? extends Foo>`).

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ExtendedBindingProperties.java
```diff
@@ -22,10 +22,38 @@
...
+	 *
+	 * Extended properties class against which default extended producer and consumer properties
+	 * are resolved. It is expected that this class has two properties - one called producer
+	 * and another called consumer that contains the extended properties for producer and
+	 * consumer respectively.
+	 *
+	 * @return extended properties class that contains extended producer/consumer properties
+	 * @since 2.1.0
+	 */
+	Class<?> getExtendedPropertiesEntryClass();
```
garyrussell: Perhaps better to use `Class<? extends SomeInterface>` and access the properties via getters?

olegz: Actually, after discussing it with Soby, I am ok with how it is at the moment given that `MergableProperties` is already reflection heavy and that the whole thing happens on initialization, but most importantly introducing a new interface (i.e., `ProducerConsumerAware`) would require changes to all binders (i.e., `RabbitBindingProperties`)
That said, I do agree that from API design perspective we would need to address it at some point.

garyrussell: But it has nothing to do with the Binder, its the `xxxBindingProperties` - I don't see a big burden of making, say, `RabbitBindingProperties` implement something like `ProducerConsumerBinderSpecificPropertiesProvider`.

We could still use reflection to invoke `getProducer()` - my objection is using the field name in the reflection, rather than a getter.

garyrussell: Actually, with

```java
public interface BinderSpecificPropertiesProvider {

	Object getConsumer();

	Object getProducer();

}
```

I don't think you'd need to use reflection at all (since you are only looking for an `Object`, and it's ok for the implementation to narrow the type, since it's still `Object`.

Issue #1478, opened 2018-09-17
Fix FunctionInvoker to convert message payload to byte[] when returning a message. 


dturanski:
Investigating the case where I have a POJO payload and need to work with the headers. 
```
@Bean
public Function<Message<?>, Message<Foo>> fooFunction() {
	return m -> {
			Foo foo = new Foo();
			foo.setBar(m.getPayload().toString());
			return MessageBuilder.withPayload(foo).setHeader("foo","foo").build();
	};
}
```

The current behavior in `GreenfieldFunctionEnableBindingTests`  

```
@Test
public void testPojoReturn() {
	try (ConfigurableApplicationContext context = new SpringApplicationBuilder(
		TestChannelBinderConfiguration.getCompleteConfiguration(FooTransform.class)).web(
			WebApplicationType.NONE).run(
"--spring.cloud.stream.function.definition=fooFunction", "--spring.jmx.enabled=false")) {
			MessageChannel input = context.getBean("input", MessageChannel.class);
			OutputDestination target = context.getBean(OutputDestination.class);
			input.send(MessageBuilder.withPayload("bar").build());
			byte[] payload = target.receive(10000).getPayload();
		}
	}

        @EnableAutoConfiguration
	@EnableBinding(Source.class)
	public static class FooTransform {

		@Bean
		public MessageChannel input() {
			return new DirectChannel();
		}

		@Bean
		public IntegrationFlow flow() {

			return IntegrationFlows.from(input()).bridge().channel(Source.OUTPUT).get();
		}

		@Bean
		public Function<Message<?>, Message<Foo>> fooFunction() {
			return m -> {
				Foo foo = new Foo();
				foo.setBar(m.getPayload().toString());
				return MessageBuilder.withPayload(foo).setHeader("foo","foo").build();
			};
		}
	}

	static class Foo {
		String bar;

		public String getBar() {
			return bar;
		}

		public void setBar(String bar) {
			this.bar = bar;
		}
	}
```

`java.lang.ClassCastException: org.springframework.cloud.stream.function.GreenfieldFunctionEnableBindingTests$Foo cannot be cast to [B`

Since the function is now invoked from the Binder, we need to rethink this. Even the function returns a Message, The payload should always be converted to a byte[].

ilayaperumalg:
Fixed via a0b4617ac771109033d98664e2c1699157d2c1fe

PR #1479, opened 2018-09-18
DO NOT MERGE: Test demonstrating the issue

dturanski:
Demonstrates #1478


olegz:
After discussing it with @dturanski we can close it as it was already addressed with a set of previous commits related to the problem

PR #1481, opened 2018-09-18
Add function support for sink

### LINKED ISSUE #1475: Add support for composing function(s) into Processors before and after
Based on the current binding support added to channel binders we now need to be able to handle Processors which are kind of unique given that the function could be composed `before` and/or `after`. This would require additional property. Based on the internal discussion we all seem to agree that the `*function.definition...` will default to `after` and something like `*function.definition.before..` would be added for cases where one need to add function definitions before the processor.

### LINKED ISSUE #1480: Add function support for sink and processor at the binder level
Along with the source channel binding addressed in #1465, we need to support sink and processor configurations at the binder level for the function support.

ilayaperumalg:
- Update IntegrationFlowFunctionSupport with the necessary changes to bind function, consumer, supplier
 - Add changes to AbstractMessageChannelBinder to add appropriate input/output message channel configuration to accommodate
function support
 - Upate tests

Resolves #1480
Resolves #1475

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -18,10 +18,13 @@ package org.springframework.cloud.stream.binder;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.commons.logging.Log;
+import org.reactivestreams.Publisher;
 
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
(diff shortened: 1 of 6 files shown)


dturanski:
I think it resolves #1478 as well

olegz:
@dturanski I think so too

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -767,16 +784,51 @@ private void registerComponentWithBeanFactory(String name, Object component) {
...
+		}
+		return (SubscribableChannel) inputChannel;
+	}
+
+	private void moveChannelInterceptors(AbstractMessageChannel existingMessageChannel,
+			AbstractMessageChannel actualMessageChannel) {
+		for (ChannelInterceptor channelInterceptor : existingMessageChannel.getChannelInterceptors()) {
+			actualMessageChannel.addInterceptor(channelInterceptor);
+			existingMessageChannel.removeInterceptor(channelInterceptor);
+		}
```
dturanski: Love it! I had raised a concern previously about apps that might put an interceptor on output. I actually tried this in one the sftp source iterations and it's not safe to assume the it will play nice with SCSt interceptors. i.e.,  the payload was already converted to byte.  IMO, if it comes up, we should advise against stream app devs putting interceptors on bound channels.

PR #1491, opened 2018-09-27
Add (javax.)Validation Support for Stream Listener

dharrigan:
This change allows StreamListener method arguments annotated with @Valid to be
validated against a registered Validator.

An example of its usage can be found here:

https://github.com/dharrigan/gh1382

-=david=-


olegz:
@dharrigan Could you please add a test? Otherwise looks good to merge.

dharrigan:
Hi @olegz Thank you for your comment. I'm not sure how to test this as the setup and use of Spring in this regard is foreign to me. My concern is also that if no validator is on the path, then I believe it fails - so somehow I have to make the absence of a validator not cause any issue. I reached out to you gitter, so happy to talk with you about how I might go about adding a test if some pointers are forthcoming :-)

-=david=-

Issue #1497, opened 2018-10-09
Error handling for destination with ':' character


b-dzoba:
I tried to use @ServiceActivator annotation for error handling as described in reference guide, but it doesn't seem to work when RabbitMQ exchange name contains colon character. Application fails to start with an exception:
```
org.springframework.jmx.export.UnableToRegisterMBeanException: Unable to register MBean [my:destination.myGroup.errors] with key 'org.springframework.integration:type=MessageChannel,name=my:destination.myGroup.errors'; nested exception is javax.management.MalformedObjectNameException: Invalid character `:' in value
```
What is the correct way to register error handler in this case?

Sample code and properties:
```
@SpringBootApplication
@EnableBinding(Sink.class)
public class DemoApplication {

	public static void main(String[] args) {
		SpringApplication.run(DemoApplication.class, args);
	}

	@StreamListener(Sink.INPUT)
	public void handle(Object value) {
		throw new RuntimeException("BOOM!");
	}


	@ServiceActivator(inputChannel = "my:destination.myGroup.errors")
	public void error(Message<?> message) {
		System.out.println("Handling ERROR: " + message);
	}
}
```
spring.cloud.stream.bindings.input.destination=my:destination
spring.cloud.stream.bindings.input.group=myGroup

garyrussell:
Can you show the full stack trace, please (always a good idea when reporting issues like this).

garyrussell:
Sorry - I didn't see you were explicitly declaring the channel via the service activator.

garyrussell:
Fixed in [Spring Integration](https://github.com/spring-projects/spring-integration/pull/2593).

Will push a test after that is merged.

garyrussell:
The fix was back ported to Spring Integration 5.0.9 which will be in the next boot 2.0.x release.

However, the user will need to add `/META-INF/spring.integration.properties` containing `spring.integration.jmx.quote.names=true`.

Stiuil06:
Hi all.
I have similar exception when I try run application to listening mqtt broker.
Im trying to run this code:
https://docs.spring.io/spring-integration/docs/5.2.0.BUILD-SNAPSHOT/reference/html/mqtt.html#configuring-with-the-java-dsl
**Configuring with the Java DSL**

for **Configuring with Java Configuration** everything works.

My stacktrace
```
[...]

PR #1501, opened 2018-10-10
GH-1497: Add test for colon in destination

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/1497

Fix was in Spring Integration.


garyrussell:
The fix was back ported to Spring Integration 5.0.9 which will be in the next boot 2.0.x release.

However, the user will need to add `/META-INF/spring.integration.properties` containing `spring.integration.jmx.quote.names=true`.

I don't think it's necessary to backport this test case.

Issue #1517, opened 2018-10-29
Avro Union Types import schema issue when we have multiple avro files


SercanKaraoglu:
Currently, when we have multiple avro files that depends on some other avro file, there is no way to configure that dependency. For example maven plugin has the following convention

```
<imports>
        <import>${project.basedir}/src/main/resources/avro/Email.avro</import>
         <import>${project.basedir}/src/main/resources/avro/Sms.avro</import>                                  
         <import>${project.basedir}/src/main/resources/avro/PushNotification.avro</import>
</imports>
```

spring cloud stream can also have similar configuration, in the application.yml file we should be able to configure property like schema-imports.

PR #1519, opened 2018-10-29
GH-1517 Avro Union Types import schema issue when we have multiple av…

SercanKaraoglu:
…ro files

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/avro/AvroMessageConverterProperties.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/avro/AvroMessageConverterProperties.java
@@ -22,6 +22,7 @@ import org.springframework.util.Assert;
 
 /**
  * @author Vinicius Carvalho
+ * @author Sercan Karaoglu
  */
 @ConfigurationProperties(prefix = "spring.cloud.stream.schema.avro")
 public class AvroMessageConverterProperties {
@@ -30,7 +31,21 @@ public class AvroMessageConverterProperties {
 
 	private Resource readerSchema;
 
+	/**
+	 * The source directory of Apache Avro schema. This schema is used by this
+	 * converter. If this schema depends on other schemas consider defining those
+	 * those dependent ones in the {@link #schemaImports}
+	 * @parameter
+	 */
 	private Resource[] schemaLocations;
+
+	/**
+	 * A list of files or directories that should be loaded first thus making
+	 * them importable by subsequent schemas. Note that imported files
+	 * should not reference each other.
+	 * @parameter
+	 */
+	private Resource[] schemaImports;
 
 	private String prefix = "vnd";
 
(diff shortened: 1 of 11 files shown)


SercanKaraoglu:
with this commit one can do following;
in the application.yml
```
spring:
  cloud:
    stream:
      bindings:
        input:
          destination: sensor-topic
      schema-registry-client:
        endpoint: http://localhost:8990
      schema:
        avro:
          schema-locations:
            - classpath:avro/Command.avro
          schema-imports:
            - classpath:avro/Sms.avro
            - classpath:avro/Email.avro
            - classpath:avro/PushNotification.avro
```
in the src/main/resources/avro
Command.avro =>
```
{
   "namespace":"example.avro",
   "name":"Command",
   "type":"record",
   "fields":[
      {
         "name":"type",
         "type":"string"
      },
      {
         "name":"correlationId",
         "type":"string"
      },
      {
         "name":"payload",
         "type":["Sms", "Email", "PushNotification"]
      }
   ]
}
```
Email.avro =>
```
{
 "namespace":"example.avro",
 "name": "Email",
 "type": "record",
 "fields":[
     {
        "name":"addressTo",
        "type":"string"
     },
     {
        "name":"title",
        "type":"string"
     },
     {
        "name":"text",
        "type":"string"
     }
 ]
}
```
..etc.

spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/avro/AvroMessageConverterProperties.java
```diff
@@ -32,6 +33,8 @@
 
 	private Resource[] schemaLocations;
 
+	private Resource[] schemaImports;
```
olegz: I am confused. Why do we need this? Isn't it the same as `schemaLocations`?
```
"--spring.cloud.stream.schema.avro.schema-locations=classpath:schemas/Command.avsc",
"--spring.cloud.stream.schema.avro.schema-imports=classpath:schemas/imports/Sms.avsc, classpath:schemas/imports/Email.avsc, classpath:schemas/imports/PushNotification.avsc" };
```

SercanKaraoglu: Imports should be parsed before Locations, you can have a look at maven avro plugin as a reference https://github.com/apache/avro/blob/master/lang/java/maven-plugin/src/main/java/org/apache/avro/mojo/AbstractAvroMojo.java#L162

SercanKaraoglu: When you have dependent schemas like the one in the tests (command requires sms,email,push noticlfication) schema should first load the dependencies so that it can parse the root. If you remove imports and change the ordering when you do clean install you should not be able to generate classes. Let's assume you generated classes somehow even If you try this example when you use schema registry client you will see that it throws exception

olegz: Ok, fair enough. So would you be able to provide a javadoc-like short sentence for both `schemaLocations` and `schemaImports`, so I can include them as part of the polishing/merging. This will help further to avoid the confusion.

SercanKaraoglu: @olegz okay done, please check it out now

PR #1557, opened 2018-12-14
Polish around default binding IllegalStateException

phillipuniverse:
If no binders were found the exception should reflect that instead of indicating that multiple binders were found.

Use case:

- `spring-cloud-stream-test-support` is on the classpath
- An `@Configuration` class contains `@EnableBinding` that comes into play in the test
- Either auto configuration is disabled altogether or via a test slice, like in my case with `@DataMongoTest`

In this case the fix is that you need to manually `@Import(TestSupportBinderAutoConfiguration.class)`. But the `IllegalStateException` that you get is confusing:

```console
...
...
Caused by: java.lang.IllegalStateException: A default binder has been requested, but there is more than one binder available for 'org.springframework.integration.channel.DirectChannel' : , and no default binder has been set.
	at org.springframework.cloud.stream.binder.DefaultBinderFactory.getBinder(DefaultBinderFactory.java:127)
	at org.springframework.cloud.stream.binding.BindingService.getBinder(BindingService.java:308)
	at org.springframework.cloud.stream.binding.BindingService.bindProducer(BindingService.java:208)

```

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
@@ -122,8 +122,12 @@ public class DefaultBinderFactory implements BinderFactory, DisposableBean, Appl
 						this.defaultBinderForBindingTargetType.put(bindingTargetType.getName(), configurationName);
 					}
 					else {
-						throw new IllegalStateException("A default binder has been requested, but there is more than "
-								+ "one binder available for '" + bindingTargetType.getName() + "' : "
+						String countMsg = (candidatesForBindableType.size() == 0)
+								? "are no binders"
+								: "is more than one binder";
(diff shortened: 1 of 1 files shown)


phillipuniverse:
Dunno what's up here with Travis, it's got a checkstyle issue with `ContentTypeTckTests` which is not within the purview of the changes I made.

I briefly merged this into master to see if there was anything required there, but it looked like it merged cleanly even with the differences in `DefaultBinderFactory`. And it still only hits this code when there is not exactly 1 binder found and it goes back to `doGetBinding()` legacy behavior.

Let me know if there's anything you all need me to change to get this merged in!

olegz:
@phillipuniverse can you please re-issue this PR on the master instead of 2.0.x branch and then I can back port it, otherwise its a bit more complicated with merge conflicts.

phillipuniverse:
@olegz no problem, superseded by #1561.

Issue #1564, opened 2018-12-17
Delegate type conversion to MessageConverters 


olegz:
Currently due to the issue described [here](https://jira.spring.io/browse/SPR-17503) we have a few places where for certain condition we perform add-hock type conversion. This also resulted in GH-1554 which was done wrong and needs to go in its entirety.

PR #1565, opened 2018-12-17
GH-1564[DO NOT MERGE] Initial type conversion simplification

olegz:


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryConfiguration.java
```diff
@@ -161,12 +166,32 @@ public CompositeMessageChannelConfigurer compositeMessageChannelConfigurer(
...
 		DefaultMessageHandlerMethodFactory messageHandlerMethodFactory = new DefaultMessageHandlerMethodFactory();
 		messageHandlerMethodFactory.setMessageConverter(compositeMessageConverterFactory.getMessageConverterForAllRegistered());
-		messageHandlerMethodFactory.setCustomArgumentResolvers(ahmar.getResolvers());
+		
+		/*
+		 * We essentially do the same thing as the DefaultMessageHandlerMethodFactory.initArgumentResolvers(..).
+		 * We can't do it as custom resolvers for two reasons. 
+		 * 1. We would have two duplicate (compatible) resolvers and would hope fopr the right order
+		 * 2. DefaultMessageHandlerMethodFactory.initArgumentResolvers(..) puts MessageMethodArgumentResolver 
+		 * before custom converters thus not allowing an override which kind of proves #1.
```
garyrussell: I don't think "hope fopr[sic]" is the right phrase; the resolvers are in a list so the order is guaranteed.

It **is** a bit odd, though, that `PayloadAR` can be superseded, but not the others.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/SmartMessageMethodArgumentResolver.java
```diff
@@ -0,0 +1,125 @@
...
+			throw new MethodArgumentTypeMismatchException(message, parameter, "Actual message type '" +
+					ClassUtils.getDescriptiveType(message) + "' does not match expected type '" +
+					ClassUtils.getQualifiedName(targetMessageType) + "'");
+		}
+
+		
+		Class<?> targetClass = parameter.getParameterType();
+		Class<?> payloadClass = message.getPayload().getClass();
+
+		if (ClassUtils.isAssignable(payloadClass, targetClass)) {
```
garyrussell: ?? `targetClass == targetMessageType`; unlikely that the payload is a message.

Shouldn't this check be testing whether the parameter payload type (e.g. `Foo` in `Message<Foo>` ) is assignable from the actual payload type? (which is what the standard resolver is doing).

I can't get my head around `isAssignable(actualClass, wantedClass)`; surely it's `isAssignable(wantedClass, actualClass)`.

PR #1591, opened 2019-01-24
Implement PollableMessageSourcePoller

### LINKED ISSUE #1433: Support Poller or PeriodicTrigger config in PollableMessageSource
Currently the MessageHandler must provide this logic. It would be nice to be able to autowire a Poller or maybe a simple trigger.

dturanski:
This fixes #1433 (the branch named incorrectly ATM).

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
@@ -59,6 +59,7 @@ import org.springframework.util.Assert;
  *
  * @author Gary Russell
  * @author Oleg Zhurakousky
+ * @author David Turanski
  * @since 2.0
  *
  */
@@ -70,7 +71,7 @@ public class DefaultPollableMessageSource implements PollableMessageSource, Life
 		dummyChannel.setBeanName("dummy.required.by.nonnull.api");
 	}
 
-	protected static final ThreadLocal<AttributeAccessor> attributesHolder = new ThreadLocal<AttributeAccessor>();
+	protected static final ThreadLocal<AttributeAccessor> attributesHolder = new ThreadLocal<>();
 
 	private final List<ChannelInterceptor> interceptors = new ArrayList<>();
 
@@ -277,7 +278,12 @@ public class DefaultPollableMessageSource implements PollableMessageSource, Life
 	 * Receives Message from the source and converts its payload to a provided type.
 	 * Can return null
 	 */
-	private Message<?> receive(ParameterizedTypeReference<?> type) {
+	@Override
+	public Message<?> receive(ParameterizedTypeReference<?> type) {
+		//Prevent a race condition if this is called before setSource().
+		if (this.source == null) {
+			return null;
+		}
 		Message<?> message = this.source.receive();
 		if (message != null && type != null && this.messageConverter != null) {
 			Class<?> targetType = type == null ? Object.class :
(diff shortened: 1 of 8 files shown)


dturanski:
Thinking about something like

```
@PollableMessageSource(poller=@Poller{...}, payloadType=String.class)
@Bean
public MessageHandler() {
...
} 
```

garyrussell:
At first glance, this seems to be orthogonal to the original driver for `PollableMessageSource`, which was specifically designed for on-demand polling - i.e. fetching a record only when the application is ready to handle one.

It's not clear what the use case is here.

In any case, since the `DefaultPollableMessageSource` is a wrapper around an SI `MessageSource`, much of this implementation is already available out-of-the box with the DSL; something like...

```java
    @Bean
    public IntegrationFlow flow() {
        IntegrationFlows.from(myKafkaMessageSource(), e -> e.poller=Pollers.fixedDelay(Duration.ofSeconds(5))
                             .channel(someChanelWithStreamInterceptors)
                             .handle("someBean", "someMethod") // <<<<< service activator
                             .get();
    }
```

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -277,7 +278,12 @@ protected boolean shouldRequeue(Exception e) {
 	 * Receives Message from the source and converts its payload to a provided type.
 	 * Can return null
 	 */
-	private Message<?> receive(ParameterizedTypeReference<?> type) {
+	@Override
+	public Message<?> receive(ParameterizedTypeReference<?> type) {
+		//Prevent a race condition if this is called before setSource().
+		if (this.source == null) {
+			return null;
+		}
```
garyrussell: How can that be possible? The poller won't start until all beans have been populated.

dturanski: I thought that too, but there is definitely a race condition.  I was seeing a couple of NPEs at first when running the test.

dturanski:
Point taken, the periodic trigger doesn't help if you're not ready for the message.  What is the meaning of `.channel(someChanelWithStreamInterceptors)`

garyrussell:
>What is the meaning of...

I was just suggesting that adding the stream interceptors to a channel wired into a DSL flow would provide this functionality without writing any extra code (except to add the interceptors).

garyrussell:
Actually - it's even simpler than I thought, we can just use `@Scheduled` with no framework changes at all..

```java
@Scheduled(fixedDelay = 5_000)
public void poll() {
	System.out.println("Polling...");
	this.source.poll(m -> {
		System.out.println(m.getPayload());

	}, new ParameterizedTypeReference<Foo>() { });
}
```


Full app...

```java
@SpringBootApplication
@EnableScheduling
@EnableBinding(Foo.class)
public class PmsScheduledApplication {

	final PollableMessageSource source;

	@Autowired
	public PmsScheduledApplication(PollableMessageSource source) {
		this.source = source;
	}

	@Scheduled(fixedDelay = 5_000)
	public void poll() {
		System.out.println("Polling...");
		this.source.poll(m -> {
			System.out.println(m.getPayload());

		}, new ParameterizedTypeReference<Foo>() { });
	}

	@Bean
	public ApplicationRunner runner(KafkaTemplate<byte[], byte[]> template) {
		return args -> IntStream.range(0, 5)
				.forEach(i -> template.send("source", ("{\"bar\":\"baz" + i + "\"}").getBytes()));
	}

		SpringApplication.run(PmsScheduledApplication.class, args);
	}

	public static class Foo {

		private String bar;

		public String getBar() {
			return this.bar;
		}

		public void setBar(String bar) {
			this.bar = bar;
		}

		@Override
		public String toString() {
			return "Foo [bar=" + this.bar + "]";
		}

	}

}

interface Foo {

	@Input
	PollableMessageSource source();

}
```

```
Foo [bar=baz0]
Polling...
Foo [bar=baz1]
Polling...
Foo [bar=baz2]
Polling...
Foo [bar=baz3]
Polling...
Foo [bar=baz4]
```

dturanski:
Closing - feature not needed.

PR #1602, opened 2019-02-06
Support KafkaNull in payload conversion

garyrussell:
See https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/539

We need to map payloads of type `KafkaNull` to `null` when mapping method
arguments (`@Payload(required = false) Foo foo`.


garyrussell:
FYI; this is already handled in SI with the `NullAwarePayloadArgumentResolver` - so maybe when the SPR issue is resolved, the SI resolvers can be used.

PR #1688, opened 2019-04-15
Remove usage of deprecated code

artembilan:
The `HandlerMethodArgumentResolversHolder` has been deprecated in SI-5.1
and now it is removed in SI-5.2

To make current Spring cloud Stream compatible with the latest Spring Boot,
which is pulled by the latest Spring Cloud, remove usage of that
deprecated class in the current version in favor of internal logic
applied in the SI

Looks like the `BinderFactoryConfiguration` is deprecated as well and
is going to be removed in the future release anyway


sobychacko:
Looks good. Merged upstream.

PR #1697, opened 2019-04-24
Make class public

G-Lem:
I'd like to implement my own avro msg converter but I can't reuse OriginalContentTypeResolver


olegz:
Original Content Type is no longer used by the framework. It is there only for backward compatibility with older versions of spring-cloud-stream. This also means that this class and all references to original-content-type will be removed all together. 

Consider looking at `AbstractAvroMessageConverter`, `AvroSchemaMessageConverter`, `AvroSchemaRegistryClientMessageConverter` which are specifically designed as extension points to customize conversion

PR #1705, opened 2019-05-08
Mismatch with server receive parameter format

zhouzhou19950825:
The problems I found in the data interaction test between client and server
.line91,line 104,line 151,line 178  corrected it. And method getSubjectVersion(...) no need;


olegz:
Please clean up commented code and provide at least one test to validate the changes

PR #1706, opened 2019-05-08
Mismatch with server receive parameter format

zhouzhou19950825:
The problems I found in the data interaction test between client and server
.line91,line 104,line 151,line 178  corrected it. And method getSubjectVersion(...) no need;

### DIFF
--- a/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/ConfluentSchemaRegistryClient.java
+++ b/spring-cloud-stream-schema/src/main/java/org/springframework/cloud/stream/schema/client/ConfluentSchemaRegistryClient.java
@@ -75,22 +75,17 @@ public class ConfluentSchemaRegistryClient implements SchemaRegistryClient {
 	public SchemaRegistrationResponse register(String subject, String format,
 			String schema) {
 		Assert.isTrue("avro".equals(format), "Only Avro is supported");
-		//String path = String.format("/subjects/%s/versions", subject);
 		HttpHeaders headers = new HttpHeaders();
 		headers.put("Accept", ACCEPT_HEADERS);
 		headers.add("Content-Type", "application/json");
 		Integer version = null;
 		Integer id = null;
 		String payload = null;
-		//add map
 		Map<String,String> maps=new HashMap<>();
         maps.put("subject",subject);
         maps.put("format",format);
         maps.put("definition",schema);
 		try {
-			//payload = this.mapper
-			//		.writeValueAsString(Collections.singletonMap("schema", schema));
-			//map exchange string (request data)
 			payload = this.mapper.writeValueAsString(maps);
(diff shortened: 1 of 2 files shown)


zhouzhou19950825:
ConfluentSchemaRegistryClientTests  need to update

zhouzhou19950825:
Fix the schema register to server

sobychacko:
@zhouzhou19950825 Could you elaborate a bit further about the issues you faced that made this PR necessary? It's not entirely clear if you got any errors or if these are improvements.

olegz:
@zhouzhou19950825 As Soby asked earlier, could you please elaborate on the purpose of this PR as it appears to change the existing functionality and may affect other users.

zhouzhou19950825:
> @zhouzhou19950825 As Soby asked earlier, could you please elaborate on the purpose of this PR as it appears to change the existing functionality and may affect other users.

Hi,olegz.Simply put, the client's request does not match the server's interface.For example,the server code` @RequestMapping(
        method = {RequestMethod.POST},
        path = {"/"},
        consumes = {"application/json"},
        produces = {"application/json"}
    )
    public synchronized ResponseEntity<Schema> register(@RequestBody Schema schema, UriComponentsBuilder builder) {....}`,but the ConfluentSchemaRegistryClient register path` public SchemaRegistrationResponse register(String subject, String format, String schema) {
        Assert.isTrue("avro".equals(format), "Only Avro is supported");
        **String path = String.format("/subjects/%s/versions", subject);**
        HttpHeaders headers = new HttpHeaders();
        headers.put("Accept", ACCEPT_HEADERS);
        headers.add("Content-Type", "application/json");`,it's not match. And if server controller completed register,you can return version without requiring another request `if (registeredEntities != null && registeredEntities.size() != 0) {
                result = validator.match(registeredEntities, schema.getDefinition());
                if (result == null) {
                    schema.setVersion(((Schema)registeredEntities.get(registeredEntities.size() - 1)).getVersion() + 1);
                    result = (Schema)this.repository.save(schema);
                }
            } else {
                schema.setVersion(1);
                result = (Schema)this.repository.save(schema);
            }` but in client ` public SchemaRegistrationResponse register(String subject, String format, String schema) {....HttpEntity<String> request = new HttpEntity(payload, headers);
            ResponseEntity<Map> response = this.template.exchange(this.endpoint + path, HttpMethod.POST, request, Map.class, new Object[0]);
            id = (Integer)((Map)response.getBody()).get("id");
            version = this.getSubjectVersion(subject, payload);...}`.so,So I made some changes, not changing the original method interface, not affecting user use.
                                                                                                                                    Thanks!

sobychacko:
@zhouzhou19950825 Merged this PR. We will move your changes over to the new [schema registry repository](https://github.com/spring-cloud/spring-cloud-schema-registry).

PR #1730, opened 2019-06-12
[DO NOT MERGE UNTIL RELEVANT KAFKA/RABBIT/SI changes are available]GH-1729 Added support for DLQ to functional programming model

### LINKED ISSUE #1729: Add support for DLQ error handling within functional programming model
I'm been reading available documentation and I am very confused about the correct approach to get an error topic working when using the Kafka binder. 

The [current documentation](https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/#spring-cloud-stream-overview-error-handling) states that the property `spring.cloud.stream.bindings.error.destination` can be used to configure a global error topic, but this support appears to have been removed in [this commit](https://github.com/spring-cloud/spring-cloud-stream/commit/a82c906913d9dbf0c42810443adce3ec1b2b5eb6).

I have also tried to implement using properties for the Kafka binder specifically; the code below demonstrates my attempt, including the configuration.

My Maven BOM is `org.springframework.cloud:spring-cloud-dependencies:Greenwich.SR1` and my dependencies are:

```
    org.springframework.boot:spring-boot-starter-web
    org.springframework.cloud:spring-cloud-stream
    org.springframework.cloud:spring-cloud-starter-stream-kafka
```

**It would be great if it could be clarified how exactly a global error topic can be set up**. So far, with the code below, I can see the message being handled multiple times via retries, but nothing is sent to the error topic (DLQ) that I have configured.

```
@SpringBootApplication
public class Application {

    public static void main(final String[] args) {
        SpringApplication.run(Application.class, args);
    }

}
```
```
@Configuration
public class AppConfig {

    @Bean
    public ServiceImpl service() {
        return new ServiceImpl();
    }

    @Bean
    public Function<String, String> execute(ServiceImpl service) {
        return service::doBusinessLogic;
    }
}
```
```
public class ServiceImpl {

    public String doBusinessLogic(String message) {
        System.out.println("Handling message");
        if (message.startsWith("foo")) {
            throw new RuntimeException("This is a test");
        }
        return message;
    }

}
```
```
spring:
  cloud:
    stream:
      kafka:
        binder:
          brokers: localhost:9091
          autoCreateTopics: false
        bindings:
          input:
            consumer:
              enableDlq: true
              dlqName: myErrorTopic

      bindings:
        input:
          destination: myInputTopic
          group: myConsumerGroup
        output:
          destination: myOutputTopic

      function:
        definition: execute
```

olegz:
- Change FunctionInvoker.onError() to send to binding error channel instead of global one
- Binders contain relevant changes as well

Resolves #1729

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionInvoker.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionInvoker.java
@@ -27,11 +27,13 @@ import org.apache.commons.logging.LogFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
+import org.springframework.beans.factory.BeanFactory;
 import org.springframework.cloud.function.context.FunctionCatalog;
 import org.springframework.cloud.function.context.FunctionType;
 import org.springframework.cloud.function.context.catalog.FunctionInspector;
 import org.springframework.cloud.stream.binder.ConsumerProperties;
 import org.springframework.cloud.stream.binder.ProducerProperties;
+import org.springframework.cloud.stream.config.BindingProperties;
 import org.springframework.cloud.stream.config.BindingServiceProperties;
 import org.springframework.cloud.stream.converter.CompositeMessageConverterFactory;
 import org.springframework.integration.support.MessageBuilder;
@@ -72,7 +74,7 @@ class FunctionInvoker<I, O> implements Function<Flux<Message<I>>, Flux<Message<O
 
 	private final CompositeMessageConverter messageConverter;
 
-	private final MessageChannel errorChannel;
+	private final BeanFactory beanFactory;
 
 	private final boolean isInputArgumentMessage;
 
@@ -95,7 +97,7 @@ class FunctionInvoker<I, O> implements Function<Flux<Message<I>>, Flux<Message<O
 	FunctionInvoker(StreamFunctionProperties functionProperties,
 			FunctionCatalog functionCatalog, FunctionInspector functionInspector,
 			CompositeMessageConverterFactory compositeMessageConverterFactory,
-			MessageChannel errorChannel) {
+			BeanFactory beanFactory) {
 
 		this.functionProperties = functionProperties;
 		Object originalUserFunction = functionCatalog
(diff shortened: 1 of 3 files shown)


olegz:
https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/pull/250

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionInvoker.java
```diff
@@ -141,10 +143,19 @@
 	}
 
 	private void onError(Throwable t, Message<I> originalMessage) {
-		if (this.errorChannel != null) {
-			ErrorMessage em = new ErrorMessage(t, (Message<?>) originalMessage);
+		String inputDestinationName = functionProperties.getInputDestinationName();
+		BindingProperties bindingProperties = functionProperties.getBindingServiceProperties().getBindings().get(inputDestinationName);
+		String destinationName = bindingProperties.getDestination();
+		String groupName = bindingProperties.getGroup();
+		String bindingErrorChannelName = destinationName + "." + groupName + ".errors";
```
garyrussell: Perhaps change `AbstractMessageChannelBinder.errorsBaseName()` to `public static` and call it here?

olegz: There is a `C` type for consumer properties which is not really used, but I would need to change it to Object or something. . . I'll keep it as is for now since there may be more refactoring soon

garyrussell: ok

AbdulZafar-Smarsh:
Hey @olegz. Any plans on merging this PR in the future to extend DLQ functionality to support reactive frameworks?

sobychacko:
@AbdulZafar-Smarsh There are no current plans for doing so. It is hard to do that since the reactive pipeline bootstrapping happens at startup, and Spring Cloud Stream cannot access the reactive context at runtime. We recommend that applications programmatically handle errors. For example, applications can use a `KafkaTemplate` or `StreamBridge` to send the record in error to a DLT.

PR #1738, opened 2019-06-19
Support Batch Listeners in Core (Function)

garyrussell:
- Support `batchMode` in function properties so, for binders that support batch listeners, a message
  payload of `List<byte[]>` can be properly converted.
- In batch mode, handle
    `public Function<List<Person>, Person> func()`
    `public Function<List<List<Person>>, Person> func()`
    `Function<Message<List<Person>>, Person> func()`
- Also use `ParameterizedType` as a conversion hint for `public Function<List<Person>, Person> func()`
  when not in batch mode (json: `[{\"name\":\"bob\"},{\"name\":\"jill\"}]`). Previously, just the'
  class was used without generic type information.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/1742.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-stream.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-stream.adoc
@@ -587,6 +587,8 @@ support.
 [[spring_cloud_function]]
 ==== Spring Cloud Function support
 
+===== Overview
+
 Since Spring Cloud Stream v2.1, another alternative for defining _stream handlers_ and _sources_ is to use build-in
 support for https://cloud.spring.io/spring-cloud-function/[Spring Cloud Function] where they can be expressed as beans of
  type `java.util.function.[Supplier/Function/Consumer]`.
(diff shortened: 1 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamFunctionProperties.java
```diff
@@ -40,6 +40,8 @@
 
 	private String outputDestinationName = Processor.OUTPUT;
 
+	private boolean batchMode;
```
garyrussell: I didn't notice that the `FunctionInvoker` has access to the `ConsumerProperties`. I think we should move this there; that way, the same flag will be used by both the invoker and the binder (if the binder supports batching).

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -712,7 +714,19 @@ For example, the above composition could be defined as such (if both functions p
 --spring.cloud.stream.function.definition=reactiveUpperCase|wrapInQuotes
 ----
 
+===== Batch Consumers
+
+When using a `MessageChannelBinder` that supports batch listeners, and the feature is enabled for the consumer binding, you can set `spring.cloud.stream.function.definition` to `true` to enable the entire batch of messages to be passed to the function in a `List`.
```
olegz: Hmm, `spring.cloud.stream.function.definition`? Did you mean `spring.cloud.stream.bindings.input.consumer.batch-mode=true`?

olegz:
This is interesting but given the work that is happening https://github.com/spring-cloud/spring-cloud-function/tree/lazy-func-catalog-multi where type/message conversion will be handled by s-c-function, I now wonder if we should just use tests and whatever relevant configuration presented in this PR, but leave the actual conversion bits to s-c-function providing the aforementioned branch merged relatively soon

PR #1753, opened 2019-06-30
Bypassing target type binding 

### LINKED ISSUE #1751: Enable target binding by downstream binders
When using the functional model, there are cases in which the binders want to take over the target binding responsibility. For example, when `EnableBinding` is removed from the application, there are no trivial ways for the core framework to know the target types (Unless we can include them at compile time which is not possible for extended binders like Kafka Streams binder since we don't want to include those types in the core framework). An alternative way is to let the binders do the target binding by bypassing it in the core framework. The bypassing only happens when `EnableBinding` is not found and the binders themselves indicate that they are capable of doing the target binding.

sobychacko:
When functional model is used and no EnableBinding is provided, this PR allows
core Spring Cloud Stream to skip any target type binding if the downstream
binder is capable of such binding on the target type. In that case, the target
binders provide their own BindableTargetProxyFactory.

Introduce additional properties in StreamFunction properties to allow overriding the
input/output bindings.

Resolves #1751

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindableProvider.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindableProvider.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2019-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.config;
+
+/**
+ * If downstream binders want to take over the responsibility of binding the target types (such as the Kafka Streams binder),
+ * then they can implement this functional interface to signal the core framework to bypass any binding.
+ *
+ * @author Soby Chacko
+ * @since 3.0.0
+ */
+@FunctionalInterface
+public interface BindableProvider {
+
+	/**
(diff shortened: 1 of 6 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindableProvider.java
```diff
@@ -0,0 +1,37 @@
...
+public interface BindableProvider {
+
+	/**
+	 * Based on the type provided, the implementation can determine whether it is capable of
+	 * binding this target type.
+	 *
+	 * @param clazz target type to bind
+	 * @return true if capable of binding
+	 */
+	boolean canBind(Class<?> clazz);
```
olegz: I don't think we need this method. In fact we simply need a marker interface where the individual binders tell the core "I got it"

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BinderFactoryAutoConfiguration.java
```diff
@@ -260,14 +260,35 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
...
+
+						boolean bindDownstream = false;
+						try {
+							Map<String, BindableProvider> bindableProviders = beanFactory.getBeansOfType(BindableProvider.class);
+							Class<?> inputTypeWrapper = inspector.getInputWrapper(definedFunction);
+							if (bindableProviders != null) {
+								Collection<BindableProvider> values = bindableProviders.values();
+								for (BindableProvider bindableProvider : values) {
+									if (bindableProvider.canBind(inputTypeWrapper)) {
+										bindDownstream = true;
```
olegz: As described in the previous comment, you can actually have annotation (e.g., `BindingProvider`) and annotate the configuration class provided by the binder itself. Then we can simply do:
```
Class<?>[] configurationClasses = bfac.getAll().values().iterator().next().getConfigurationClasses();
		boolean bindingProvider = Stream.of(configurationClasses)
				.filter(clazz -> AnnotationUtils.findAnnotation(clazz, BindingProvider.class) != null)
				.findFirst().isPresent();
. . .
if (!bindingProvider) {
. . .
}
```

PR #1759, opened 2019-07-05
Fix pollable consumer auto-startup

### LINKED ISSUE #1758: Auto-startup setting not honored for Pollable Consumers
Check to see if the consumer should be auto-started is missing when binding pollable consumers.

https://github.com/spring-cloud/spring-cloud-stream/blob/35dff8762a643efc0c1f3d6f1724630f0141e7db/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java#L490-L492

Walliee:
resolves #1758


Walliee:
Thanks @olegz! Any chance this can be included in 2.1.x and 2.2.x patch releases?

olegz:
I back-ported it to 2.2.x

PR #1807, opened 2019-09-17
[REVIEW ONLY] - Add initial support for functions with multiple inputs/outputs as well…

olegz:
… as multiple functions

- Currently there is no multi-out support for Supplier and I am not really sure what the use case would be so holding off.
- The logic in FunctionConfiguration effectively split in bootstrapping simple functions (e.g., Function<String, String>) vs. multi-in/out (e.g., Function<Tuple2<Flux<String>, Flux<String>>, Flux<String>>). This is temporary given that multi-in/out is still considered WIP. Once it becomes stable we can merge the two two for consistency.
- Added multiple input/output support for TestBinder

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/BindableFunctionProxyFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/BindableFunctionProxyFactory.java
@@ -22,31 +22,41 @@ import org.springframework.cloud.stream.binding.BoundTargetHolder;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.SubscribableChannel;
 import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
 
 /**
  * {@link FactoryBean} for creating inputs/outputs destinations to be bound to
  * function arguments. It is an extension to {@link BindableProxyFactory} which
  * operates on Bindable interfaces (e.g., Source, Processor, Sink) which internally
- * define inputs and output channels. Unlike BindableProxyFactory, this class simply
- * operates based on the count of provided inputs and outputs and the names of inputs and outputs
- * are based on convention.
+ * define inputs and output channels. Unlike BindableProxyFactory, this class
+ * operates based on the count of provided inputs and outputs deriving the binding
+ * (channel) names based on convention - {@code `<function-definition>_ + <in/out> + _<index>`}
+ * <br>
+ * For example, `myFunction_in_0` - is the binding for the first input argument of the
+ * function with the name `myFunction`.
  *
  * @author Oleg Zhurakousky
  *
- *
- *
  * @since 3.0
  */
-public class BindableFunctionProxyFactory extends BindableProxyFactory {
+class BindableFunctionProxyFactory extends BindableProxyFactory {
 
 	private final int inputCount;
 
 	private final int outputCount;
 
-	public BindableFunctionProxyFactory(int inputCount, int outputCount) {
+	private final String functionDefinition;
+
+	private final boolean nameBasedOnFunctionName;
+
+	private boolean multiple;
+
+	BindableFunctionProxyFactory(String functionDefinition, int inputCount, int outputCount, boolean nameBasedOnFunctionName) {
 		super(null);
 		this.inputCount = inputCount;
 		this.outputCount = outputCount;
+		this.functionDefinition = functionDefinition;
+		this.nameBasedOnFunctionName = nameBasedOnFunctionName;
 	}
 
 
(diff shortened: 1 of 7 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/BindableFunctionProxyFactory.java
```diff
@@ -55,25 +65,73 @@ public void afterPropertiesSet() {
...
+	}
+
+	protected String getFunctionDefinition() {
+		return this.functionDefinition;
+	}
+
+	protected String getInputName(int index) {
+		return CollectionUtils.isEmpty(this.getInputs())
+				? null
+						: this.getInputs().toArray(new String[] {})[index];
```
artembilan: Looks like ` buildInputNameForIndex(int index)` produces exactly the same, but doesn't create an extra array in between.
Anyway it is recommended to use `new String[0] for `toArray()`...

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionConfiguration.java
```diff
@@ -99,83 +101,87 @@
...
-
 		return integrationFlow;
 	}
 
+	/*
+	 * Creates a publishing trigger to ensure Supplier does not begin publishing until binding is created
+	 */
+	private Publisher<Object> setupBindingTrigger(GenericApplicationContext context) {
+		AtomicReference<MonoSink<Object>> triggerRef = new AtomicReference<>();
+		Publisher<Object> beginPublishingTrigger = Mono.create(emmiter -> {
```
artembilan: Consider to use `MonoProcessor` instead. The `onComplete()` will be called from the event instead of this hacks.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionConfiguration.java
```diff
@@ -224,111 +230,181 @@ private IntegrationFlowBuilder integrationFlowFromProvidedSupplier(Supplier<?> s
...
+				return this.enhancePublisher(MessageChannelReactiveUtils.toPublisher(inputChannel), inputBindingName);
+			}).toArray(Publisher[]::new);
+
+
+			Object resultPublishers = function.apply(inputPublishers.length == 1 ? inputPublishers[0] : Tuples.fromArray(inputPublishers));
+			if (resultPublishers instanceof Iterable) {
+				Iterator<String> outputBindingIter = outputBindingNames.iterator();
+				((Iterable) resultPublishers).forEach(publisher -> {
+					MessageChannel outputChannel = context.getBean(outputBindingIter.next(), MessageChannel.class);
+					Flux.from((Publisher) publisher).doOnNext(message -> outputChannel.send((Message) message)).then().subscribe();
```
artembilan: Why do we need that `then()`? Why just `subscribe()` doesn't work?

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
```diff
@@ -234,7 +234,7 @@ public void testFiniteFluxSupplierSimple() {
 			assertThat(new String(target.receive(2000).getPayload())).isEqualTo("6");
 
 			assertThat(context.getBean("supplierInitializer")).isNotEqualTo(null);
-			assertThat(context.getBean("functionInitializer")).isEqualTo(null);
+//			assertThat(context.getBean("functionInitializer")).isEqualTo(null);
```
artembilan: What made you to comment this out, please?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionConfiguration.java
```diff
@@ -426,37 +535,35 @@ public void afterPropertiesSet() throws Exception {
...
-						this.inputCount = 1;
-						this.outputCount = 0;
-					}
-					else {
-						this.inputCount = 1;
-						this.outputCount = 1;
+				String[] functionDefinitions = streamFunctionProperties.getDefinition().split(";");
+				boolean nameBasedOnFunctionName = functionDefinitions.length > 1;
+				for (String functionDefinition : functionDefinitions) {
+					RootBeanDefinition functionBindableProxyDefinition = new RootBeanDefinition(BindableFunctionProxyFactory.class);
```
artembilan: Consider to use `BeanDefinitionBuilder` instead, so you wouldn't need to call `getConstructorArgumentValues()` all the time.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/BindableFunctionProxyFactory.java
olegz: True, but `getInputName` returns the cached version, while he other one builds a new string so it could be added to `input/outputHolders` (existing functionality from BindableProxyFactory)

spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/SourceToFunctionsSupportTests.java
olegz: It is no longer true so I'll remove it all together

PR #1810, opened 2019-09-25
Add Spring Integration as function sample

olegz:
Add Spring Integration as function docs

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-stream.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-stream.adoc
@@ -106,8 +106,7 @@ To run a Spring Cloud Stream application in production, you can create an execut
 === The Binder Abstraction
 
 Spring Cloud Stream provides Binder implementations for https://github.com/spring-cloud/spring-cloud-stream-binder-kafka[Kafka] and https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit[Rabbit MQ].
-Spring Cloud Stream also includes a https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-test-support/src/main/java/org/springframework/cloud/stream/test/binder/TestSupportBinder.java[TestSupportBinder], which leaves a channel unmodified so that tests can interact with channels directly and reliably assert on what is received.
-You can also use the extensible API to write your own Binder.
+Spring Cloud Stream also includes a test binder for integration testing of your applications as spring-cloud-stream application. See <<Testing>> section for more details.
 
 Spring Cloud Stream uses Spring Boot for configuration, and the Binder abstraction makes it possible for a Spring Cloud Stream application to be flexible in how it connects to middleware.
 For example, deployers can dynamically choose, at runtime, the destinations (such as the Kafka topics or RabbitMQ exchanges) to which channels connect.
@@ -208,7 +207,7 @@ NOTE: To set up a partitioned processing scenario, you must configure both the d
 To understand the programming model, you should be familiar with the following core concepts:
 
 * *Destination Binders:* Components responsible to provide integration with the external messaging systems.
-* *Destination Bindings:* Bridge between the external messaging systems and application provided _Producers_ and _Consumers_ of messages (created by the Destination Binders).
+* *Bindings:* Bridge between the external messaging systems and application provided _Producers_ and _Consumers_ of messages (created by the Destination Binders).
 * *Message:* The canonical data structure used by producers and consumers to communicate with Destination Binders (and thus other applications via external messaging systems).
 
 image::{github-raw}/docs/src/main/asciidoc/images/SCSt-overview.png[width=800,scaledwidth="75%",align="center"]
@@ -221,14 +220,14 @@ This integration is responsible for connectivity, delegation, and routing of mes
 invocation of the user code, and more.
 
 Binders handle a lot of the boiler plate responsibilities that would otherwise fall on your shoulders. However, to accomplish that, the binder still needs
-some help in the form of minimalistic yet required set of instructions from the user, which typically come in the form of some type of configuration.
+some help in the form of minimalistic yet required set of instructions from the user, which typically come in the form of some type of _binding_ configuration.
 
 While it is out of scope of this section to discuss all of the available binder and binding configuration options (the rest of the manual covers them extensively),
-_Destination Binding_ does require special attention. The next section discusses it in detail.
+_Binding_ as a concept, does require special attention. The next section discusses it in detail.
 
-=== Destination Bindings
+=== Bindings
 
-As stated earlier, _Destination Bindings_ provide a bridge between the external messaging system and application-provided _Producers_ and _Consumers_.
+As stated earlier, _Bindings_ provide a bridge between the external messaging system (e.g., queue, topic etc.) and application-provided _Producers_ and _Consumers_.
 
 The following example shows a fully configured and functioning Spring Cloud Stream application that receives the payload of the message
 as a `String` type (see <<Content Type Negotiation>> section), logs it to the console and sends it down stream after converting it to upper case.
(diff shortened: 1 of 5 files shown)


docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+
+For those who are familiar with SI you can see we are simply defining a bean of type `IntegrationFlow` where we
+declaring an integration flow we want to expose as a `Function<String, String>` (using SI DSL) by the name `uppercase`.
+The `MessageFunction` interface here simply allows us to explicitly declare the type of the inputs and outputs for
+the purpose of proper type conversion.
+For cases where you simply want to receive raw input you could use `from(Function.class, ...)`.
+
+The resulting function will be bound to the input and output destination exposed by the target binder.
+For simple cases you don't need to now the names of such destinations, however for most cases where you want to configure
+destination's Producer/Consumer properties you do need to know such names. For cases where this is the only function
```
artembilan: Doesn't look like names concern for properties should be mentioned here.
Looking at this I just confused how that `from(Function)` is relevant to names concern.

It might be better to mention that you need to declare your own `Function` extension interface to make multi-binding via `from(Class<?>)` as well.

The names convention should be mentioned in some other place.
Although we definitely might need to link there from here.

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+that you intend to use for bindings, the names are `input` and `output`. For cases where you may have multiple bindings (<<Multiple functions in a single application>>)
+the binding naming conventions is as follows:
+
+*Binding naming convention:*
+
+* input - `<functionName> + .in. + <index>`
+* output - `<functionName> + .out. + <index>`
+
+So for single binding cases an example of configuration property for input is `spring.cloud.stream.bindings.input.destination=myQueue` and for
+multiple-binidng cases ``spring.cloud.stream.bindings.uppercase.in.0.destination=myQueue``.
```
artembilan: Looks like typo: why double back-ticks ?

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+	}
+
+	public interface MessageFunction extends Function<Message<String>, Message<String>> {
+
+	}
+}
+----
+
+For those who are familiar with SI you can see we are simply defining a bean of type `IntegrationFlow` where we
+declaring an integration flow we want to expose as a `Function<String, String>` (using SI DSL) by the name `uppercase`.
```
Buzzardo: Change to "declare an integration flow that we want to expose as a `Function<String, String>` (using SI DSL) called"

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+The `MessageFunction` interface here simply allows us to explicitly declare the type of the inputs and outputs for
```
Buzzardo: Change to "The `MessageFunction` interface lets us explicitly declare the type of the inputs and outputs for"

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+the purpose of proper type conversion.
```
Buzzardo: Remove "the purpose of"

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+For cases where you simply want to receive raw input you could use `from(Function.class, ...)`.
```
Buzzardo: Change to "To receive raw input, you can use `from(Function.class, ...)`."

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+
+The resulting function will be bound to the input and output destination exposed by the target binder.
```
Buzzardo: Change "will be" to "is"

Is the destination the same? If not, it should be "destinations" (plural).

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+For simple cases you don't need to now the names of such destinations, however for most cases where you want to configure
```
Buzzardo: Change to "For simple cases, you need not know the names of such destinations. However, for most cases, where you want to configure the"

docs/src/main/asciidoc/spring-cloud-stream.adoc
Buzzardo: Change to "destination's Producer or Consumer properties, you do need to know such names. For cases where this is the only function"

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+destination's Producer/Consumer properties you do need to know such names. For cases where this is the only function
+that you intend to use for bindings, the names are `input` and `output`. For cases where you may have multiple bindings (<<Multiple functions in a single application>>)
```
Buzzardo: Add "see" before the link (inside the parentheses).

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+the binding naming conventions is as follows:
```
Buzzardo: Change "is" to "are"

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+
+*Binding naming convention:*
+
+* input - `<functionName> + .in. + <index>`
```
Buzzardo: Change to "Input: `<functionName> + .in. + <index>`" (If "input" is code, format it as code. Keep the colon either way.)

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+* output - `<functionName> + .out. + <index>`
```
Buzzardo: Change to "Output: `<functionName> + .out. + <index>`" (If "output" is code, format it as code. Keep the colon either way.)

docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -709,7 +709,60 @@ public Function<List<Person>, Person> findFirstPerson() {
...
+
+So for single binding cases an example of configuration property for input is `spring.cloud.stream.bindings.input.destination=myQueue` and for
```
Buzzardo: Add a comma after "so", hyphenate "single-binding", and add a comma after "cases".

docs/src/main/asciidoc/spring-cloud-stream.adoc
Buzzardo: Also, "multiple-binidng" should be "multiple-binding".

PR #1811, opened 2019-09-26
Add documentation for polling suppliers

olegz:


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/ImplicitFunctionBindingTests.java
```diff
@@ -394,4 +416,17 @@ public IntegrationFlow uppercaseFlow() {
...
 	}
 
+
+	@EnableAutoConfiguration
+	public static class SupplierWithExplicitPollerConfiguration {
+
+		@Bean
+		public Supplier<String> supplier() {
+			return () -> {
+				System.out.println("Invoking supplier");
```
artembilan: I think we don't need this SOUT in the code at all.

PR #1815, opened 2019-10-15
custom converter priority

wonwoo:
I think custom converter should be applied priority.
 
What do you think?


olegz:
Can you please explain your intention and also attach a test that would fail before your suggested PR?

olegz:
This has actually been fixed as part of https://github.com/spring-cloud/spring-cloud-stream/issues/1851

PR #1824, opened 2019-10-17
Publishing template customizer

ArtyomGabeev:
Add PublishingTemplateCustomizer which should be used to customize underlying template, e.g. RabbitTemplate to add sleuth post processor (https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/265).

As I understand this PR needs to be merged first, before we can integrate customizer into specific binders implementations.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-stream.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-stream.adoc
@@ -1952,6 +1952,23 @@ See `<<spring-cloud-stream-overview-error-handling>>` for more information.
 +
 Default: `false`.
 
+==== Advanced Producer Configuration
+
+For advanced configuration of the underlying publishing template, add a single `PublishingTemplateCustomizer` bean to the application context.
+It will be invoked after the above properties have been applied and can be used to set additional properties.
+
+The following is an example for the RabbitMQ binder:
+
+====
+[source, java]
+----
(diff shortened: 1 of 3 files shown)


artembilan:
See also here: #1828 

When we make an `AmqpOutboundEndpoint.getRabbitTemplate()` as `public`, such a requirement will be covered by that `ProducerMessageHandlerCustomizer`

ArtyomGabeev:
hi @artembilan
It makes sense, thanks!

Do I need to change rabbit template visibility after https://github.com/spring-cloud/spring-cloud-stream/pull/1828 is merged?

artembilan:
I think you can change it even right now. Those features don't depend on each other.

artembilan:
So, after merging more general solution with the `ProducerMessageHandlerCustomizer`, this one becomes redundant.
Of course, we need to fix Spring Integration for the mentioned `AmqpOutboundEndpoint.getRabbitTemplate()`, but that's already different story and not Spring Cloud Stream responcibility.
Whenever that fix is available, you'll get access to features you want to customize with already existing `ProducerMessageHandlerCustomizer`.

So, vote to close as it is redundant.

Thanks for understanding

artembilan:
See here: https://github.com/spring-projects/spring-integration/pull/3088

PR #1828, opened 2019-10-22
Add `ProducerMessageHandlerCustomizer` support

artembilan:
Related to https://github.com/spring-cloud/spring-cloud-stream-binder-aws-kinesis/issues/107

In some cases there is not enough configuration properties to setup
a producing `MessageHandler` in the binder.
Or we just can't do that via properties because some real object is
required for particular MH option.
For example in the AWS Kinesis Binder end-user would like to provide
a custom `AsyncHandler` which definitely cannot be expressed via properties

* Introduce a `ProducerMessageHandlerCustomizer` similar to existing
`MessageSourceCustomizer` and `ListenerContainerCustomizer`
* Add an `AbstractMessageChannelBinder.setProducerMessageHandlerCustomizer()`
to avoid breaking changes with an additional constructor arg
* Use a provided `handlerCustomizer` in the `customizeProducerMessageHandler()`
called from the `doBindProducer()` on the current `MH` before its
`afterPropertiesSet()`
* It is a target binder configuration responsibility to inject a
`ProducerMessageHandlerCustomizer` into the binder bean


artembilan:
See usage in the relevant PR for Kinesis Binder: https://github.com/spring-cloud/spring-cloud-stream/pull/1828

PR #1836, opened 2019-10-31
Ft multiple partition support

mmichailidis:


sobychacko:
@mmichailidis We are already in an RC phase of a major release. What prompted this PR? Is there an associated issue/bug that you are trying to resolve?

mmichailidis:
@sobychacko Great! is there anything I can help with?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -118,13 +119,35 @@ public BindingService(BindingServiceProperties bindingServiceProperties,
...
 						: doBindConsumer(input, inputName, binder, consumerProperties,
-								target);
-
-				bindings.add(binding);
+						target);
+
+					bindings.add(binding);
+				}
+				else {
+					for (Integer index : consumerProperties.getInstanceIndexList()) {
```
garyrussell: Needs to check the `multiplex` setting. When multiplex is true, there should be one binding that listens to all queues (will need a change in the Rabbit binder).

mmichailidis: Should the multiplex be checked in here before "streaming" the index-keys to the binder? Instead of changing the binder the multiplex will be taken care from the parent. Is that something that can apply in Kafka? From the code that I already went through in rabbit this will work as expected. (Which is the old behavior)

garyrussell: See how we deal with multiple destinations...

```java
		String bindingTarget = this.bindingServiceProperties
				.getBindingDestination(inputName);

		if (consumerProperties.isMultiplex()) {
			bindings.add(doBindConsumer(input, inputName, binder, consumerProperties,
					bindingTarget));
		}
		else {
			String[] bindingTargets = StringUtils
					.commaDelimitedListToStringArray(bindingTarget);
```

When multiplex is true we create one binding and the binder busts up the destination list and configures the listener container to listen to multiple queues.

We need to do the same here too (binding per instanceIndex Vs. one binding that listens to all of the instanceIndexes).

It's further complicated if you have multiple destinations **and** multiple instance indexes.

mmichailidis: I am currently testing a scenario in order to understand how the multiplex should work. 
```yml
      bindings:
        first-requests:
          destination: first-requests,second-requests
          group: execution
          durable-subscription: true
          consumer:
            max-attempts: 1
            partitioned: true
            multiplex: false
            instance-index-list:
              - 3
              - 4
              - 5
```

So setting my yml like this and spawning 6 jobs in an other project ( ThreadPoolExecutor#submit ) I fill the "first-requests" and "second-requests" with messages.

the Log of the listener is this 
```

2019-11-01 11:21:29.025  INFO 23866 --- [s.execution-4-1] DemoLogger                               : Thread : first-requests.execution-4-1 , RoutingKey : first-requests-4 , Content : Hello World! from 18
2019-11-01 11:21:29.025  INFO 23866 --- [s.execution-3-1] DemoLogger                               : Thread : second-requests.execution-3-1 , RoutingKey : second-requests-3 , Content : Hello World! from 18
2019-11-01 11:21:29.038  INFO 23866 --- [s.execution-5-1] DemoLogger                               : Thread : first-requests.execution-5-1 , RoutingKey : first-requests-5 , Content : Hello World! from 18
2019-11-01 11:21:29.038  INFO 23866 --- [s.execution-5-1] DemoLogger                               : Thread : second-requests.execution-5-1 , RoutingKey : second-requests-5 , Content : Hello World! from 18
2019-11-01 11:21:29.038  INFO 23866 --- [s.execution-3-1] DemoLogger                               : Thread : first-requests.execution-3-1 , RoutingKey : first-requests-3 , Content : Hello World! from 18
2019-11-01 11:21:29.038  INFO 23866 --- [s.execution-4-1] DemoLogger                               : Thread : second-requests.execution-4-1 , RoutingKey : second-requests-4 , Content : Hello World! from 18

2019-11-01 11:21:30.026  INFO 23866 --- [s.execution-4-1] DemoLogger                               : Thread : first-requests.execution-4-1 , RoutingKey : first-requests-4 , Content : Hello World! from 19
2019-11-01 11:21:30.026  INFO 23866 --- [s.execution-3-1] DemoLogger                               : Thread : second-requests.execution-3-1 , RoutingKey : second-requests-3 , Content : Hello World! from 19
2019-11-01 11:21:30.039  INFO 23866 --- [s.execution-3-1] DemoLogger                               : Thread : first-requests.execution-3-1 , RoutingKey : first-requests-3 , Content : Hello World! from 19
2019-11-01 11:21:30.039  INFO 23866 --- [s.execution-5-1] DemoLogger                               : Thread : first-requests.execution-5-1 , RoutingKey : first-requests-5 , Content : Hello World! from 19
2019-11-01 11:21:30.039  INFO 23866 --- [s.execution-4-1] DemoLogger                               : Thread : second-requests.execution-4-1 , RoutingKey : second-requests-4 , Content : Hello World! from 19
2019-11-01 11:21:30.039  INFO 23866 --- [s.execution-5-1] DemoLogger                               : Thread : second-requests.execution-5-1 , RoutingKey : second-requests-5 , Content : Hello World! from 19
```
The empty line was added manually by me to explain. The listener is as simple as 
```java
    @StreamListener("first-requests")
    @SendTo("first-response")
    public Message<String> doFirst(final String fromStuff, @Header(AmqpHeaders.RECEIVED_ROUTING_KEY) final String mc) throws InterruptedException {
        Thread.sleep(1000L);
        la.info("Thread : {} , RoutingKey : {} , Content : {}",
                Thread.currentThread().getName(),
                mc,
                fromStuff);
        return MessageBuilder.withPayload(fromStuff).build();
    }
```
So this gap that I added to the logs is the Thread#sleep as expected. 6 messages read in parallel and in correct order per thread. (6 channels in rabbit ) 

And the question is if I change multiplex to true what should happen? As all those queues are already routed at the same listener. Just changing the property from false to true will ignore the indexList and will spawn only 1 queue per destination with the naming "<comma-separated-destination>-<instance-index>" (first-requests.execution-0) and 1 channel in total in rabbit which listens on both queues. This is a side-effect because the call will be done 3 times but all 3 times the name would be the same.
In my understanding it should spawn again 6 queues with 1 channel listening to all of them. Is that correct? 

In order to achieve that this line should be provided with something like "first-requests.execution-3,first-requests.execution-4,first-requests.execution-5,second-requests.execution-3,second-requests.execution-4,second-requests.execution-5"
```java
consumerDestination = new RabbitConsumerDestination( StringUtils
  .arrayToCommaDelimitedString(provisionedDestinations),
  null);
```
which can either be created in parent in the multiplex==true branch or on both projects by adding the multiplex==true branch on parent and a array to array multiplication before creating the consumer. ( [first-req,second-req] * [3,4,5] )

garyrussell:
@mmichailidis We are planning a second release candidate on Thursday and would like to get this in. But, we can't include this unless we have a corresponding change in the rabbit binder to support the multiplex case. If you can't submit a PR for that today or tomorrow; we can take it over.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
```diff
@@ -118,13 +120,35 @@ public BindingService(BindingServiceProperties bindingServiceProperties,
 			String[] bindingTargets = StringUtils
 					.commaDelimitedListToStringArray(bindingTarget);
 			for (String target : bindingTargets) {
-				Binding<T> binding = input instanceof PollableSource
+				if (ObjectUtils.isEmpty(consumerProperties.getInstanceIndexList())) {
```
garyrussell: I believe we should only do this for partitioned consumers.

				if (!consumerProperties.isPartitioned() || ObjectUtils.isEmpty(consumerProperties.getInstanceIndexList())) {

garyrussell:
Merged as 8838d922 with my suggested change and some docs.

Thanks for the contribution!!

Issue #1841, opened 2019-11-11
Add ConsumerEndpointCustomizer


garyrussell:
Allow post processing of the consumer endpoint - e.g. setting `onPartitionsAssignedSeekCallback` on the Kafka adapter.

PR #1842, opened 2019-11-11
GH-1841: Add ConsumerEndpointCustomizer

garyrussell:


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -402,6 +422,7 @@ protected void afterUnbindProducer(ProducerDestination destination,
 			if (properties.isAutoStartup() && consumerEndpoint instanceof Lifecycle) {
 				((Lifecycle) consumerEndpoint).start();
 			}
+			this.consumerCustomizer.configure(consumerEndpoint, name, group);
```
artembilan: @garyrussell ,

Sorry, I don't notice this because everything else is OK.
So, I think this is already too late to customize it after `afterPropertiesSet()` and `start()`.

Please, fix it in subsequent commit.

garyrussell: Doh

PR #1857, opened 2019-11-29
Fixing functional binding

tkvangorder:
The functional binding requires the binding to be "log-in-0" rather than "input"


tkvangorder:
Ran into this while working through the example. The functional bindings are cool btw.

PR #1904, opened 2020-02-06
Add missing properties in DefaultPollerProperties

### LINKED ISSUE #1903: Add cron and initialDelay properties in DefaultPollerProperties

sobychacko:
cron and initialDelay properties in DefaultPollerProperties

Resolves #1903


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/DefaultPollerProperties.java
```diff
@@ -60,4 +80,19 @@ public void setMaxMessagesPerPoll(long maxMessagesPerPoll) {
 		this.maxMessagesPerPoll = maxMessagesPerPoll;
 	}
 
+	public String getCron() {
+		return cron;
```
artembilan: Don't we need to have a `this.` prefix whenever we access to class properties?
I'm not sure what is your Checkstyle config, but it is not allowed without prefix in my projects.

Thanks

PR #1918, opened 2020-02-27
GH-1912 Added support for non-reactive-source messaging bridge

### LINKED ISSUE #1912: Provide ability to have supplier that does not expose Reactor's internals
Currently we require the use of the `EmitterProcessor` to tunnel external events to the output destination.

olegz:
Currently the use of reactive EmitterProcessor is the only way to connect foreign sources such as HTTP with spring-cloud-stream
The newly added StreamBridgeUtils class provides a simpler mechanism to brideg non-stream application with stream application without requiring reactor API while also honoring the same type conversion and partitioning contract used by functions

Added docs

Resolves #1912

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-stream.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-stream.adoc
@@ -615,7 +615,7 @@ For example `--spring.cloud.stream.poller.fixed-delay=2000` sets the poller inte
 ===== Foreign event-driven sources
 
 There are cases where the actual source of data may be coming from the external (foreign) system that is not a binder. For example, the
-source of the data may be a classic REST endpoint. How do we bridge such source with the functional mechanism used by sopring-cloud-stream?
+source of the data may be a classic REST endpoint. How do we bridge such source with the functional mechanism used by spring-cloud-stream?
 
 Spring Cloud Stream provides two mechanisms, so let's look at them in more details
 
--- a/docs/src/main/asciidoc/spring-cloud-stream.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-stream.adoc
@@ -614,10 +614,55 @@ For example `--spring.cloud.stream.poller.fixed-delay=2000` sets the poller inte
 
 ===== Foreign event-driven sources
 
-There are cases where the actual source of data may be coming from outside system that is not a binder. For example, the
-source of the data may be a classic web endpoint. How do we bridge such source with the functional Supplier?
+There are cases where the actual source of data may be coming from the external (foreign) system that is not a binder. For example, the
+source of the data may be a classic REST endpoint. How do we bridge such source with the functional mechanism used by sopring-cloud-stream?
 
-Let's look at a simple example:
+Spring Cloud Stream provides two mechanisms, so let's look at them in more details
+
+Here, for both samples we'll use a standard MVC endpoint method called `delegateToSupplier` bound to the root web context, 
+delegating incoming requests to stream via two different mechanisms - 
+imperative (via StreamBridgeUtils) and reactive (via EmitterProcessor).
+
+====== Using StreamBridgeUtils
+
+[source, java]
+----
+@SpringBootApplication
+@Controller
+public class WebSourceApplication {
+
+	public static void main(String[] args) {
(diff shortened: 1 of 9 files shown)


docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -614,10 +614,55 @@ For example `--spring.cloud.stream.poller.fixed-delay=2000` sets the poller inte
 
 ===== Foreign event-driven sources
 
-There are cases where the actual source of data may be coming from outside system that is not a binder. For example, the
-source of the data may be a classic web endpoint. How do we bridge such source with the functional Supplier?
+There are cases where the actual source of data may be coming from the external (foreign) system that is not a binder. For example, the
+source of the data may be a classic REST endpoint. How do we bridge such source with the functional mechanism used by sopring-cloud-stream?
```
sobychacko: typo: sopring-cloud-stream

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamBridgeUtils.java
```diff
@@ -0,0 +1,140 @@
...
+ * This utility class allows user to do just that - <i>bridge non-spring-cloud-stream applications
+ * with spring-cloud-stream</i> by providing a mechanism (bridge) to send data to an output binding while
+ * maintaining the  same invocation contract (i.e., type conversion, partitioning etc) as if it was
+ * done through a declared function.
+ *
+ * @author Oleg Zhurakousky
+ * @since 3.0.3
+ *
+ */
+public final class StreamBridgeUtils implements SmartInitializingSingleton {
```
sobychacko: If we name an API class as Utils, it might give the impression that the class contains static utility methods. But we provide `StreamBridgeUtils` as part of the SCSt API and intend to be used for injection.  Can we name it to a first class `StreamBridge`?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/PartitionAwareFunctionWrapper.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.cloud.stream.binder.PartitionHandler;
+import org.springframework.cloud.stream.binder.ProducerProperties;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.expression.spel.support.StandardEvaluationContext;
+import org.springframework.integration.expression.ExpressionUtils;
+import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.Message;
+
+/**
+ * hHis class is effectively a wrapper which is aware of the stream related partition information
```
sobychacko: typo

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamBridgeUtils.java
```diff
@@ -0,0 +1,140 @@
...
+	 * Sends 'data' to an output binding specified by 'bindingName' argument while
+	 * using the content type specified by the 'outputContentType' argument to deal
+	 * with output type conversion (if necessary).
+	 * @param bindingName the name of the output binding
+	 * @param data the data to send
+	 * @param outputContentType content type to be used to deal with output type conversion
+	 * @return true if data was sent successfully, otherwise false or throws an exception.
+	 */
+	@SuppressWarnings("unchecked")
+	public boolean send(String bindingName, Object data, MimeType outputContentType) {
```
sobychacko: If the mime type is not the default one, then are we asking applications to specify the `MimeType`? Is that the reason why this is public?

olegz: Yes, basically this is feature of s-c-function. If there is no output content type then no conversion will happen at all, so if you look at FunctionConfiguration we always execute lookup version with output CT to ensure that the outgoing message is converted back to `Message<byte[]>`

But for this case it's even more important since this would allow user skip the creation fo the entire message just to provide custom CT header.

PR #1955, opened 2020-05-05
GH-1949 Fix logic to deal with multiple binder multi-initialization

### LINKED ISSUE #1949: With Kafka Streams Binder, creation of multiple application contexts causes problems with spring-cloud-vault-aws-config
We tried to run a Spring Boot app with spring-cloud-stream-binder-kafka-streams for using a compacted Kafka Topic as GlobalKTable and spring-cloud-vault-config-aws for obtaining temporary AWS credentials (we set up Vault to create temporary AWS IAM users)  

While it basically works, we observed that not only one temporary AWS IAM user was created by spring-cloud-vault, but five. We debugged into the app and noticed that our @SpringBootApplication class' main() method was actually called five times, first without parameters, then four times with one parameter each time, apparently corresponding to the different Kafka Streams topologies, so we suspected the Kafka Streams Binder to be responsible for this.

Now, as spring-cloud-vault-config-aws operates during bootstrap phase, creating five contexts and thus obviously having five bootstrap phases, five temporary AWS IAM users are created. However, on application shutdown, only the first user (created when Application.main() was called without parameters), is actually removed, with four orphaned IAM users remaining in the AWS account.

How come the app's main() method is called more than once? Is this really necessary? It does not seem intuitive. 

Currently we use spring-vault with @VaultPropertySource and EnvironmentVaultConfiguration which does not have this problem. However we'd prefer spring-cloud-vault as it seems more elegant to initialize AWS at bootstrap and providing AWS credentials as properties at that point, so they can be used in the app config without having to worry about property sources ordering etc.

olegz:
Fixed logic which was creating an instance of every available binder configuration. This is primarily an issue for
Kafka Streams binder which as a module brings 3 different binder configurations - KTable, KStream, GlobalKTable.

Resolves #1949

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
@@ -48,6 +48,7 @@ import org.springframework.core.convert.support.GenericConversionService;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.StandardEnvironment;
+import org.springframework.messaging.MessageChannel;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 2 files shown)


sobychacko:
LGTM. Merged upstream.

PR #1988, opened 2020-06-12
Provide no-op AcknowledgementCallback implementation if header is not present

dturanski:
I hit an NPE in testing a PollableMessageSource with the TestChannelBinder.  This is intended to prevent that.

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 the original author or authors.
+ * Copyright 2018-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,6 +22,8 @@ import java.util.function.BiConsumer;
 
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.aop.support.NameMatchMethodPointcutAdvisor;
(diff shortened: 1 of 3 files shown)


spring-cloud-stream/src/test/java/org/springframework/cloud/stream/binder/PollableConsumerTests.java
```diff
@@ -414,6 +415,43 @@ public void testRequeue() {
...
+		final AtomicInteger count = new AtomicInteger();
+		try {
+			assertThat(pollableSource.poll(received -> {
+				count.incrementAndGet();
+				throw new RequeueCurrentMessageException("test retry");
+			})).isTrue();
+		}
+		catch (Exception e) {
+			// no op
+		}
```
garyrussell, line 450: This test passes when the DPMS is reverted (ignores the NPE). Shouldn't we just remove the try/catch?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultPollableMessageSource.java
```diff
@@ -204,6 +209,11 @@ public boolean poll(MessageHandler handler, ParameterizedTypeReference<?> type)
 
 		AcknowledgmentCallback ackCallback = StaticMessageHeaderAccessor
 				.getAcknowledgmentCallback(message);
+
+		if (ackCallback == null) {
+			ackCallback = status -> log.warn("No AcknowledgementCallback defined. Status: " + status.name() + " " + message);
+		}
```
garyrussell, line 215: Wouldn't it be better to fix the `TestChannelBinder` as well?

dturanski: I'll take a look

dturanski: I'm not sure how to fix `TestChannelBinder` . It is the responsibility of the `MessageSource` to set the header. TCB doesn't enter into it AFAICT.  I could override `bindPollableConsumer` if you think it's necessary.

garyrussell: ?

```java
	private MessageSource<?> messageSourceDelegate = () -> new GenericMessage<>(
			"polled data",
			Collections.singletonMap(MessageHeaders.CONTENT_TYPE, "text/plain"));
```

Just add a no-op ack to the canned message that the test binder's message source emits.

dturanski: Ok. I wasn't thinking about that since my test overrides the message source.  The custom message source can be configured to cause an exception  - how I discovered this issue in the retry loop. I'm not sure that this one fails. I'll take another look.

PR #2036, opened 2020-10-20
Configure consumerEndpoint

0x006EA1E5:
Set ApplicationContext, BeanFactory and ApplicationEventPublisher if appropriate

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -202,15 +202,16 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 
 	/**
 	 * Apply {@link AutowireCapableBeanFactory#initializeBean(Object, String) initializeBean} to the
-	 * {@code target} with the empty string {@code ""} as the {@code beanName} parameter.
+	 * {@code target}.
 	 * @param target the {@code Object} to initialize
+	 * @param name the name to provide
 	 * @see AutowireCapableBeanFactory#initializeBean(Object, String)
 	 */
-	protected void applyInitializeBean(Object target) {
+	protected void applyInitializeBean(Object target, String name) {
 		ApplicationContext applicationContext = getApplicationContext();
 		if (applicationContext != null) {
 			applicationContext.getAutowireCapableBeanFactory()
-				.initializeBean(target, "");
+				.initializeBean(target, name);
 		}
 	}
 }
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -199,4 +199,11 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 		return rt;
 	}
 
+	protected void applyBeanPostProcessors(Object target, String name) {
+		ApplicationContext applicationContext = getApplicationContext();
+		if (applicationContext != null) {
+			applicationContext.getAutowireCapableBeanFactory()
+				.applyBeanPostProcessorsBeforeInitialization(target, name);
+		}
+	}
 }
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
@@ -24,6 +24,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.config.AutowireCapableBeanFactory;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.cloud.stream.annotation.StreamRetryTemplate;
 import org.springframework.context.ApplicationContext;
@@ -199,11 +200,17 @@ public abstract class AbstractBinder<T, C extends ConsumerProperties, P extends 
 		return rt;
 	}
 
-	protected void applyBeanPostProcessors(Object target, String name) {
+	/**
+	 * Apply {@link AutowireCapableBeanFactory#initializeBean(Object, String) initializeBean} to the
+	 * {@code target} with the empty string {@code ""} as the {@code beanName} parameter.
+	 * @param target the {@code Object} to initialize
+	 * @see AutowireCapableBeanFactory#initializeBean(Object, String)
+	 */
+	protected void applyInitializeBean(Object target) {
 		ApplicationContext applicationContext = getApplicationContext();
 		if (applicationContext != null) {
 			applicationContext.getAutowireCapableBeanFactory()
-				.applyBeanPostProcessorsBeforeInitialization(target, name);
+				.initializeBean(target, "");
 		}
 	}
 }
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -27,7 +27,6 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
 import org.apache.commons.logging.Log;
 
-import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.BeanFactoryAware;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.InitializingBean;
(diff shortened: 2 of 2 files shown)


0x006EA1E5:
Hopefully this is acceptable

0x006EA1E5:
Would this be better done using `IntegrationFlowBeanPostProcessor#postProcessBeforeInitialization`?

garyrussell:
At first glance, I think this belongs in the individual binders; for example, the Kafka binder has two components on the consumer side that needs the context propagation.

It would be odd to have split responsibilities, I think.

But I could be persuaded otherwise.

0x006EA1E5:
> At first glance, I think this belongs in the individual binders, not the abstract claaa; for example, the Kafka binder has two components on the consumer side that needs the context propagation (endpoint and listener container).

This was my first thought too, and in fact I did submit a PR https://github.com/spring-cloud/spring-cloud-stream-binder-aws-kinesis/pull/143

But it does seem nicer if configuring endpoints that implement such common interfaces as `ApplicationContextAware`, to just reuse some `BeanPostProcessor` to run through and configure everything properly in one place?

I find it a bit confusing understanding if these things are actually Beans or not - I guess they aren't when used here, but are when used straight up in integration?

garyrussell:
I have proposed several times that these objects should indeed be registered as beans so they are fully initialized by Spring instead of having all of this hand-wiring.

Before the commit I referenced above, I previously added a comment about it:

https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/a5e569a9047e7647a61df2b0d86fcdd0b3ea8d26/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L626-L636

0x006EA1E5:
Would it be acceptable to do this:

```
			ApplicationContext applicationContext = getApplicationContext();
			if (applicationContext != null) {
				applicationContext.getAutowireCapableBeanFactory()
					.applyBeanPostProcessorsBeforeInitialization(consumerEndpoint, name);
			}
```
?

garyrussell:
Yes, that would be cleaner, I guess; but it still doesn't solve the problem such as the Kafka binder needing to do the same for the listener container. Maybe add a helper method to the abstract class (probably the super class of this one - `AbstractBinder`) that can also be called from subclasses that need to initialize other "beans".

0x006EA1E5:
And if we did this on both `doBindProducer` and `doBindConsumer`?

It seems a bit wrong to do this, but I don't see a better option, except hand-wring as you say.

For example, `IntegrationObjectSupport` looks to implement `BeanNameAware`, but the value we are passing for the name is not actually a Bean name.  It is better to pass `null` as the `name` parameter to `applyBeanPostProcessorsBeforeInitialization`?

artembilan (changes requested):
Isn't it better to use this instead:
```
/**
	 * Initialize the given raw bean, applying factory callbacks
	 * such as {@code setBeanName} and {@code setBeanFactory},
	 * also applying all bean post processors (including ones which
	 * might wrap the given raw bean).
	 * <p>Note that no bean definition of the given name has to exist
	 * in the bean factory. The passed-in bean name will simply be used
	 * for callbacks but not checked against the registered bean definitions.
	 * @param existingBean the existing bean instance
	 * @param beanName the name of the bean, to be passed to it if necessary
	 * (only passed to {@link BeanPostProcessor BeanPostProcessors};
	 * can follow the {@link #ORIGINAL_INSTANCE_SUFFIX} convention in order to
	 * enforce the given instance to be returned, i.e. no proxies etc)
	 * @return the bean instance to use, either the original or a wrapped one
	 * @throws BeansException if the initialization failed
	 * @see #ORIGINAL_INSTANCE_SUFFIX
	 */
	Object initializeBean(Object existingBean, String beanName) throws BeansException;
```
?
See `AutowireCapableBeanFactory`

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
```diff
@@ -406,6 +407,7 @@ protected void afterUnbindProducer(ProducerDestination destination,
 			}
 			consumerEndpoint = createConsumerEndpoint(destination, group, properties);
 			consumerEndpoint.setOutputChannel(inputChannel);
+			applyBeanPostProcessors(consumerEndpoint, name);
 			this.consumerCustomizer.configure(consumerEndpoint, name, group);
 			if (consumerEndpoint instanceof InitializingBean) {
```
0x006EA1E5, line 412: If we use AutowireCapableBeanFactory#initializeBean can we also get rid of this call to `afterPropertiesSet()`, as it will get called at `AbstractAutowireCapableBeanFactory#invokeInitMethods`, followed by `applyBeanPostProcessorsAfterInitialization(...)`

artembilan: That's correct, but then we need to fix all the binders which call that `afterPropertiesSet()`

0x006EA1E5:
> Isn't it better to use this instead:
> 
> ```
> 	Object initializeBean(Object existingBean, String beanName) throws BeansException;
> ```

Should `this.consumerCustomizer.configure` happen before or after that?

artembilan:
> Should `this.consumerCustomizer.configure` happen before or after that?

It must happen before `initializeBean()` bean, indeed. We need to ensure that bean is not committed to its initialized state until we apply all the options to it.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
0x006EA1E5: > That's correct, but then we need to fix all the binders which call that afterPropertiesSet()


🤔 But `((InitializingBean) consumerEndpoint).afterPropertiesSet();` is called here already (and also in `doBindProducer`), and they are `final` methods.  

This change just removes it from these two methods and lets it get called via `applyInitializeBean` -> `initializeBean` -> `invokeInitMethods `instead.

Actually, `AbstractAutowireCapableBeanFactory#invokeInitMethods` does a load of security stuff I don't understand, but it looks like it is trying to achieve the same thing.

0x006EA1E5: Sorry, the above change is outdated now, but the same applies to the most recent commit

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
```diff
@@ -199,4 +200,17 @@ protected RetryTemplate buildRetryTemplate(ConsumerProperties properties) {
...
+	 * Apply {@link AutowireCapableBeanFactory#initializeBean(Object, String) initializeBean} to the
+	 * {@code target} with the empty string {@code ""} as the {@code beanName} parameter.
+	 * @param target the {@code Object} to initialize
+	 * @see AutowireCapableBeanFactory#initializeBean(Object, String)
+	 */
+	protected void applyInitializeBean(Object target) {
+		ApplicationContext applicationContext = getApplicationContext();
+		if (applicationContext != null) {
+			applicationContext.getAutowireCapableBeanFactory()
+				.initializeBean(target, "");
```
0x006EA1E5, line 213: Is the empty string for the `beanName` okay here?

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
artembilan: That's correct. It is better to defer to Spring `ApplicationContext` to initialize bean properly rather than trying to do that manually. 
And that's why I said that we probably need to revise all the binders since we do that stuff correct already here on the core level.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractBinder.java
artembilan: We probably need to think making some bean name over here since it is going to be propagated down to the `setBeanName()` on the target component.
I suppose we can reply on the provided destination name and group. And add some prefix or suffix to make it clear that it is consumer or producer.

0x006EA1E5: > We probably need to think making some bean name over here since it is going to be propagated down to the `setBeanName()` on the target component.
> I suppose we can reply on the provided destination name and group. And add some prefix or suffix to make it clear that it is consumer or producer.

I have taken a random guess, but I feel in no way qualified to decide on a naming scheme here.  Should we be using the class name, maybe?

I'm not sure what the consequences are of giving a name here which isn't in the BeanFactory... I can say from personal experience that it can be very confusing to see these objects appear to be Beans when debugging, but not actually be able to get them from the context.  I feel like the name should give some kind of clue where it came from, and that it _isn't_ a Bean, if possible 🤷

olegz:
Make sure you run Rabbit/Kafka binder builds. I just tried with Rabbit and there are test failures

olegz:
Are there any updates on this 0x006EA1E5? There seem to be no action since the last comment so I am poised to rejected it and close.

0x006EA1E5:
I guess reject it, then.  I can't give it the attention it would require right now :-/

PR #2046, opened 2020-11-06
Don't check bridge and synthetic methods in postProcessAfterInitialization()

stsypanov:
At debug time I've found out that in `postProcessAfterInitialization()` we check all the methods from target bean class, some of them are bridge and synthetic methods. `@StreamListener` is obviously cannot be put over those methods so we can filter them out.

Also I did some minor clean-ups to reduce allocations.


olegz:
I'll review, but just as an FYI. . .
`@StreamListener` and the entire annotation model has been deprecated and will be removed in future versions. We've been promoting functional programming model for the last two years. Also, in the current documentation we removed all references to annotation-based programming model.

Issue #1541, opened 2018-11-26
A custom `ConnectionNameStrategy` is ignored if there is a custom binder configuration


kinjelom:
A custom `ConnectionNameStrategy` is ignored if there is a custom binder configuration.
Configuration without a custom binder works ok.

### Configuration and code to reproduce this case

Spring Boot: 2.0.6.RELEASE
Spring Cloud: Finchley.RELEASE

```yaml
spring.cloud.stream:
  binders:
    demoinout:
      type: rabbit
      defaultCandidate: false
      inheritEnvironment: false
      environment.spring.rabbitmq:
        host: localhost
        port: 5672
        virtual-host: demo
        username: demo
        password: demo
  bindings:
    input:
      binder: demoinout
      destination: in-demo-a-dest
      group: in-demo-a-group
    output:
      binder: demoinout
      destination: out-demo-b-dest
      group: out-demo-b-group
```

```java
@Configuration
public class MessagingConfiguration {

	@Bean
	public ConnectionNameStrategy defineConnectionNameStrategy() {
		return connectionFactory -> "demo#"+UUID.randomUUID().toString().substring(0, 8);
	}

}
```

```java
@Service
@EnableBinding(Processor.class)
public class MessagingProcessorService {

	@Transformer(inputChannel = Processor.INPUT, outputChannel = Processor.OUTPUT)
	public Message<byte[]> demoProcessor(@Payload final byte[] body, @Headers Map<String, Object> headers) {
		return MessageBuilder.withPayload(body).copyHeaders(headers).build();
	}
}

```

-----
In this case Spring creates two `CachingConnectionFactory` (publisher: false, true) and set them connection strategy to one defined as a bean `ConnectionNameStrategy`. 
After that `DefaultBinderFactory` creates another two `CachingConnectionFactory` but **doesn't set defined `ConnectionNameStrategy`**:
```
at   org.springframework.boot.builder.SpringApplicationBuilder.run(SpringApplicationBuilder.java:137)
--
at   org.springframework.cloud.stream.binder.DefaultBinderFactory.getBinderInstance(DefaultBinderFactory.java:206)
at   org.springframework.cloud.stream.binder.DefaultBinderFactory.getBinder(DefaultBinderFactory.java:139)
at   org.springframework.cloud.stream.binding.BindingService.getBinder(BindingService.java:308)
at   org.springframework.cloud.stream.binding.BindingService.bindProducer(BindingService.java:208)
[...]

PR #2051, opened 2020-11-18
Muti binder customization

sobychacko:
Enable applications with multi binders to provide customizations for
the binders. This is accomplished by providing a new API called
BinderCustomizer. When an implementation is found for this interface,
the customizations are applied against the binders before providing access to them.

Tests and docs.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/1541

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderCustomizer.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderCustomizer.java
@@ -34,6 +34,7 @@ public interface BinderCustomizer {
 	 * In the case of a single binder, the use of this customizer is redundant.
 	 *
 	 * @param binder to be customized
+	 * @param binderName binder name to distinguish between two different instances of the same binder type
 	 */
-	void customize(Binder<?, ConsumerProperties, ProducerProperties> binder);
+	void customize(Binder<?, ConsumerProperties, ProducerProperties> binder, String binderName);
 }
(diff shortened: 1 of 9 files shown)


spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderCustomizer.java
```diff
@@ -0,0 +1,39 @@
...
+	 * enables the application to properly apply customizations in all the binders.
+	 * By providing an implementation of this interface, the binders, although reside
+	 * in different application contexts, will receive the customization.
+	 * Spring Cloud Stream ensures that the customizations take place before the binders are
+	 * accessed. The user must check for the binder type and then apply the necessary customizations.
+	 * In the case of a single binder, the use of this customizer is redundant.
+	 *
+	 * @param binder to be customized
+	 */
+	void customize(Binder<?, ConsumerProperties, ProducerProperties> binder);
```
garyrussell, line 38: We need to include the binder name - they might want to customize 2 different Kafka binders differently.

garyrussell (changes requested):
Looks good - pretty simple 😄  but I do think we need to know the name.

spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderCustomizer.java
```diff
@@ -34,6 +34,7 @@
 	 * In the case of a single binder, the use of this customizer is redundant.
 	 *
 	 * @param binder to be customized
+	 * @param binderName binder name to distinguish between two different instances of the same binder type
```
garyrussell, line 37: A bit pedantic, perhaps, but it should say "...distinguish between multiple instances of..." (there might be more than 2).

garyrussell (changes requested):

garyrussell (approved):

PR #2084, opened 2021-01-15
Spell correct in preface doc

imechemi:


imechemi:
The same test on CI is failing on master branch. Can someone help?

MrMoon:
can you elaborate on the cause of the test fail, because I can't see how fixing a spelling mistake made a test fail

olegz:
@imechemi thanks for your fix. Don't worry about the failing test, i'll look. It obviously has nothing to do with your fix.

PR #2089, opened 2021-01-20
StreamBridge: add cache

GreenRover:
Add the case does:
- remove unnecessary log.info messages for each send message
- solve: https://github.com/spring-cloud/spring-cloud-function/issues/628
- increase performance: For 100_000 executions, the runtime is reduced from 3726ms before to 1926ms after
- reduce the GC overhead what again improves the performance

### DIFF
--- a/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/BindingMimeTypeKey.java
+++ b/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/BindingMimeTypeKey.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2018-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


olegz:
@GreenRover Can you please clarify as to what value does it add considering we already have caching mechanism in place - https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamBridge.java#L108

GreenRover:
@olegz currently exists: Channel cache
new added: function cache

Mening the "functionToInvoke" will be cached.
To avoid `this.functionCatalog.lookup(STREAM_BRIDGE_FUNC_NAME, outputContentType.toString());` for each message.
Because this is always the same but is costly and creates a log.info each time.

olegz:
Thank you, I understand now, but if you look at implementation of SimpleFunctionRegistry, there is caching mechanism there as well, so the whole initialization/composition only happens on the first lookup and then it is cached, so each subsequent call to lookup is pulling function from cache.
```
	<T> T doLookup(Class<?> type, String functionDefinition, String[] expectedOutputMimeTypes) {
		FunctionInvocationWrapper function = this.wrappedFunctionDefinitions.get(functionDefinition);

		if (function == null) {
			function = this.compose(type, functionDefinition);
		}

		if (function != null   && !ObjectUtils.isEmpty(expectedOutputMimeTypes)) {
			function.expectedOutputContentType = expectedOutputMimeTypes;
		}
		else if (logger.isDebugEnabled()) {
			logger.debug("Function '" + functionDefinition + "' is not found in cache");
		}

		function = this.wrapInAroundAviceIfNecessary(function);

		return (T) function;
	}
```
Does your approach provides an additional value?

GreenRover:
@olegz you are right. It seams that you/someone fixed the issue in meant time.
So i reject this issues.

olegz:
Yes, there was major refactoring done on SimpleFunctionRegistry several month ago. That said, I truly appreciate you looking at things as I am sure there are things to improve!!!

PR #2150, opened 2021-04-13
Fixed spring-cloud-stream.adoc documentation rendering

IlyasYOY:
Hello there! 

I was reading docs and saw that the rendering went wrong. (Image attached below)
There was a problem in the code block, so I fixed it. One error caused everything else below to be broken too.

- Fixed a few ADOC lines 
- Added punctuation mark 
 May be removed, it was IDEA's autocorrection, I'm not native speaker myself, so I may be wrong here

image

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-stream.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-stream.adoc
@@ -1694,22 +1694,23 @@ The user must check for the binder type and then apply the necessary customizati
 
 Here is an example of providing a `BinderCustomizer` bean.
 
-```
+[source, java]
+----
 @Bean
 public BinderCustomizer binderCustomizer() {
     return (binder, binderName) -> {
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/spring-cloud-stream.adoc
```diff
@@ -1694,22 +1694,23 @@ The user must check for the binder type and then apply the necessary customizati
...
-            ((KafkaMessageChannelBinder) binder).setRebalanceListener(...);
+            ((KafkaMessageChannelBinder) binder).setRebalanceListener(/*...*/);
         }
         else if (binder instanceof KStreamBinder) {
-            ...
+            // ...
         }
         else if (binder instanceof RabbitMessageChannelBinder) {
-            ...
+            // ...
```
olegz, line 1709: It is implied that this is a pseudo code (incomplete code). There is no reason to escape it.

Issue #2266, opened 2022-01-14
GlobalChannelInterceptors are added multiple time to MessageChannels when dynamic-destination-cache-size is smaller than the number of bindings


snussbaumer:
If you have more bindings than what is declared in `spring.cloud.stream.dynamic-destination-cache-size` + a GlobalChannelInterceptor, then the ChannelInterceptor is added to the MessageChannel **each time** streamBridge.send is called. This results in a List of interceptors that grows and grows and grows after each message that is sent : it's a memory leak IMHO.

This caused a huge performance issue in our application.

I have a [pull request](https://github.com/spring-cloud/spring-cloud-stream/pull/2267) with a failing tests. If you increase dynamic-destination-cache-size the test pases.

This is reproductible on main, but also with Spring Boot 2.5.8 + Spring Cloud 2020.0.0.4

snussbaumer:
I've just created a second pull requests for a failing tests that is a bit simpler : https://github.com/spring-cloud/spring-cloud-stream/pull/2269

I've closed the first PR, but if you prefer that one for the test, just let me know.

Issue #2249, opened 2021-11-23
When no partition-key-expression is defined for a producer, a partition-key-expression from another producer may be picked


snussbaumer:
This leads unsurprisingly to "Partition key cannot be null" exceptions, and anyways it's quite a random behaviour.

A workaround is to force a partition-key-expression on something (payload for instance), but this should not be necessary.

I have a simple repro here : https://github.com/snussbaumer/repro-partition-key-expression-dj/blob/master/src/test/java/com/example/reproinputdestinationwithexception/ReproPartitionKeyExpressionDJ.java

olegz:
@snussbaumer Thank you for the test that reproduces the issue. Next time if it's easier you can submit a failing test as a PR as it will cut down some time on fixing it ;)

PR #2269, opened 2022-01-17
GH-2266 Add test to ensure GlobalChannelInterceptor is not added multiple times to the MessageChannels

snussbaumer:
This pull request contains a test that reproduces issue 2266

In the parameters, if you specify "--spring.cloud.stream.dynamic-destination-cache-size=2" instead of 1 the test passes. 

The test should pass whatever the value of "dynamic-destination-cache-size"


olegz:
So, if I understand correctly you just added a failing test to showcase the issue, right?

snussbaumer:
yes, as suggested by you here : https://github.com/spring-cloud/spring-cloud-stream/issues/2249#issuecomment-988604524

PR #2275, opened 2022-02-03
Fix NPE  in StreamBridge.resolveDestination()

ncheema:
resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2276


ncheema:
@olegz thank you for fixing the checkstyle violation. I will ensure to  run full build next time 🙏

PR #2277, opened 2022-02-08
Language improvement spring-cloud-stream docs

plopp:
Improved language in the Bindings chapter.


plopp:
Minor language fix so CLA should not be needed.

Issue #2293, opened 2022-03-10
Supporting reactor-kafka


sobychacko:
@ghahramani commented on [Tue May 04 2021](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1075)

Hi,

After some discussion in the Gitter spring cloud stream channel, I realized to have a sequential pause/resume situation in reactive Kafka we need to use a back pressure mechanism that is not supported by web flux and Stream right now as we do not have any binder for `reactor-kafka` yet. I would like to raise this feature request for supporting `reactor-kafka` in spring cloud stream so we can easily use its functionalities.

---

@garyrussell commented on [Tue May 04 2021](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1075#issuecomment-832186087)

@sabbyanandan @olegz This may not be trivial, but I think if we can pull it off, it should become the only way to support reactive applications in SCSt. 

Reactor apps on top of non-reactive binders (especially Kafka) is a fool's errand in my opinion.

There is also a `reactor-rabbitmq` project.

Perhaps we should spike this?

---

@olegz commented on [Wed May 05 2021](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1075#issuecomment-832678915)

@garyrussell I completely agree and believe made it clear internally and externally. The same I believe would need to be made for Rabbit https://github.com/reactor/reactor-rabbitmq. This would also allow us to partition the binder API on reactive and imperative where reactive can still support the imperative model but not the other way around as it doesn't make any sense (as today)

---

@sabbyanandan commented on [Tue May 18 2021](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1075#issuecomment-843247547)

@ghahramani: When you get a chance, can you describe your use-case, and why the backpressure and being fully reactive would help the said scenarios? What are you trying to achieve and what are you OK to compromise live with? 

I am curious to learn more about the real-world scenarios to validate and back my own hypothesis. 

---

@ghahramani commented on [Tue May 18 2021](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/1075#issuecomment-843282359)

> @ghahramani: When you get a chance, can you describe your use-case, and why the backpressure and being fully reactive would help the said scenarios? What are you trying to achieve and what are you OK to compromise live with?
> 
> I am curious to learn more about the real-world scenarios to validate and back my own hypothesis.

[...]

PR #2294, opened 2022-03-10
GH-2293: Initial Commit Reactor-Kafka Binder

garyrussell:
I squashed the commits from the old repo into a single commit and added a second commit to add lifecycle to the adapters.

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
@@ -1,0 +1,273 @@
+/*
+ * Copyright 2021-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.binder.reactorkafka;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
+import org.apache.kafka.common.serialization.ByteArraySerializer;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.Sinks;
+import reactor.kafka.receiver.KafkaReceiver;
+import reactor.kafka.receiver.ReceiverOptions;
+import reactor.kafka.sender.KafkaSender;
+import reactor.kafka.sender.SenderOptions;
+import reactor.kafka.sender.SenderRecord;
+import reactor.kafka.sender.SenderResult;
+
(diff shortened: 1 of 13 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
```diff
@@ -0,0 +1,327 @@
...
+			protected void doStart() {
+				Flux<Message<Object>> flux = receiver
+						.receive()
+						.map(record -> (Message<Object>) converter.toMessage(record, null, null, null));
+				subscribeToPublisher(flux);
+			}
+
+			@Override
+			protected void doStop() {
+				// TODO how do we dispose of the Disposable created by subscribeToPublisher?
```
garyrussell, line 146: @artembilan question for you - it doesn't keep a reference to the `Disposable`.

artembilan: It does it by itself:
```Flux<? extends Message<?>> messageFlux =
				Flux.from(publisher)
						.map(this::trackMessageIfAny)
						.doOnComplete(this::stop)
						.doOnCancel(this::stop)
						.takeWhile((message) -> isActive());
```

That `TakeWhileSubscriber` does this:
```
if (!b) {
	s.cancel();

	onComplete();

	return;
}
```
So, nothing to dispose.

garyrussell: Thanks.

sobychacko:
PR merged upstream.

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
garyrussell: @artembilan There is still a problem here; there is nothing to signal the flux completion.

See https://github.com/spring-cloud/spring-cloud-stream/blob/0f213fedf67d9269e7f984a806d00c5156ac3780/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderTests.java#L109-L110

If you put a breakpoint on the unbind and step over, you see that the `"reactive-kafka-foo-1"` thread is still running.

We normally shutdown the consumer by calling `dispose()` on the `Disposable` returned by the `subscribe()`...

https://github.com/spring-cloud/spring-cloud-stream/blob/dcaa621b3c72bfcce2423622a0e272513cb463c5/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java#L140

I think we need a hook to be able to explicitly perform the `dispose()` when the binding is stopped or unbound; perhaps that should be done in `super.doStop()` ?

artembilan: That `dispose()` is called in the `FluxMessageChannel.destroy()`.

I'm not sure why `KafkaConsumer` is not stopped when a `cancel()` is called for its `Flux`.
Perhaps that's something has to be improved in `reactor-kafka`?

I'll try to debug your test later.

Thanks

garyrussell: II might be wrong but I don’t think unbinding a binding destroys the channel. Stopping the binding certainly doesn’t, but that should cancel the disposable.

artembilan: I have figured out the problem, Gary.

So, that `TakeWhileSubscriber` does an action we need only when there is a data streaming from a `Publisher`.
In your test we have only one record published to Kafka. And then we just go to the `stop()` immediately.

We really need an additional hook to unsubscribe from the source `Publsher` when endpoint is stopped externally.
It would be great if we would have something what would signal the subscriber to cancel...

Not sure which `Flux` operator to use here...

```
.doOnSubscribe(subs -> this.subscription = subs);
```
?

garyrussell: Thanks; that works - https://github.com/spring-cloud/spring-cloud-stream/pull/2295

PR #2295, opened 2022-03-12
GH-2293: Cancel Subscription on Stop

garyrussell:


binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
```diff
@@ -131,15 +132,26 @@ protected MessageProducer createConsumerEndpoint(ConsumerDestination destination
...
 			private final KafkaReceiver<Object, Object> receiver = KafkaReceiver.create(opts);
 
+			private volatile Subscription subscription;
+
 			@SuppressWarnings("unchecked")
 			@Override
 			protected void doStart() {
 				Flux<Message<Object>> flux = receiver
 						.receive()
+						.doOnSubscribe(subs -> this.subscription = subs)
```
artembilan, line 145: Thanks, Gary, for a quick turn around, but my point was to have such a fix in SI directly.
Apparently that `takeWhile()` is wrong in most cases.

Will PR it shortly over there.
Perhaps we would need then extra commit over here to make the code more cleaner...

garyrussell: OK; makes sense; I just wanted a quick fix here.

Issue #2299, opened 2022-03-14
In concurrent scenarios the "scst_partition" header may be randomly omitted


snussbaumer:
Reproduction is quite simple : 
- Have a few threads that send messages with streamBridge with a partitionKey
- Some of the messages will miss the scst_partition header 

The number of messages without the header is variable but around a value that seems dependent on the number of threads. For example with 10 threads  around 10% of the messages are missing the header with my repro.

Respect of partition keys is **critical** in our case. We use it abundantly throughout our applications, mostly to avoid having multiple threads changing the same data at the same time.

I can reproduce this on main, but in our real world case we encounter the problem with Spring Boot 2.5.9 and Spring Cloud 2020.0.5, with a kafka cluster.

I'll submit a pull request with a failing test shortly.

I found no workaround for this problem, but really looking for one in the meantime (other that adding "synchronized" each time I use streamBridge)

snussbaumer:
I see a commit related to this ticket ... is it considered done ? Is it possible to have it backported to 3.1.x as it's quite critical ?

olegz:
@snussbaumer the 3.1 is out of support and EOL. There is still a bit of a window for commercial clients, but that also mean any request has to come thru proper channels. 
Why aren't you using 3.2.x?

snussbaumer:
Hum we're on the next to last version of spring cloud : 2020.0.5 (illford) and haven't have time to migrate to 2021.0.1 yet.

illford is already EOL ?

snussbaumer:
high @olegz : any news on this issue ? it's still open yet there's a commit that has been pushed related to it ... and if you could tell me about the support/EOL that would be great ? don't you support Spring Cloud N-1 ?

snussbaumer:
@olegz : are you sure it's been released ? I do not see a related commit on 3.2.x

snussbaumer:
hum ok, sorry the code is there, it was commited with other things on 3.2.x : "GH-SCF_856 Fix propper Cloud Event header prefix in StreamBridge"

snussbaumer:
but I still don't get the "3.1 is out of support and EOL" you said in an earlier post 
According to https://github.com/spring-cloud/spring-cloud-release/wiki/Supported-Versions :
> Spring Cloud follows the [Pivotal OSS support policy](https://pivotal.io/support/oss) and supports major versions for 3 years from the release date.
[...]

PR #2300, opened 2022-03-14
GH-2299 Add test to ensure scst_partition header is always set in concurrent scenarios

snussbaumer:
Repro for https://github.com/spring-cloud/spring-cloud-stream/issues/2299


olegz:
Is this test intended to be a failing test to showcase the issue? I mean it is failing for me all the time hence I am wondering

olegz:
@snussbaumer yes yes yes, and thank you!!! I was just confirming that this is a failing test to demonstrate the issue. For same reason it wasn't clear to me. But yes this is a preferred way, so thank your once again

Issue #2307, opened 2022-01-10
Multi-binding declaration after multilple rabbitmq connection attempts.


sebastienbinda:
Hello,

When rabbitmq server is not available at startup of a SCS service, and the configuration of a consumer allows multiple bindings to the same queue, only one binding is saved in the "autoDeclareContext".

I produced this issue by starting my rabbitmq server after my service with a multiple binding between a DLX and a DLQ.

Here is the source code of rabbit binder where i think the problem is :
https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/blob/97c6f79fff15d985434e24c6b48c85caa962a4e6/spring-cloud-stream-binder-rabbit-core/src/main/java/org/springframework/cloud/stream/binder/rabbit/provisioning/RabbitExchangeQueueProvisioner.java#L635

Note : Based on a stackoverflow ticket : https://stackoverflow.com/questions/70651340/spring-cloud-stream-rabbit-queue-creation-when-rabbit-server-unavailable

Thanks.

PR #2309, opened 2022-03-22
GH-2307: Fix Problems with Rabbit Declarables

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2307

- when creating `Declarables` beans for redeclaration after a connetion failure
  2 DLX/DLQ bindings used the same bean name so the second one was never recovered/
- removing consumer `Declarable` beans from the `autoDeclareContext` was incomplete
- removing producer `Declarable` beans was not performed at all

- fix bean names; fix cleanup; add tests for all config scenarios
- check cleanup in all `RabbitBinderTests`

**cherry-pick to 3.2.x when the mono repo is established**


olegz:
I've already cherry-picked it 4e66d2b513b43e52e8aebf4898ce469e83329a43

Issue #2345, opened 2022-04-04
Unable to use ListenerContainerWithDlqAndRetryCustomizer


mattdupuy:
Hi,

I have been trying to use the new ListenerContainerCustomizer "ListenerContainerWithDlqAndRetryCustomizer" for Kafka Binder.
I want to use it to perform a complex retry process.
After following the documentation https://docs.spring.io/spring-cloud-stream-binder-kafka/docs/3.2.2/reference/html/spring-cloud-stream-binder-kafka.html#retry-and-dlq-processing, when I start my application, my customizer is never call, so I have performed some debugging to trying to figure out what is happening.

I have found a strange behavior in the class KafkaMessageChannelBinder, in the method "createConsumerEndpoint", line 748 and line 797.
```
if (!(customizer instanceof ListenerContainerWithDlqAndRetryCustomizer)
			|| ((ListenerContainerWithDlqAndRetryCustomizer) customizer)
					.retryAndDlqInBinding(destination.getName(), group)) {
		kafkaMessageDrivenChannelAdapter
				.setRetryTemplate(buildRetryTemplate(extendedConsumerProperties));
		kafkaMessageDrivenChannelAdapter
				.setRecoveryCallback(errorInfrastructure.getRecoverer());
}

if (customizer instanceof ListenerContainerWithDlqAndRetryCustomizer) {
	BiFunction<ConsumerRecord<?, ?>, Exception, TopicPartition> destinationResolver = createDestResolver(
			extendedConsumerProperties.getExtension());
	BackOff createBackOff = extendedConsumerProperties.getMaxAttempts() > 1
			? createBackOff(extendedConsumerProperties)
			: null;
	((ListenerContainerWithDlqAndRetryCustomizer) customizer)
			.configure(messageListenerContainer, destination.getName(), consumerGroup, destinationResolver,
					createBackOff);
}
```
The code is checking if the customizer is an instance of "ListenerContainerWithDlqAndRetryCustomizer", however, even when the customizer if an instance of "ListenerContainerWithDlqAndRetryCustomizer", the result if false. After investigating, I have found out that the customizer is wrap inside a Anonymous implementaton of ListenerContainerCustomizer, declare inside the DefaultBinderFactory class, line 425.

![image](https://user-images.githubusercontent.com/49486505/161563890-003025b3-7b02-4a41-9715-f7c0ea276b96.png)

```
[...]

PR #2348, opened 2022-04-07
Add setter for Container Customizer on AbstractMessageChannelBinder

mattdupuy:
Based on suggestion in #2345, add a setter for Container Customizer and update documentation accordingly


sobychacko:
@mattdupuy Did you verify that these changes worked for your scenario? I wonder whether we need a test to verify.

mattdupuy:
@sobychacko It works with the use of reflection in the current version of the framework, as mentioned in issue #2345, so adding the setter will allow me to remove the reflection and have the expected behavior.
If you have an idea of a test to add, let me know.

sobychacko:
Ok, sounds good. Since it is only a setter change, I think we can merge it without the test.

sobychacko:
Merged upstream. Thanks for the PR!

Issue #2328, opened 2022-03-30
Wrong Default for Compression Level


sobychacko:
@garyrussell commented on [Wed Feb 09 2022](https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/353)

Documented as:

>**spring.cloud.stream.rabbit.binder.compressionLevel**

>The compression level for compressed bindings. See java.util.zip.Deflater.

>Default: 1 (BEST_LEVEL).

It's actually 0 (`NO_COMPRESSION`), which overrides spring-amqp's default, which is `BEST_SPEED` (1).

See https://stackoverflow.com/questions/71054400/with-spring-cloud-stream-rabbitmq-is-it-possible-to-zip-outbound-and-unzip-inbo/71054642#71054642

Change the property to `Integer` and only override spring-amqp's default if not null.

PR #2354, opened 2022-04-13
Change compression level to integer and only set if not null. Resolve…

ben-enfuse-io:
…s https://github.com/spring-cloud/spring-cloud-stream/issues/2328

In this commit we changed the compression level to an Integer from an int and only set in the configuration if it is not null, as per #2328.

Thanks for taking time to look at it!


sobychacko:
@ben-enfuse-io Thanks for the PR. Since you touched on two Java classes, please update the copyright years and add your name as an author. Add as an additional commit, and then we can squash them together on merge.

PR #2357, opened 2022-04-20
Make the lag metric return correct values even if there are no committed offsets

lbilger:
Currently, if the group has no committed offset for some or all partitions of a topic and the topic has limited message retention, the calculated lag will be too high. It will be the equal to the log-end-offset even if there are fewer or no messages left in the partition. If there is no committed offset, the log-start-offset should be used instead.


sobychacko:
@lbilger Thank you for the PR. It seems like a valid fix. Could you add your name as an author on the classes you changed and update the copyright years? Then we will further review and merge.

lbilger:
@sobychacko thanks for having a look so quickly! I added the requested changes. When this is merged, will it automatically be back ported to 3.2.x or should I cherry-pick and create another pull request?

sobychacko:
We will backport to `3.2.x` once it is merged upstream.

sobychacko:
Merged upstream to `main` and backported to `3.2.x`.

PR #2358, opened 2022-04-21
Use testcontainers for Rabbit binder integration tests

### LINKED ISSUE #2341: Consider using TestContainer for running Rabbit binder integration tests

onobc:
There is now no need to manually start Rabbit for CI. We can remove that next.

Fixes #2341


binders/rabbit-binder/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/LocalizedQueueConnectionFactoryIntegrationTests.java
```diff
@@ -34,20 +35,22 @@
...
 /**
  * @author Gary Russell
+ * @author Chris Bono
  */
 public class LocalizedQueueConnectionFactoryIntegrationTests {
 
+	private static final RabbitMQContainer RABBITMQ = RabbitTestContainer.sharedInstance();
+
 	@RegisterExtension
-	public static RabbitTestSupport rabbitAvailableRule = new RabbitTestSupport(true);
```
onobc, line 41: I moved all `RabbitTestSupport` instance variables to private non-final scope. There is no need for public. There is no big win for `static final` and it actually makes one think that it has to be final to prevent Rabbit from starting every test method - but that is not the case and that is controlled in `RabbitTestContainer`. 

Based on the above I made this consistent in all of these tests.

binders/rabbit-binder/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitBinderCleanerTests.java
```diff
@@ -184,34 +186,8 @@ private void waitForConsumerStateNot(String queueName, long state) throws Interr
 		assertThat(cleanedExchanges).hasSize(6);
 	}
 
-	public static class AmqpQueue {
```
onobc, line 187: This was dead code.

binders/rabbit-binder/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitBinderCleanerTests.java
```diff
@@ -184,34 +186,8 @@ private void waitForConsumerStateNot(String queueName, long state) throws Interr
...
-		protected boolean isDurable() {
-			return durable;
-		}
-
-		protected void setDurable(boolean durable) {
-			this.durable = durable;
-		}
-
+	private static Map<String, List<String>> doClean(RabbitBindingCleaner cleaner, String entity, boolean isJob) {
+		return cleaner.clean(RABBITMQ.getHttpUrl() + "/api", "guest", "guest", "/", BINDER_PREFIX, entity, isJob);
```
onobc, line 190: Had to use `clean` method that accepted "all the things" as we needed to configure the port - the cleaner default is hardcoded to `15672`.

binders/rabbit-binder/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitTestContainer.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Provides a static {@link RabbitMQContainer} that can be shared across test classes.
+ *
+ * @author Chris Bono
+ */
+public class RabbitTestContainer {
+
+	private static final RabbitMQContainer RABBITMQ;
+	static {
+		String image = "rabbitmq:management";
```
onobc, line 32: I swtiched off the custom Pivotal image that was being used in [AbstractIntegrationTest](https://github.com/spring-cloud/spring-cloud-stream/pull/2358/files#diff-6631b85f70208404bd3e3682dea09de101fc54eb219264abeec90998645fc118L34). Unless there is a glaring reason to use it I would suggest we move to a standard image configured w/ streams enabled (as below). Also note this allows us to `RabbitMQContainer` for port/url access rather than `GenericContainer`.

binders/rabbit-binder/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitTestContainer.java
```diff
@@ -0,0 +1,52 @@
...
+		String cache = System.getenv().get("IMAGE_CACHE");
+		if (cache != null) {
```
onobc, line 34: I skipped the [spring.rabbit.use.local.server](https://github.com/spring-cloud/spring-cloud-stream/pull/2358/files#diff-6631b85f70208404bd3e3682dea09de101fc54eb219264abeec90998645fc118L34) condition here as well. Is that actually used for anything? If so, we can incorporate here as well.

bom/spring-cloud-starter-parent/pom.xml
```diff
@@ -7,6 +7,7 @@
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
 		<version>3.0.0-M2</version>
+		<relativePath/>
```
onobc, line 10: **[UNRELATED]** was failing in IDEA due to invalid parent path.

binders/rabbit-binder/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/RabbitTestContainer.java
olegz: There is a great chance that it is some type of leftover as you've already discovered with dead code.

PR #2362, opened 2022-04-25
Remove Kafka binder dependence on KafkaAutoConfiguration

### LINKED ISSUE #2349: Remove hard dependency on KafkaAutoConfiguration
There are situations where we need to stick to a certain version of the Kafka client library. With spring-boot we can opt-out of it applying its `KafkaAutoConfiguration` and set `${kafka.version}` and `${spring-kafka.version}` as needed while still benefitting from other spring-boot features in the rest of the application.

However, in `KafkaBinderConfiguration` (and also the streams binder variants) there is a hard `@Import` for `KafkaAutoConfiguration` which by-passes the support for exclusion and ordering of AutoConfigurations. Would it be possible to use `@AutoConfigureAfter` or, preferably, create a dedicated `KafkaConfiguration` class in `spring-cloud-stream-binder-kafka-core` which would bail out if the required beans are already provided by AutoConfiguration? That way the dependency on spring-boot becomes optional.

https://github.com/spring-cloud/spring-cloud-stream/blob/6c8423365adebac2bdee4c3ccb0ca5c01a8d00e9/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java#L84

onobc:
See [here](https://github.com/spring-cloud/spring-cloud-stream/issues/2349#issuecomment-1108893702) for details.

Fixes #2349

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderConfiguration.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderConfiguration.java
@@ -18,7 +18,6 @@ package org.springframework.cloud.stream.binder.reactorkafka;
 
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
-import org.springframework.boot.autoconfigure.kafka.KafkaAutoConfiguration;
 import org.springframework.boot.autoconfigure.kafka.KafkaProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.stream.binder.Binder;
@@ -28,12 +27,16 @@ import org.springframework.cloud.stream.binder.kafka.provisioning.AdminClientCon
 import org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Import;
 
+/**
+ * Binder configuration for ReactorKafka.
+ *
+ * @author Gary Russell
+ * @author Chris Bono
+ */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnMissingBean(Binder.class)
-@Import({ KafkaAutoConfiguration.class })
-@EnableConfigurationProperties({ KafkaExtendedBindingProperties.class })
+@EnableConfigurationProperties({ KafkaProperties.class, KafkaExtendedBindingProperties.class })
 public class ReactorKafkaBinderConfiguration {
 
 	@Bean
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderIntegrationTests.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderIntegrationTests.java
@@ -24,14 +24,16 @@ import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.producer.ProducerRecord;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import reactor.core.publisher.Flux;
 import reactor.kafka.receiver.ReceiverRecord;
 
-import org.springframework.boot.SpringApplication;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.WebApplicationType;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.integration.support.MessageBuilder;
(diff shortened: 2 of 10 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderConfiguration.java
```diff
@@ -28,12 +27,16 @@
...
+ * Binder configuration for ReactorKafka.
+ *
+ * @author Gary Russell
+ * @author Chris Bono
+ */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnMissingBean(Binder.class)
-@Import({ KafkaAutoConfiguration.class })
-@EnableConfigurationProperties({ KafkaExtendedBindingProperties.class })
+@EnableConfigurationProperties({ KafkaProperties.class, KafkaExtendedBindingProperties.class })
```
onobc, line 39: Same approach that you will see throughout the binder configs "Replace the KAC w/ KP".

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderIntegrationTests.java
```diff
@@ -42,76 +44,86 @@
...
 
 /**
+ * Integration tests for {@link ReactorKafkaBinder}.
+ *
  * @author Soby Chacko
  * @author Gary Russell
+ * @author Chris Bono
  */
+@ExtendWith(SpringExtension.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
```
onobc, line 63: Went from static single instance to instance per test (there are only 2 tests). The reason is that we now have 2 tests (w/ and w/o KAC excluded) and each of them writes records. Additionally `KafkaTestUtils` does not have an easy way to "get last record" and the 2nd tests always fails w/ "more than 1 record found". To get the last record is possible but the tradeoff on complexity of doing that vs. keeping the test simple and isolated wins. If there were > 2 tests I would consider going the "static kafka" route.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderConfigurationTest.java
```diff
@@ -16,42 +16,25 @@
...
-import org.springframework.util.ReflectionUtils;
+import org.springframework.test.util.ReflectionTestUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * @author Ilayaperumal Gopinathan
+ * @author Chris Bono
  */
-@ExtendWith(SpringExtension.class)
```
onobc, line 37: We need to do a later sweep and remove `@ExtendWith(SpringExtension.class)` from all `@SpringBootTest` as SBT includes that annotation.

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderIntegrationTests.java
```diff
@@ -42,76 +44,86 @@
...
 @EmbeddedKafka(topics = { "uppercased-words", "lowercased-words" })
-public class ReactorKafkaBinderIntegrationTests {
-
-	private static final EmbeddedKafkaBroker embeddedKafka = EmbeddedKafkaCondition.getBroker();
+class ReactorKafkaBinderIntegrationTests {
```
onobc, line 65: Throughout the PR I have moved to package-protected access modifier for tests and test methods as that is one of the JUnit5 features. On the surface it seems trivial but it promotes: 
1. the "Less is more" mindset
2. Principle of Least Privileged as a default mode of operation

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
```diff
@@ -78,11 +77,12 @@
  * @author Oleg Zhurakousky
  * @author Artem Bilan
  * @author Aldo Sinanaj
+ * @author Chris Bono
  */
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnMissingBean(Binder.class)
-@Import({ KafkaAutoConfiguration.class, KafkaBinderHealthIndicatorConfiguration.class })
```
artembilan, line 84: I think this was done especially for multi-binder scenario, when we even have several Kafka binder instances in the same stream application.
So, every binder got its own auto-configured beans based exactly on their specific properties.
Therefore how it is going to work from now on?

Thanks

artembilan (changes requested):
And if we go this way, please, consider to make similar changes to the `KinesisBinderConfiguration`.
See its `@Import({ ContextCredentialsAutoConfiguration.class, ContextRegionProviderAutoConfiguration.class })` for the same reason as we have `@Import({ KafkaAutoConfiguration.class })`  - multi-binder environment.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
onobc: @artembilan , yep this is for the multi-binder case and described [here](https://github.com/spring-cloud/spring-cloud-stream/issues/2349#issuecomment-1108893702). However, the only beans that we actually use are KafkaProperties, none of the other beans that are brought in by KAC are actually used (AFAICT).

onobc:
> KinesisBinderConfiguration
I must have missed this one @artembilan as I did look in the Kinesis binder module. I will amend this PR to include it as well.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java
onobc: I believe @garyrussell comment [here](https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/issues/187#issuecomment-440777907) also points to only using the properties for the binder.

artembilan: Sure! I see now. After closer look indeed Kafka binder doesn't use an auto-configured `KafkaTemplate` and friends like `ProducerFactory` or `ConsumerFactory`.
So, yeah... It might rally work for Kafka binder, but that is not the case for Kinesis. It does rely on the auto-configured `AWSCredentialsProvider` and `RegionProvider`. Sorry for misleading.
We may revise that in favor of "manual" auto-configuration against respective properties like it is done for Kafka binder.

The Rabbit binder does rely on the `ConnectionFactory rabbitConnectionFactory`, though. See `RabbitMessageChannelBinderConfiguration`.
That might be revised respectively.

I guess different issue to stop polluting here 😄 

Thank you!

artembilan: OK. I think I have missed a couple rounds. I see now there is this one `RabbitConfiguration`. And looks like the `RabbitBinderConfiguration` does not import any auto-configurations any more.
Sorry for a noise.

onobc: > Sorry for a noise.

Nws. It is not noise to me and, well, it used to import _RabbitAutoConfiguration_ but I added [this to SB](https://github.com/spring-projects/spring-boot/pull/26982) a while back to break the auto-config dependence. 

> but that is not the case for Kinesis. It does rely on the auto-configured AWSCredentialsProvider and RegionProvider. Sorry for misleading.
We may revise that in favor of "manual" auto-configuration against respective properties like it is done for Kafka binder.

Ahh good to know. Do you mind creating a ticket to break dependency on KAC for the Kinesis binder and I will follow on w/ that work once the dust settles on this proposal?

artembilan (approved):

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderConfigurationTest.java
onobc: To be handled in #2371

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderIntegrationTests.java
onobc: To be handled in #2371

onobc:
> > KinesisBinderConfiguration
> 
> I must have missed this one @artembilan as I did look in the Kinesis binder module. I will amend this PR to include it as well.

I created #2372 to extend this to the Kinesis binder.

PR #2366, opened 2022-04-27
Fix Poller Config for ambiguity

artembilan:
Related to https://stackoverflow.com/questions/72023445/spring-cloud-stream-mutually-exclusive-property-issue

The `PollerConfigEnvironmentPostProcessor` has a flaw not checking user-provided properties
and does not check for mutually exclusive before setting default

* Fix `PollerConfigEnvironmentPostProcessor` to check for other mutual poller
properties before setting default for `fixed-delay`
* Same for `max-messages-per-poll`
* Verify that fix is working in the `PollableSourceTests` where poller
options make sense in the Spring Cloud Stream scenario

**Cherry-pick to `3.2.x`**

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/PollerConfigEnvironmentPostProcessor.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/PollerConfigEnvironmentPostProcessor.java
@@ -68,9 +68,7 @@ public class PollerConfigEnvironmentPostProcessor implements EnvironmentPostProc
 
 			streamPollerProperties.putIfAbsent(INTEGRATION_PROPERTY_PREFIX + "fixed-delay", "1s");
 		}
-		if (!environment.containsProperty(INTEGRATION_PROPERTY_PREFIX + "max-messages-per-poll") &&
-			streamPollerProperties.containsKey(INTEGRATION_PROPERTY_PREFIX + "max-messages-per-poll")) {
-
+		if (!environment.containsProperty(INTEGRATION_PROPERTY_PREFIX + "max-messages-per-poll")) {
 			streamPollerProperties.putIfAbsent(INTEGRATION_PROPERTY_PREFIX + "max-messages-per-poll", "1");
 		}
 
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/PollerConfigEnvironmentPostProcessor.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/PollerConfigEnvironmentPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021-2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/PollerConfigEnvironmentPostProcessor.java
```diff
@@ -56,15 +56,28 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
...
+			!environment.containsProperty(INTEGRATION_PROPERTY_PREFIX + "cron") &&
+			!environment.containsProperty(INTEGRATION_PROPERTY_PREFIX + "fixed-rate") &&
+			!environment.containsProperty(INTEGRATION_PROPERTY_PREFIX + "fixed-delay")) {
 
-		environment.getPropertySources()
-			.addLast(new MapPropertySource("spring.integration.poller", streamPollerProperties));
+			streamPollerProperties.putIfAbsent(INTEGRATION_PROPERTY_PREFIX + "fixed-delay", "1s");
+		}
+		if (!environment.containsProperty(INTEGRATION_PROPERTY_PREFIX + "max-messages-per-poll") &&
+			streamPollerProperties.containsKey(INTEGRATION_PROPERTY_PREFIX + "max-messages-per-poll")) {
```
onobc, line 72: Should this be `!`, otherwise I don't see how line 74 "putIfAbsent" would ever invoke.

core/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/PollableSourceTests.java
```diff
@@ -42,18 +46,23 @@ void test() {
 				.getCompleteConfiguration(PollableAppSampleConfiguration.class))
 						.web(WebApplicationType.NONE).run(
 								"--spring.jmx.enabled=false",
-								"--spring.cloud.stream.pollable-source=blah")) {
+								"--spring.cloud.stream.pollable-source=blah",
+								"--spring.integration.poller.cron=*/2 * * * * *")) {
```
onobc, line 50: Wdyt about adding tests that cover the other exceptional env properties such as `fixed-rate`, `fixed-delay` and also the case for `max-messages-per-poll`?

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/PollerConfigEnvironmentPostProcessor.java
artembilan: This one is an error.
We don't to do any check here since the next `putIfAbsent()` covers us.

That's actually a good point to think about more tests since I'm making so many mistakes 😄

core/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/PollableSourceTests.java
```diff
@@ -57,9 +62,66 @@ void test() {
...
 
+	@Test
+	void testPollerDefaultFixedDelay() {
+		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(TestChannelBinderConfiguration
+			.getCompleteConfiguration(PollableAppSampleConfiguration.class))
+			.web(WebApplicationType.NONE).run()) {
+
+			PollerMetadata pollerMetadata = context.getBean(PollerMetadata.class);
+			assertThat(pollerMetadata.getTrigger()).isInstanceOf(PeriodicTrigger.class);
+			assertThat(TestUtils.getPropertyValue(pollerMetadata.getTrigger(), "fixedRate")).isEqualTo(false);
```
onobc, line 77: Not sure if you knew about this or not, but AssertJ has really great support for "digging at fields" via `assertThat(<thing>).hasFieldOrPropertyWithValue(`.

onobc (approved):

core/spring-cloud-stream/src/test/java/org/springframework/cloud/stream/function/PollableSourceTests.java
artembilan: Yep! Know that one.
It didn't work for me in some scenario and then I just keep forgetting to give it a second chance 😄

sobychacko:
This PR is now merged upstream to the main branch via https://github.com/spring-cloud/spring-cloud-stream/commit/3b32bea133bea32fcc724f12956e8a3d02e820d7. Changes are also backported to `3.2.x`.

PR #2377, opened 2022-05-05
Add docker compose for Kafka cluster

onobc:
Moving the Kafka docker compose over to SCS mono-repo from SCS-samples repo

cc: @sobychacko

### DIFF
--- a/tools/kafka/docker-compose/README.adoc
+++ b/tools/kafka/docker-compose/README.adoc
@@ -1,0 +1,64 @@
+== Docker Compose scripts for Kafka
+
+=== Kafka Cluster
+Runs a multi-node Kafka cluster w/ 3 brokers available at `locahost:9091,localhost:9092,localhost:9093`.
+
+To start the brokers run the following command:
+[source,shell]
+----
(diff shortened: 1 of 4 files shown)


onobc:
@olegz 

If you have not yet used these docker-composed Kafka "scripts" you should check them out, they will actually change your life. LOL - they did mine anyways. 

The main question we have is should we put these at the root such as is this proposal, or should we put them under samples? 

## Tradeoffs
### Under '/'
➕ These tools are good beyond the samples, I use them daily, we simply had them stuffed in the samples repo as we did not have a clear home for them. The new samples will be a user of these tools as well. 
➕ These tools are kafka focused today but could easily be extended to include rabbit, kinesis, etc.. later (another reason to go under root)
➖ Putting them at the root does make them more visible and feels like we would have to support them more "visibly"; aka they feel like they are part of the main line now, not just some tool used by the samples. 

### Under '/samples'
Opposite of above

sobychacko:
@onobc Chatted with Oleg offline, and he is fine with merging this. Merging now.

PR #2393, opened 2022-05-11
Migrating Kafka native serialization sample

sobychacko:


onobc:
@sobychacko should we skip bringing this sample in now that native encoding is the default? Rather bring a valid sample showing SCS encoding?

onobc (approved):
@sobychacko do you mind now approving/reviewing my updates to the PR?

PR #2395, opened 2022-05-11
Cleanup SerdeResolverUtils

### LINKED ISSUE #2370: SerdeResolver cleanup
This is a follow on ticket to make some minor improvements around the Serdes resolver that were highlighted during the initial PR #2367

- [x] `SerdeResolverUtils`
    - [x] invert the if (short circuit) 
    - [x] slightly change how the "is default serdes" is implemented
    - [x] take [this comment](https://github.com/spring-cloud/spring-cloud-stream/pull/2367#discussion_r862886550) into consideration as well
    - [x] Consider using [ObjectProvider](https://github.com/spring-cloud/spring-cloud-stream/pull/2367#discussion_r862892496) to lookup serdes

- [x] `SerdeResolverUtilsTest`
    - [x] add a more complicated Serde type that has 2 level generics
    - [x] add test for edge cases

- [x] ~Reduce duplication in [SerdesProvidedAsBeansTests](
https://github.com/spring-cloud/spring-cloud-stream/pull/2367#discussion_r862580900)~ (chose not to do this as its only repeated 2x. If it were 3x I would factor out the commonality)

onobc:
* Refactor SerdeResolverUtils logic
* Add more tests in SerdeResolverUtilsTests
* Fixed some random test warnings

Fixes #2370

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KeyValueSerdeResolver.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KeyValueSerdeResolver.java
@@ -207,7 +207,7 @@ public class KeyValueSerdeResolver implements ApplicationContextAware {
 	 * @return {@link Serde} for the state store key.
 	 */
 	public Serde<?> getStateStoreKeySerde(String keySerdeString) {
-		return getKeySerde(keySerdeString, (Map<String, ?>) null);
+		return getKeySerde(keySerdeString, null);
 	}
 
 	/**
@@ -217,7 +217,7 @@ public class KeyValueSerdeResolver implements ApplicationContextAware {
 	 */
 	public Serde<?> getStateStoreValueSerde(String valueSerdeString) {
 		try {
-			return getValueSerde(valueSerdeString, (Map<String, ?>) null);
+			return getValueSerde(valueSerdeString, null);
 		}
 		catch (ClassNotFoundException ex) {
 			throw new IllegalStateException("Serde class not found: ", ex);
@@ -242,20 +242,18 @@ public class KeyValueSerdeResolver implements ApplicationContextAware {
 	}
 
 	private Serde<?> getKeySerde(String keySerdeString, ResolvableType resolvableType, Map<String, ?> extendedConfiguration) {
-		Serde<?> keySerde = null;
+		Serde<?> keySerde;
 		try {
 			if (StringUtils.hasText(keySerdeString)) {
 				keySerde = Utils.newInstance(keySerdeString, Serde.class);
 			}
 			else {
+				keySerde = Serdes.ByteArray();
 				if (resolvableType != null &&
 						(isResolvalbeKafkaStreamsType(resolvableType) || isResolvableKStreamArrayType(resolvableType))) {
-					ResolvableType generic = resolvableType.isArray() ? resolvableType.getComponentType().getGeneric(0) : resolvableType.getGeneric(0);
+					ResolvableType targetType = resolvableType.isArray() ? resolvableType.getComponentType().getGeneric(0) : resolvableType.getGeneric(0);
 					Serde<?> fallbackSerde = getFallbackSerde("default.key.serde");
-					keySerde = SerdeResolverUtils.resolveForType(this.context, generic, fallbackSerde);
-				}
-				if (keySerde == null) {
-					keySerde = Serdes.ByteArray();
+					keySerde = SerdeResolverUtils.resolveForType(this.context, targetType, fallbackSerde);
 				}
 			}
 			keySerde.configure(combineStreamConfigProperties(extendedConfiguration), true);
@@ -298,24 +296,19 @@ public class KeyValueSerdeResolver implements ApplicationContextAware {
 				: Serdes.ByteArray();
 	}
 
-	@SuppressWarnings("unchecked")
 	private Serde<?> getValueSerde(String valueSerdeString, ResolvableType resolvableType, Map<String, ?> extendedConfiguration)
 			throws ClassNotFoundException {
-		Serde<?> valueSerde = null;
+		Serde<?> valueSerde;
 		if (StringUtils.hasText(valueSerdeString)) {
 			valueSerde = Utils.newInstance(valueSerdeString, Serde.class);
 		}
 		else {
-
+			valueSerde = Serdes.ByteArray();
 			if (resolvableType != null && ((isResolvalbeKafkaStreamsType(resolvableType)) ||
 					(isResolvableKStreamArrayType(resolvableType)))) {
+				ResolvableType targetType = resolvableType.isArray() ? resolvableType.getComponentType().getGeneric(1) : resolvableType.getGeneric(1);
 				Serde<?> fallbackSerde = getFallbackSerde("default.value.serde");
-				ResolvableType generic = resolvableType.isArray() ? resolvableType.getComponentType().getGeneric(1) : resolvableType.getGeneric(1);
-				valueSerde = SerdeResolverUtils.resolveForType(this.context, generic, fallbackSerde);
-			}
-			if (valueSerde == null) {
-
-				valueSerde = Serdes.ByteArray();
+				valueSerde = SerdeResolverUtils.resolveForType(this.context, targetType, fallbackSerde);
 			}
 		}
 		valueSerde.configure(combineStreamConfigProperties(extendedConfiguration), false);
@@ -329,7 +322,7 @@ public class KeyValueSerdeResolver implements ApplicationContextAware {
 
 	private Map<String, ?> combineStreamConfigProperties(Map<String, ?> extendedConfiguration) {
 		if (extendedConfiguration != null && !extendedConfiguration.isEmpty()) {
-			Map<String, Object> streamConfiguration = new HashMap(this.streamConfigGlobalProperties);
+			Map<String, Object> streamConfiguration = new HashMap<>(this.streamConfigGlobalProperties);
 			streamConfiguration.putAll(extendedConfiguration);
 			return streamConfiguration;
 		}
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
@@ -16,15 +16,13 @@
 
 package org.springframework.cloud.stream.binder.kafka.streams;
 
-import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.commons.logging.Log;
@@ -32,11 +30,12 @@ import org.apache.commons.logging.LogFactory;
 import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.common.serialization.Serdes;
 
-import org.springframework.beans.factory.annotation.AnnotatedBeanDefinition;
+import org.springframework.beans.factory.BeanFactoryUtils;
+import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.core.ResolvableType;
 import org.springframework.kafka.support.serializer.JsonSerde;
-import org.springframework.util.ClassUtils;
+import org.springframework.lang.Nullable;
 
 /**
  * Utility class that contains various methods to help resolve {@link Serde Serdes}.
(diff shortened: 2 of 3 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KeyValueSerdeResolver.java
```diff
@@ -207,7 +207,7 @@ public Serde<?> getOutboundValueSerde(ProducerProperties producerProperties,
 	 * @return {@link Serde} for the state store key.
 	 */
 	public Serde<?> getStateStoreKeySerde(String keySerdeString) {
-		return getKeySerde(keySerdeString, (Map<String, ?>) null);
```
onobc, line 210: **[UNRELATED-WARNING-FIX]**

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KeyValueSerdeResolver.java
```diff
@@ -217,7 +217,7 @@ public Serde<?> getStateStoreKeySerde(String keySerdeString) {
 	 */
 	public Serde<?> getStateStoreValueSerde(String valueSerdeString) {
 		try {
-			return getValueSerde(valueSerdeString, (Map<String, ?>) null);
```
onobc, line 220: **[UNRELATED-WARNING-FIX]**

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KeyValueSerdeResolver.java
```diff
@@ -242,20 +242,18 @@ private Serde<?> getKeySerde(String keySerdeString, Map<String, ?> extendedConfi
 	}
 
 	private Serde<?> getKeySerde(String keySerdeString, ResolvableType resolvableType, Map<String, ?> extendedConfiguration) {
-		Serde<?> keySerde = null;
```
onobc, line 245: **[UNRELATED-WARNING-FIX]**

onobc, line 245: **[UNRELATED-WARNING-FIX]**

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -49,118 +48,142 @@ abstract class SerdeResolverUtils {
...
 		}
 		List<SerdeWithSpecificityScore> matchingSerdes = new ArrayList<>();
-
-		context.getBeansOfType(Serde.class).forEach((beanName, serdeBean) -> {
-			final Class<?> beanConfigClass = ClassUtils.resolveClassName(((AnnotatedBeanDefinition)
-					context.getBeanFactory().getBeanDefinition(beanName))
-					.getMetadata().getClassName(),
-				ClassUtils.getDefaultClassLoader());
+		ResolvableType serdeType = ResolvableType.forClassWithGenerics(Serde.class, targetType);
+		String[] serdeBeanNames = BeanFactoryUtils.beanNamesForTypeIncludingAncestors(context.getBeanFactory(),
```
onobc, line 173: @artembilan I was not able to use `context.getBeanProvider` as it did not handle sorting (it does respect `@Order` but that will not work here as the order is dependent on the target type, not the bean definition) - it otherwise returns them in order they are defined in the config class. However, I was able to use similar `BeanFactoryUtils.beanNamesForTypeIncludingAncestors` to get the proper bean names and then grab the bean and bean definition so I can gets its actual type and sort by it. 

This let me simplify things 2 ways:
1. no reflection on the config class to get the return type of the bean method (I use the bean definition for this now)
2. no need to be the final arbiter w/ manual call of `serdeBeanGeneric.isAssignableFrom` (the `beanNamesForTypeIncludingAncestors` handles all of that internally now).

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtilsTests.java
```diff
@@ -17,76 +17,355 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.mockito.Mockito.mock;
 
 /**
  * Unit tests for {@link SerdeResolverUtils}.
  *
  * @author Chris Bono
  */
+@SuppressWarnings({ "rawtypes", "NewClassNamingConvention", "unchecked" })
```
onobc, line 50: These are expected as we do all sort of different types of generic combos in this test.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtilsTests.java
```diff
@@ -17,76 +17,355 @@
...
 class SerdeResolverUtilsTests {
```
onobc, line 51: I ❤️ using nested in cases like this as it helps organize the tests logically. Take a look at the output in IDEA when I run this test suite:
Screen Shot 2022-05-11 at 5 05 18 PM

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -49,118 +48,142 @@ abstract class SerdeResolverUtils {
...
-						matchingSerdes.add(new SerdeWithSpecificityScore(calculateScore(targetType, serdeBeanGeneric), serdeBean));
-					}
-				});
-			}
-			catch (Exception e) {
-				if (LOG.isTraceEnabled()) {
-					LOG.trace("Failed to introspect Serde bean method '" + serdeBean + "'", e);
+				BeanDefinition beanDefinition = context.getBeanFactory().getMergedBeanDefinition(beanName);
+				ResolvableType serdeBeanGeneric = beanDefinition.getResolvableType().getGeneric(0);
+				Serde<?> serdeBean = context.getBean(beanName, Serde.class);
```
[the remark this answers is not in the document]
onobc: Yes, this thought crossed my mind as well as sounded something like this....

"We are not even using this ordered list of matching beans. Now we have this complicated set of tests to look at / maintain. We should only add things when we actually need them".   

1. However, we do have to go through them all in order to find the "closest/best match".
2. I wanted to test the sort/ordering from above so I ended up w/ the test permutations.

Since at this point the code was already there I chose to leave it in rather than pare it back. To limit any public use of the N-based API we could instead create a ` public Serde<?> findClosestMatchingSerde` API which delegates to the existing one and move the existing one to package-protected. 

I think this would be a good thing. Wdyt?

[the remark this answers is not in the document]
artembilan: Not sure how the suggested refactoring is relevant to my concern about beans loading overhead.

According the rest of your logic, you do propagate `Serde` beans in the `SerdeWithSpecificityScore`, but you do not use them in the score calculation and then you don't use the in compare function.
Not sure how dropping bean reference at this point in favor if just its name may break all your tests...

artembilan (changes requested):

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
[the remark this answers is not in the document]
onobc: Ahh ok. I misunderstood the concern.

Yes, since we do not actually use the bean for anything other than returning to the caller, we could instead just return the beanName(s) and let the caller load the bean(s), OR if we pared the API down to return only a single "top match" we could just load that single serde bean.

One thing to remember is that this operation is during the initial configuration so not loading the handful of serde beans probably configured will not be a big performance gain.

> Not sure how dropping bean reference at this point in favor if just its name may break all your tests..

Let's not worry about that. I will update the tests based on the direction we choose. (we tell them what to do ;P )

sobychacko (approved):

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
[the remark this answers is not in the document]
artembilan: Well, if it's used only on configuration phase, then OK: let's your logic a is.
Plus there is no too much to worry about since the API in the end is not `public`.

sobychacko:
@onobc The PR looks good. I am ok with merging it. We can make any further changes through subsequent changes if need be.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -49,118 +48,126 @@ abstract class SerdeResolverUtils {
...
+		}
+		Class<?> serdeClass = serde.getClass();
+		return Serdes.Integer().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Long().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Short().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Double().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Float().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.ByteArray().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.String().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.UUID().getClass().isAssignableFrom(serdeClass);
```
artembilan, line 140: This will fall under another SonarCube rule - long boolean expression, but I don't find it very reasonable 😄 

On the other thought I wonder why plain `serde instance Serdes.Integer()` doesn't work instead? Or even `equals()`...
Just because those `Serdes` classes are all `final`, therefore cannot be extension or so:
```
static public final class IntegerSerde extends WrapperSerde<Integer> {
```

artembilan (changes requested):

onobc:
> @onobc The PR looks good. I am ok with merging it. We can make any further changes through subsequent changes if need be.

Sounds good @sobychacko . The changes were straight forward to update though. Let me push the commit and we can decide which API to keep.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
onobc: Yeh, I had the same thoughts about isAssignable here. 

Either way though, I do prefer the concise "long boolean expression" way of doing this as it keeps it easier to read.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -49,123 +48,130 @@ abstract class SerdeResolverUtils {
...
+		if (String.class.isAssignableFrom(genericRawClazz)) {
+			return Serdes.String();
 		}
-		else {
-			// If the type is Object, then skip assigning the JsonSerde and let the fallback mechanism takes precedence.
-			if (!genericRawClazz.isAssignableFrom((Object.class))) {
-				serde = new JsonSerde(genericRawClazz);
-			}
+		if (UUID.class.isAssignableFrom(genericRawClazz)) {
+			return Serdes.UUID();
```
artembilan, line 123: How about `Serdes.serdeFrom(Class<T> type)` instead of all of these ifs ?

Then you catch their `IllegalArgumentException` and return `null` (or whatever logic you'd like to do)

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -49,123 +48,130 @@ abstract class SerdeResolverUtils {
...
+		}
+		Class<?> serdeClass = serde.getClass();
+		return Serdes.Integer().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Long().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Short().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Double().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.Float().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.ByteArray().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.String().getClass().isAssignableFrom(serdeClass) ||
+			Serdes.UUID().getClass().isAssignableFrom(serdeClass); //NOSONAR
```
artembilan, line 140: I think this `//NOSONAR` has to be on a first line of the "failed" code block:
https://github.com/spring-projects/spring-integration/blob/main/spring-integration-http/src/main/java/org/springframework/integration/http/multipart/UploadedMultipartFile.java#L67 - direct array argument storage.
thttps://github.com/spring-projects/spring-integration/blob/main/spring-integration-http/src/main/java/org/springframework/integration/http/multipart/UploadedMultipartFile.java#L67 - too long method.

artembilan (changes requested):

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
onobc: Got it - I thought it was at the end of the statement. I will adjust and then if ti fails during Jenkins I will adjust that.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
onobc: Yeh I was originally looking in that class to see if there was some built-in methods that would work for us. They do handle one more type than we do IIRC but I will take a scan at using it.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
artembilan: Yeah... I'm not sure there is SonarCube involved in Spring Cloud Stream CI plan.
@sobychacko ?

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
artembilan: Sure! Let it be then as it is! Probably not related to this PR anyway.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
artembilan: So, @onobc , are you going to move that `//NOSONAR`?
Or let it just live as is?

sobychacko: Yes, @artembilan, there is one on our Jenkins CI plans.

onobc: Moving, give me a second. Plus, now I am tweaking around w/ those methods. I will commit shortly.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtilsTests.java
```diff
@@ -16,77 +16,349 @@
...
+			@MethodSource("kafkaStreamsBuiltInTypes")
+			void returnsStandardSerdeForKafkaStreamsBuiltInType(Class<?> builtInType, Serde<?> expectedBuiltInSerde) {
+				contextRunner.run((context) ->
+					assertThat(SerdeResolverUtils.resolveForType(context, ResolvableType.forClass(builtInType), fallback))
+						.isInstanceOf(expectedBuiltInSerde.getClass()));
+			}
+
+			static Stream<Arguments> kafkaStreamsBuiltInTypes() {
+				return Stream.of(
+					arguments(String.class, Serdes.String()),
```
onobc, line 95: I did re-order these and make them match the new standard default classes in SRU

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -48,124 +47,120 @@ abstract class SerdeResolverUtils {
 
 	private static final Log LOG = LogFactory.getLog(SerdeResolverUtils.class);
 
+	/** Classnames of the standard built-in Serdes supported in {@link Serdes#serdeFrom(Class)} */
```
onobc, line 50: I went ahead and squashed/rebased . The changes since last review round are here and the 2 related standard default methods.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -48,124 +47,120 @@ abstract class SerdeResolverUtils {
...
-		}
-		else if (Short.class.isAssignableFrom(genericRawClazz)) {
-			serde = Serdes.Short();
-		}
-		else if (Double.class.isAssignableFrom(genericRawClazz)) {
-			serde = Serdes.Double();
-		}
-		else if (Float.class.isAssignableFrom(genericRawClazz)) {
-			serde = Serdes.Float();
+		List<String> matchingSerdes = SerdeResolverUtils.beanNamesForMatchingSerdes(context, targetType);
```
artembilan, line 87: Nit: don't need to point to itself if you call your own methods.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
```diff
@@ -48,124 +47,120 @@ abstract class SerdeResolverUtils {
...
+		Serdes.String().getClass().getName(),
+		Serdes.Short().getClass().getName(),
+		Serdes.Integer().getClass().getName(),
+		Serdes.Long().getClass().getName(),
+		Serdes.Float().getClass().getName(),
+		Serdes.Double().getClass().getName(),
+		Serdes.ByteArray().getClass().getName(),
+		Serdes.ByteBuffer().getClass().getName(),
+		Serdes.Bytes().getClass().getName(),
+		Serdes.UUID().getClass().getName());
```
artembilan, line 61: Why names? Why just classes not enough?
Looks like you have the target class as well to check its existence in this set.
Otherwise you do extra step to take the name of that class.

artembilan (changes requested):

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
onobc: Yep, it is intentional and the reason is fear of holding on to a class reference vs. a reference to a string. Probably paranoia though.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/SerdeResolverUtils.java
onobc: It is definitely a style thing as some Sonar complain w/o reference to it (even when static). However, I had it w/o before and I have no strong preference. Updating...

artembilan: That sound suspicious: a couple lines below you use a `getStandardDefaultSerde()` without `SerdeResolverUtils.` prefix...

artembilan (approved):

onobc:
Screen Shot 2022-05-13 at 10 52 00 AM

Finally! Haha just kidding. Great review - thanks for all the feedback @artembilan

sobychacko:
@onobc checkstyle failed on me during the local build. Do you mind fixing it?

onobc:
@sobychacko yep, my bad. I forgot to do final local build after last changes. Adjusting now.

onobc:
Resolved via https://github.com/spring-cloud/spring-cloud-stream/commit/cc05160cc05825914a827c5511503770bde24d4d

PR #2398, opened 2022-05-16
Make confluent-schema-registry-integration sample poms consistent

onobc:
@sobychacko I noticed when review the Kafka native sample that the Confluent schema registry sample was a bit inconsistent. I believe the original PR had the `avro-samples` intermediate level pom removed for simplicity and I think this may have been a "miss" from that. 

Anyways, comments inline.

### DIFF
--- a/samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer1/pom.xml
+++ b/samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer1/pom.xml
@@ -17,22 +17,18 @@
     <dependencies>
         <dependency>
             <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-actuator</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.avro</groupId>
-            <artifactId>avro</artifactId>
-            <version>${avro.version}</version>
         </dependency>
         <dependency>
             <groupId>org.springframework.cloud</groupId>
             <artifactId>spring-cloud-stream-binder-kafka</artifactId>
         </dependency>
-        <dependency>
+	    <dependency>
+		    <groupId>org.apache.avro</groupId>
+		    <artifactId>avro</artifactId>
+		    <version>${avro.version}</version>
+	    </dependency>
+	    <dependency>
             <groupId>io.confluent</groupId>
             <artifactId>kafka-avro-serializer</artifactId>
             <version>${confluent.version}</version>
--- a/samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer1/pom.xml
+++ b/samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer1/pom.xml
@@ -14,42 +14,6 @@
 		<version>4.0.0-SNAPSHOT</version>
 	</parent>
 
-    <dependencies>
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-web</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.springframework.cloud</groupId>
-            <artifactId>spring-cloud-stream-binder-kafka</artifactId>
-        </dependency>
-	    <dependency>
-		    <groupId>org.apache.avro</groupId>
-		    <artifactId>avro</artifactId>
-		    <version>${avro.version}</version>
-	    </dependency>
-	    <dependency>
-            <groupId>io.confluent</groupId>
-            <artifactId>kafka-avro-serializer</artifactId>
-            <version>${confluent.version}</version>
-            <exclusions>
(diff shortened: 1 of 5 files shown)


samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer1/pom.xml
```diff
@@ -15,24 +15,20 @@
 	</parent>
 
     <dependencies>
-        <dependency>
```
onobc, line 18: I noticed we have actuator in all the samples by default. I know there are some SCS things in there, but should we keep the pom light and focused on what we are demoing? Or is there some value in getting our SCS actuator code to execute during startup to possible show any issues w/ specific configurations/samples?

samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer1/pom.xml
```diff
@@ -15,24 +15,20 @@
...
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-actuator</artifactId>
-        </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
         </dependency>
-        <dependency>
```
onobc, line 26: I just moved all of the avros below, together in the 3 app poms.

samples/confluent-schema-registry-integration/pom.xml
```diff
@@ -1,12 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
-	<groupId>org.springframework.cloud</groupId>
 	<artifactId>confluent-schema-registry-integration</artifactId>
 	<version>4.0.0-SNAPSHOT</version>
 	<packaging>pom</packaging>
 	<name>confluent-schema-registry-integration</name>
-	<description>confluent-schema-registry-integration</description>
+	<description>Confluent Schema Registry Integration Sample App</description>
```
onobc, line 8: Should we try to put somethng descriptive here for all of the samples or simply omit the element altogether if we just end up copyinhg the "name" value into this?

samples/pom.xml
```diff
@@ -18,7 +18,7 @@
 	<properties>
 		<avro.version>1.11.0</avro.version>
 		<confluent.version>7.0.1</confluent.version>
-		<spring-cloud.version>2022.0.0-SNAPSHOT</spring-cloud.version>
+		<spring-cloud.version>2022.0.3-SNAPSHOT</spring-cloud.version>
```
onobc, line 21: I think this is the latest value now.

samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer2/pom.xml
```diff
@@ -15,23 +15,19 @@
...
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.avro</groupId>
-            <artifactId>avro</artifactId>
-            <version>${avro.version}</version>
-        </dependency>
         <dependency>
```
onobc, line 22: Should we move all of the common deps up to the parent confluent-schema-registry-integration/pom.xml for these 3 apps? Or do we have them duplicated in here so peepz can easily see what each app has/requires? 

I had moved the avro deps up to that middle `avro-samples/pom.xml` in the original PR for the reason of de-duplicating. I am ok either way, just would like to know if its intentional or not.

samples/confluent-schema-registry-integration/pom.xml
sobychacko: Descriptive is better, IMO.

samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer1/pom.xml
sobychacko: LGTM.

samples/confluent-schema-registry-integration/pom.xml
onobc: Sounds good. Lets keep an eye out on each one as we migrate over and try to make sure we give them something descriptive.

samples/confluent-schema-registry-integration/confluent-schema-registry-integration-producer2/pom.xml
sobychacko: I think having it at a common location is fine. No need to duplicate. It was probably again an oversight to have them all over the place.

samples/pom.xml
onobc: I was mistaken - I got confused w/ 2021. I am adjusting in commit shortly.

PR #2403, opened 2022-05-18
Migrate stream-bridge-avro sample app

onobc:


samples/stream-bridge-avro/src/main/resources/application.yml
```diff
@@ -0,0 +1,72 @@
+spring:
+  cloud:
+    stream:
+      function:
+        definition: supplier;receiveAndForward;receive
+      # this default config avoids having to set the property on every binding (see commented section below)
+      default:
```
onobc, line 7: So part of me wanted to **not** put in the default properties as it makes it less clear where they would be configured on each binding otherwise. To show both the factored-out-optimized config and the individiual-config I put both and left the latter commented out w/ a comment describing this fact.

sobychacko (approved):

sobychacko:
LGTM - Merged upstream.

PR #2408, opened 2022-05-26
add missing backoff multiplier

amseager:
Currently, when `KafkaTransactionManager` is present, `BackOff` for `DefaultAfterRollbackProcessor` and `ListenerContainerWithDlqAndRetryCustomizer` is being configured without multiplier (it uses the default value 1.5). 

With this PR, it's going to be filled from the config properties as `InitialInterval` and `MaxInterval` do.


sobychacko:
Merged via https://github.com/spring-cloud/spring-cloud-stream/commit/90e5d5ee8deb316209bc317c4cf9a8f39988594f

sobychacko:
`Backported to 3.2.x as well`.  Thanks for the PR!

PR #1203, opened 2018-01-31
#1202 Inconsistent use of this.prefix in AvroSchemaRegistryClientMessageConverter.resolveSchemaForWriting

jmax01:
Replaced hard coding with `this.prefix`

PR #2410, opened 2022-05-27
Clean up resources on close. Allow clean application termination.

tszmytka:
An issue submitted in the former repository #1203 has not been identified correctly (at least partly). 
There still is a bug where cleanly closing your application by invoking `org.springframework.context.ConfigurableApplicationContext#close` is not possible if an `ExecutorService` has been initiated in `org.springframework.cloud.stream.binder.kafka.KafkaBinderMetrics#bindTo`. Other Spring resources get closed but the scheduler remains hanging and keeps the application JVM process running.
This PR fixes that and allows proper termination.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,3 +12,7 @@ insert_final_newline = true
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.java]
+ij_java_names_count_to_use_import_on_demand = 999
+ij_java_class_count_to_use_import_on_demand = 999
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
@@ -63,6 +63,7 @@ import org.springframework.util.ObjectUtils;
  * @author Thomas Cheyney
  * @author Gary Russell
  * @author Lars Bilger
+ * @author Tomek Szmytka
  */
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -12,3 +12,7 @@ insert_final_newline = true
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.java]
+ij_java_names_count_to_use_import_on_demand = 999
+ij_java_class_count_to_use_import_on_demand = 999
```
sobychacko, line 18: Why is this change necessary?

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderMetrics.java
```diff
@@ -64,7 +65,7 @@
  * @author Lars Bilger
  */
```
sobychacko, line 66: Please add your name as an author. Also, make sure that any Java classes modified have the updated copyright years. Thanks!

sobychacko (changes requested):

.editorconfig
tszmytka: As the commit message says: this instructs the IDE to not try to bundle multiple imports into one with a star (default behavior in Intellij IDEA) - this seems to be the repo's preference. Admittedly it has no direct connection with the feature but keeps you from messing up when you use automatic code formatting.

sobychacko: That is fine. The checkstyle rules would prevent using asterisks for multiple imports.

sobychacko:
Thanks for the PR addressing this issue. I squashed your commits, polished and then merged upstream. Backported to `3.2.x` as well.

PR #2447, opened 2022-06-26
Make InteractiveQueryService only return used state stores

### LINKED ISSUE #2445: InteractiveQueryService throwing exceptions on Kafka Streams 3.1 with multiple Kafka Streams applications
**Describe the issue**

When upgrading to Spring Boot 2.7.0 with Kafka 3.1.1 I came across an issue with `InteractiveQueryService`. Our application uses functional binding with multiple Kafka Streams applications, each with it's own state store defined as a `StoreBuilder` bean. When trying to query the stores, the `ReadOnlyKeyValueStore.get(key)` started to throw 

```
org.apache.kafka.streams.errors.InvalidStateStoreException: The state store, <store>, may have migrated to another instance.
```

The issue appears to be caused by a slight behavior change of `org.apache.kafka.streams.KafkaStreams#store(StoreQueryParameters<T>...)` called by `InteractiveQueryService` - instead of returning just the stores actually used in the one topology, it checks all stores in the topology. This, in combination with Spring Cloud Stream `AbstractKafkaStreamsBinderProcessor` adding all the `StoreBuilder` beans in context to every `StreamsBuilder`, is causing the `KafkaStreams#store(StoreQueryParameters<T>...)` method to always return a store no matter the Kafka Streams instance, but one that cannot be used, unless the right Kafka Streams instance happens to be the first one checked.

I'm not sure if this is more of an SCS or Kafka Streams issue, but in Kafka Streams does exactly as specified in the method documentation

> Throws:
>   UnknownStateStoreException – If the specified store name does not exist in the topology.

**To Reproduce**

See the repository at https://github.com/Xrixcis/kstreams-error-demo

**Version of the framework**

Spring Boot 2.7.0 / Kafka Streams 3.1.1
Spring Cloud Stream 2021.0.3
Spring Cloud Stream Kafka Streams Binder 3.2.4

**Expected behavior**

Stores can be queried without exceptions as on Kafka Streams 3.0

onobc:
Fixes #2445

### Background
In KafkaStreams 3.0 the mechanism used to answer _“does this topology have this store?“_ was changed.

In 3.0 it asked the actual built topology which would not return a state store unless the KafkaStreams instance was actually using it:
```java
public <T> T store(StoreQueryParameters<T> storeQueryParameters) {
    this.validateIsRunningOrRebalancing();
    String storeName = storeQueryParameters.storeName();
    if (this.taskTopology != null && this.taskTopology.hasStore(storeName) || this.globalTaskTopology != null && this.globalTaskTopology.hasStore(storeName)) {
        return this.queryableStoreProvider.getStore(storeQueryParameters);
    } else {
        throw new UnknownStateStoreException("Cannot get state store " + storeName + " because no such store is registered in the topology.");
    }
}
```

However in 3.2 it asks the topology metadata which will answer yes if any of its state factories have the state store, regardless if it used or not. 

```java
public <T> T store(final StoreQueryParameters<T> storeQueryParameters) {
    validateIsRunningOrRebalancing();
    final String storeName = storeQueryParameters.storeName();
    if (!topologyMetadata.hasStore(storeName)) {
        throw new UnknownStateStoreException(
            "Cannot get state store " + storeName + " because no such store is registered in the topology."
        );
    }
    return queryableStoreProvider.getStore(storeQueryParameters);
}
```
Because SCSt adds all state stores beans to all KafkaStreams apps, we now get back unwanted/unused state stores.

### Not Chosen Solution
One solution would be to add an extra param to `InteractiveQueryService.store` for the application id to differentiate which KafkaStreams app the caller is interested in. However, this is an API change and rather intrusive. 

### Chosen Solution
Instead we have added an additional check against the streams state store metadata to see if it the state store is actually being used. This returns us back to the 3.0 behavior. 

### Caveat
⚠️  The streams state store metadata check will fail if the topology is a named topology. This should not be an issue w/ the streams created by SCSt. If it does become a problem later, we will then add the additional optional app id param to `InteractiveQueryService.store`.

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
@@ -16,12 +16,10 @@
 
 package org.springframework.cloud.stream.binder.kafka.streams;
 
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
@@ -33,6 +31,7 @@ import org.apache.kafka.streams.KeyQueryMetadata;
 import org.apache.kafka.streams.StoreQueryParameters;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.InvalidStateStoreException;
+import org.apache.kafka.streams.errors.UnknownStateStoreException;
 import org.apache.kafka.streams.state.HostInfo;
 import org.apache.kafka.streams.state.QueryableStoreType;
 import org.apache.kafka.streams.state.StreamsMetadata;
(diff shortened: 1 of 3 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
```diff
@@ -105,32 +100,34 @@ public <T> T getQueryableStore(String storeName, QueryableStoreType<T> storeType
...
-					throwable);
+			throw new IllegalStateException("Error retrieving state store: " + storeName, throwable);
 		});
 	}
 
+	private <T> T getStateStoreFromKafkaStreams(KafkaStreams kafkaStreams, String storeName, QueryableStoreType<T> storeType) {
+		// Check KafkaStreams app knows about the state store
+		T store = kafkaStreams.store(StoreQueryParameters.fromNameAndType(storeName, storeType));
+
+		// Check KafkaStreams app actually has the state store
```
onobc, line 123: This is the extra check that fixes the issue. If you comment out these lines you will see the newly added tests fail.

PR #2448, opened 2022-06-27
Make InteractiveQueryService only return used state stores (backport)

onobc:
Backport of https://github.com/spring-cloud/spring-cloud-stream/commit/0cb59496fef59226afafbda464230231b38b0386

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryServiceMultiStateStoreTests.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryServiceMultiStateStoreTests.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright 2022-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryServiceMultiStateStoreTests.java
```diff
@@ -0,0 +1,235 @@
...
+								+ embeddedKafka.getBrokersAsString())
+		) {
+			waitForRunningStreams(context.getBean(KafkaStreamsRegistry.class));
+
+			InteractiveQueryService queryService = context.getBean(InteractiveQueryService.class);
+
+			assertThatThrownBy(() -> queryService.getQueryableStore(STORE_1_NAME, QueryableStoreTypes.keyValueStore()))
+					.isInstanceOf(IllegalStateException.class)
+					.hasMessage("Error retrieving state store: " + STORE_1_NAME)
+					.hasRootCauseInstanceOf(UnknownStateStoreException.class);
```
onobc, line 120: You will notice that I backed off of the specific message verification as was done in main. This is because our message is different than the message spit out by KafkaStreams 3.0. Same exception type, slightly different message. 
```
SCSt 3.2.x + KafkaStreams 3.0 -> old message
SCSt3.2.x + KafkaStreams 3.1+ -> new message
```
I think this is fine as an exception message should not be considered public API.

PR #2489, opened 2022-08-20
Remove kafka scala dependency management

onobc:
### What
This PR removes the `org.apache.kafka:kafka_2.13` dependencies from the KafkaBinder. 

### Why
This was causing the `SC 2021.0.x` + `SB 2.7.x` + `SCSt 3.2.x` compatibility tests to fail. Any tests that were using `@EmbeddedKafka/EmbeddedKafkaBroker` (eg `KafkaBinderMeterRegistryTest`) would get the `NoSuchMethodError` down in the KafkaServer Scala code.  

`SB 2.7.x` uses `spring-kafka:2.8.8` which in turn uses `kafka-clients:3.1.1` which in turn requires a `org.apache.kafka:kafka_2.13:jar:3.1.1:test` **NOT** a `org.apache.kafka:kafka_2.13:jar:3.0.0:test` for the embedded server to work properly.

SCSt directly pulls in the Kafka Scala libs for test and uses the `kafka.version` property to manage it’s version numbers - [which is 3.0.0]( https://github.com/spring-cloud/spring-cloud-stream/blob/168184b370df31f76ddb9898cf73c37accb0a2ba/binders/kafka-binder/pom.xml#L25) in SCS 3.2.x.

It is not enough to simply switch the SB version to 2.7.3 as these Kafka scala test jars are not version aligned / managed via a BOM so it seems. Bumping SB to 2.7.3 does cover most things, but not these. We end up w/ the variant ending in `3.0.0`. 

Because `spring-kafka-test` brings these dependencies in transitively to allow use of `EmbeddedKafka` and SCSt was only using these during tests, we can safely remove them altogether. 

### Concerns
We were specifying the lib at both test and [compile](https://github.com/spring-cloud/spring-cloud-stream/pull/2489/files#diff-fe33fe2a739aec23d729b97252302c77e4a4ed485cde1b000352f6068edeba16L67) scopes which would transitively piggyback this jar to the SCSt library consumers. This change will stop that transitivity. 

My understanding is that the only reason for having this lib around besides for `@EmbeddedKafka` is  for the older Kafka clients. I do not think this is valid any longer w/ the versions of Kafka that spring-kafka supports. @garyrussell can you ACK or NACK my assumption here?

### DIFF
--- a/binders/kafka-binder/pom.xml
+++ b/binders/kafka-binder/pom.xml
@@ -20,10 +20,6 @@
 	</scm>
 	<properties>
 		<java.version>1.8</java.version>
-		<spring-kafka.version>2.8.5</spring-kafka.version>
-		<spring-integration-kafka.version>5.5.8</spring-integration-kafka.version>
-		<kafka.version>3.0.0</kafka.version>
-		<spring-cloud-stream.version>3.2.5-SNAPSHOT</spring-cloud-stream.version>
 		<maven-checkstyle-plugin.failsOnError>true</maven-checkstyle-plugin.failsOnError>
 		<maven-checkstyle-plugin.failsOnViolation>true</maven-checkstyle-plugin.failsOnViolation>
 		<maven-checkstyle-plugin.includeTestSourceDirectory>true</maven-checkstyle-plugin.includeTestSourceDirectory>
@@ -34,34 +30,6 @@
 		<module>spring-cloud-stream-binder-kafka-core</module>
 		<module>spring-cloud-stream-binder-kafka-streams</module>
 	</modules>
-
-	<dependencyManagement>
-		<dependencies>
-			<dependency>
-				<groupId>org.springframework.kafka</groupId>
-				<artifactId>spring-kafka</artifactId>
-				<version>${spring-kafka.version}</version>
-			</dependency>
-			<dependency>
-				<groupId>org.apache.kafka</groupId>
-				<artifactId>kafka-streams</artifactId>
-				<version>${kafka.version}</version>
-				<exclusions>
-					<exclusion>
-						<groupId>org.slf4j</groupId>
-						<artifactId>slf4j-log4j12</artifactId>
-					</exclusion>
-				</exclusions>
-			</dependency>
-			<dependency>
-				<groupId>org.apache.kafka</groupId>
-				<artifactId>kafka-clients</artifactId>
-				<version>${kafka.version}</version>
-				<classifier>test</classifier>
-				<scope>test</scope>
-			</dependency>
-		</dependencies>
-	</dependencyManagement>
 
 	<dependencies>
 		<dependency>
--- a/binders/kafka-binder/pom.xml
+++ b/binders/kafka-binder/pom.xml
@@ -55,33 +55,6 @@
 			</dependency>
 			<dependency>
 				<groupId>org.apache.kafka</groupId>
-				<artifactId>kafka_2.13</artifactId>
-				<version>${kafka.version}</version>
-				<scope>test</scope>
-			</dependency>
-			<dependency>
-				<groupId>org.apache.kafka</groupId>
-				<artifactId>kafka_2.13</artifactId>
-				<classifier>test</classifier>
-				<scope>test</scope>
-				<version>${kafka.version}</version>
-				<exclusions>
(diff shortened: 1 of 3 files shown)


spencergibb (approved):

binders/kafka-binder/pom.xml
```diff
@@ -53,33 +53,6 @@
 					</exclusion>
 				</exclusions>
 			</dependency>
-			<dependency>
-				<groupId>org.apache.kafka</groupId>
-				<artifactId>kafka_2.13</artifactId>
-				<version>${kafka.version}</version>
```
garyrussell, line 59: We should remove this property altogether and let Boot manage the versions of all dependencies.

garyrussell (changes requested):
>It is not enough to simply switch the SB version to 2.7.3 as these Kafka scala test jars are not version aligned / managed via a BOM so it seems. Bumping SB to 2.7.3 does cover most things, but not these. We end up w/ the variant ending in 3.0.0.

Historically, Boot did not manage the versions for these two jars; hence the need here. That was fixed some time ago.

binders/kafka-binder/pom.xml
onobc: Sure enough, the SB BOM covers all of them in [2.6.x](https://github.com/spring-projects/spring-boot/blob/6d80b601b69f02059bcda618049039a0cbea550c/spring-boot-project/spring-boot-dependencies/build.gradle#L1057) as well as [2.7.x](https://github.com/spring-projects/spring-boot/blob/392aec4a789ece4341cb3c8e7fac13b474874d65/spring-boot-project/spring-boot-dependencies/build.gradle#L1088). 

I have removed the `kafka.version` and the `dependencyManagement` altogether and tested against `2.7x` and `2.6x` and all is well. 

This eases the concerns of whether or not the `kafka_213` is required or not. However, for future reference, @garyrussell do you know if that jar is required any longer? (from my concern in original description).

binders/kafka-binder/pom.xml
```diff
@@ -20,10 +20,6 @@
 	</scm>
 	<properties>
 		<java.version>1.8</java.version>
-		<spring-kafka.version>2.8.5</spring-kafka.version>
```
onobc, line 23: 2 of these were unused prior to this MR (`spring-integration-kafka.version` and `spring-cloud-stream.version`). 
The other 2 are no longer needed as we defer to SB BOM for dependency mgmt.

binders/kafka-binder/pom.xml
```diff
@@ -35,61 +31,6 @@
 		<module>spring-cloud-stream-binder-kafka-streams</module>
 	</modules>
 
-	<dependencyManagement>
```
onobc, line 38: This is to allow SB BOM full control of dep. mgmt.

binders/kafka-binder/pom.xml
garyrussell: Yes, it is needed by the embedded broker code (`TestUtils.createServer()`).

onobc: Right, but other than for testing, it is not needed by consumers of SCSt-kafka. Correct? 

> My understanding is that the only reason for having this lib around besides for @EmbeddedKafka is for the older Kafka clients. I do not think this is valid any longer w/ the versions of Kafka that spring-kafka supports

garyrussell: I am not sure what you mean; it has nothing to do with the clients; it is required by the embedded broker.

https://github.com/spring-projects/spring-kafka/blob/51db0b6ecca9514686b4c138de69a1d9efe80664/spring-kafka-test/src/main/java/org/springframework/kafka/test/EmbeddedKafkaBroker.java#L350

The broker uses some common code from the clients, so everything has to be the same version.

onobc: To close this thread up, we chatted offline and the summary was this for the `kafka_213` scala jar:

> The one classified as `test` has never been used by anything other than the embedded broker; 
The one not classified as `test` might have been used by very early versions of the clients but now all common code (used by both the broker and clients) has been moved to Java and is located in kafka-clients jar

IOW - it would be safe to remove the one not classified as `test`.

sobychacko (approved):

Issue #2491, opened 2022-08-24
Schema registry migration


sobychacko:
Migrate Schema Registry from spring-cloud back to Spring Cloud Stream.

It is still desirable to have a schema registry implementation that caters to the needs of Spring Cloud Stream use cases. Our current (deprecated) Spring Cloud Schema Registry can be better managed and maintained if the scope is limited to Spring Cloud Stream use cases. This is how it was before, hence with the `4.0.0` version of Spring Cloud Stream, we will attempt to bring it back as part of Spring Cloud Stream. 

https://github.com/spring-cloud/spring-cloud-schema-registry

PR #2492, opened 2022-08-24
Schema Registry Migration

sobychacko:
Initial attempt at migrating schema registry from https://github.com/spring-cloud/spring-cloud-schema-registry
to Spring Cloud Stream.

Schema Registry within the scope of Spring Cloud Stream will only support Spring Cloud Stream
specific use cases in the context of binders.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2491


sobychacko:
@olegz This is an initial take on the schema registry migration. Once this is reviewed/merged, we will have more changes, refactoring, docs, samples, etc. to follow.

Issue #2486, opened 2022-08-18
Don't Create an AdminClient if `autoCreateTopics` is False


garyrussell:
While commenting on this question https://stackoverflow.com/questions/73408817/kafka-version-3-0-1-kafka-admin-clients-created-repeatedly-memory-leak I noticed that, if topic creation is disabled, we still go ahead and create an `AdminClient`, only to immediately close it. We should skip creating the client if we're not going to use it.

PR #2494, opened 2022-08-25
AdminClient usage if autoCreateTopics is disabled

sobychacko:
Don't Create an AdminClient if autoCreateTopics is False

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2486


binders/kafka-binder/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2018 the original author or authors.
+ * Copyright 2014-2012 the original author or authors.
```
garyrussell, line 2: 2022

garyrussell (changes requested):
Just one nit.

PR #2528, opened 2022-10-09
Smarter state store resolution for InteractiveQueryService

### LINKED ISSUE #2523: Kafka stream state stores cannot be found after upgrade to 3.2.5
**Describe the issue**
The following code works with SCS 3.2.4:
```kotlin
@Service
class TechnischerplatzAdresseHandler(private val interactiveQueryService: InteractiveQueryService) {
    private val technischerplatzAdresseStore: ReadOnlyKeyValueStore<String, TechnischerplatzAdresse>
        get() = interactiveQueryService.getStore("technischerplatzadresse-store")

    suspend fun get(request: ServerRequest) = withContext(Dispatchers.IO) {
        technischerplatzAdresseStore["1234"]?.let {
            ServerResponse.ok().build().awaitSingle()
        } ?: ServerResponse.notFound().buildAndAwait()
    }
}
```

But it raises an exception with SCS 3.2.5:
```
Caused by: org.apache.kafka.streams.errors.UnknownStateStoreException: Cannot get state store technischerplatzadresse-store because no such store is registered in the topology.
	at org.apache.kafka.streams.KafkaStreams.store(KafkaStreams.java:1603)
	at org.springframework.cloud.stream.binder.kafka.streams.InteractiveQueryService.getStateStoreFromKafkaStreams(InteractiveQueryService.java:121)
	at org.springframework.cloud.stream.binder.kafka.streams.InteractiveQueryService.lambda$getQueryableStore$0(InteractiveQueryService.java:109)
	at org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:329)
	at org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:209)
	at org.springframework.cloud.stream.binder.kafka.streams.InteractiveQueryService.getQueryableStore(InteractiveQueryService.java:88)
	at at.wrwks.bef.projekt.domain.technischerplatzadresse.TechnischerplatzAdresseHandler$get$2.invokeSuspend(TechnischerplatzAdresseHandler.kt:39)
```

**To Reproduce**
Try to load a state store as described

**Version of the framework**
Spring Cloud ~~2020.0.4~~ 2021.0.4 with Spring Cloud Stream 3.2.5

**Expected behavior**
**Screenshots**

**Additional context**
With debugging I got down to `StreamsMetadataState#isInitialized()` where `clusterMetadata.topics().isEmpty() == true` which it shouldn't be. It is located in `kafka-streams:3.1.2` on line 321.
The topology hasn't changed and the store is there.

onobc:
This proposal modifies the mechanism used to resolve (choose) state store in `InteractiveQueryService#getQueryableStore`. 

The current mechanism relies on the `KafkaStreams.StreamsMetadataState` which is not reliable unless the `application.server` streams config property is set. For a KafkaStreams app w/ a single instance this should not have to be set. 

The new mechanism does a couple of things:
1. only goes into the "resolve the store" logic when in fact there are multiple state store candidates
2. does not use the above mechanism (StreamsMetadataState) but rather the `topologyMetadata.sourceTopics`

#### Commit 1 
The 1st commit does not change anything other than modify the existing tests so that they surface the bug. They previously were not verifying this.

#### Commit 2
Actual code change described above.

Fixes #2523

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.stream.binder.kafka.streams;
 
+import java.lang.reflect.Field;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -32,6 +34,7 @@ import org.apache.kafka.streams.StoreQueryParameters;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.InvalidStateStoreException;
 import org.apache.kafka.streams.errors.UnknownStateStoreException;
+import org.apache.kafka.streams.processor.internals.TopologyMetadata;
 import org.apache.kafka.streams.state.HostInfo;
 import org.apache.kafka.streams.state.QueryableStoreType;
 import org.apache.kafka.streams.state.StreamsMetadata;
@@ -41,6 +44,7 @@ import org.springframework.retry.RetryPolicy;
 import org.springframework.retry.backoff.FixedBackOffPolicy;
 import org.springframework.retry.policy.SimpleRetryPolicy;
 import org.springframework.retry.support.RetryTemplate;
+import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 
 /**
@@ -53,6 +57,7 @@ import org.springframework.util.StringUtils;
  * @author Renwei Han
  * @author Serhii Siryi
  * @author Nico Pommerening
+ * @author Chris Bono
  * @since 2.1.0
  */
 public class InteractiveQueryService {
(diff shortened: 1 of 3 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
```diff
@@ -104,28 +114,51 @@ public <T> T getQueryableStore(String storeName, QueryableStoreType<T> storeType
...
-		if (kafkaStreams.streamsMetadataForStore(storeName).stream()
-				.noneMatch((sm) -> sm.stateStoreNames().contains(storeName))) {
-			throw new UnknownStateStoreException("Store (" + storeName + ") not available to Streams instance");
+	private boolean storeActuallyAvailable(KafkaStreams kafkaStreams, String storeName) {
+		try {
+			TopologyMetadata topologyMetadata = (TopologyMetadata) ReflectionUtils.getField(topologyMetadataField, kafkaStreams);
+			return !topologyMetadata.sourceTopicsForStore(storeName, null).isEmpty();
+		}
+		catch (Exception ex) {
+			LOG.error("Unable to determine if store (" + storeName + ") is available to the app due to: " + ex.getMessage(), ex);
```
onobc, line 159: When/if this fails, we will not surface exception but rather log and let resolve process continue.

PR #2531, opened 2022-10-12
Remove Usage of RabbitMQ http-client

garyrussell:
Replace with `WebClient`.


olegz:
Resolved with 48c855075

Issue #2520, opened 2022-10-03
Error in KafkaTopicProvisioner For Consumer


omercelikceng:
Hello, you have solved this error before. Thank you so much. 
But you just solved this for **ProducerProvisioner(provisionProducerDestination method)** .  (https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/888 )
I am currently getting the same error for **ConsumerProvisioner**. 
Could you please add the **retry mechanism in ConsumerProvisioner?** Or I can add it and open it in pull request.

You have identified the problem very well in this link. That's why I didn't give a long explanation. But I can give a longer explanation if you want. (https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/888 )

Your solution for the producer:

![ProducerProvisioner](https://user-images.githubusercontent.com/34758411/193551301-844eb2ff-fa5d-4286-8f44-bcd27de2e0c0.PNG)

But still no retry mechanism has been added for consumer provisioner.

![ConsumerProvisioner](https://user-images.githubusercontent.com/34758411/193552207-7b821ee5-ed42-4580-92ca-1ac6356bd30f.PNG)

I am getting the following error from the KafkaTopicProvisioner:

```
[2022-10-03T12:39:05.005Z] [org.springframework.cloud.stream.binding.BindingService] [main] [200] [ERROR] Failed to create consumer binding; retrying in 30 seconds
org.springframework.cloud.stream.provisioning.ProvisioningException: Provisioning exception encountered for person-topic-1 nested exception is java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.UnknownTopicOrPartitionException: This server does not host this topic-partition.
	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.doProvisionConsumerDestination(KafkaTopicProvisioner.java:252)
	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.provisionConsumerDestination(KafkaTopicProvisioner.java:202)
	at org.springframework.cloud.stream.binder.kafka.provisioning.KafkaTopicProvisioner.provisionConsumerDestination(KafkaTopicProvisioner.java:86)
	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder.doBindConsumer(AbstractMessageChannelBinder.java:421)
[...]

PR #2536, opened 2022-10-16
KafkaTopicProvisioner Retry Mechanism Improvement For Consumer

omercelikceng:
Use a retry template for topic description method call through admin client when provisioning consumer destinations. We are retrying because in the event this operation gets failed, it is retried with the default retry settings in the provisioner.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2520


sobychacko (approved):

sobychacko:
Looks good. I had to make some checkstyle and test fixes. Squashed the commits and merged upstream to `main`. Backported to `3.2.x`.

PR #2578, opened 2022-12-05
Fix typo in ListenerContainerWithDlqAndRetryCustomizer example of kafka binder

snimmagadda1:
noticed this when trying to implement


sobychacko:
Thanks for the PR! It is now merged upstream.

Issue #2576, opened 2022-12-05
ProducerMessageHandlerCustomizer doesn't allow to configure the Micrometer Tracing Observation


codependent:
The issue is thouroughly described in the following Stackoverflow question: https://stackoverflow.com/questions/74666562/how-to-instrument-spring-boot-3-x-with-spring-cloud-stream-4-x-producers-and-con --> UPDATES 1 & 2

**Version of the framework** Spring Boot 3 - Spring Cloud Stream 4.0.0-RC3

sobychacko:
Hi, @codependent, Thanks for reporting the issue and the context on the SO thread. Any chance you can create a simple application where we can reproduce the issue and try some fixes?

codependent:
> Hi, @codependent, Thanks for reporting the issue and the context on the SO thread. Any chance you can create a simple application where we can reproduce the issue and try some fixes?

Sure!

codependent:
I've uploaded a sample project reproducing the problem: https://github.com/codependent/kotlin-kafka-correlation

It just needs a local Kafka cluster on the default port.

After starting the application just invoke `curl localhost:8080/publish`

Relevant code:

The controller is wrapped with `withObservationContext(observationRegistry) {` to setup the coroutine context with the tracing information used in MDC:

```
        withObservationContext(observationRegistry) {
            eventProducer.send(RandomStringUtils.randomAlphanumeric(10), RandomStringUtils.randomAlphanumeric(10))
        }
```

In order to propagate the tracing information in the Kafka Producer I had to define the following `ProducerMessageHandlerCustomizer` as indicated by Gary Russell.
```
@Configuration
class KafkaProducerConfiguration {

    @Bean
    fun kafkaProducerObservationCustomizer (applicationContext: ApplicationContext, observationRegistry: ObservationRegistry) : ProducerMessageHandlerCustomizer<KafkaProducerMessageHandler<String, String>> {
        return ProducerMessageHandlerCustomizer<KafkaProducerMessageHandler<String, String>> {
                handler, destinationName ->
            handler.kafkaTemplate.setObservationEnabled(true)
            handler.kafkaTemplate.setApplicationContext(applicationContext)
            handler.kafkaTemplate.afterSingletonsInstantiated()
        }
    }
}
```

This additional calls:

```
            handler.kafkaTemplate.setApplicationContext(applicationContext)
            handler.kafkaTemplate.afterSingletonsInstantiated()
```
[...]

PR #2582, opened 2022-12-08
Observation related changes in Kafka binder

sobychacko:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2576

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -508,12 +508,17 @@ public class KafkaMessageChannelBinder extends
 			kafkaTemplate.setObservationEnabled(true);
 		}
 		kafkaTemplate.setApplicationContext(getApplicationContext());
-		customizeProducerMessageHandler(handler, destination.getName());
-		((SmartInitializingSingleton) kafkaTemplate).afterSingletonsInstantiated();
 
 		return handler;
 	}
 
+	@Override
+	@SuppressWarnings("rawtypes")
+	protected void customizeProducerMessageHandler(MessageHandler producerMessageHandler, String destinationName) {
+		super.customizeProducerMessageHandler(producerMessageHandler, destinationName);
+		KafkaTemplate<?, ?> kafkaTemplate = ((KafkaProducerMessageHandler) producerMessageHandler).getKafkaTemplate();
+		((SmartInitializingSingleton) kafkaTemplate).afterSingletonsInstantiated();
+	}
 
 	@Override
 	protected void postProcessOutputChannel(MessageChannel outputChannel,
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -51,6 +51,7 @@ import org.apache.kafka.common.header.internals.RecordHeader;
 import org.apache.kafka.common.header.internals.RecordHeaders;
 
 import org.springframework.beans.factory.DisposableBean;
+import org.springframework.beans.factory.SmartInitializingSingleton;
 import org.springframework.cloud.stream.binder.AbstractMessageChannelBinder;
 import org.springframework.cloud.stream.binder.BinderHeaders;
 import org.springframework.cloud.stream.binder.BinderSpecificPropertiesProvider;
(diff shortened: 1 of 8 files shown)


garyrussell:
LGTM - after further thought; we should just document that this property must be used rather than `handler.getKafkaTemplate().setObservationEnabled()` in a customizer. Otherwise, (s)he must call `afterSingletonsInstantiated()`

sobychacko:
@garyrussell Because of this [change](https://github.com/spring-cloud/spring-cloud-stream/pull/2582/files#diff-fd4ebeed7803453faa88bd22484e805d5f74c0b2b8da355f2fd1b37bb0357afeR512), the binder is the one now calls `afterSingletonsInstantiated` regardless of whether the new properties are used, or observations enabled through the customizer. If using the customizer option, I believe that the only thing that the user must do is enable observation. `KMCB` sets the application context on `KafkaTemplate,` then call the customizer and the `afterSingletonsInstantiated.` For `KMCB` to call the customizer, I had to change the `customizeProducerMessageHandler` method access level to `protected` (from `private`). Do you think that I am missing something?

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -619,6 +628,11 @@ protected MessageProducer createConsumerEndpoint(
 								? new ContainerProperties(Pattern.compile(topics[0]))
 								: new ContainerProperties(topics)
 						: new ContainerProperties(topicPartitionOffsets);
+
+		if (this.configurationProperties.isEnableObservation() || extendedConsumerProperties.getExtension().isEnableObservation()) {
```
garyrussell, line 632: I guess it's possible for them to enable it at the binder level and then disable it at a specific binding level.

OTOH, I wonder if we need binding level properties at all - they could always configure multiple binders if they want that level of control.

garyrussell (commented):
LGTM; just one comment for your consideration.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -502,6 +503,14 @@ public void fromHeaders(MessageHeaders headers, Headers target) {
 
 		}
 		handler.setHeaderMapper(mapper);
+
+		if (this.configurationProperties.isEnableObservation() || producerProperties.getExtension().isEnableObservation()) {
+			kafkaTemplate.setObservationEnabled(true);
+		}
+		kafkaTemplate.setApplicationContext(getApplicationContext());
+		customizeProducerMessageHandler(handler, destination.getName());
```
garyrussell, line 511: This will cause it to be customized twice, right?

garyrussell: Perhaps override it, call super, then aSI().

sobychacko: Ya, from what I could see, it is an idempotent invocation. Do you see any issues with calling it twice? This way, we can call the `afterSingletonsInstantiated` from `KMCB` properly. If the user decides to enable observation using the customizer. , we will miss the `enableObservation` check in `KafkaTempate` `aSI` method without calling the customizer here. We can't call `aSI` on `KafkaTemplate` in `AMCB`.

sobychacko: > Perhaps override it, call super, then `handler.getKT().aSI()`.

Done.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -502,9 +503,22 @@ public void fromHeaders(MessageHeaders headers, Headers target) {
...
+
 		return handler;
 	}
 
+	@Override
+	@SuppressWarnings("rawtypes")
+	protected void customizeProducerMessageHandler(MessageHandler producerMessageHandler, String destinationName) {
+		super.customizeProducerMessageHandler(producerMessageHandler, destinationName);
+		KafkaTemplate<?, ?> kafkaTemplate = ((KafkaProducerMessageHandler) producerMessageHandler).getKafkaTemplate();
+		((SmartInitializingSingleton) kafkaTemplate).afterSingletonsInstantiated();
```
garyrussell, line 520: Unnecessary cast.

garyrussell (changes requested):

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
sobychacko: My bad. Done now.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
garyrussell: Just to close the loop. Yes, it is likely idempotent, but we have no idea what user code might do. It might mutate objects outside of the template in a non-idempotent manner.

garyrussell (approved):

PR #2583, opened 2022-12-09
code optimization

oswardlx:
It is unnecessary to continue the cycle when defaultCandidatesExist = true;


oswardlx:
This is not important, or is there something wrong with my question? This is my first PR and I'm a bit inexperienced

oswardlx:
@sobychacko

PR #2594, opened 2022-12-18
Fix a message array index in kafka binder test code

ByungjunYou:
fix a message array index in kafka binder test code, so that the third message can be tested.

### message send/receive related code
https://github.com/spring-cloud/spring-cloud-stream/blob/f1d4f8a53ef4110ae70b9bd2ce99cc26ef3feee1/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java#L1761-L1768


sobychacko:
@ByungjunYou Thanks for catching this. Merging the PR.

Issue #2608, opened 2023-01-03
Kafka streams binder, way to tune StoreQueryParameters is required


raydac:
[InteractiveQueryService#getQueryableStore](https://github.com/spring-cloud/spring-cloud-stream/blob/af2df8ba0eff2e81c3a8fce358a09a54abb3a974/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java#L89) should provide way to tune StoreQueryParameters object. StoreQueryParameters has different options (for instance enableStaleStore), but that's impossible to tune them.

PR #2618, opened 2023-01-10
Allow fine tuning of StoreQueryParameters

sobychacko:
When using InteractiveQueryService in Kafka Streams binder, it is requested to customize the StoreQueryParameters beyond what is done through the InteractiveQueryService itself. Provide a StoreQueryParametersCustomizer that enables this customization.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2608


binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/InteractiveQueryService.java
```diff
@@ -298,10 +306,17 @@ public <K> KafkaStreams getKafkaStreams(String store, K key, Serializer<K> seria
 	public List<HostInfo> getAllHostsInfo(String store) {
 		return kafkaStreamsRegistry.getKafkaStreams()
 				.stream()
-				.flatMap(k -> k.allMetadataForStore(store).stream())
+				.flatMap(k -> k.streamsMetadataForStore(store).stream())
```
onobc, line 309: How is this change related?

sobychacko: It is unrelated; I just saw that the original call was deprecated and decided to clean up inline.

onobc: I would leave it as-is for this PR. The reason being is that it is deprecated in kafka-streams 3.3.1 but I think there is slight behavior change in the 2 methods across earlier versions of kafka-streams.

onobc (approved):
@sobychacko LGTM.

Issue #2613, opened 2023-01-05
Kafka Streams Binder: deserialization exception handling mechanism not invoked when routing configured based on event types


flemaitre:
When routing is configured based on event types (*), the configured deserialization exception handler (**) is not invoked, but the deserialization exception is handled by the default StreamsUncaughtExceptionHandler handler.

(*) https://docs.spring.io/spring-cloud-stream/docs/current/reference/html/spring-cloud-stream-binder-kafka.html#_event_type_based_routing_in_kafka_streams_applications
(**) https://docs.spring.io/spring-cloud-stream/docs/current/reference/html/spring-cloud-stream-binder-kafka.html#_handling_deserialization_exceptions_in_the_binder

Kafka Streams Binder 3.2.6

sobychacko:
@flemaitre Any chance you can create a very minimal sample application where we can reproduce this issue with some instructions to reproduce?

flemaitre:
@sobychacko Please find two demo apps (one for the Producer and one for the Consumer) here: https://github.com/flemaitre/scs-issue-2613-sample-app

sobychacko:
@flemaitre Thanks for the sample app. I looked into this area of the binder after observing your configuration. Here is the problem. When event type routing is enabled, we will use a byte[] Serde by default regardless of the value-serde you configure or the one the binder infers. Then, later on, if we have an event-type match, we will deserialize using the proper Serde. When using a byte serde, the deserialization works almost all the time, and no error is thrown; When later on you get the exception from the proper Serde, the issue is that it is now too late to hook into Kafka Stream's deserialization exception handling mechanism as we are now deep through the binder stack. The deserialization exception handling only works at the front end of the Kafka Streams stack, i.e, when the record is deserialized by Kafka Streams originally. 

To fix this, we can introduce a new flag in which you force the proper value serde at the outset to use by Kafka Streams rather than a byte[] Serde. There is a slight performance hit here, as we may unnecessarily deserialize all the incoming records using a complex Serde rather than a pass-through Serde implementation such as the `ByteSerde`. 

Let me know if it makes sense and then we can go ahead and try implementing this solution.

flemaitre:
[...]

PR #2633, opened 2023-01-24
Event type routing and deserialization issues

sobychacko:
In Kafka Streams binder, deserialization exception handler does not take effect when event type routing is enabled. Fixing this issue by allowing the applications to use the configured or inferred Serde rather than the byte[] Serde used by the event type router initially.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2613

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/AbstractKafkaStreamsBinderProcessor.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/AbstractKafkaStreamsBinderProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2022 the original author or authors.
+ * Copyright 2019-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -409,7 +409,7 @@ public abstract class AbstractKafkaStreamsBinderProcessor implements Application
 		}
 	}
 
-	@SuppressWarnings({"rawtypes", "unchecked"})
+	@SuppressWarnings({"unchecked"})
 	protected KStream<?, ?> getKStream(String inboundName, BindingProperties bindingProperties, KafkaStreamsConsumerProperties kafkaStreamsConsumerProperties,
 									StreamsBuilder streamsBuilder, Serde<?> keySerde, Serde<?> valueSerde, Topology.AutoOffsetReset autoOffsetReset, boolean firstBuild) {
 		if (firstBuild) {
@@ -428,8 +428,7 @@ public abstract class AbstractKafkaStreamsBinderProcessor implements Application
 		}
 
 		KStream<?, ?> stream;
-		final Serde<?> valueSerdeToUse = StringUtils.hasText(kafkaStreamsConsumerProperties.getEventTypes()) ?
-				new Serdes.BytesSerde() : valueSerde;
+		final Serde<?> valueSerdeToUse = getValueSerdeToUse(kafkaStreamsConsumerProperties, valueSerde);
 		final Consumed<?, ?> consumed = getConsumed(kafkaStreamsConsumerProperties, keySerde, valueSerdeToUse, autoOffsetReset);
 
 		if (this.kafkaStreamsExtendedBindingProperties
@@ -455,11 +454,19 @@ public abstract class AbstractKafkaStreamsBinderProcessor implements Application
 			// Branching based on event type match.
 			final KStream<?, ?>[] branch = stream.branch((key, value) -> matched.getAndSet(false));
 			// Deserialize if we have a branch from above.
-			final KStream<?, Object> deserializedKStream = branch[0].mapValues(value -> valueSerde.deserializer().deserialize(
-					topicObject.get(), headersObject.get(), ((Bytes) value).get()));
+			final KStream<?, Object> deserializedKStream = !kafkaStreamsConsumerProperties.isUseConfiguredSerdeWhenRoutingEvents() ?
+				branch[0].mapValues(value -> valueSerde.deserializer().deserialize(
+					topicObject.get(), headersObject.get(), ((Bytes) value).get())) : (KStream<?, Object>) branch[0];
 			return getkStream(bindingProperties, deserializedKStream, nativeDecoding);
 		}
 		return getkStream(bindingProperties, stream, nativeDecoding);
+	}
+
+	private Serde<?> getValueSerdeToUse(KafkaStreamsConsumerProperties kafkaStreamsConsumerProperties, Serde<?> valueSerde) {
+		if (StringUtils.hasText(kafkaStreamsConsumerProperties.getEventTypes())) {
+			return kafkaStreamsConsumerProperties.isUseConfiguredSerdeWhenRoutingEvents() ? valueSerde : new Serdes.BytesSerde();
+		}
+		return valueSerde;
 	}
 
 	private KStream<?, ?> getkStream(BindingProperties bindingProperties, KStream<?, ?> stream, boolean nativeDecoding) {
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/kafka/kafka-streams.adoc
```diff
@@ -1552,6 +1552,14 @@ For instance, if we want to change the header key on this binding to `my_event`
 
 `spring.cloud.stream.kafka.streams.bindings.process-in-0.consumer.eventTypeHeaderKey=my_event`.
 
+When using the event routing feature in Kafkfa Streams binder, it uses the byte array `Serde` to deserialze all incoming records.
+If the record headers match the event type, then only it uses the actual `Serde` to do a proper deserialization using either the configured or the inferred `Serde`.
+This introduces issues if you set a deserialization exception handler on the binding as the expected deserialization only happens down the stack causing unexpected errors.
+In order to address this issue, you can set the following property on the binding to force the binder to use the configured or inferred `Serde` instead of byte array `Serde`.
+This way, the application can detect deserialization issues right away when using the event routing feature and can take appropriate handling decisions.
+
+`spring.cloud.stream.kafka.streams.bindings.<process-in-0>.consumer.useConfiguredSerdeWhenRoutingEvents`
```
onobc, line 1561: **[NIT]** I thinik moving the property up under the line that says "the following property" and then moving the `This way, the application can detect deserialization issues right away when using the event routing feature and can take appropriate handling decisions.` would be easier to follow.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/properties/KafkaStreamsConsumerProperties.java
```diff
@@ -71,6 +71,12 @@ public class KafkaStreamsConsumerProperties extends KafkaConsumerProperties {
 	 */
 	private String consumedAs;
 
+	/**
+	 * When event type based routing is enabled, the binder uses the byte[] Serde by default.
+	 * Use this property to override this default behavior by forcing the binder to use the configured or inferred Serde.
+	 */
+	private boolean useConfiguredSerdeWhenRoutingEvents;
```
onobc, line 78: **[NIT]** The name is a bit long, but I don't really have a shorter better alternate recommendation.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/AbstractKafkaStreamsBinderProcessor.java
```diff
@@ -455,13 +454,21 @@ protected Serde<?> getValueSerde(String inboundName, KafkaStreamsConsumerPropert
 			// Branching based on event type match.
 			final KStream<?, ?>[] branch = stream.branch((key, value) -> matched.getAndSet(false));
 			// Deserialize if we have a branch from above.
-			final KStream<?, Object> deserializedKStream = branch[0].mapValues(value -> valueSerde.deserializer().deserialize(
-					topicObject.get(), headersObject.get(), ((Bytes) value).get()));
+			final KStream<?, Object> deserializedKStream = !kafkaStreamsConsumerProperties.isUseConfiguredSerdeWhenRoutingEvents() ?
+				branch[0].mapValues(value -> valueSerde.deserializer().deserialize(
+					topicObject.get(), headersObject.get(), ((Bytes) value).get())) : (KStream<?, Object>) branch[0];
```
onobc, line 459: Why do we not have to deserialize w/ the configured value serde in the case where `isUseConfiguredSerdeWhenRoutingEvents=true`? You are just passing the branch as-is in that case?

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/EventTypeRoutingWithInferredSerdeTests.java
```diff
@@ -0,0 +1,166 @@
...
+		consumer = cf.createConsumer();
+		embeddedKafka.consumeFromEmbeddedTopics(consumer, "foo-2");
+	}
+
+	@AfterAll
+	public static void tearDown() {
+		consumer.close();
+	}
+
+	//For more context, https://github.com/spring-cloud/spring-cloud-stream/issues/2613
```
onobc, line 82: I think having the link in the class comment is sufficient.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/EventTypeRoutingWithInferredSerdeTests.java
```diff
@@ -0,0 +1,166 @@
...
+				template.send(producerRecord4);
+
+				senderProps.put("value.serializer", StringSerializer.class);
+				DefaultKafkaProducerFactory<Integer, String> pf1 = new DefaultKafkaProducerFactory<>(senderProps);
+				KafkaTemplate<Integer, String> template1 = new KafkaTemplate<>(pf1, true);
+				template1.setDefaultTopic("foo-1");
+				final ProducerRecord<Integer, String> producerRecordBar = new ProducerRecord<>("foo-1", 0, 67, "foobar", headers);
+				// This send should throw a deserializer exception by the Kafka Streams processor.
+				// Since the processor is provisioned with an exception handler of logAndContinue, it shouldn't affect anything.
+				template1.send(producerRecordBar);
```
onobc, line 133: W/o the fix, what does this test do at this point? I am guessing its fine, but the consumer will barf when attempting to pull the invalid data from the routed topic?

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/AbstractKafkaStreamsBinderProcessor.java
sobychacko: Yes, that is right. If it is `true`, it already got deserialized with that `Serde`.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/properties/KafkaStreamsConsumerProperties.java
sobychacko: Ya, let me know if you have a better alternative. :-)

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/EventTypeRoutingWithInferredSerdeTests.java
sobychacko: W/o the fix, the deserialization will fail from that branch ^^ and run into a runtime error. With the fix, apps can set the property to `true` and provide an exception handler (such as `logAndContinue` or `sendToDlq`).

sobychacko: Basically, w/o the fix, the test, as it is, will fail at runtime.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/properties/KafkaStreamsConsumerProperties.java
onobc: If there was an `event-routing` sub-properties that contained `supported-types` and `header-key` it would fit in nicely as `use-configured-serde`. However, I think w/ the flat nature of the props currently this is fine.

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/EventTypeRoutingWithInferredSerdeTests.java
onobc: Got it, when that exception handler is not in play it would never reach the next line..

onobc (approved):
LGTM @sobychacko !

PR #2660, opened 2023-03-01
[actuator] It is not possible to pause/resume correctly a binding with multiple destinations

### LINKED ISSUE #2658: It is not possible to pause/resume correctly a binding with multiple destinations by actuator endpoint
**Describe the issue**
When setting up a binding with multiple destinations as shown below:
```
spring:
  cloud:
    function:
      definition: foo
    stream:
      bindings:
        foo-in-0:
          group: my-group
          destination: topic1,topic2,topic3
```
Spring Cloud Stream creates a binding with the same name for each destination correctly.

The problem is that through the actuator endpoint, when making a POST request to stop/resume the binding in question, only the first binding it encounters, does so as indicated in the following [line ](https://github.com/spring-cloud/spring-cloud-stream/blob/af2df8ba0eff2e81c3a8fce358a09a54abb3a974/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingsLifecycleController.java#L181) of the `BindingsLifecycleController.java` class, is paused/resumed.

**To Reproduce**
1. Start the attached demo application ([demoscskafka.zip](https://github.com/spring-cloud/spring-cloud-stream/files/10824773/demoscskafka.zip))
2. Access the endpoint http://localhost:8080/demo-scs-kafka/actuator/bindings and check that there are three bindings with `status: running`.
3. Execute the `foo-in-0` binding pause: `curl --location 'http://localhost:8080/demo-scs-kafka/actuator/bindings/foo-in-0' --header 'Content-Type: application/json' --data '{"state": "PAUSED"}'`
4. Access the endpoint http://localhost:8080/demo-scs-kafka/actuator/bindings and check that only the first binding has the `status: paused`.

**Version of the framework**
Spring Boot 3.0.2
Spring Cloud 2022.0.1

**Expected behavior**
All bindings are paused/resumed.

ferblaca:
resolves #2658

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/BindingsEndpoint.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/BindingsEndpoint.java
@@ -53,7 +53,7 @@ public class BindingsEndpoint {
 	}
 
 	@ReadOperation
-	public Binding<?> queryState(@Selector String name) {
+	public List<Binding<?>> queryState(@Selector String name) {
 		return this.lifecycleController.queryState(name);
 	}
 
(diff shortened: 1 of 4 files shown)


core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/endpoint/BindingsEndpoint.java
```diff
@@ -53,7 +53,7 @@ public List<?> queryStates() {
 	}
 
 	@ReadOperation
-	public Binding<?> queryState(@Selector String name) {
+	public List<Binding<?>> queryState(@Selector String name) {
```
ferblaca, line 56: A noteworthy change is that the queryState operation will now return a list instead of a single element, which modifies the json generated by the response.

olegz: That makes sence

PR #2689, opened 2023-04-03
GH-2655: AOT support user declared binders

### LINKED ISSUE #2655: AOT support for multiple binders
I am trying to run a springboot3 native app with spring-cloud-stream-binder-rabbit dependency. Our application has multiple binders (e.g. usersevent, emailevent). 
The normal application is working fine as expected but in native or aot-mode I am facing following issue.
When I tried with spring boot 3.0.0 and 3.0.2. the application is always taking the binder name as rabbit and connecting to localhost and not connecting to the url or binders defined in my property file
When I tried with spring boot 3.0.3-SNAPSHOT. The build failed with error

Pre-creating binder child context (AOT) for rabbit
Exception in thread "main" java.lang.IllegalArgumentException: Code generation is not supported for bean definitions declaring an instance supplier callback : Root bean: class [org.springframework.integration.config.MessagingAnnotationPostProcessor]; scope=singleton; abstract=false; lazyInit=null; autowireMode=0; dependencyCheck=0; autowireCandidate=true; primary=false; factoryBeanName=null; factoryMethodName=null; initMethodNames=null; destroyMethodNames=null
        at org.springframework.beans.factory.aot.BeanDefinitionMethodGenerator.<init>(BeanDefinitionMethodGenerator.java:82)

Can someone please help in this regards.
Thanks

onobc:
- Adds support for AOT child context generation for user-declared binder configurations.
- Add AOT processing test to verify single and multi-binder cases'

Fixes #2655

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderChildContextInitializer.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderChildContextInitializer.java
@@ -30,6 +30,10 @@ import org.springframework.beans.factory.aot.BeanRegistrationAotContribution;
 import org.springframework.beans.factory.aot.BeanRegistrationAotProcessor;
 import org.springframework.beans.factory.aot.BeanRegistrationCode;
 import org.springframework.beans.factory.support.RegisteredBean;
+import org.springframework.boot.context.properties.bind.Bindable;
+import org.springframework.boot.context.properties.bind.Binder;
+import org.springframework.cloud.stream.config.BindingServiceConfiguration;
+import org.springframework.cloud.stream.config.BindingServiceProperties;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ApplicationContextInitializer;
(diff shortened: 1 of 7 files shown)


core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderChildContextInitializer.java
```diff
@@ -94,6 +104,13 @@ public BeanRegistrationAotContribution processAheadOfTime(RegisteredBean registe
 		return null;
 	}
 
+	private BindingServiceProperties createBindingServiceProperties() {
```
onobc, line 107: At this stage of AOT processing, the env/yml is not available by default and as such the BindingServiceProperties passed into the DefaultBinderFactory have defaults and not the user-declared binders.

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -160,27 +172,37 @@ else if (binders.size() > 1) {
 	}
 
 	private <T> Binder<T, ConsumerProperties, ProducerProperties> doGetBinder(String name, Class<? extends T> bindingTargetType) {
-		if (CollectionUtils.isEmpty(this.binderChildContextInitializers)) {
-			return this.doGetBinderConventional(name, bindingTargetType);
+		// If child initializers - use AOT lookup
+		if (!CollectionUtils.isEmpty(this.binderChildContextInitializers)) {
+			return this.doGetBinderAOT(name, bindingTargetType);
```
onobc, line 177: Moved the AOT flavor or "doGetBinder" into its own method.

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -160,27 +172,37 @@ else if (binders.size() > 1) {
...
+
+	private <T> Binder<T, ConsumerProperties, ProducerProperties> doGetBinderAOT(String name, Class<? extends T> bindingTargetType) {
+		// If neither name nor default given - return single or fail when > 1
+		if (!StringUtils.hasText(name) && !StringUtils.hasText(this.defaultBinder)) {
+			if (this.binderChildContextInitializers.size() == 1) {
 				String configurationName = this.binderChildContextInitializers.keySet().iterator().next();
+				this.logger.debug("No specific name or default given - using single available child initializer '" + configurationName + "'");
 				return this.getBinderInstance(configurationName);
 			}
-			else if (this.defaultBinder != null && this.binderChildContextInitializers.size() > 1) {
```
onobc, line 171: This improves the existing impl and now behaves as follows (inline comments hopefully describe this as well):
- if no name or no default specified then fail if > 1 binders or return single binder if 1 binder available
- if name specified be sure it exists in child initializers (else fail)
- if default specified be sure it exists in child initializers (else fail)
- else fail

sobychacko:
Looks good. Merged upstream.

Issue #2706, opened 2023-04-13
StreamBridge partition calculation mechanism does not update used partition count at runtime when using kafka broker


oliverfuehrertsystems:
We are using StreamBridge to produce records against different topics using 'partition-key-expression' mechanism.
If we change the partition count within kafka broker we have to restart the application to force the application to uses new partition count for partition calculation.

When we use Spring KafkaTemplate instead of StreamBridge it is possible to configure partition count updates during runtime using parameter 'spring.kafka.producer.properties.metadata.max.age.ms'. The updated partition count will be available in application after specified time in that case.

We are using spring-cloud-stream 3.2.7.

sobychacko:
@oliverfuehrertsystems Did you try it on the latest GA versions? `4.0.2` is the latest GA.

oliverfuehrertsystems:
We will test with a newer spring-cloud version (and spring boot) soon, but after having a look into the current code I can see that the necessary method org.springframework.cloud.stream.binder.PartitionHandler#setPartitionCount (see  https://github.com/spring-cloud/spring-cloud-stream/blob/main/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/PartitionHandler.java) is still not called anywhere in the code after initial binding.

oliverfuehrertsystems:
Our producer properties are ok.

It's not so easy to reproduce. We used our loadgenerater software (as producer) to continuosly send  records in a loop to local kafka. We changed partion count in our kafka during that run and recognized that new partitions are never used until we restart our loadgenerator. 

If you look into the code you see that the used PartitionHandler stores the partitionCount in a member variable. This member variable is only updated during initialization phase and never called later during runtime.

Kafka libs may support an metatdata update mechanism (KafkaTemplate spring libraries use e.g. parameter 'spring.kafka.producer.properties.metadata.max.age.ms' for that) where a change of partition count for a topic is received. 

This mechanism should be also implemented within SCS and within that code the method PartitionHandler::setPartitionCount should be called.

oliverfuehrertsystems:
I think about it, but the AbstractMessageChannel architecture is something I have to dive in deep ;)

oliverfuehrer:
I created a fix implementing a new KafkaPartitionHandler (mostly a copy of existing PartitionHandler), which is called within KafkaMessageChannelBinder.ProducerConfigurationMessageHandler and gets an updated partitionCount from KafkaTemplate.

[...]

PR #2720, opened 2023-05-02
Update partition count changes dynamically during runtime. 

oliverfuehrer:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2706


binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1516,6 +1524,19 @@ public void addProducerConfigCustomizer(ProducerConfigCustomizer producerConfigC
 				setSendTimeoutExpression(producerProperties.getExtension().getSendTimeoutExpression());
 			}
 			this.producerFactory = producerFactory;
+
+			/*
+			 	Activate kafkaPartitionHandler if 'getPartitionKeyExpression' was used to recalculate/override
+			 	partition with current partition size from kafkaTemplate just before message is send.
+			 	PartitionKeyExpression 'payload' is not supported here,
+			 	because of OutboundContentTypeConvertingInterceptor was called before.
```
sobychacko, line 1532: This comment needs to be clarified. Could you tell me why you are guarding the `payload` key expression? What will happen if we specify `payload` as the `partition-key-expression`? Not sure how the `OutboundContentTypeConvertingInterceptor` is related here.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaPartitionHandler.java
```diff
@@ -0,0 +1,197 @@
...
+import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+import org.springframework.util.ReflectionUtils;
+import org.springframework.util.StringUtils;
+
+
+/**
+ * Based(nearly copy) on org.springframework.cloud.stream.binder.PartitionHandler but with partitionCount as parameter.
+ */
+public class KafkaPartitionHandler {
```
sobychacko, line 40: `PartitionHandler` is a public class that is open for extension. Can you extend from `PartitionHandler` and only add the `determinePartition` method with the extra `partitionCount` parameter? That way, we can avoid duplicating all this code in `KafkaPartitionHandler`.

sobychacko (changes requested):

sobychacko:
@oliverfuehrer I started with some comments. Please take a look at them inline above. We don't need a new sample app for this enhancement. I suggest removing it altogether and then adding a test instead in `KafkaBinderTests.`  In your test, you may want to change the topic partition dynamically and verify that the new partition count is applied since that is the crux of the requirement with these changes. Also, I suggest adding a few sentences around these enhancements as part of the Kafka binder docs. These will completely remove the need for a new sample app. What do you think?

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
oliverfuehrer: I had to add a KafkaPartitionHandler within handleMessage because of architectual reasons (see my next/other reply). The disadvantage of this is that the payload of a message is already encoded by OutboundContentTypeConvertingInterceptor before and can't be read anymore here. A producer configuration with payload as PartitionKeyExpression can't be handled by my fix therefore.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaPartitionHandler.java
oliverfuehrer: My first try was to extend existing PartitionHandler, but this doesnt't work because of SCS architecture. The necessary KafkaTemplate class (which contains updated partition count) is stored within 'KafkaMessageChannelBinder.ProducerConfigurationMessageHandler' and this class cannot be accessed from existing PartitionHandler. You get a cyclic maven dependency error if you try access it from spring-cloud-stream-core. I see no other possibilty as to add the kafka partition handling code directly to the othe kafka code within 'spring-cloud-stream-binder-kafka'. Maybe I missed an other soloution!?

oliverfuehrer:
The partition count update is hidden within Spring KafkaTemplate which registeres to kafka libraries for metadata updates. As far as I understand, these updates will send/requested usiing kafka protocol from broker. I have no Idea how to write a Unit-Test which somehow simulates this process using a real kafka. This is the reason why I wrote a sample app to manually test the whole SCS->Spring->Kafka integration.

oliverfuehrer:
I think ee need a general discussion how to change the SCS library to support kafka internal information (like an updated partition count) during SCS partition handling.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaPartitionHandler.java
sobychacko: I wasn't proposing you modify `PartitionHandler` in `spring-cloud-stream-core.` Why can't you keep the `KafkaPartitionHandler` in the Kafka binder (as it is right now) and then extend it from `PartitionHandler`. ? You then call the super constructor from the new extension. Later on, in the `handleMessage()` call, you call `determinePartition(message, partitionCount)`, which is only available in `KafkaPartitionHandler.` I don't see how this may introduce a cycle.

sobychacko (changes requested):

sobychacko:
How is the sample application verifying that the new partition count is taken into consideration? I still don't think we need a full-blown sample app for this enhancement in the repository. This can be somehow verified using some automated testing + docs.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaPartitionHandler.java
oliverfuehrer: Ok I got it. My code was not meant to be part of a pull request from the beginning. I will rewrite the KafkaPartitionHandler and will remove the sample application (which was more for my own purpose). Still no idea how to manually test without using my sample program.

sobychacko:
@oliverfuehrer Looks much cleaner now. A couple of more suggestions.
1. Can you add your name as the @author on the classes you modified or added?
2.  I think you can add a small unit test for the `KafkaPartitonHandler` by forcing an output binding through the Kafka binder. Look at `KafkaBinderTests` or several other tests in the binder if you need some templates. Then use some mock/spy mechanism on the `KafkaPartitionHandler` object to make sure that the correct `determinePartition` method is called as expected. What do you think about that? We don't need to test the complex scenario where you dynamically change the topic partitions, but as long as you exercise the new code somehow, I am ok with that. 
3. Could you please add a sentence or two about this enhancement in the docs?

oliverfuehrer:
I simplified the code (use default PartitionHandler instead of creating new class) and added simple test and some comments. No idea where to document the new behavior - especially the not supported 'payload' expression.

oliverfuehrer:
Added some documentation - please review. Thinking about a configuration parameter / feature flag to enable/disable my new calculation - what do you think?

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1548,6 +1575,32 @@ public boolean isRunning() {
 			return this.running;
 		}
 
+		@Override
+		public void handleMessage(Message<?> message) {
+
+			// if we use our own partition handler to update partition count we recalculate partition
+			if (kafkaPartitionHandler != null) {
+				kafkaPartitionHandler.setPartitionCount(getKafkaTemplate().partitionsFor(this.topic).size());
```
sobychacko, line 1583: This is nice. I wasn't thinking yesterday that we could do this. Much simpler conceptually now.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
```diff
@@ -1548,6 +1575,32 @@ public boolean isRunning() {
...
+				super.handleMessage(newMessage);
+			}
+			else {
+				super.handleMessage(message);
+			}
+		}
+
+		PartitionHandler getKafkaPartitionHandler() {
+			return kafkaPartitionHandler;
+		}
```
sobychacko, line 1599: Do we need this accessor and setter below? I don't see where they are called from.

oliverfuehrer: I invented this getter for the test I wrote.

sobychacko: Ok, I see that now. I wonder if we could find a different way to do those assertions. We don't typically add production code if it is only required by the tests, as that seems like a slight code smell. It is ok to use `ReflctionUtils` in the test to access the private fields.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
```diff
@@ -4005,6 +4011,36 @@ void testObservationEnabledThroughProducerMessageHandlerCustomizer() throws Exce
...
+			createProducerBindingProperties(createProducerProperties()));
+
+		Binding<MessageChannel> producerBinding = binder.bindProducer("foo",
+			outputChannel, properties);
+
+		KafkaMessageChannelBinder.ProducerConfigurationMessageHandler kafkaProducerMessageHandler =
+			(KafkaMessageChannelBinder.ProducerConfigurationMessageHandler) TestUtils.getPropertyValue(
+				producerBinding, "lifecycle", KafkaProducerMessageHandler.class);
+
+		assertThat(kafkaProducerMessageHandler.getKafkaPartitionHandler()).isNotNull();
```
sobychacko, line 4032: Do we need this assertion since if the partition handler is really null, the subsequent verifications on the spy will fail?

docs/src/main/asciidoc/kafka/kafka_partitions.adoc
```diff
@@ -65,6 +65,8 @@ You can override this default by using the `partitionSelectorExpression` or `par
 Since partitions are natively handled by Kafka, no special configuration is needed on the consumer side.
 Kafka allocates partitions across the instances.
 
+NOTE: The partitionCount for a kafka topic may change during runtime (e.g. due to an adminstration task). The calculated  partitions will be different after that (e.g. new partitions will be used). Since SCS 4.0.3 runtime changes of partition count will be supported. Due to some limitations this is not possible if the used 'partition-key-expression' directly uses the 'payload' of the message (see also https://github.com/spring-cloud/spring-cloud-stream/issues/2706).
```
sobychacko, line 68: Instead of `SCS,` say it fully qualified - `Spring Cloud Stream.`  
```
Since version 4.0.3 of Spring Cloud Stream, the Kafka binder supports dynamic partition count changes during runtime without requiring a restart of the application.
```

sobychacko:
> 

This is definitely a good idea. This way, we can only enable this feature, if the application really demands it. For example, some applications may want to continue without checking if the partitions have changed or not in order to avoid a round trip through the Kafka admin during message processing.

sobychacko:
@oliverfuehrer Changes look great. I added some more comments. Please take a look. 

Could you also rebase your PR branch against the latest `main` branch? I see that you brought in some changes from upstream mixed with your commits.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
oliverfuehrer: I got your point. Will try to use Reflection instead.

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/KafkaBinderTests.java
oliverfuehrer: It could be null in some usecases. Is there a rule or is it your personal preference?

sobychacko: There is no rule. If this is possible to go null, sure, you can do the assertion.

oliverfuehrer:
> 

I had problems rebasing my branch against main (maybe related to the GitHub-Outage today when I tried to push), but I merged changes from synchronized forked main into my branch. I will add a property to disable this message handling by default and will adapt the documentation also.

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
```diff
@@ -81,6 +81,11 @@ public class ProducerProperties {
 
 	private PollerProperties poller;
 
+	/**
+	 * A flag which enables/disables kafka partition count updates during runtime.
+	 */
+	private boolean kafkaPartitionCountUpdateEnabled = false;
```
sobychacko, line 87: Actually, since this is a binder-specific feature, I would suggest making this part of the extended binding properties in `KafkaProducerProperties.` We don't want to leak this field to the core `ProducerProperties.` Also, how about `dynamicPartitionUpdatesEnabled` for the property name? Or something similar? Once the changes are in the `KafkaProducerProperties`, we don't need the prefix `kafka`.

docs/src/main/asciidoc/kafka/kafka_partitions.adoc
```diff
@@ -65,7 +65,8 @@ You can override this default by using the `partitionSelectorExpression` or `par
 Since partitions are natively handled by Kafka, no special configuration is needed on the consumer side.
 Kafka allocates partitions across the instances.
 
-NOTE: The partitionCount for a kafka topic may change during runtime (e.g. due to an adminstration task). The calculated  partitions will be different after that (e.g. new partitions will be used). Since 4.0.3 of Spring Cloud Stream runtime changes of partition count will be supported. Due to some limitations this is not possible if the used 'partition-key-expression' directly uses the 'payload' of the message (see also https://github.com/spring-cloud/spring-cloud-stream/issues/2706).
+NOTE: The partitionCount for a kafka topic may change during runtime (e.g. due to an adminstration task). The calculated  partitions will be different after that (e.g. new partitions will be used). Since 4.0.3 of Spring Cloud Stream runtime changes of partition count will be supported. See also parameter 'spring.metadata.max.age.ms'.
```
sobychacko, line 68: Can you put one sentence per line? That is usually the `asciidoc` standard we follow.

docs/src/main/asciidoc/kafka/kafka_partitions.adoc
```diff
@@ -65,7 +65,8 @@ You can override this default by using the `partitionSelectorExpression` or `par
...
+Due to some limitations this is not possible if the used 'partition-key-expression' directly uses the 'payload' of the message (see also https://github.com/spring-cloud/spring-cloud-stream/issues/2706). The update behaviour is disabled by default and can be enabled using configuration parameter 'producer.kafkaPartitionCountUpdateEnabled=true'.
```
sobychacko, line 69: The new property needs to be moved to the proper configuration section in the docs.

sobychacko (changes requested):

sobychacko:
>I had problems rebasing my branch against main (maybe related to the GitHub-Outage today when I tried to push), but I merged changes from synchronized forked main into my branch. 

I will rebase the branch against `main` before I merge.

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
oliverfuehrer: I'm not sure if it is in general really a Kafka related feature, but there will be only my Kafka implementation for now. KafkaProducerProperties are mostly related to real Kafka properties from DefaultKafkaProducerFactory, which is not the case here. I rename it to  'dynamicPartitionUpdatesEnabled' and it is currently only implemented for KafkaBinders. I'm not sure about other technologies and if a partitionCount could also be changed at runtime. This should a topic for a architecture discussion.

docs/src/main/asciidoc/kafka/kafka_partitions.adoc
oliverfuehrer: Ok. Will change that.

oliverfuehrer:
I improved the test a lot.

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
sobychacko: Ok, that is a fair point. Since it is adding to the core producer properties, let me discuss this with our internal engineering team. I will get back to you asap.

sobychacko:
> I improved the test a lot. Starts to make fun if you get used to it ;)

That is really nice.

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/ProducerProperties.java
oliverfuehrer: Will do that.

sobychacko:
@oliverfuehrer I think we are getting closer to merging the PR. I tried to rebase your branch locally but ran into issues. Could you try a `git rebase` against the latest `main` branch? If rebase doesn't work, now that we are clear on the final shape of the changes, you can create a new commit in a separate branch and then issue a new PR. If we go with that approach, we can refer to this PR from there to preserve all the comment history.

oliverfuehrer:
See new pull request: https://github.com/spring-cloud/spring-cloud-stream/pull/2725

sobychacko:
Closing this PR in favor of https://github.com/spring-cloud/spring-cloud-stream/pull/2725 since this PR has some rebasing issues.

Issue #2729, opened 2023-05-10
Provide a Mechanism to Receive SenderResults


garyrussell:
The reactor-kafka binder currently has an undocumented feature of setting the `sendResult` header, containing a `Mono<RecordMetadata>`.

Change this to return `Mono<SenderResult>` instead - the record metadata is not very useful without the correlation metadata in the `SenderResult`.

https://stackoverflow.com/questions/76221809/how-to-get-correlation-id-in-spring-cloud-stream/76221941#76221941

kitkars:
I was about to open this issue...You did already :)

To the original issue, The header approach sounds good. I can understand. But will it be ok for an extremely high throughput application?

```
@Bean
Supplier<Flux<Message<Integer>>> someProducer(){
	return () -> Flux.range(1, 10_000_000).map(i -> convertToMessage(i));
}
```
So, I was thinking, if we could send the results to a separate consumer as shown here?

```
@Bean
Consumer<Flux<SenderResult>> resultConsumer(){
	return flux -> flux.doOnNext(...).subscribe();
}
```

```
spring:
  cloud:
    function:
      definition: someProducer;
    stream:
      default-binder: reactorKafka
      bindings:
        producer-out-0:
          destination: order-events
          sendResultsTo: resultConsumer // function bean which accepts the result
```

sorry if it sounds ridiculous :)

kitkars:
That would be great @garyrussell Thank you so much.

PR #2730, opened 2023-05-11
GH-2729: Reactor Kafka Binder SenderResult Support

garyrussell:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2729

Allow configuration of a `FluxMessageChannel` to receive `SenderResult`s. Add `SenderResultMessageHandler` to consume from that channel.

Remove undocumented `sendResult` header, which has no value without the sender result correlation metadata.

**Do not merge yet; initial review only; needs docs and a full integration test**


binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
```diff
@@ -169,7 +170,13 @@ protected MessageHandler createProducerMessageHandler(ProducerDestination destin
 				SenderOptions.create(configs));
 		// TODO bean for converter; MCB doesn't use one on the producer side.
 		RecordMessageConverter converter = new MessagingMessageConverter();
-		return new ReactorMessageHandler(opts, converter, destination.getName());
+		AbstractApplicationContext applicationContext = getApplicationContext();
+		FluxMessageChannel resultChannel = null;
+		String channelName = producerProperties.getExtension().getRecordMetadataChannel();
+		if (channelName != null && applicationContext.containsBean(channelName)) {
+			resultChannel = applicationContext.getBean(channelName, FluxMessageChannel.class);
```
artembilan, line 177: The `KafkaProducerProperties.recordMetadataChannel` must reflect this requirement. Kinda just enough to say `The bean name of a FluxMessageChannel...`

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
```diff
@@ -316,35 +323,40 @@ private static class ReactorMessageHandler extends AbstractMessageHandler implem
...
-						(ProducerRecord<Object, Object>) converter.fromMessage(message, topic), uuid);
-				Flux<SenderResult<UUID>> result = sender.send(Flux.just(sr));
-				result.subscribe(res -> sink.emitValue(res.recordMetadata(), null));
-			}
-			else {
-				sink.emitError(new IllegalStateException("Handler is not running"), null);
+				SenderRecord<Object, Object, Object> sr = SenderRecord.create(
+						(ProducerRecord<Object, Object>) converter.fromMessage(message, topic), correlation);
+				Flux<SenderResult<Object>> result = sender.send(Flux.just(sr));
+				result.subscribe(res -> {
```
artembilan, line 353: This is better to have like this:
```
				if (this.results != null) {
					this.results.subscribeTo(
							result.map(res ->
									MessageBuilder.withPayload(res)
											.copyHeaders(message.getHeaders())
											.build()));
				}
				else {
					result.subscribe();
				}
```
So, the `FluxMessageChannel` would subscribe to that `Flux` in its own manner.

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/SenderResultMessageHandler.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Message handler for {@link SenderResult}s.
+ *
+ * @param <T> the correlation metadata type.
+ *
+ * @author Gary Russell
+ * @since 4.0.3
+ *
+ */
+public abstract class SenderResultMessageHandler<T> extends AbstractReactiveMessageHandler {
```
artembilan, line 34: Not sure why do we need this extra adapter. Why just plain `MessageHandler` is not enough?
Or in most cases I guess that is going to be like this:
```
@ServiceActivator(inputChannel = "myRecordMetadataChannel")
void(SenderResult<MyData> result) {
    ...
}
```

artembilan (changes requested):

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/SenderResultMessageHandler.java
garyrussell: My thought was this would be convenient to use in the DSL.

```java
IntegrationFlows.from(MessageChannels.flux)
    .handle(new SenderResultMessageHandler<Integer>() {...})
    .get();
```

garyrussell: That would work, but it would have to be `.<SenderResult<Integer>>handle(...)`.

Ok; will remove it.

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
garyrussell: But wouldn't that cause a new subscription for each send? How would it be disposed of?

artembilan: There is nothing wrong with new subscription for each send.
Essentially that is exactly what you do here with this `result.subscribe()`.
Since this is limited `Flux` from `sender.send()` result, then it disposes itself after completion.

However I see the problem in the `FluxMessageChannel`: we always pile `Disposable`s into a `Disposable.Composite` and clean it up only on `FluxMessageChannel.destroy()`. 
So, this is a bug leading to memory leak.

Either way: do we talk about the same matter or you mean something else about them?

garyrussell: Yes, that's what I meant. But what is wrong with using send here? It works as expected by emitting via the sink.

artembilan: Yeah... Sure! Never mind. Technically it is really the same.
Unless we don't have a memory leak with your current solution 😉

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderIntegrationTests.java
```diff
@@ -107,34 +109,30 @@ void endToEndReactorKafkaBinder(boolean excludeKafkaAutoConfig) {
...
-			}
-			finally {
-				pf.destroy();
-			}
+			StreamBridge streamBridge = context.getBean(StreamBridge.class);
+			streamBridge.send("words1", MessageBuilder.withPayload("foobar").setCorrelationId(42).build());
+			streamBridge.send("words2", MessageBuilder.withPayload("BAZQUX").setCorrelationId(43).build());
+
+			assertThat(KafkaTestUtils.getSingleRecord(consumer1, "uppercased-words"))
+				.isNotNull()
```
onobc, line 124: **[INFO-SHARING]** The `isNotNull()` in many cases is redundant as almost all of the other AssertJ assertions all do a `isNotNull` before anything else. You may already know this and prefer the `isNotNull` here to be explicit. Just sharing in case you did not know about this.

garyrussell: Thanks for the tip, I was not aware, but I do think I prefer the explicit method call.

garyrussell:
Added integration test and docs; removed draft status.

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/test/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinderIntegrationTests.java
onobc: The other interesting AssertJ redundant check (that is more valuable and does not provide much value being there explicitly) is the context assert when used in say `ApplicationContextRunner`. 
```java
assertThat(context)
  .hasNotFailed()
  .hasSingleBean("foo").hasBlah()....
```
It is somewhat expected that the context has not failed and when you see it 100 times in 100 tests in a single file it is a bit redundant. 

Anyways, in case you run into that one.

binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/SenderResultMessageHandler.java
artembilan: Any further motivation to keep this?

artembilan (changes requested):

artembilan (approved):

PR #2731, opened 2023-05-13
reactor kafka sender result consumer draft

kitkars:
Hi @garyrussell @artembilan ,
I just wanted to propose this for the sending the SenderResult.

It is just a draft to give an idea. I tested locally for my application and it seems to work fine.
I was thinking we could add a `Consumer<SenderResult<T>>` as part of **_SenderOptionsCustomizer_**.

```
default Consumer<SenderResult<UUID>> senderResultConsumer() {
	return sr -> {
               // default NO-OP
	};
}

```

Users like us can use it like this.

```
private final Sinks.Many<SenderResult<UUID>> sink = Sinks.many().unicast().onBackpressureBuffer();

@Bean
public SenderOptionsCustomizer<String, String> customizer(){
    return new SenderOptionsCustomizer<String, String>() {
        
        @Override
        public SenderOptions<String, String> apply(String s, SenderOptions<String, String> stringStringSenderOptions) {
            return stringStringSenderOptions;
        }

        @Override
        public Consumer<SenderResult<UUID>> senderResultConsumer() {
            return sink::tryEmitNext;
        }
    };
}
```
and somewhere else
```
public void printSentMessageIds(){
        sink.asFlux()
                .doOnNext(id -> print(id))
                .subscribe();
}
```

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-reactive/src/main/java/org/springframework/cloud/stream/binder/reactorkafka/ReactorKafkaBinder.java
@@ -21,16 +21,14 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 import java.util.regex.Pattern;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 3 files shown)


artembilan:
@garyrussell has already issued some PR with possible fix: https://github.com/spring-cloud/spring-cloud-stream/pull/2730.

Technically and logically your ideas are similar: Reactor Sink or `MessageChannel` - the outcome for handling those results are the same.

we probably need voting to determine what direction is better to go…

kitkars:
I really apologize. I did not check Gary's PR. I just checked after your comment. I really like Gary's new PR. 👍 
Thanks @artembilan @garyrussell  for your work and helping us :)

PR #2735, opened 2023-05-15
Kafka binder partitioning doc improvements

sobychacko:


docs/src/main/asciidoc/kafka/kafka_partitions.adoc
```diff
@@ -54,7 +54,13 @@ spring:
             partition-count: 12
 ----
 
-IMPORTANT: The topic must be provisioned to have enough partitions to achieve the desired concurrency for all consumer groups.
+IMPORTANT: It is important to keep in mind that since Apache Kafka supports partitioning natively, there is no need to rely on binder partitioning as described above unless you are using custom partition keys as in the example or an expression that involves the payload itself.
```
garyrussell, line 57:
```suggestion
IMPORTANT: It is important to keep in mind that, since Apache Kafka supports partitioning natively, there is no need to rely on binder partitioning as described above unless you are using custom partition keys as in the example or an expression that involves the payload itself.
```

docs/src/main/asciidoc/kafka/kafka_partitions.adoc
```diff
@@ -54,7 +54,13 @@ spring:
...
+The binder-provided partitioning selection is otherwise intended for middleware technologies that do not support native partitioning.
+Note that we are using a custom key called `partitionKey` in the above example, that will be the determining factor for the partition, thus in this case it is appropriate to use binder partitioning.
+When using native Kafka partitioning, i.e, when you do not provide the `partition-key-expression`, then Apache Kafka will select a partition, which by default will be the hash value of the record key over the available number of partitions.
```
garyrussell, line 60:
```suggestion
When using native Kafka partitioning, i.e, when you do not provide the `partition-key-expression`, then Apache Kafka will select a partition, which by default will be the hash value of the record key over the available number of partitions.
To add a key to an outbound record, set the `KafkaHeaders.KEY` header to the desired key value in a spring-messaging `Message<?>`.
```

docs/src/main/asciidoc/kafka/kafka_partitions.adoc
```diff
@@ -54,7 +54,13 @@ spring:
...
+When no record key is provided, Apache Kafka will choose a partition based on a round-robin strategy.
```
garyrussell, line 61:
```suggestion
By default, when no record key is provided, Apache Kafka will choose a partition based on the logic described in the https://kafka.apache.org/documentation/#producerconfigs_partitioner.class[Apache Kafka Documentation].
```

garyrussell (changes requested):

garyrussell (approved):

garyrussell:
@sobychacko FYI, when committing ` ```suggestion`s to a PR, you can batch them into a single commit.

Not a big deal when there's only a few but when there are lots...

PR #2740, opened 2023-05-22
Fix FunctionUnit derivation

jamfor352:
Given the following definition:

```
spring:
  cloud:
    function:
      definition: >
        function1;
        function2;
        function3
```

the Kafka Stream binder does *not* work, as after a split, each function will contain a leading whitespace, so when it later on goes to do a comparison to declared function method names via `kafkaStreamsMethodNames.contains`, it doesn't actually create the bindings when it should:

```
if (kafkaStreamsMethodNames.contains(functionUnit)) {
						ResolvableType[] resolvableTypes = new ResolvableType[]{getResolvableTypes().get(functionUnit)};
						RootBeanDefinition rootBeanDefinition = new RootBeanDefinition(
								KafkaStreamsBindableProxyFactory.class);
						registerKakaStreamsProxyFactory(registry, functionUnit, resolvableTypes, rootBeanDefinition);
					}
```

The current workaround is to place everything on one line, which is ugly if you have lots of functions.

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderUtils.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderUtils.java
@@ -66,6 +66,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Soby Chacko
  * @author Gary Russell
+ * @author James Forward
  */
 public final class KafkaStreamsBinderUtils {
 
@@ -84,6 +85,14 @@ public final class KafkaStreamsBinderUtils {
 	 */
 	public static Optional<Method> findMethodWithName(String key, Method[] methods) {
 		return Arrays.stream(methods).filter(m -> m.getName().equals(key)).findFirst();
+	}
+
+	public static String[] deriveFunctionUnits(String definition) {
+		if(!StringUtils.hasText(definition)) {
+			return new String[]{};
+		}
+		final String[] rawSplitDefinition = definition.split(";");
+		return Arrays.stream(rawSplitDefinition).map(String::trim).toArray(String[]::new);
 	}
 
 	static void prepareConsumerBinding(String name, String group,
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/KafkaStreamsFunctionBeanPostProcessor.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/KafkaStreamsFunctionBeanPostProcessor.java
@@ -52,10 +52,11 @@ import org.springframework.cloud.stream.binder.kafka.streams.KafkaStreamsBinderU
 import org.springframework.cloud.stream.function.StreamFunctionProperties;
 import org.springframework.core.ResolvableType;
 import org.springframework.util.ClassUtils;
-import org.springframework.util.StringUtils;
 
 /**
  * @author Soby Chacko
+ * @author James Forward
+ *
  * @since 2.2.0
  */
 public class KafkaStreamsFunctionBeanPostProcessor implements InitializingBean, BeanFactoryAware {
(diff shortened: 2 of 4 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionUnitDeriver.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class FunctionUnitDeriver {
+	static String[] getFunctionUnits(String definition) {
+		if(!StringUtils.hasText(definition)) {
+			return new String[]{};
+		}
+		final String[] rawSplitDefinition = definition.split(";");
+		return Arrays.stream(rawSplitDefinition).map(String::trim).toArray(String[]::new);
+	}
+}
```
sobychacko, line 15: We probably don't need a new class since this functionality is trivial. How about adding this as a new static utility method in `KafkaStreamsBinderUtils`?

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionUnitDeriverTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.util.StringUtils;
+
+import java.util.Arrays;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.jupiter.api.Assertions.assertAll;
+
+class FunctionUnitDeriverTest {
+	@Test
+	void testFunctionNameDerivation() {
```
sobychacko, line 13: Maybe a bit more descriptive, as the scenario under test is a function definition with explicit spaces in between?

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionUnitDeriver.java
```diff
@@ -0,0 +1,15 @@
+package org.springframework.cloud.stream.binder.kafka.streams.function;
+
+import org.springframework.util.StringUtils;
+
+import java.util.Arrays;
+
+public class FunctionUnitDeriver {
+	static String[] getFunctionUnits(String definition) {
```
sobychacko, line 8: How about `deriveFunctionUnits` as `get...` has a JavaBean feel?

sobychacko (changes requested):

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionUnitDeriver.java
jamfor352: KafkaStreamsBinderUtils and its methods are all package protected, so breaking the encapsulation might be a problem - Unless we're okay with making it public.

(I still don't get why there isn't a "package and subpackage" visibility in Java, it would make things so much nicer 😭  )

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/test/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionUnitDeriverTest.java
jamfor352: > Maybe a bit more descriptive, as the scenario under test is a function definition with explicit spaces in between?

Some of the functions have varied length spaces between them, two of them have none - the intention was just to show the fact they are fully trimmed regardless was always consistent, but I'll update the name of the test

binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/function/FunctionUnitDeriver.java
sobychacko: We actually made `KafkaStreamsBinderUtils` a public utility class for some other reasons.

jamfor352: Thanks @sobychacko , my remote was based on the comment before this one - I've rebased now

jamfor352:
Hi @sobychacko , I've made the requested changes. Does this look good, now?

sobychacko:
Thanks, @jamfor352. Looks good. Polished the commits (some checkstyle issues, assert4j assertions needed, etc.). Merged upstream via https://github.com/spring-cloud/spring-cloud-stream/commit/14a444885f004b4c56d532b2b74fce962ce3bc78. You are welcome to contribute more PRs with bug fixes, enhancements, etc. Thanks again!

jamfor352:
@sobychacko Thanks for this!

Down the line, what would you think of having a @ConfigurationProperties class for `spring.cloud.function`, with `definition` being a List of Strings? Might be a cleaner approach than having to trim/have a delimiter in the first place. So someone could do something like:

```
spring:
  cloud:
    function:
      definition:
        - function1
        - function2
        - function3|function4
        - function5
```

etc

sobychacko:
@jamfor352 Could you please open a new issue with these details? It could certainly be up for discussion. Having it inside a closed PR makes it difficult to track.

PR #2748, opened 2023-06-01
Fix for function auto-detection when input/output-bindings are defined

Shapoval1van:
This issue was discovered in the case of explicit binding creation but in ApplicationContext already exist some functional beans that should not be treated as binding function.
Already present [fix](https://github.com/spring-cloud/spring-cloud-stream/commit/2ab71f2f320bad3ce7af6e059d205f8230ba1d08) for case when I have single input/output-bindings but it will not work when I have several explicit binding.


olegz:
Great fix, thank you!

PR #2757, opened 2023-06-15
Initial migration of Pulsar binder

sobychacko:
- Moving Pulsar binder from the old 4.1.x branch

### DIFF
--- a/binders/pulsar-binder/pom.xml
+++ b/binders/pulsar-binder/pom.xml
@@ -1,0 +1,170 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<artifactId>spring-cloud-stream-binder-pulsar-parent</artifactId>
+	<version>4.1.0-SNAPSHOT</version>
+	<packaging>pom</packaging>
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-stream-binders</artifactId>
+		<version>4.1.0-SNAPSHOT</version>
+	</parent>
+
+	<properties>
+		<java.version>17</java.version>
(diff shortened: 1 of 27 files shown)


binders/pulsar-binder/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+		<maven-checkstyle-plugin.includeTestSourceDirectory>true</maven-checkstyle-plugin.includeTestSourceDirectory>
+	</properties>
+	<modules>
+		<module>spring-cloud-stream-binder-pulsar</module>
+	</modules>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.junit.vintage</groupId>
+			<artifactId>junit-vintage-engine</artifactId>
```
onobc, line 26: Is this needed?

binders/pulsar-binder/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
+	</profiles>
+	<repositories>
+		<repository>
+			<id>spring-snapshots</id>
+			<name>Spring Snapshots</name>
+			<url>https://repo.spring.io/libs-snapshot-local</url>
```
onobc, line 118: Unrelated to this particular PR but these should be pointing to `/snapshot, /milestone` and `/release` replaced w/ maven central. Otherwise users would be required to auth to Artifactory to build.

binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/main/java/org/springframework/cloud/stream/binder/pulsar/PulsarBinderHeaderMapper.java
```diff
@@ -0,0 +1,70 @@
...
+		pulsarHeaders.remove(MessageHeaders.TIMESTAMP);
+		pulsarHeaders.remove(IntegrationMessageHeaderAccessor.DELIVERY_ATTEMPT);
+		pulsarHeaders.remove(BinderHeaders.NATIVE_HEADERS_PRESENT);
+		return pulsarHeaders;
+	}
+
+	@Override
+	public MessageHeaders toSpringHeaders(Message<?> pulsarMessage) {
+		var springHeaders = this.delegate.toSpringHeaders(pulsarMessage);
+		if (!springHeaders.isEmpty()) {
```
olegz, line 61: Consider using `CollectionUtils.isEmpty(Map)` which also checks for null value and it is more idiomatic to Spring way of doing these things

PR #2767, opened 2023-07-06
Schema registry doc corrections

vicziani:
I found some outdated information in the Schema Registry Server documentation, which made it difficult to try out. I will try to correct it.


sobychacko:
Thanks for the PR! Merged upstream.

Issue #2626, opened 2023-01-19
Explicit binding definition must NOT modify the name of the binding


olegz:
During the implicit binding definitions we derive binding names form the function name.
For example
```
@Bean
Function<String, String> uppercase() {..}
```
... will result in two bindings: `uppercase-in-0` and `uppercase-out-0`. We use the name of the function with suffix `in/out` to distinguish between the input or output binding and add an index of the input or output for cases where function take multiple arguments.

However, when explicitly defining bindings - bindings that are not backed by a function, we still treat it as an implicit binding and modify its name. 
For example:
```
spring.cloud.stream.output-bindings=baz
```
... will result in `baz-out-0`. 

This is wrong and confusing for several reasons.
First, it's pretty evident from the property name that this is an output binding and framework will treat it as such, so there is no need to add `in` or `out` suffix. The same for the index since in the context of explicit bindings you can simply define as many bindings as you want using names that you want. 
Also, given that users often map binding names to a more meaningful names for cases with explicit bindings where users can already have a meaningful binding name they are forced to do such mapping again resulting in something like this which adds to the confusion:
```
spring.cloud.stream.output-bindings=myMeaningfulBindingName
spring.cloud.stream.function.bindings.myMeaningfulBindingName-out-0=myMeaningfulBindingName
```

Basically we should treat explicit binding names as they are - unchanged, This also ensures that it follows the same functionality we had with annotation-based programming model when defining explicit binings.

Shapoval1van:
@olegz so this [docs](https://docs.spring.io/spring-cloud-stream/docs/current/reference/html/spring-cloud-stream.html#_sending_arbitrary_data_to_an_output_e_g_foreign_event_driven_sources) already not actuall

RomainWilbert:
I think it should it be : 
`spring.cloud.stream.function.bindings.myMeaningfulBindingName=myMeaningfulBindingName-out-0`
instead of
`spring.cloud.stream.function.bindings.myMeaningfulBindingName-out-0=myMeaningfulBindingName`

jonenst:
Hi,
can you confirm that we have to use either

```
spring.cloud.stream.bindings.MYBINDING-out-0.destination=MYQUEUE
spring.cloud.stream.output-bindings: MYBINDING-out-0
# in code: streambridge.send("MYBINDING-out-0", message)
```
Or

```
spring.cloud.stream.bindings.MYBINDING.destination=MYQUEUE
[...]

PR #2768, opened 2023-07-06
Fix documentation related to explicit binding name change

Shapoval1van:
After https://github.com/spring-cloud/spring-cloud-stream/issues/2626 some documtations ware not updated according to changes


Shapoval1van:
@olegz

sobychacko:
The changes make sense. Merged upstream and back-ported to `4.0.x`.

PR #2778, opened 2023-07-18
Added example to identify if kafka error is in brokers or in topic

Claudio-code:
At the company where I work we had to implement a similar solution to set up monitoring alerts and help better identify where the root cause of the problem is, I think this example can help other teams as well.

### DIFF
--- a/docs/src/main/asciidoc/kafka/kafka_overview.adoc
+++ b/docs/src/main/asciidoc/kafka/kafka_overview.adoc
@@ -1017,12 +1017,13 @@ public KafkaBinderHealth kafkaBinderHealthIndicator() {
 }
 ```
 
-[[check-if-the-connection-to-the-brokers-is-in-error-or-in-topic]]
-=== Check if the connection to the brokers is in error or in topic
+[[custom-kafka-binder-health-indicator-example]]
+=== Custom kafka Binder Health Indicator Example
 
-To help identify if the cluster brokers are down or if it is just a specific topic that is having a connection problem,
-in complex scenarios where the service connects with several topics this verification helps to identify the severity of the problem.
+Here is the pseudo-code for writing a custom Kafka binder HealthIndicator. In this example, we try to override the binder provided Kafka HealthIndicator by specifically checking first for cluster connectivity and then followed by topic-related issues.
 
+
+1. First we need create a custom implementation of the `KafkaBinderHealth` interface.
 
 ```
 public class KafkaBinderHealthImplementation implements KafkaBinderHealth {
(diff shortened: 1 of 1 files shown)


Claudio-code:
I used this post as a basis for this code, it would be nice to have it in this doc.
https://www.baeldung.com/apache-kafka-check-server-is-running

sobychacko:
@Claudio-code Thanks for the PR! Although this is a docs-only change, the current `KafkaBinderHealthIndicator` implementation already provides the exact results you provided in the PR, i.e., if the broker is down, you will know it via the health indicator, the same goes for topics. See the [implementation](https://github.com/spring-cloud/spring-cloud-stream/blob/main/binders/kafka-binder/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/common/AbstractKafkaBinderHealthIndicator.java) for more details.  I don't know if we need to provide all the code samples in the PR since there are multiple ways to accomplish the logic. Plus, we have to be mindful of any Kafka-level API changes to make sure that the code samples are up to date. I suggest providing only the pseudo-code around your code sample by eliminating the implementation details. The point is that if someone wants to override the binder-provided health indicator implementation, it is possible to do so by applications providing an implementation and adding it as a `KafkaBinderHealth` bean. How the implementation is done is totally up to the application. The binder framework does not have any control over that process. Please let us know what you think about this.

Claudio-code:
I think you are right, by adding this code to the doc we would have to keep updating it with every change to the Kafka client and I also agree that there are other ways to solve this problem. I only made the suggestion because it was very laborious to think of this solution and if there was a path in the doc, the task would be easier. So, what would this pseudo code look like? Can you show me an example?

docs/src/main/asciidoc/kafka/kafka_overview.adoc
```diff
@@ -1016,3 +1016,56 @@ public KafkaBinderHealth kafkaBinderHealthIndicator() {
...
+        } catch (ExecutionException e) {
+            logger.error("ExecutionException {}", e.getMessage());
+        } catch (InterruptedException e) {
+            logger.error("InterruptedException {}", e.getMessage());
+            Thread.currentThread().interrupt();
+        }
+        return false;
+    }
+
+    public boolean checkTopicConnection() {
```
sobychacko, line 1058: Remove this implementation here as well. Add a small comment about what should happen here. You may want to point to the Kafka API docs for `describeTopics`.

docs/src/main/asciidoc/kafka/kafka_overview.adoc
```diff
@@ -1016,3 +1016,56 @@ public KafkaBinderHealth kafkaBinderHealthIndicator() {
     };
 }
 ```
+
+[[check-if-the-connection-to-the-brokers-is-in-error-or-in-topic]]
+=== Check if the connection to the brokers is in error or in topic
```
sobychacko, line 1021: Custom Kafka Binder Health Indicator Example

docs/src/main/asciidoc/kafka/kafka_overview.adoc
```diff
@@ -1016,3 +1016,56 @@ public KafkaBinderHealth kafkaBinderHealthIndicator() {
...
+
+To help identify if the cluster brokers are down or if it is just a specific topic that is having a connection problem,
+in complex scenarios where the service connects with several topics this verification helps to identify the severity of the problem.
```
sobychacko, line 1024: Simply state that `Here is the pseudo-code for writing a custom Kafka binder HealthIndicator. In this example, we try to override the binder provided Kafka HealthIndicator by specifically checking first for cluster connectivity and then followed by topic-related issues.`

sobychacko (changes requested):

sobychacko:
@Claudio-code I added some review comments in-line on the docs changes.

Claudio-code:
Cool until the weekend I'll make the corrections

Claudio-code:
@sobychacko I made changes if you can review again

docs/src/main/asciidoc/kafka/kafka_overview.adoc
```diff
@@ -1016,3 +1016,58 @@ public KafkaBinderHealth kafkaBinderHealthIndicator() {
...
+
+    @Override
+    public Health health() {
+        if (!checkBrokersConnection()) {
+            logger.error("Error when connect brokers");
+        }
+		if (!checkTopicConnection()) {
+			logger.error("Error when trying to connect with specific topic");
+		}
+        return Health.up().build();
```
sobychacko, line 1048: The code does not return a down status if the above calls return `false.` Is that intended? Instead of logging the error, why don't you return `Health.down()..`?

docs/src/main/asciidoc/kafka/kafka_overview.adoc
```diff
@@ -1016,3 +1016,58 @@ public KafkaBinderHealth kafkaBinderHealthIndicator() {
...
+    }
+
+	// The document has an example on how to implement these checks
+	// https://www.baeldung.com/apache-kafka-check-server-is-running
```
sobychacko, line 1052: You might want to point to the official Apache Kafka docs on this instead of an external blog.

sobychacko (changes requested):

docs/src/main/asciidoc/kafka/kafka_overview.adoc
```diff
@@ -1049,7 +1049,7 @@ public class KafkaBinderHealthImplementation implements KafkaBinderHealth {
     }
 
 	// The document has an example on how to implement these checks
-	// https://www.baeldung.com/apache-kafka-check-server-is-running
+	// https://docs.spring.io/spring-kafka/reference/html/#kafka-validation
```
sobychacko, line 1054: This link is probably not what we want since that is unrelated. I suggest dropping the link altogether and let the users decide how to write the implementation.

docs/src/main/asciidoc/kafka/kafka_overview.adoc
```diff
@@ -1016,3 +1016,60 @@ public KafkaBinderHealth kafkaBinderHealthIndicator() {
     };
 }
 ```
+
+[[custom-kafka-binder-health-indicator-example]]
+=== Custom kafka Binder Health Indicator Example
+
+Here is the pseudo-code for writing a custom Kafka binder HealthIndicator. In this example, we try to override the binder provided Kafka HealthIndicator by specifically checking first for cluster connectivity and then followed by topic-related issues.
```
sobychacko, line 1023: Can you put only one sentence per line? That's the asciidoc standard we follow.

sobychacko (changes requested):

sobychacko:
Thanks, @Claudio-code, for this docs PR. Now merged upstream with the squashed commits.

Issue #2796, opened 2023-08-25
Clarify documentation about partitioning with Kafka Binder and Functional style


welovelain:
Spring Cloud Stream version 3.29

https://docs.spring.io/spring-cloud-stream/docs/3.2.9/reference/html/spring-cloud-stream.html#_producer_properties
It is written:

> partitionCount
> The number of target partitions for the data, if partitioning is enabled. Must be set to a value greater than 1 if the producer is partitioned. On Kafka, it is interpreted as a hint. The larger of this and the partition count of the target topic is used instead.

While that's true for `@StreamListener` annotations, with `PartitioningInterceptor`/`DefaultPartitioningInterceptor` this is correct.
But if one uses `PartitionAwareFunctionWrapper` and functions, the `PartitionAwareFunctionWrapper` always takes `partitionCount` value from the properties, never taking it from the Kafka topic itself, which leads to `scst_partition` header always be 0. 
Because key.hash % partitionCount (which is 1) is always equal to 0.

sobychacko:
@welovelain Although the `partitionCount` is taken from the properties, the `KafkaTopicProvisioner` will check the topic partitions on the broker even when you are using `PartitionAwareFunctionWrapper.` This happens at the initial time when it bootstraps the binding. From that point onwards, throughout the application execution, the proper partition count is used, i.e., if it finds a higher number from the topic, then that will be used. If this is not what you observe, please clarify and give us a path to reproduce the issue.

sobychacko:
@welovelain Sorry about my earlier [comment](https://github.com/spring-cloud/spring-cloud-stream/issues/2796#issuecomment-1707257622) ^^. I verified that this is indeed a bug and that we are not updating the partition count to the one with which the Kafka topic is provisioned. We will fix this.

welovelain:
Sorry I couldn't provide an example, glad to be helpful though and thanks for fixing it!

PR #2807, opened 2023-09-13
GH-2796: Fix partition count related issues

sobychacko:
If a target middleware (Kafka for ex) topic has more partitions than what is set on the partition-count producer property, the binder never updates the original producer partition-count property to the partition count from the middleware if it is higher than what was given through the property. Because of this, te PartitionAwareFunctionWrapper which evaluates and assigns the partition header in Spring Cloud Stream does not compute the correct partition since the hash operation still using the original partition-count from the producer binding property. This commit is addressing this issue.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2796


placaze:
I think we are maybe hitting this bug with Spring Cloud Stream 4.0.4. Does it need to be backported to the `4.0.x` branch for the next Spring Cloud 2022 release?

Issue #2806, opened 2023-09-13
Ensure Pulsar binder has ExtendedBindingHandlerMappingsProviderConfiguration 


onobc:
Be sure the Pulsar-specific binding props are able to merge default into bindings. 

See similar issue [here](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/2151).

Other impls: 
- BindingHandlerAdvise  in core
- ExtendedBindingHandlerMappingsProviderConfiguration
- ExtendedBindingHandlerMappingsProviderConfiguration

PR #2811, opened 2023-09-18
GH-2806: Pulsar binder defaults properties config

sobychacko:
- Ensure that the Pulsar binder default properties can be properly expressed via spring.cloud.stream.pulsar.default property prefix.
 - Add the binder child context bean with the name binderName_binderProducingContext into the parent application context so that individual beans from the binder context can be easily queried.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2806

### DIFF
--- a/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/test/java/org/springframework/cloud/stream/binder/pulsar/PulsarExtendedBindingDefaultPropertiesTests.java
+++ b/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/test/java/org/springframework/cloud/stream/binder/pulsar/PulsarExtendedBindingDefaultPropertiesTests.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2023-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.stream.binder.pulsar;
+
+import java.util.function.Function;
+
+import org.apache.pulsar.common.schema.SchemaType;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+import org.springframework.cloud.stream.binder.pulsar.config.ExtendedBindingHandlerMappingsProviderConfiguration;
+import org.springframework.cloud.stream.binder.pulsar.properties.PulsarExtendedBindingProperties;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link ExtendedBindingHandlerMappingsProviderConfiguration}.
+ *
+ * @author Soby Chacko
+ */
+class PulsarExtendedBindingDefaultPropertiesTests implements PulsarTestContainerSupport {
+
+	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+		.withUserConfiguration(DefaultPropertiesTestApp.class)
+		.withPropertyValues(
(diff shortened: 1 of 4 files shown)


core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
```diff
@@ -482,6 +482,10 @@ public void onApplicationEvent(ApplicationEvent event) {
 
 		if (refresh) {
 			binderProducingContext.refresh();
+			if (this.context != null) {
```
onobc, line 485: I noticed this was added just now (therefore not actually needed by the other binders). Is this just so that we can get access to the context in the tests? If so, could we not just verify the properties are set on the actual consumer/producer like the Kafka Binder ITs do for this?

### Example
The Producer key serializer gets set via default property here
https://github.com/spring-cloud/spring-cloud-stream/blob/e06d4eedfd70302f0ba7241adb3e3e3d8d8b00d5/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/integration/KafkaBinderExtendedPropertiesTest.java#L63-L64

and then is verified that it is on the binding properties https://github.com/spring-cloud/spring-cloud-stream/blob/e06d4eedfd70302f0ba7241adb3e3e3d8d8b00d5/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/integration/KafkaBinderExtendedPropertiesTest.java#L92-L97

binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/test/java/org/springframework/cloud/stream/binder/pulsar/PulsarExtendedBindingDefaultPropertiesTests.java
```diff
@@ -0,0 +1,114 @@
...
+ * Tests for {@link ExtendedBindingHandlerMappingsProviderConfiguration}.
+ *
+ * @author Soby Chacko
+ */
+class PulsarExtendedBindingDefaultPropertiesTests implements PulsarTestContainerSupport {
+
+	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+		.withUserConfiguration(DefaultPropertiesTestApp.class)
+		.withPropertyValues(
+			"--spring.pulsar.client.service-url=" + PulsarTestContainerSupport.getPulsarBrokerUrl(),
```
onobc, line 44: I ❤️ the tests. 

One question, is the discrepancy in the property pair prefix intended? (some have `--`)

onobc (changes requested):
Nice work @sobychacko ! A couple of minor comments to address.

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
sobychacko: @onobc We can do that, but I like to avoid those casting and the usage of DBF to gain access to the binder, etc. I had a discussion with @olegz on this the other day, and we decided to make this change at the core level to add the binder context to the parent context as a separate isolated context for anyone to get this type of read-only access to the binder context. Post 4.1 release, we were even thinking of removing this parent-child hierarchy altogether for scenarios that don't need that type of relationship (multi-binder is the only thing that requires it). If you are okay with that, l say, let's go with this approach. Please let me know.

binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/test/java/org/springframework/cloud/stream/binder/pulsar/PulsarExtendedBindingDefaultPropertiesTests.java
sobychacko: Hmm, it was not intended. Didn't cause an issue. Do you want me to take care of them? or will you adjust on merge?

core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
onobc: Makes sense. I do also ❤️ the idea of scrapping the parent/child when not needed as it add complexity.

binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/test/java/org/springframework/cloud/stream/binder/pulsar/PulsarExtendedBindingDefaultPropertiesTests.java
onobc: If you can take care of them that would be good as I was hoping to just merge from Github for this one.

onobc (approved):
LGTM! Thanks for the updates @sobychacko . I will merge soon.

Issue #2662, opened 2023-03-01
Kafka message no longer processed after upgrading from Spring Cloud Stream 3.2.4 to 4.0.1


gnagy:
**Describe the issue**
I have a simple Spring Boot app in Kotlin using Spring Cloud Stream that has:

- REST endpoint that publishes a kafka message using streamBridge
- Consumer<> writing said message from same kafka topic to a database
- 2 test cases for valid and invalid messages, both passing with SCS 3.2.4 but failing with 4.0.1:
  - happy test case: message never arrives to consumer, test times out after 5sec
  - error case: Exception "Dispatcher has no subscribers for channel 'spring-example.message-saver-in-0'."

**To Reproduce**
See tests in https://github.com/vacuumlabs/spring-example/pull/1:
- `new transaction - valid`
- `new transaction - valid, nonexistent account number`

I tried debugging the issue and checking any changes from 3->4; looking at the Wiki and GitHub issues but can't see what might be the problem:

https://github.com/spring-cloud/spring-cloud-stream/wiki/4.0-Migration-guide
https://github.com/spring-cloud/spring-cloud-stream/issues/2610
https://github.com/spring-cloud/spring-cloud-stream/issues/2632

**Version of the framework**
Going from 3.2.4 to 4.0.1

**Expected behavior**
Message is published and consumed, or found on DLQ

sobychacko:
@gnagy Could you provide a simple Java application where we can reproduce the issue easily? Also, could you include a simple README with step-by-step instructions to set up the app and the necessary components?

gnagy:
Hi @sobychacko,

The issue occurred in the simple `spring-example` project I linked above. You can reproduce it by running the tests with `./gradlew test`. It uses testcontainers so you'll also need docker running.

https://github.com/vacuumlabs/spring-example/pull/1

sobychacko:
I wonder if it is anything Kotlin-related? Can you compare the same setup with a Java application? If we were missing data, as you describe, it would have been a major issue on the 4.0 line, and we would have already seen it. I was unable to run your application due to some environment-related issues. If you can provide a similar Java project, I would be happy to look into it further.

sobychacko:
@gnagy Any update on this issue?

gnagy:
Hi, yes, I will find some time to see if I can reproduce same issue in Java.

gnagy:
Hi @sobychacko,

[...]

PR #2812, opened 2023-09-19
GH-2662: Kafka binder DLQ root cause message

sobychacko:
- Becasue NestedRuntimeException from Spring Framework core 6.x removed the getMessage method that included the detailMessage with cause in it, the Kafka binder DLQ records no longer include the cause message. Fix this issue by including the cause in the exception message.

See this comment for more details:
 - https://github.com/spring-cloud/spring-cloud-stream/issues/2662#issuecomment-1722849892 This fix is based on the following Spring Kafka commit.
 - https://github.com/spring-projects/spring-kafka/commit/6f585058a60090f2940c563bca0a4b9208b7e44e

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2662


garyrussell (commented):
LGTM, but the PR build failed for an unrelated reason.

garyrussell:
A bunch of test failures; click on details. in the "All checks have failed".

sobychacko:
Ah, I think you can ignore those for the time being. Jenkins main CI is all good. This is a new ci-pr we added last week that needs to be cleaned up with a few things.

garyrussell (approved):

Issue #2805, opened 2023-09-12
StreamBridge send to bindingName & binderName ignores binding property's content type


KevinLiLu:
**Describe the issue**
When using StreamBridge's `send(String bindingName, @Nullable String binderName, Object data)` API, it ignores the binding's configured content type and instead forces `application/json`.

https://github.com/spring-cloud/spring-cloud-stream/blob/0c256343c7b063d9ef44b7b96df95a0f8b1f357c/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamBridge.java#L150

**To Reproduce**
1. Configure binding to use something other than `application/json` content type
2. Call StreamBridge's `send(String bindingName, @Nullable String binderName, Object data)` API

**Version of the framework**
Latest 4.0.4

**Expected behavior**
If content type is not provided in the API, then it should first try to use the binding's configured content type (if binding exists and content type is configured). If binding does not exist or content type is not configured, then it should default to `application/json`. This is the behavior seen in the other send API which does not have content type as a parameter: https://github.com/spring-cloud/spring-cloud-stream/blob/0c256343c7b063d9ef44b7b96df95a0f8b1f357c/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamBridge.java#L138-L142

**Screenshots**
None

**Additional context**
None

KevinLiLu:
I'm willing to raise a PR if we agree on the expected behavior!

KevinLiLu:
@sobychacko @olegz Thoughts?

sobychacko:
@KevinLiLu We will get back to you soon on this. Thanks!

sobychacko:
@KevinLiLu This was a bug, and thanks for pointing it out. The fix was straightforward and already merged upstream. Please take it for a spin and see if it works for your use case.

sobychacko:
@KevinLiLu My sincere apologies. I saw that earlier, but while investigating the issue, I realized that the fix is so simple that I decided to issue a PR directly from my triaging. We highly encourage you to contribute PRs if you see issues like these. Just ping on the issue and let us know that you are working on it so that we are aware of it. Sorry again for the miscommunication on this matter. Your willingness to help is greatly appreciated and looking forward to your contributions.

PR #2813, opened 2023-09-19
GH-2805: StreamBridge send and custom content-type

sobychacko:
- When StreamBridge#send is called with binder-name and custom content-type, it does not honor the content-type value, but default to application/json. Fixing this issue for this call path by explicitly checking for any custom content-type provided on the binding.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2805


olegz:
Merged and backported to 4.0.x

PR #2833, opened 2023-10-18
Add DltAwareProcessor trace logging

StevenPG:
Since exception is not propogated, adding a trace log allows an optional way for a developer to utilize the defaultRecoverer and still do some basic review of a given exception.


sobychacko:
@StevenPG Looks good. Could you add your name as an author in `DltAwareProcessor`?

Issue #2828, opened 2023-10-11
User's defined cloud stream binder are not recognized by AOT stage


petretiandrea:
**Describe the issue**
This introduce a bug where AOT doesn't recognize user defined binders.
https://github.com/spring-cloud/spring-cloud-stream/blob/f811ed0b7ca7108f16e57f4e6716ebfc00ba4705/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/BinderChildContextInitializer.java#L111
Given this configuration the AOT pre-creates only "kafka" (the default one) not for test1 and test2
```yaml
spring:
    cloud:
        function:
            definition: test1;test2
        stream:
            default-binder: test1
            bindings:
                test1-in-0:
                    destination: topic-test-1
                    group: topic-test1-group
                    binder: test1
                test2-in-0:
                    destination: topic-test-2
                    group: topic-test2-group
                    binder: test2
            binders:
              test1:
                type: kakfa
              test2:
                type: kakfa
```
By editing the configuration in this way:
```yaml
spring:
    cloud:
        function:
            definition: test1;test2
        stream:
            default-binder: test1
            bindings:
                test1-in-0:
                    destination: topic-test-1
                    group: topic-test1-group
                    binder: test1
                test2-in-0:
                    destination: topic-test-2
                    group: topic-test2-group
                    binder: test2
            binders:
              binders:
                test1:
                  type: kakfa
                test2:
                  type: kakfa
```
Yes, with two `binders` keyword, the AOT pre-creates `test1` and `test2`. 

I'm doing something wrong or is an unexpected bug?

**To Reproduce**
Try to use this simple cloud stream configuration
```yaml
spring:
    cloud:
        function:
            definition: test1;test2
        stream:
            default-binder: test1
            bindings:
                test1-in-0:
                    destination: topic-test-1
                    group: topic-test1-group
                    binder: test1
                test2-in-0:
                    destination: topic-test-2
                    group: topic-test2-group
                    binder: test2
            binders:
              test1:
                type: kakfa
              test2:
                type: kakfa
```

**Version of the framework**
2023.0.0-SNAPSHOT
**Expected behavior**
No two `binders` keywork required and AOT recognize user defined binders.
**Screenshots**

**Additional context**

PR #2834, opened 2023-10-19
GH-2828: Custom binders not recognized by AOT

sobychacko:
- When using multi-binders with custom binders and running in AOT mode, there is a regression that is causing some issues for propertly identifying the binders during the AOT phase. It forces the users to provide property in the form of `spring.cloud.stream.binders.binders...` Fixing this issue by properly binding the custom binders in BinderChildContextInitializer.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2828


petretiandrea:
Why has been closed?

sobychacko:
The issue is addressed via the above-mentioned commit on the `main` branch upstream.

Issue #2675, opened 2023-03-23
Kafka binder used with pollables makes fail the bindings actuator endpoint


JordiMartinezVicent:
**Describe the issue**
Given an application with **two instances** and:

- Uses the kafka binder
- Declares a pollable

The actuator endpoint `/actuator/bindings` lasts around 100s if the call is done to the instance which has no assigned partitions. 

**To Reproduce**

[scs-pollables.zip](https://github.com/spring-cloud/spring-cloud-stream/files/11051045/scs-pollables.zip)

With the attached application the behavior can be reproduced. In order to simulate two instances of the same application, it declares two pollables with the same group pointing to the same topic. 

To startup the application, a kafka broker is needed. 

To reproduce:

- Start up the attached application
- Call to the endpoint `http://localhost:8080/actuator/bindings`

**Expected behavior**
The call to the endpoint should not last around 100s

**Version of the framework**

Spring Boot: 3.0.4
spring-cloud-dependencies: 2022.0.1

**Additional context**

The issue is due to the class `org.springframework.integration.kafka.inbound.KafkaMessageSource` is synchronized and used for polling and to retrieve the binding state from the bindings endpoint. 

On the one hand, it defines the following timeouts: 

```java
this.pollTimeout = Duration.ofMillis(consumerProperties.getPollTimeout());
this.assignTimeout = Duration.ofMillis(Math.max(this.pollTimeout.toMillis() * 20, MIN_ASSIGN_TIMEOUT)); // NOSONAR - magic 		
```

Note that the default `pollTimeout` is 5 seconds, so the `assignTimeout ` will be 100s. 

And it uses them here at the method `pollRecord`:

```java 
ConsumerRecords<K, V> records = this.consumer
  poll(this.assignedPartitions.isEmpty() ? this.assignTimeout : this.pollTimeout);
```

Note that the `assignTimeout` is used if there are no assignedPartitions

On the other hand, when calling the endpoint, the method of the same class `isRunning` is called (which is syncronized).

Maybe to avoid the synchronization point, the state running and stoped could be manage at the class `DefaultBinding`, same as it does with the "paused" state. In that manner, the polling and the retrieving of the state would not collision.

JordiMartinezVicent:
Hi guys, any update with this? 
@olegz @garyrussell

sobychacko:
Closing this in lieu of https://github.com/spring-projects/spring-integration/issues/8778

PR #2838, opened 2023-10-25
GH-2675: DefaultBinding isRunning enhancements

sobychacko:
- Instead of relying on the lifecycle object's isRunning method which could possibly be synchronized and cause a contention for the lock,  maintain the running state separately inside DefaultBinding via a field (similar to how it tracks the pausable state).

   Without this change, in the pollable Kafka consumer based applications, there is a possiblity for a longer wait when queyring the binding status in DefaultBinding (for example via the binding actuator endpoint) due to this aforementioned lock contention.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2675

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/AbstractMessageChannelBinder.java
@@ -228,7 +228,7 @@ public abstract class AbstractMessageChannelBinder<C extends ConsumerProperties,
 	}
 
 	private String resolveBinderName(String bindingName, BindingServiceProperties bindingServiceProperties) {
-		String binder = bindingServiceProperties == null ? null : bindingServiceProperties.getBindings().get(bindingName).getBinder();
+		String binder = getBinderName(bindingName, bindingServiceProperties);
 		if (!StringUtils.hasText(binder)) {
 			return resolveFromDefaultBinder();
 		}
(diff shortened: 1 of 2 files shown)


sobychacko:
@garyrussell @artembilan What do you think about temporarily having this change here? If it's not worth it, we will close this PR here.

artembilan:
I'm sorry, I fail to determine what is wrong and what should be fixed in the `KafkaMessageSource`.
Note: in `6.2` we have moved to an internal `Lock` instead of `synchronized` on those methods.
Maybe that can fix it? Then we simply can back-port the fix to previous versions.
But again: where is a dead lock, please?
Thanks

sobychacko:
@artembilan The deadlock happens when the `/binding` actuator endpoint makes a call to the `KafkaMessageSource#isRunning()` method. Because of a timeout set there, there is a code path, that a thread may wait for the lock to release, which happens only after the timeout. This only happens for pollable consumers. See this issue for more details. https://github.com/spring-cloud/spring-cloud-stream/issues/2675

sobychacko:
Closing this in lieu of https://github.com/spring-projects/spring-integration/issues/8778

garyrussell:
I think we can just change the lifecycle synchronization to use an `AtomicBoolean` for `running`.

I am not sure why it is set to true after creating the consumer (in `doReceive()`) - looks like that was added by an external contributor.

artembilan:
The `doReceive()` is `synchronized` in the Spring Integration version and it is blocked waiting for result from the `pollRecord()` which ends up to be waiting for `consumer.poll()` where no partitions assigned to this consumer since the only one existing has been assigned to another one.
That's how it is dead-locking.
I think we can just narrow a synchronized/locked block in the `doReceive()` to exclude that `pollRecord()` call.

garyrussell:
Right; we. should also call `consumer.wakeup()` in `stopConsumer()` in case the poller is stuck here.

PR #2900, opened 2024-02-13
GH-2893: Enable AutoConfiguration of TestChannelBinderConfiguration

### LINKED ISSUE #2893: Delay in autoconfiguration of TestBinder
With the update in #2566, it is no longer necessary to import `TestChannelBinderConfiguration` to use TestBinder.

However, if no message handler is registered (i.e., not defining Supplier/Consumer/Function), it seems that the autoconfiguration of `TestChannelBinderConfiguration` is delayed. In this case, the injection of `OutputDestination` fails. Therefore, it is still necessary to import it.
```java
@SpringBootTest
public class AppTest {
  @Autowired
  private OutputDestination outputDestination; // Injection failure
```
Log excerpt
```log
[ERROR] Errors:
[ERROR]   AppTest.test » UnsatisfiedDependency Error creating bean with name 'com.example.AppTest': Unsatisfied dependency expressed through field 'outputDestination': No qualifying bean of type 'org.springframework.cloud.stream.binder.test.OutputDestination' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
```
Is this the intended behavior? Is it not possible to completely eliminate the import?

## Demo and workaround
[Demo and workaround](https://github.com/kutmk/spring-cloud-stream-test-binder-issue-demo): This demo intends to test an application that uses Test Binder, `StreamBridge`, and `OutputDestination` without registering a message handler.

## Environment
* Java 17
* Spring Boot 3.1.1
* Spring Cloud 2022.0.3
* Spring Cloud Stream 4.0.3

## Cause
The `TestChannelBinderConfiguration` is autoconfigured when `DefaultBinderFactory#initializeBinderContextSimple` is executed.

If a message handler is registered, this method is called before the injection of the test class's member variables due to `OutputBindingLifecycle` or `InputBindingLifecycle`.

However, if a message handler is not registered, this method is not called until during the application’s execution when a Binder is needed (such as during the execution of `StreamBridge#send`). As a result, the Bean is not created until then, and the injection fails.

kutmk:
- Resolved the issue where the auto-configuration of TestChannelBinderConfiguration is delayed and injection fails if a message handler is not registered.
- Always enable the AutoConfiguration of TestChannelBinderConfiguration.

Resolves #2893

### DIFF
--- a/core/spring-cloud-stream-test-binder/src/main/java/org/springframework/cloud/stream/binder/test/TestChannelBinderConfiguration.java
+++ b/core/spring-cloud-stream-test-binder/src/main/java/org/springframework/cloud/stream/binder/test/TestChannelBinderConfiguration.java
@@ -20,7 +20,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 import org.springframework.cloud.stream.binder.Binder;
 import org.springframework.cloud.stream.binder.ConsumerProperties;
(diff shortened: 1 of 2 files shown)


sobychacko:
Looks good. Merged upstream.

Issue #2902, opened 2024-02-14
Expose `RabbitBinderConfigurationProperties.enableObservation` property similar to the one in `KafkaBinderConfigurationProperties`


artembilan:
Right now we do enable observation in RabbitMQ Binder endpoints unconditionally when an `ObservationRegistry` bean in present in the application context.
For consistency it is better to leave the choice up to end-user independently of the environment.
See `TODO`  in the `RabbitMessageChannelBinder`.

Issue #2901, opened 2024-02-14
Micrometer context propagation does not work in multi-binder scenarios when binder types are the same (e.g., Rabbit and Rabbit)


Toxa45:
Good afternoon.
When using SteamBridge with RabbitMQ, the traceId of the sender and recipient are different.
Here is an example based on what was attached in https://github.com/spring-cloud/spring-cloud-stream/issues/2639, but the traceId are different.
https://github.com/Toxa45/TestRabbitTraceId/tree/master/TestRabbitTraceId
![image](https://github.com/spring-cloud/spring-cloud-stream/assets/57655161/5fb27a54-8f56-4a04-93f0-ea12ec9a1a2d)
I also noticed that this happens when the “environment” is filled. Initially, I made a typo and wrote “environment” - and, to my surprise, everything worked locally.

Toxa45:
@olegz Good afternoon. When is 4.2 planned to be released? Or maybe there is some kind of temporary solution?

Toxa45:
@olegz Good afternoon. The example that I posted is just an example that was transferred from a real project. The binders settings in the example are just to reproduce the error locally. You don’t have binders, so how do you suggest specifying the required host and credentials to connect to rabbit?


    spring:
        cloud:
            stream:
                binders:
                    rabbit:
                        type: rabbit
                        environment:
                            spring:
                                rabbitmq:
                                    password: pass
                                    username: username
                                    port: 5672
                                    virtual-host: /
                                    deliveryMode: 2
                                    host: RABBIT_HOST

PR #2903, opened 2024-02-14
Rework an observation for Rabbit Binder

artembilan:
The observation propagation doesn't work in multi-binder configuration

* Remove `ObservationAutoConfiguration` since it is not visible in case of multi-binder configuration
* Instead move `setObservationEnabled` flag setting to the `RabbitMessageChannelBinder`
* Add `io.micrometer.observation.ObservationRegistry` into `shared.beans` to make it visible for binder-specific application context
* Add `RabbitMultiBinderObservationTests` integration test where Rabbit Binder is in multi-binder environment

As a side effect this fixes an observation propagation for Kafka binder as well in the multi-binder environment. Its configuration is OK, but an `ObservationRegistry` must make it visible for the binder-specific application context. See the mentioned `shared.beans`

Related to https://github.com/spring-cloud/spring-cloud-stream/issues/2901 Also see https://github.com/spring-cloud/spring-cloud-stream/issues/2902 for possible evolution


sobychacko:
@artembilan Looks good. One small comment - some classes are missing copyright updates to `2024` and your author tag.

Issue #2915, opened 2024-03-07
Pulsar binder integ test failure on JDK 21


sobychacko:
[PulsarBinderIntegrationTests#binderAndBindingPropsAreAppliedAndRespected](https://github.com/spring-cloud/spring-cloud-stream/blob/main/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/test/java/org/springframework/cloud/stream/binder/pulsar/PulsarBinderIntegrationTests.java#L79) is failing on JDK 21 because of some issues with the property [max-pending-messages](https://github.com/spring-cloud/spring-cloud-stream/blob/main/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/test/java/org/springframework/cloud/stream/binder/pulsar/PulsarBinderIntegrationTests.java#L91). Please investigate, why this property has issues on JDK 21, but not on JDK 17.

onobc:
## TL;DR
The error has been there from inception and is only noticed because the `ProducerConfigurationData.setMaxPendingMessages` and `ProducerConfigurationData.setMaxPendingMessagesAcrossPartitions` are called in different order in Java17 and 21. 

If the failing test would have instead set only the `maxPendingMessagesAcrossPartitions` we would have seen the same issue in Java17. 

## Details
The binder allows users to configure the underlying producer used by the binder. 

The config props are layered from `base -> binder -> binding` and we only include those properties that have changed from the defaults. 

When hydrating the ProducerConfigurationData with the modified props, the `setMaxPendingMessagesAcrossPartitions` validates that it is `>= the maxPendingMessages`. 

In the case where only the `maxPendingMessages` is set, the `maxPendingMessagesAcrossPartitions` has a default of 0. 
If the `setMaxPendingMessagesAcrossPartitions` is called after `setMaxPendingMessages` then the validation will fail.

In Java17, the serialized JSON and merged set of props were put into a HashMap which just so happened to result in `maxPendingMessagesAcrossPartitions` being called before `maxPendingMessages` and we did not notice the issue in the test.

In Java21, the resulting props in the HashMap just so happen to result in `maxPendingMessages` being called before `maxPendingMessagesAcrossPartitions` and we see the validation error.

## Solution
[...]

PR #2918, opened 2024-03-13
Fix Pulsar binder for Java 21

onobc:
Running the Pulsar binder with Java 21 surfaced a bug where the layered binder/binding config props are not properly applied to the targeted consumer/producer builders.

The bug existed before Java 21 and is unrelated.

The binder/binding config props are converted to a Pulsar ProducerConfigurationData object via Jackson. This conversion process iterates over a map of config props. The iteration order of the map changed between Java 17 and 21. This results in object setters getting called in different orders which in turn causes cross field validation errors.

See #2915 for more details

### DIFF
--- a/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/main/java/org/springframework/cloud/stream/binder/pulsar/PulsarBinderUtils.java
+++ b/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/main/java/org/springframework/cloud/stream/binder/pulsar/PulsarBinderUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023-2023 the original author or authors.
+ * Copyright 2023-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/main/java/org/springframework/cloud/stream/binder/pulsar/PulsarBinderUtils.java
```diff
@@ -91,65 +100,70 @@ static Map<String, Object> mergeModifiedProducerProperties(ProducerConfigPropert
...
 
-		Map<String, Object> newOrModifiedBindingProps = extractNewOrModifiedProperties(bindingProps, baseProps);
+		Map<String, Object> newOrModifiedBindingProps = extractNewOrModifiedProperties(bindingProps, globalProps);
 		LOGGER.trace(() -> "New or modified binding props: %s".formatted(newOrModifiedBindingProps));
 
 		Map<String, Object> mergedProps = new HashMap<>(newOrModifiedBinderProps);
 		mergedProps.putAll(newOrModifiedBindingProps);
-		LOGGER.trace(() -> "Final merged props: %s".formatted(mergedProps));
 
+		// Add in default properties for any props not customized by the user
```
onobc, line 162: Ultimately this is the "fix" in that it ensures that all extended properties values are passed into the object setters.

sobychacko (approved):

PR #2920, opened 2024-03-15
#2919 Load eagerly the configuration, even the lazy-init enabled

### LINKED ISSUE #2919: Enabling lazy initialization breaks Kafka Listeners
I've been trying to optimize startup times by enabling **spring.main.lazy-initialization**. However, after making this change, I've encountered an issue with Kafka listeners. Specifically, the listeners stop receiving messages.

I attempted to explicitly set the `@Lazy(false)` annotation on the Kafka listener components, this hasn't resolved the problem. While it appears that the listener beans are created at startup *(as expected with `@Lazy(false)`)*, they still do not receive any messages from Kafka.

I use **Spring Boot 3.2.2** with **Spring Cloud Stream 4.1.0** and I configure my Kafka listeners from the property yml like this:

```yml
spring:
  cloud:
    config:
      allowOverride: true
      overrideNone: true
      overrideSystemProperties: false
    function:
      definition: >-
        listenExample;
    stream:
      bindings:
        listenExample-in-0:
          destination: example-topic
          group: example-group
          consumer:
            use-native-decoding: true
      kafka:
        bindings:
          listenExample-in-0:
            consumer:
              enableDlq: true
              dlqName: example-error-topic
              dlqPartitions: 1
```

The component class:

```java
@Component("listenExample")
@Lazy(false)
@RequiredArgsConstructor
@Transactional
public class ExampleListener implements Consumer<Message<String>> {

    @Override
    public void accept(Message<String> message) {
        // do something
    }
}
```

The expected behavior could be the possibility of loading these stream components in a lazy initialization environment using a config property, for example.

PatrikScully:
Load eagerly the `FunctionConfiguration`, even with `spring.main.lazy-initialization` property enabled, because the listeners must initialize at startup, otherwise, they won't work.

fixes #2919

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionConfiguration.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionConfiguration.java
@@ -128,6 +128,7 @@ import org.springframework.util.StringUtils;
  * @author Chris Bono
  * @author Byungjun You
  * @author Ivan Shapoval
+ * @author Patrik Péter Süli
  * @since 2.1
  */
 @Lazy(false)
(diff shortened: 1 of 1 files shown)


PatrikScully:
> @PatrikScully, Can you add your name as an author to the file?

Done, thank you.

Issue #2951, opened 2024-05-16
Batch Consumer With ListenerContainerWithDlqAndRetryCustomizer should disable binder DLQ


sobychacko:
Batch consumer with `ListenerContainerWithDlqAndRetryCustomizer` should disable binder-based DLQ. 

Related to: https://stackoverflow.com/questions/78485425/spring-cloud-stream-kafka-binder-retries-not-working-when-using-dlq-in-batch-m.

PR #2952, opened 2024-05-16
GH-2951: Batch consumer and DLQ issues

sobychacko:
Batch consumer with `ListenerContainerWithDlqAndRetryCustomizer` should disable binder-based DLQ.

Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2951

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/BatchWithDlqCustomizerDisablesBinderDlqTests.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/BatchWithDlqCustomizerDisablesBinderDlqTests.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2024-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/BatchWithDlqCustomizerDisablesBinderDlqTests.java
```diff
@@ -0,0 +1,78 @@
...
+			new ExtendedConsumerProperties<>(kafkaConsumerProperties);
+		extendedConsumerProperties.setBatchMode(true);
+
+		MessageHandler errorMessageHandler =
+			kafkaMessageChannelBinder.getErrorMessageHandler(null, null, extendedConsumerProperties);
+		// verifies that binder does not create a message handler for errors, which otherwise creates a handler for DLQ.
+		assertThat(errorMessageHandler).isNull();
+	}
+
+	@SpringBootApplication
```
artembilan, line 62: It is dangerous to use this annotation in tests. It comes with `@ComponentScan`, so unexpected beans might appear in the context from the same package as this test class. Consider to use `@EnableAutoConfiguration` and `@Configuration` pair instead . You can see them from the mentioned `@SpringBootApplication`

artembilan (changes requested):

binders/kafka-binder/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/BatchWithDlqCustomizerDisablesBinderDlqTests.java
sobychacko: Thanks, Artem, for spotting that. I was copying from an existing test.

artembilan (changes requested):

PR #2978, opened 2024-07-26
Remove redundant `delaySubscription` from `FunctionConfiguration`

artembilan:
Related to: https://github.com/spring-projects/spring-integration/issues/9362

After the fix in Spring Integration: https://github.com/spring-projects/spring-integration/commit/bdcb856a9081bc091d56be16d51f0f4d561bc9ce we ended up in a deadlock situation with a `beginPublishingTrigger` in the `FunctionConfiguration` used for the `delaySubscription()` on an original `Publisher`. The `FluxMessageChannel` uses its own `delaySubscription()` until the channel has its subscribers. Apparently the logic before was with errors, so the `FluxMessageChannel` was marked as active even if its subscriber is not ready yet, leading to famous `Dispatcher does not have subscribers` error. So, looks like this `beginPublishingTrigger` was introduced back in days in Spring Cloud Stream to mitigate that situation until we really emit a `BindingCreatedEvent`.

The deadlock (and the flaw, respectively) is with the `setupBindingTrigger()` method implementation where `FluxMessageChannel` now "really" delays a subscription to the provided `Publisher`, therefore not triggering that `Mono.create()` fulfilment immediately. The `BindingCreatedEvent` arrives earlier, than we have a subscriber on the channel, but `triggerRef.get()` is `null`, so we don't `success()` it and in the end don't subscribe to an original `Publisher` since `delaySubscription()` on it is never completed.

Since `FunctionConfiguration` fully relies on `IntegrationFlow.from(Publisher)`, which ends up with the mentioned  `FluxMessageChannel.subscribeTo()` and its own `delaySubscription()` (which, in turn, apparently fixed now), we don't need our own `delaySubscription()` any more. Therefore the fix in this PR is to propose to remove `beginPublishingTrigger` logic altogether.

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionConfiguration.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/FunctionConfiguration.java
@@ -40,7 +40,6 @@ import org.apache.commons.logging.LogFactory;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
-import reactor.core.publisher.MonoSink;
 import reactor.util.function.Tuples;
 
 import org.springframework.beans.BeansException;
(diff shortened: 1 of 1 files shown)


sobychacko (approved):

sobychacko:
Merged via https://github.com/spring-cloud/spring-cloud-stream/commit/3b5bb2e0d9a89c3b9ae9152c99268460ca7ed442.

@olegz We need to discuss this when you are back. I merged it to address some CI test failures.

wrwksexahatvani:
@artembilan @sobychacko Any update on the release date?

sobychacko:
Hi, this will be part of the `2024.0.0-M2` release of spring-cloud scheduled for 10/03/24. 

https://calendar.spring.io/

PR #2983, opened 2024-08-01
Missing Headers Problem

omercelikceng:
Hello,

While working on a development related to Spring Cloud Stream, I noticed a situation that is also relevant to Spring Cloud Function.

To explain the issue clearly, we will first examine a method from Spring Cloud Function. The following method exhibits different behaviors. In some cases, it generates the output message using the input header, while in other cases, it does not consider the input header at all. This results in some variations in the code.
In other words, the code below can sometimes directly set the headers provided in the input to the output header. Sometimes, it creates the output headers without using the input headers.

Screenshot 2024-08-01 233525

Now, let's examine the method affected by the above code. I have written some tests, and based on the results, the target-protocol header is sometimes set and sometimes not. Therefore, I have made the relevant development.

However, I still have some questions that I do not fully understand. I would greatly appreciate your help. Is the target-protocol header a value that should be set for every message? If so, there are many places that we need to correct. Could you please provide me with information on this matter?

I am a bit confused, so I wanted to ask you before continuing with the development. Thank you.

### DIFF
--- a/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/function/HeaderTests.java
+++ b/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/function/HeaderTests.java
@@ -42,7 +42,9 @@ import static org.assertj.core.api.Assertions.assertThat;
 /**
  * @author Omer Celik
  */
+
 public class HeaderTests {
+
 	@BeforeAll
 	public static void before() {
 		System.clearProperty("spring.cloud.function.definition");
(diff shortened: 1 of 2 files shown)


core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/function/HeaderTests.java
```diff
@@ -0,0 +1,159 @@
...
+
+			Message<byte[]> result = output.receive(1000, "uppercase-out-0");
+			MessageHeaders headers = result.getHeaders();
+			assertThat(headers).isNotNull();
+			assertThat(headers.get(MessageUtils.TARGET_PROTOCOL)).isEqualTo("kafka");
+			assertThat(headers.get(MessageHeaders.CONTENT_TYPE)).isEqualTo("application/json");
+		}
+	}
+
+	@Test
```
omercelikceng, line 112: If the development I made is not implemented, only this test fails. The other tests complete successfully.

omercelikceng:
@olegz, hello. Thank you very much for your interest. However, there are other situations where these headers are not set. If this target_protocol needs to be set in all cases, I will need to make further developments accordingly. Could you inform me about this?

olegz:
Please do

PR #2990, opened 2024-08-16
Code cleanup

omercelikceng:


binders/rabbit-binder/spring-cloud-stream-binder-rabbit/src/test/java/org/springframework/cloud/stream/binder/rabbit/integration/RabbitBinderModuleTests.java
```diff
@@ -193,7 +193,7 @@ void parentConnectionFactoryInheritedByDefault() throws Exception {
 	private void checkCustomizedArgs() throws MalformedURLException, URISyntaxException, InterruptedException {
 		List<Map<String, Object>> bindings = RestUtils.getBindingsBySource(client, uri, "/", "process-in-0");
 		int n = 0;
-		while (n++ < 100 && bindings == null || bindings.size() < 1) {
```
olegz, line 196: I have not looked at this code, but usually `< 1` implies that there may be two or more bindings, for which there would be additional handling. Please take a look

omercelikceng: Thank you very much @olegz . I hadn't looked at the situation from this perspective. I'm reverting my change right away.

PR #2998, opened 2024-09-03
Avoid duplicate error logging when multiple destinations per binding

hightea:
When using multiple destinations for one binding, the errors are logged as many times as the number of destinations.
For an example, when an error occurs in the following binding : 
```
spring.cloud.stream.bindings.process-in-0.consumer.max-attempts=1
spring.cloud.stream.bindings.process-in-0.destination=one,two,three
```

This results in three successive logs from the `org.springframework.integration.handler.LoggingHandler`

This pull request fixes this by bridging only once the bindingErrorHandler whith the globalErrorHandler

### DIFF
--- a/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/binder/ErrorBindingTests.java
+++ b/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/binder/ErrorBindingTests.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.stream.binder;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
+import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.Test;
 
 import org.springframework.boot.WebApplicationType;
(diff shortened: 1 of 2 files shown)


olegz:
I like the PR, but I wonder if we should assume that some destination will get successful delivery, a single log could be confusing. What do you think exposing your code via some boolean flag? I can even agree that the default path wil be to a single log, but having a flag would allow users to change back to the original behavior.
Let me know what you think. . .

hightea:
This is only for consumers (`registerErrorInfrastructure(ConsumerDestination, ...)`).
Every consumer for the binding (one per "destination") share the same binderErrorChannel. With the current behavior, every error is logged as many times as the number of destinations, but in our case the "destinations" are inputs so a message originates from **one** destination only, and when an error occurs, it's logged several times.
Note that the logs are emmitted successively without any intermediate processing

However, I can add a flag if you think it's a better option, but I don't see any case where we'll want multiple logs.

olegz:
I see it now. I'll merge.

hightea:
Thank you !

PR #3011, opened 2024-09-26
Ensuring Thread-Safety for BindingServiceProperties.bindings

kurthong:
First, thank you for the excellent StreamBridge feature! I’ve found it very useful.
However, I rarely encounter the following exception during race conditions:
```
java.lang.NullPointerException: null
    at java.base/java.util.TreeMap.rotateRight(TreeMap.java:2240)
    at java.base/java.util.TreeMap.fixAfterInsertion(TreeMap.java:2272)
    at java.base/java.util.TreeMap.put(TreeMap.java:580)
    at org.springframework.cloud.stream.config.BindingServiceProperties.bindToDefault(BindingServiceProperties.java:397)
    at org.springframework.cloud.stream.config.BindingServiceProperties.bindIfNecessary(BindingServiceProperties.java:381)
    at org.springframework.cloud.stream.config.BindingServiceProperties.getBindingProperties(BindingServiceProperties.java:301)
    at org.springframework.cloud.stream.function.StreamBridge.send(StreamBridge.java:149)
```
It seems the BindingServiceProperties.bindings property should be thread-safe when utilizing StreamBridge on the producer side to avoid such issues in concurrent scenarios.

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java
@@ -20,7 +20,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.TreeMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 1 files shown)


olegz:
Thank you for your contribution!!!

olegz:
Back-ported to 4.1.x

stratozero:
I think it got lost before 5.0
I'm using 5.0.3 and the issue is again back

Issue #2939, opened 2024-04-23
Feature Request: Add delivery attempt count to RabbitMQ DLQ messages


samragu:
RabbitMQ sets **x-death** message headers that had a **count** parameter incremented each time a message was delivered to DLQ. The messages from DLQ went back to normal queue after TTL expired and we relied on the count parameter to retry a message so many times. Starting with RabbitMQ 3.13, RabbitMQ has discontinued support for this feature and making all x- headers as server headers. See https://github.com/rabbitmq/rabbitmq-server/issues/10709 for details. Can SCSt set a similar header and increment the value each time the message is delivered to DLQ? The exception stack trace is already being set on these messages which are very useful for troubleshooting.

artembilan:
We currently don’t set any such a header and fully rely on whatever broker gives us in that `x-death` header. So, why is the request, please?

samragu:
As mentioned above, RabbitMQ 13.0 has stopped incrementing **count** parameter in **x-death** header. So there is no way for application to determine how many times a messages has landed in DLQ

artembilan:
OK. I see now. So, reopening respective Spring AMQP issue: https://github.com/spring-projects/spring-amqp/issues/2688.

Not sure, though, what you mean about fixing Rabbit Binder, but probably that fails just to the docs as we discussing in that issue.
So, keeping this opened until we got a fix.
If you have any ideas what has to be done, feel free to come back to us in that Spring AMQP issue.
However the docs over here for Rabbit Binder can get the fix in parallel since it looks like they talk about some custom strategy based on `x-death` for now: https://github.com/search?q=repo%3Aspring-cloud%2Fspring-cloud-stream-binder-rabbit%20x-death&type=code

Thanks

artembilan:
@samragu ,

See the comment from RabbitMQ team member: https://github.com/spring-projects/spring-amqp/issues/2688#issuecomment-2076548597.

So, an `x-death` is still going to be there and its `count` will be incremented anyway, but apparently that is going to be done only on the server side when message is moved back from DLQ after expiration.

Either way something has to be fixed. What is good that it is only docs on Spring Cloud Stream side.

dvovney:
I wanted to talk about a couple of issues that popped up because of something I did. 

I've been using this code to give something a few tries:

```kotlin
@Service
class EventHandleService {

    fun handleEvent(
        message: Message<out Any>,
        maxAttempts: Int,
        handler: () -> Unit
    ) {
[...]

PR #3019, opened 2024-10-07
GH-2939: Rely on custom `AmqpHeaders.RETRY_COUNT` for server retries

artembilan:
Fixes: https://github.com/spring-cloud/spring-cloud-stream/issues/2939

The RabbitMQ 4.0 does not deal with client side `x-*` headers. Therefore, an `x-death.count` is not incremented anymore when message is re-published from client back to the broker.

* Spring AMQP 3.2 has introduced an `AmqpHeaders.RETRY_COUNT` custom header. Use `messageProperties.incrementRetryCount()` in the `RabbitMessageChannelBinder` when we re-published message back to the broker for server-side retries
* Fix docs respectively

### DIFF
--- a/binders/rabbit-binder/spring-cloud-stream-binder-rabbit/pom.xml
+++ b/binders/rabbit-binder/spring-cloud-stream-binder-rabbit/pom.xml
@@ -99,6 +99,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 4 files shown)


artembilan:
NOTE: this might fail for a while until Spring Cloud could bring for us Spring AMQP and Spring Integration SNAPSHOTs.

Issue #2869, opened 2023-12-11
Spring cloud stream can not close MessageChannel


liuxuzxx:
**Describe the issue**
When define dynamic-destination-cache-size ,the channelCache execute removeEldestEntry method,the bindingService can found key

The reason is ,when we create a new MessageChannel object put channelCache,the code is:
```java
    this.bindingService.bindProducer(messageChannel, destinationName, true, binder);
    if (StringUtils.hasText(binderName)) {
        this.channelCache.put(binderName + ":" + destinationName, messageChannel);
    }
    else {
        this.channelCache.put(destinationName, messageChannel);
    }
```
the bindingService use key: desinationName(Topic,not include binderName)
the channelCache use key: binderName + ":" + destinationName (When we use send set the binderName is not null)

but the remove method is,for example when we use StreamBridge.send("dev_create_just_1","nansha","The message body"):
```java
protected boolean removeEldestEntry(Map.Entry<String, MessageChannel> eldest) {
    boolean remove = size() > bindingServiceProperties.getDynamicDestinationCacheSize();
    if (remove) {
        if (logger.isDebugEnabled()) {
	    logger.debug("Removing message channel from cache " + eldest.getKey());
	}
        //the eldest.getKey() return "nansha:dev_create_just_1" but the bindingService put key is: "dev_create_just_1"
	bindingService.unbindProducers(eldest.getKey());
    }
    return remove;
}
```
so the bindingService.unbindProducers execute is:
```java
public void unbindProducers(String outputName) {
    Binding<?> binding = this.producerBindings.remove(outputName);//outputName is :"nansha:dev_create_just_1"
    //so the binding is null
    if (binding != null) {
        binding.stop();
	//then
	binding.unbind();
    }
    else if (this.log.isWarnEnabled()) {
        this.log.warn("Trying to unbind '" + outputName + "', but no binding found.");
    }
}
```

liuxuzxx:
So If you want to solve this problem，We must modify the resolveDestination method to :

from
```java
                                this.bindingService.bindProducer(messageChannel, destinationName, true, binder);
				if (StringUtils.hasText(binderName)) {
					this.channelCache.put(binderName + ":" + destinationName, messageChannel);
				}
				else {
					this.channelCache.put(destinationName, messageChannel);
				}
```
to
```java

				if (StringUtils.hasText(binderName)) {
                                        this.bindingService.bindProducer(messageChannel, binderName + ":" + destinationName, true, binder);
[...]

PR #3026, opened 2024-11-02
Close message channel issue

omercelikceng:
Resolves https://github.com/spring-cloud/spring-cloud-stream/issues/2869

### DIFF
--- a/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/binding/BindingServiceTests.java
+++ b/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/binding/BindingServiceTests.java
@@ -90,12 +90,13 @@ import static org.mockito.Mockito.when;
  * @author Chris Bono
  * @author Artem Bilan
  * @author Kotaro Matsumoto
+ * @author Omer Celik
  */
 class BindingServiceTests {
 
 	@SuppressWarnings({ "unchecked", "rawtypes" })
 	@Test
-	void defaultGroup() throws Exception {
+	void defaultGroup() {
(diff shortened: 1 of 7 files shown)


omercelikceng:
Alright, @olegz. I’ll explain in detail, but I'm currently on holiday. I'll provide a detailed explanation on Sunday."

omercelikceng:
> @omercelikceng The issue #2869 marked as closed. Can you please explain what you are adding with this PR?

Hi @olegz .  I'd like to explain using the tests I've written. Assume we want to produce data by providing a binder to a dynamic destination. 

image

In this example, when data is sent, it is put into the **channelCache** map within the **StreamBridge** class under the **key** "**kafka1:dynamicTopic**". The relevant code is shown in the image below.

bindProducer

Additionally, the **this.bindingService.bindProducer(..)** method is also invoked here. Let’s examine this method. When this method is called, the data is put into the **producerBindings** map within the **BindingService** class under the **key "dynamicTopic".**  The relevant code is shown in the image below.

bind2


Moreover, all classes derived from **AbstractBindableProxyFactory** populate the **outputHolders map,** where the **key** is solely the **output name.** These classes also have the capability to invoke the **unbindProducer** method.

Subsequently, when the application is shutting down, the **destroy** method of the **StreamBridge** class is executed. There, all relevant values in the **channelCache** map are unbound (via channelCache.keySet().forEach(bindingService::unbindProducers);). However, since the keys in channelCache differ from those in producerBindings, the unbind operation is not performed.

In other words, the key values in these maps are not identical. To resolve this issue during the unbind process, I made several enhancements to correct the inconsistencies. 

**If you run the tests I have written with the previous code, you will see that the error occurs.**

Issue #3078, opened 2025-02-11
NullPointerException in StreamBridge when functionToInvoke returns null


StevenPG:
**Describe the issue**

NullPointerException in StreamBridge when `functionToInvoke` returns null in StreamBridge. This happens reliably when an exception occurs when sending a message using StreamBridge with an object that fails to find a converter. I have a related issue in spring-cloud-function. 

This happens for example when attempting to pass a `NestedRuntimeException` which has a self reference and throws a self-directed error. This results in no converter matching, and a null returned from functionToInvoke.

https://github.com/spring-cloud/spring-cloud-function/issues/1237

I believe this is a valid issue because the RecordRecoverableProcessor added recently expects the following error handling BiConsumer `BiConsumer<Record<KIn, VIn>, Exception>`. Because the BiConsumer expects `Exception`, this means all exceptions handled that extend spring's `NestedRuntimeException` will throw this error if the caller decides to try to send the exception through StreamBridge.

It happens in postProcessResult on the following line

```java
if (this.messageConverter != null && this.CLOUD_EVENT_CLASS != null && this.CLOUD_EVENT_CLASS.isAssignableFrom(result.getClass())) {
            convertedResult = this.messageConverter.toMessage(result, input.getHeaders());
        }
```

with the error 

```java
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "result" is null
```

because result is being inserted as null.

PR submitted for context (and potential resolution) https://github.com/spring-cloud/spring-cloud-stream/pull/3079

**To Reproduce**

(Fastest way with a debugger)

0. Locate the unit test `test_2249()` in StreamBridgeTests

1. Place a breakpoint on the following line in the StreamBridge.send method using a debugger

```java
resultMessage = (Message<?>) this.functionInvocationHelper.postProcessResult(resultMessage, null);
```

2. In the spring-cloud-stream project, execute the unit test `test_2249()` in StreamBridgeTests

3. Manually override resultMessage to null, continue the test and you will be met with the following error:

```java
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "result" is null
```

(Unit test method)

This test can be added into `StreamBridgeTests` to verify

```java
@Test
void test() {
	try (ConfigurableApplicationContext context = new SpringApplicationBuilder(
		TestChannelBinderConfiguration.getCompleteConfiguration(
			EmptyConfiguration.class)).web(WebApplicationType.NONE).run(
[...]

PR #3079, opened 2025-02-11
Add Null Handling for FunctionToInvoke in StreamBridge along with Test

StevenPG:
There may be more to consider here, but I wanted to put out a PR to identify the possible "fix" to the potential issue identified by https://github.com/spring-cloud/spring-cloud-stream/issues/3078


olegz:
Thank you for your contribution! Merged.

PR #3089, opened 2025-03-01
fix #3066: default-retryable & retryable-exceptions together with transactions

### LINKED ISSUE #3066: Default-retryable setting is ignored
**Describe the issue**

Even if `default-retryable` is set to false, the messages are retried. It is the same for `retryable-exceptions`.

There also seems to be a connection with the transactions, as once the `transaction-id-prefix` is empty, it works as intended.

**To Reproduce**

SpringBootApplication
```
@Configuration
@Slf4j
public class ConsumerConfig {
    @Bean
    public Consumer<String> consumeMessage() {
        return s -> {
            log.info("Consuming {}", s);
            throw new IllegalArgumentException(s);
        };
    }
}
```
application.xml
```spring:
  cloud:
    function:
      definition: consumeMessage
    stream:
      kafka:
        binder:
          transaction:
            transaction-id-prefix: transaction-
          required-acks: all
          configuration:
            key.serializer: org.apache.kafka.common.serialization.StringSerializer
            key.deserializer: org.apache.kafka.common.serialization.StringDeserializer
        bindings:
          consumeMessage-in-0:
            consumer:
              enable-dlq: true
      bindings:
        consumeMessage-in-0:
          group: my-group
          destination: my-topic
          consumer:
            default-retryable: false
            max-attempts: 5
            back-off-initial-interval: 100
            retryable-exceptions:
              java.lang.UnsupportedOperationException: true
              java.lang.IllegalArgumentException: false
```

There is a GitHub project with the minimal setup to reproduce the issue:
https://github.com/DidierLoiseau/kafka-transactions-and-retries/tree/main

When we submit a message on my-topic, it will retry it 5 times despite the default-retryable: false and java.lang.IllegalArgumentException: false.

We have asked a question on StackOverflow but (so far) have not gotten any answers: https://stackoverflow.com/questions/79309828/how-to-configure-retryable-exceptions-for-consumers-when-kafka-transactions-are

**Version of the framework**
Spring Boot 3.4.0 and Spring Cloud 2024.0.0

**Additional context**
Diving into the Spring Cloud Stream code, we found that [KafkaMessageChannelBinder](https://github.com/spring-cloud/spring-cloud-stream/blob/3f8b7003e7bcb8412a9c8d97ceb7dc3202620516/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L703-L743) will [set a RetryTemplate configured by buildRetryTemplate(properties)](https://github.com/spring-cloud/spring-cloud-stream/blob/3f8b7003e7bcb8412a9c8d97ceb7dc3202620516/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L709) if there is no TransactionManager, but if there is one, it will configure an AfterRollbackProcessor instead, [passing it only a BackOff](https://github.com/spring-cloud/spring-cloud-stream/blob/3f8b7003e7bcb8412a9c8d97ceb7dc3202620516/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java#L741) without using the retryable exceptions configuration.

DidierLoiseau:
Implements a fix for #3066.

Known differences compared to the non-transactional case:
* custom `RetryTemplate` are still not supported
* I set `retainStandardFatal` to `true` to align with the “default true” case. Not too sure what’s best.
* `DefaultAfterRollbackProcessor` only supports `Exception` classifying and not `Throwable` so I preferred to throw an exception if someone tries to configure that (not sure why you would want to configure `Error` or other non-`Exception` types as `retryable-exception`)

I didn’t find tests for `default-retryable: false` / `retryable-exceptions` in the non-transactional case, so I didn’t add any for the new code either.

(I also simplified an `if` that IntelliJ was warning about but I can revert it if desired)

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/KafkaMessageChannelBinder.java
@@ -165,6 +165,7 @@ import org.springframework.util.backoff.FixedBackOff;
  * @author Byungjun You
  * @author Oliver Führer
  * @author Omer Celik
+ * @author Didier Loiseau
  */
 public class KafkaMessageChannelBinder extends
 		// @checkstyle:off
@@ -566,8 +567,14 @@ public class KafkaMessageChannelBinder extends
 	}
 
 	@Override
-	@SuppressWarnings("unchecked")
 	protected MessageProducer createConsumerEndpoint(
+			final ConsumerDestination destination, final String group,
+			final ExtendedConsumerProperties<KafkaConsumerProperties> extendedConsumerProperties) {
+		return createConsumerEndpointCaptureHelper(destination, group, extendedConsumerProperties);
+	}
+
+	@SuppressWarnings("unchecked")
+	private <K, V> MessageProducer createConsumerEndpointCaptureHelper(
 			final ConsumerDestination destination, final String group,
 			final ExtendedConsumerProperties<KafkaConsumerProperties> extendedConsumerProperties) {
 
@@ -577,7 +584,7 @@ public class KafkaMessageChannelBinder extends
 				"DLQ support is not available for anonymous subscriptions");
 		String consumerGroup = anonymous ? "anonymous." + UUID.randomUUID().toString()
 				: group;
-		final ConsumerFactory<?, ?> consumerFactory = createKafkaConsumerFactory(
+		final ConsumerFactory<K, V> consumerFactory = (ConsumerFactory<K, V>) createKafkaConsumerFactory(
 				anonymous, consumerGroup, extendedConsumerProperties, destination.getName() + ".consumer", destination.getName());
 		int partitionCount = extendedConsumerProperties.getInstanceCount()
 				* extendedConsumerProperties.getConcurrency();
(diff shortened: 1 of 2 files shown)


sobychacko:
@DidierLoiseau Some initial comments - please use DCO when committing. https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring. This is our replacement for CLA. Then, we will start reviewing your PR. Thanks!

DidierLoiseau:
@sobychacko done. I was confused because your README still references the old CLA, but the link is broken.

sobychacko:
@DidierLoiseau Sorry about that. We will update the README. Any chance you can add a test to verify the code changes? Thanks!

DidierLoiseau:
@sobychacko I managed to add some tests. I also added the same tests for the non-transactional case since it was not covered. Tell me if it sounds right.

sobychacko:
@DidierLoiseau It looks good! Great test coverage for the retry behavior. Merged the changes. I added [reference docs](https://github.com/spring-cloud/spring-cloud-stream/commit/b342c2157f5c6ed4aaea580ff8004d00d6e8648d) explaining the changes since they have a user-level impact. Please take a look. Also, I back-ported the changes to branches 4.2.x and 4.1.x.

DidierLoiseau:
@sobychacko Thanks a lot for your reactivity! The docs look very good and thorough, thanks for taking care of it. Maybe it could be linked to/from [Consumer Properties](https://github.com/spring-cloud/spring-cloud-stream/blob/main/docs/modules/ROOT/pages/spring-cloud-stream/binding-properties.adoc#consumer-properties) and [Error Handling](https://github.com/spring-cloud/spring-cloud-stream/blob/main/docs/modules/ROOT/pages/spring-cloud-stream/overview-error-handling.adoc)?

sobychacko:
@DidierLoiseau Snapshot docs are still being published. Feel free to update those with a PR. Otherwise, I will update those tomorrow. Thanks!

DidierLoiseau:
Hey @sobychacko, I just saw 4.2.1 just got released. Since you backported this to 4.2.x (and 4.1.x), shouldn’t it be included in the release notes?

PR #3093, opened 2025-03-07
fix(kafka-topic-provisioner): Prevent resource leak

akenra:
fix(kafka-topic-provisioner): Prevent resource leak on binding producer to KafkaMessageChannelBinder

- Producer that fetches partition info now initializes within a try-with-resources block
- If exceptions occur on calling producer.partitionsFor(topicName), it's now properly closed and resources are released

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/provisioning/KafkaTopicProvisioner.java
@@ -93,6 +93,7 @@ import org.springframework.util.StringUtils;
  * @author Yi Liu
  * @author Omer Celik
  * @author Byungjun You
+ * @author Roman Akentev
  */
 public class KafkaTopicProvisioner implements
 		// @checkstyle:off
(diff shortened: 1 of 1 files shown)


akenra:
Hey folks, during the production use of the library I've noticed that whenever an error occurs (due to improper binder configuration, such as invalid topic name, or an authorization error on a broker, for example) when trying to fetch partition information for the topic on binding a Kafka Producer to KafkaMessageChannelBinder, it leads to resource leak due to continous attempts to retry the initialization and subsequent recreation of Producer instances. 

Which basically means the Producer instances (and whatever they reference), Producer IO thread count, open file descriptors, etc. keep growing indefinitely until resource exhaustion occurs - whether it's platform threads, or heap space, or file descriptor limit. It can also cause a DDOS on the broker side due to leaking producers taking up TCP connections and never releasing them.

After some time spent debugging, I found out the root cause and felt that the solution is so glarinly obvious that I didn't bother creating a POC project and went straight for PR with a fix.

Please note that the issue isn't reproduced when fetching partition info for consumers in **KafkaTopicProvisioner#getPartitionInfoForConsumer** due to a proper use of try-with-resources.

Thanks!

sobychacko:
@akenra Thanks for the PR! Could you add your name as an author to the provisioner class you modified?

akenra:
> @akenra Thanks for the PR! Could you add your name as an author to the provisioner class you modified?

Of course! Done now.

sobychacko:
Here is the release schedule: https://github.com/spring-cloud/spring-cloud-release/milestones?direction=asc&sort=due_date&state=open

PR #3123, opened 2025-06-17
Enable proper override of binder beans with user-defined configurations in a multi-binder scenario

### LINKED ISSUE #3114: Enable proper override of binder beans with user-defined configurations in a multi-binder scenario
**Describe the issue**
Trying to disable the KafkaBinderMetrics bean of the Kafka binder that is annotated with @ConditionalOnMissingBean(KafkaBinderMetrics.class), we have realized that in a multibinder scenario it is not possible to disable it, even when creating a specific configuration for the binder using the spring.main.resources property, as shown in this sample application.

I think this happens because, in a multibinder scenario, the configurations specified in the spring.main.resources property are always loaded after initializing the  [`KafkaBinderConfiguration`](https://github.com/spring-attic/spring-cloud-stream-binder-kafka/blob/main/spring-cloud-stream-binder-kafka/src/main/java/org/springframework/cloud/stream/binder/kafka/config/KafkaBinderConfiguration.java).

**To Reproduce**  
1. Start the application located in this [repository](https://github.com/ferblaca/demoStreamKafkaObservation/tree/main)  
2. Check that the `spring_cloud_stream_binder_kafka_offset` metrics still appear in the [Prometheus endpoint](http://localhost:8080/actuator/prometheus)

**Version of the framework**  
- Spring Boot: 3.4.5  
- Spring Cloud: 2024.01

**Expected behavior**  
The @ConditionalOnMissingBean(KafkaBinderMetrics.class) condition binder bean should be work in a multi-binder scenario if a bean of same type is defined in a user-configuration from spring.main.resources property

ferblaca:
closes #3114

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binder/DefaultBinderFactory.java
@@ -510,8 +510,10 @@ public class DefaultBinderFactory implements BinderFactory, DisposableBean, Appl
 			}
 		}
 
-		// Register the sources classes to the specific binder context after configuring the environment property sources
-		List<Class> sourceClasses = new ArrayList<>(Arrays.asList(binderType.getConfigurationClasses()));
+		// Modified registration: first user classes (spring.main.sources) and then binder classes
+		List<Class> sourceClasses = new ArrayList<>();
+
+		// First register user classes if defined in spring.main.sources
 		if (binderProperties.containsKey("spring.main.sources")) {
 			String sources = (String) binderProperties.get("spring.main.sources");
 			if (StringUtils.hasText(sources)) {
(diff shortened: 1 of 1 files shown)


sobychacko:
Okay, that's good to know. I will review it again and make sure that there are no side effects. Since `DefaultBinderFactory` is a common component and a core one in Spring Cloud Stream, we need to ensure that this change does not break any other binders, but I will review it and possibly merge it soon. Thanks!

ferblaca:
> @ferblaca actually i played with your sample. It makes sense, so I'll merge it. You can still provide test cases as separate PR. Thank you

Ok @olegz, I'll take some time to add tests for this. In the previous PR, I looked into adding unit tests in the core module, but it wasn’t very clear how to approach it. It might be easier to verify this in an IT test for the Kafka binder.

olegz:
Yes, it would be difficult to do it in core, so IT test in Kafka would do. Thank you!!!

ferblaca:
> @ferblaca actually i played with your sample. It makes sense, so I'll merge it. You can still provide test cases as separate PR. Thank you

Hi @olegz, I have created PR #3127 to add the tests. Although everything compiles correctly and the tests pass locally, the PR seems to have failed in maven setup...

PR #3150, opened 2025-10-18
fix: remove unnecessary cast

mmoayyed:
The cast here is unnecessary, and in some cases can cause a ClassCastException particularly when the environment created is a different type (as is the case with Spring Cloud Bootstrap)


AkashB23:
@olegz any updates with `4.3.2` patch with the above fix ? when is it planned to be released ?

PR #3192, opened 2026-05-09
GH-3164 Support multiple StreamsBuilderFactoryBeanConfigurers

### LINKED ISSUE #3164: KafkaStreamsFunctionProcessor supports only one StreamsBuilderFactoryBeanConfigurer
The [KafkaStreamsFunctionProcessor](https://github.com/spring-cloud/spring-cloud-stream/blob/83b78fb0a88b6c0c64766ed883eaf4128dca3c43/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/KafkaStreamsBinderSupportAutoConfiguration.java#L390) supports only a single StreamsBuilderFactoryBeanConfigurer. 
Shouldn't it support a **list of configurers**?

The issue arose when upgrading to spring-boot-4.0.0 which has [KafkaMetricsAutoConfiguration](https://docs.spring.io/spring-boot/api/java/org/springframework/boot/kafka/autoconfigure/metrics/KafkaMetricsAutoConfiguration.html) now in the `spring-boot-kafka` library.
My configurer was then **silently** ignored because `getIfUnique` returned null.

**To Reproduce**

1. Consider a spring-boot-3.5.x project with a custom:
```
@Bean
public StreamsBuilderFactoryBeanConfigurer kafkaStreamsConfigurer() { ... }
```

2. Upgrade to spring-boot-4.0.0
3. Because of spring-kafka-4.0.0 a second StreamsBuilderFactoryBeanConfigurer will be in the context and both of them will be silently ignored now in the **KafkaStreamsBinderSupportAutoConfiguration**

**Obviously this can easily be mitigated by adding `@Primary` to the custom Configurer if you don't care about the metrics like me.**

**Version of the framework**
spring-cloud-2025.1.0
spring-boot-4.0.0

**Expected behavior**
Both Configurers are applied to the stream binding function.

wadhwaroh-lang:
Fixes #3164

This updates the Kafka Streams binder to apply all available `StreamsBuilderFactoryBeanConfigurer` beans instead of using `getIfUnique()`, which returns `null` when more than one configurer is present.

This allows applications to combine user-defined configurers with framework-provided configurers, such as Spring Boot Kafka metrics configuration.

Changes:
- Collect ordered `StreamsBuilderFactoryBeanConfigurer` beans from the provider.
- Apply each configurer to the `StreamsBuilderFactoryBean`.
- Add a regression test for multiple configurers.
- Update the Kafka Streams binder documentation to remove the single-configurer limitation.

Tests:
- `./mvnw -pl binders/kafka-binder/spring-cloud-stream-binder-kafka-streams -Dtest=KafkaStreamsBinderSupportAutoConfigurationTests test`

### DIFF
--- a/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/AbstractKafkaStreamsBinderProcessor.java
+++ b/binders/kafka-binder/spring-cloud-stream-binder-kafka-streams/src/main/java/org/springframework/cloud/stream/binder/kafka/streams/AbstractKafkaStreamsBinderProcessor.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.stream.binder.kafka.streams;
 
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


wadhwaroh-lang:
hi @olegz  and team if you can review please. Thanks!

PR #3200, opened 2026-06-13
Signed-off-by: Dileep Kumar Somajohassula dileepharithasa@gmail.com - Replace HashMap with ConcurrentHashMap for producer and consumer

dileepharithasa:
Signed-off-by: Dileep Kumar Somajohassula dileepharithasa@gmail.com -binding registries, and use CopyOnWriteArrayList for consumer binding lists to prevent concurrent modification issues during bind, unbind operations and concurrent iteration.

### DIFF
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/binding/BindingService.java
@@ -20,10 +20,11 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


dileepharithasa:
Hello @olegz and team,

Could you please review this PR when you have a chance?

This change makes the BindingService registries thread-safe by replacing HashMap with ConcurrentHashMap and using CopyOnWriteArrayList for consumer binding lists.

Thank you!

dileepharithasa:
Hello @olegz and team,

Could you please review this PR when you have a chance?

Thanks

olegz:
Unfortunately `ConcurrentHashMap` presents a significant performance degradation

dileepharithasa:
> Unfortunately `ConcurrentHashMap` presents a significant performance degradation

Thanks for the feedback.

Could you provide more details on the performance concerns with ConcurrentHashMap in this context?
1. Compared to what baseline or benchmark?
2. Under what workload characteristics (read-heavy, write-heavy, binding churn, etc.)?
3. Approximately how many bindings are expected to be present?
4. Do we have any throughput or latency measurements demonstrating a noticeable impact?
5. Is this registry known to be on a performance-critical execution path?

**The proposed solution:**
Uses ConcurrentHashMap to provide thread-safe access to the producer and consumer binding registries.
Uses CopyOnWriteArrayList for consumer bindings to allow safe iteration while updates occur concurrently.
Preserves the existing behavior and API contract while improving concurrency guarantees.

My understanding is that these binding registries are relatively small and are accessed concurrently by multiple threads. In this case, correctness and thread safety would seem more important than the relatively small overhead introduced by ConcurrentHashMap. Unless there is profiling or benchmark data showing a measurable regression in this specific scenario, it may be difficult to assess the significance of the performance concern.

olegz:
Closing it as not planned

PR #3210, opened 2026-07-12
GH-3120 Align KafkaListenerContainerCustomizer with DLQ customizer de…

### LINKED ISSUE #3120: KafkaListenerContainerCustomizer should implement ListenerContainerCustomizer’s only method
**Describe the issue**
`KafkaListenerContainerCustomizer` extends `ListenerContainerCustomizer`, but without overriding its only method, forcing all implementations to
* declare a class to implement it (cannot use a lambda to benefit from the additional parameter)
* also provide a no-op implementation of `ListenerContainerCustomizer.configure()` (since it will never be called)

It would be better to provide a default implementation for it instead, like `ListenerContainerWithDlqAndRetryCustomizer` does. Moreover, `KafkaListenerContainerCustomizer.configure()` shouldn’t delegate to `ListenerContainerCustomizer.configure()` since there is no point in implementing `KafkaListenerContainerCustomizer` if you don’t override it.

**Version of the framework**: 4.2.1
**Expected behavior**: be able to implement `KafkaListenerContainerCustomizer` with a lambda that gets the additional `ExtendedConsumerProperties` parameter.

**Additional context**
This interface was introduced in #3015, in the scope of #2985

wadhwaroh-lang:
## Summary
- Align `KafkaListenerContainerCustomizer` with `ListenerContainerWithDlqAndRetryCustomizer` by making the 3-arg `configure()` a no-op default and keeping the 4-arg method abstract
- Enables implementing the customizer as a lambda that receives `ExtendedConsumerProperties`
- Update docs and add a focused lambda coverage test

Fixes #3120

## Test plan
- [x] `./mvnw -pl binders/kafka-binder/spring-cloud-stream-binder-kafka -am test -Dtest=KafkaListenerContainerCustomizerTests -Dsurefire.failIfNoSpecifiedTests=false`

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Setup Maven
         uses: jvalkeal/setup-maven@v1
         with:
-          maven-version: 3.8.9
+          maven-version: 3.9.16
           maven-mirror: 'https://dlcdn.apache.org/maven/maven-3/'
 
       - name: Build and run unit tests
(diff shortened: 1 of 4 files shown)


DidierLoiseau:
LGTM.

wadhwaroh-lang:
Thanks @DidierLoiseau . 
@olegz  Can you please take a look and merge if looks good?

PR #3235, opened 2026-07-28
GH-3234 TestChannelBinder subscriber leak fix

### LINKED ISSUE #3234: TestChannelBinder subscriber leak on Spring Framework 7.0.3+ context pause/resume
### Describe the bug

When Spring Framework 7.0.3+'s new test-context cache "pause" feature (default `spring.test.context.cache.pause=ON_CONTEXT_SWITCH`) pauses and later resumes a cached `@SpringBootTest` `ApplicationContext`, `TestChannelBinder` re-subscribes its consumer's message listener onto the shared destination channel *without ever having unsubscribed the previous one*.

Each pause/resume cycle leaves one more stale subscriber registered on the channel, so a single `InputDestination.send(...)` call ends up delivering the message to every subscriber accumulated so far — not just the current one. In real test suites this shows up as consumer methods being invoked N times (or output-capture assertions counting duplicate/multiplied messages) whenever multiple `@SpringBootTest` classes with the same test-binder-backed binding run interleaved with other test classes that use a different cached context.

### Root cause

- `TestChannelBinder.createConsumerEndpoint(...)` subscribes the consumer's message listener directly onto the shared `PublishSubscribeChannel` via a plain `channel.subscribe(...)` call, made once, in the binder's endpoint-creation factory method — entirely outside the returned adapter's own `Lifecycle`.
- The returned `IntegrationBinderInboundChannelAdapter` extends `MessageProducerSupport`, whose `doStart()`/`doStop()` are no-ops by default. The adapter never overrides them to (re-)establish or tear down the channel subscription.
- Higher up, Spring Cloud Stream's own binding lifecycle (`InputBindingLifecycle` / `DefaultBinding`) is correctly `Lifecycle`-aware: it calls `stop()` on this adapter when the enclosing Spring context is paused, and `start()` when it's resumed. But because the adapter's `doStart()`/`doStop()` don't actually manage the channel subscription, calling `stop()` then `start()` on it is a no-op with respect to channel wiring — the original subscription set up at bean-construction time is never removed, and no new one needs to be added because the old one is still silently active.
- This was safe for years because, prior to Spring Framework 7.0.3, a test `ApplicationContext` had exactly two states: active, or fully destroyed. Destroying a context also discarded the channel bean and its subscriber list together, so a stale subscription could never survive past a single context's lifetime. Spring Framework 7.0.3 introduced a third state — paused-but-alive, kept around for cheap resumption — and `TestChannelBinder` was never updated to correctly support being stopped and restarted on the same long-lived channel instance.

### Minimal reproduction

[https://github.com/mattiagualtieri/test-channel-binder-bug](https://github.com/mattiagualtieri/test-channel-binder-bug)

Running `./mvnw test` fails 3 of 5 tests, with logs showing the destination channel's subscriber count climbing (1 -> 1 -> 2 -> 2 -> 3) across the run instead of staying at 1. Running with `-Dspring.test.context.cache.pause=never` (restoring pre-7.0.3 always-destroy behavior for evicted contexts) makes all 5 pass. Full explanation and annotated logs are in the repo's README.

### Environment

- Spring Boot: 4.0.7
- Spring Cloud: 2025.1.2 (spring-cloud-stream 5.0.2, spring-cloud-stream-test-binder 5.0.2, spring-cloud-function 5.0.3)
- Spring Framework: 7.0.x (bug requires >= 7.0.3, the version that introduced `spring.test.context.cache.pause`)
- JDK: 25

ggutim:
Resolves [GH-3234](https://github.com/spring-cloud/spring-cloud-stream/issues/3234)

### DIFF
--- a/core/spring-cloud-stream-test-binder/src/main/java/org/springframework/cloud/stream/binder/test/TestChannelBinder.java
+++ b/core/spring-cloud-stream-test-binder/src/main/java/org/springframework/cloud/stream/binder/test/TestChannelBinder.java
@@ -100,6 +100,7 @@ import org.springframework.util.StringUtils;
  * @author Oleg Zhurakousky
  * @author Gary Russell
  * @author Artem Bilan
+ * @author Mattia Gualtieri
  *
  */
 public class TestChannelBinder extends
@@ -158,7 +159,7 @@ public class TestChannelBinder extends
 
 		IntegrationMessageListeningContainer messageListenerContainer = new IntegrationMessageListeningContainer();
 		IntegrationBinderInboundChannelAdapter adapter = new IntegrationBinderInboundChannelAdapter(
-			messageListenerContainer);
+			siBinderInputChannel, messageListenerContainer);
 		adapter.setBeanFactory(this.beanFactory);
 		String groupName = StringUtils.hasText(group) ? group : "anonymous";
 		ErrorInfrastructure errorInfrastructure = registerErrorInfrastructure(destination,
@@ -171,8 +172,6 @@ public class TestChannelBinder extends
 			adapter.setErrorMessageStrategy(errorMessageStrategy);
 			adapter.setErrorChannel(errorInfrastructure.getErrorChannel());
 		}
-
-		siBinderInputChannel.subscribe(messageListenerContainer);
 
 		return adapter;
 	}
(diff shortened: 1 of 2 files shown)


olegz:
Perfect, i'll review and merge. Thank you

PR #3244, opened 2026-08-08
GH-3242: Key StreamBridge function cache by binding name

### LINKED ISSUE #3242: StreamBridge's hashProducerProperties produces hash collisions across different binding names, causing Partition key cannot be null
# Issue: StreamBridge's `hashProducerProperties` produces hash collisions across different binding names, causing `Partition key cannot be null`

## Metadata

- **Component**: `spring-cloud-stream` (core) — `StreamBridge`
- **Version verified**: `4.3.1` (also still present on `main` / `5.0.2`)
- **Binder**: RabbitMQ (and any binder implementing `ExtendedPropertiesBinder`; see "Root cause")
- **Type**: Bug

## Summary

`StreamBridge.hashProducerProperties(...)` computes a cache key for the shared `FunctionInvocationWrapper` used by `StreamBridge.send(...)`. The hash does **not** reliably include the binding name, so two **different** producer bindings can produce the **same** hash and therefore share (reuse) the same cached `FunctionInvocationWrapper`. When one of those bindings is partitioned and the other is not, the non-partitioned binding can receive the partition handler/`PartitionHandler` of the partitioned one, and since its message has no partition key, it fails with:

```
java.lang.IllegalArgumentException: Partition key cannot be null
```

at `PartitionHandler.extractKey(PartitionHandler.java:124)`.

## Root cause

```java
// StreamBridge.hashProducerProperties (identical in 4.3.1 and main/5.0.2)
private int hashProducerProperties(ProducerProperties producerProperties, String outputContentType) {
    int hash = outputContentType.hashCode()
            + Boolean.hashCode(producerProperties.isUseNativeEncoding())
            + Boolean.hashCode(producerProperties.isPartitioned())
            + producerProperties.getPartitionCount();

    if (producerProperties.getPartitionKeyExpression() != null && producerProperties.getBindingName() != null) {
        hash += producerProperties.getBindingName().hashCode();
    }

    return hash;
}
```

Two independent defects:

1. **The binding name is only added when the binding is partitioned** (guarded by `getPartitionKeyExpression() != null`). A **non-partitioned** binding never contributes its binding name to the hash. So a non-partitioned binding can collide with a partitioned binding that shares the same `contentType`, `useNativeEncoding`, `isPartitioned`-affecting fields, and `partitionCount`.

2. **`getBindingName()` is actually `null` in the `StreamBridge.send` path** even for partitioned bindings. `populateBindingName(...)` is only invoked inside `BindingService.bindProducer(...)` on the **copy** of the producer properties that is created when the binder is an `ExtendedPropertiesBinder` (e.g. RabbitMQ). `StreamBridge.send` calls `BindingServiceProperties.getProducerProperties(bindingName)` which returns the **original** `ProducerProperties` whose `bindingName` field is never populated. So the `getBindingName() != null` guard effectively prevents the binding name from ever being added in the `send` path.

### Concrete collision (verified arithmetically)

With the same `outputContentType`, `Boolean.hashCode(false) == 1237` and `Boolean.hashCode(true) == 1231`, so a partitioned binding and a non-partitioned binding collide whenever:

```
1231 + partitionCount_partitioned  ==  1237 + partitionCount_nonPartitioned
```

i.e. `partitionCount_partitioned == partitionCount_nonPartitioned + 6`. In our demo configuration, `partitionedBinding-out-0` used `partitionCount: 7` and `nonPartitionedBinding-out-0` used the default `partitionCount: 1` → `1231 + 7 == 1237 + 1 == 1238`. **Collision confirmed.**

The shared cached `FunctionInvocationWrapper` then received the partition `enhancer` (via `PartitionAwareFunctionWrapper.setEnhancer(...)`), and the non-partitioned binding's message (which has no partition key) was processed by `PartitionHandler`, throwing `Partition key cannot be null`.

## Workaround

Setting a `partitionCount` that cannot collide with any other binding's effective hash (here `partition-count: 15` → `1237 + 15 = 1252`, isolated from the partitioned binding's `1238`) avoids the collision:

```yaml
spring:
  cloud:
    stream:
      bindings:
        nonPartitionedBinding-out-0:
          producer:
            partition-count: 15
```

This is fragile: it depends on the hard-coded `Boolean.hashCode` values and requires every new `StreamBridge.send` binding to manually avoid collisions — easy to regress.

## Suggestion

The cache key for `StreamBridge` should uniquely identify the binding. There are two complementary fixes:

1. **Fix `getBindingName()` returning `null` in the send path.** `populateBindingName(...)` is only invoked on the **copy** of `ProducerProperties` created inside `BindingService.bindProducer(...)` for `ExtendedPropertiesBinder` (e.g. RabbitMQ), while `StreamBridge.send(...)` reads the **original** `ProducerProperties` returned by `BindingServiceProperties.getProducerProperties(...)`, whose `bindingName` is never populated. The framework should populate `bindingName` on the original object too (or ensure both paths share the same instance), so the existing `getBindingName() != null` guard can actually work.

2. **Include the binding name unconditionally in the hash.** Even with fix #1, because the guard is `getPartitionKeyExpression() != null && getBindingName() != null`, a **non-partitioned** binding would still never contribute its binding name. The hash should be derived from the `bindingName` argument passed to `StreamBridge.send("...", ...)`, which is always available, rather than relying on `ProducerProperties#getBindingName()`. For example:

```java
private int hashProducerProperties(String bindingName, ProducerProperties producerProperties, String outputContentType) {
    int hash = outputContentType.hashCode()
            + bindingName.hashCode()
            + Boolean.hashCode(producerProperties.isUseNativeEncoding())
            + Boolean.hashCode(producerProperties.isPartitioned())
            + producerProperties.getPartitionCount();
    return hash;
}
```

With both fixes, each binding gets a distinct cache key — collisions (and the resulting `Partition key cannot be null`) are eliminated.

## Reproduction

We can provide a minimal reproducer if needed. High-level steps:

1. Configure two outbound bindings through `StreamBridge.send`:
   - Binding A: partitioned (`partitionKeyExpression` set), `partitionCount: 7`.
   - Binding B: non-partitioned, default `partitionCount: 1`, same `contentType`.
2. Send a message to Binding A first (populates the cache), then send a message to Binding B.
3. Observe `java.lang.IllegalArgumentException: Partition key cannot be null` thrown from `PartitionHandler.extractKey`.

## Environment

- Spring Cloud Stream `4.3.1` (Spring Cloud `2025.0.1`), Spring Boot `3.5.x`
- Also reproducible on `main` (`5.0.2`) — same code
- RabbitMQ binder, but the defect is reproducible with any binder that implements `ExtendedPropertiesBinder`

kdelay:
Fixes #3242

### What

`StreamBridge.send(..)` caches the `FunctionInvocationWrapper` under a hash of the producer properties. The binding name only entered that hash when `partitionKeyExpression` and `ProducerProperties#getBindingName()` were both non-null, and `getBindingName()` is never populated on the instance `StreamBridge` reads: `BindingService#bindProducer` calls `populateBindingName(..)` on the extended copy it builds for an `ExtendedPropertiesBinder`, while `StreamBridge` reads the original from `BindingServiceProperties#getProducerProperties(bindingName)`.

So a partitioned binding can share its cached function with another binding. `PartitionAwareFunctionWrapper` sets the partition enhancer on that shared function and only clears it when the result carries no `scst_partition` header, so after a successful partitioned send the enhancer stays on the cached function. The next send on the colliding binding then runs through `PartitionHandler` without a partition key and fails with `IllegalArgumentException: Partition key cannot be null`.

The hash also summed its components, so `Boolean.hashCode(true) + partitionCount` and `Boolean.hashCode(false) + partitionCount` collide whenever the partitioned binding's count is 6 higher than the other one's.

### Change

The cache is keyed by a record of `outputContentType`, `useNativeEncoding`, `partitioned`, `partitionCount` and `bindingName` instead of by a computed `int`, so `ConcurrentHashMap` distinguishes bindings by equality and no hash collision can make two of them share a function. The binding name comes from the `send(..)` argument, which is always available, and is only part of the key for partitioned bindings, so non-partitioned bindings keep sharing one cached function and the behaviour asserted by `test_2783` is unchanged.

### Verification

`test_3242` in `StreamBridgeTests` reproduces the failure: a binding partitioned via `partition-key-extractor-name` (which leaves `partitionKeyExpression` null, so the old guard never added the binding name) with `partition-count: 7`, and a non-partitioned binding with `partition-count: 1`. On `main` the second send fails with `Partition key cannot be null`; with this change both sends succeed and only the partitioned message carries `scst_partition`.

`partitionedBindingIsNotSharedWithHashCollidingBinding` covers the collision case: two accepted configurations whose properties hash alike under `Objects.hash(..)` (both `682613285`). A cache keyed by that hash hands the partition-aware function to the non-partitioned send, which then fails the same way; with the value key both sends succeed and the cache holds two entries.

- `./mvnw -f core/pom.xml clean install` (the build CI runs): BUILD SUCCESS.
- `StreamBridgeTests`: 41 tests, 1 failure. The failure is `test_3033`, which fails the same way on `main` at a2bbc439, before this change (that module is currently commented out of the `core` reactor).
- Whole `spring-cloud-stream-integration-tests` module, this branch versus its base a2bbc439: 13 tests fail here and 14 on the base, and the set here is a subset of the base's. The one extra failure on the base (`MultipleInputOutputFunctionTests.multiInputSingleOutputWithCustomContentType2`) passes when run on its own, so it is flaky in the full run.

### DIFF
--- a/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/function/StreamBridgeTests.java
+++ b/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/function/StreamBridgeTests.java
@@ -47,6 +47,8 @@ import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.cloud.function.cloudevent.CloudEventMessageBuilder;
 import org.springframework.cloud.function.cloudevent.CloudEventMessageUtils;
 import org.springframework.cloud.function.context.catalog.SimpleFunctionRegistry.FunctionInvocationWrapper;
+import org.springframework.cloud.stream.binder.BinderHeaders;
+import org.springframework.cloud.stream.binder.PartitionKeyExtractorStrategy;
 import org.springframework.cloud.stream.binder.test.InputDestination;
 import org.springframework.cloud.stream.binder.test.OutputDestination;
 import org.springframework.cloud.stream.binder.test.TestChannelBinderConfiguration;
@@ -484,6 +486,67 @@ class StreamBridgeTests {
 		}
 		finally {
 			executor.shutdownNow();
+		}
+	}
+
+	/*
+	 * Two bindings whose properties hash alike must still get their own function, which is why the
+	 * cache is keyed by value rather than by a hash of those properties. This pair collides under
+	 * Objects.hash(contentType, nativeEncoding, partitioned, partitionCount, bindingName): with a
+	 * cache keyed by that hash, the non-partitioned send picks up the function left partition-aware
+	 * by the previous send and fails with "Partition key cannot be null" as in GH-3242.
+	 */
+	@SuppressWarnings("rawtypes")
+	@Test
+	void partitionedBindingIsNotSharedWithHashCollidingBinding() throws Exception {
+		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(
+			TestChannelBinderConfiguration.getCompleteConfiguration(
+				PartitionKeyExtractorConfiguration.class)).web(WebApplicationType.NONE).run(
+			"--spring.cloud.stream.source=nonPartitioned",
+			"--spring.cloud.stream.bindings[A>].producer.partition-count=120",
+			"--spring.cloud.stream.bindings[A>].producer.partition-key-extractor-name=partitionKeyExtractor",
+			"--spring.cloud.stream.bindings.nonPartitioned-out-0.producer.partition-count=1",
+			"--spring.jmx.enabled=false")) {
+			StreamBridge streamBridge = context.getBean(StreamBridge.class);
+			Field field = ReflectionUtils.findField(StreamBridge.class, "streamBridgeFunctionCache");
+			Objects.requireNonNull(field).setAccessible(true);
+			Map functionCache = (Map) field.get(streamBridge);
+
+			streamBridge.send("A>", MessageBuilder.withPayload("partitioned").setHeader("partitionKey", "key").build());
+			streamBridge.send("nonPartitioned-out-0", MessageBuilder.withPayload("nonPartitioned").build());
+
+			assertThat(functionCache.size()).isEqualTo(2);
+
+			OutputDestination output = context.getBean(OutputDestination.class);
+			assertThat(output.receive(1000, "A>").getHeaders()
+				.containsKey(BinderHeaders.PARTITION_HEADER)).isTrue();
+			assertThat(output.receive(1000, "nonPartitioned-out-0").getHeaders()
+				.containsKey(BinderHeaders.PARTITION_HEADER)).isFalse();
+		}
+	}
+
+	// See https://github.com/spring-cloud/spring-cloud-stream/issues/3242
+	@Test
+	void test_3242() {
+		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(
+			TestChannelBinderConfiguration.getCompleteConfiguration(
+				PartitionKeyExtractorConfiguration.class)).web(WebApplicationType.NONE).run(
+			"--spring.cloud.stream.source=partitioned;nonPartitioned",
+			"--spring.cloud.stream.bindings.partitioned-out-0.producer.partition-count=7",
+			"--spring.cloud.stream.bindings.partitioned-out-0.producer.partition-key-extractor-name=partitionKeyExtractor",
+			"--spring.cloud.stream.bindings.nonPartitioned-out-0.producer.partition-count=1",
+			"--spring.jmx.enabled=false")) {
+			StreamBridge streamBridge = context.getBean(StreamBridge.class);
+
+			streamBridge.send("partitioned-out-0",
+				MessageBuilder.withPayload("partitioned").setHeader("partitionKey", "key").build());
+			streamBridge.send("nonPartitioned-out-0", MessageBuilder.withPayload("nonPartitioned").build());
+
+			OutputDestination output = context.getBean(OutputDestination.class);
+			assertThat(output.receive(1000, "partitioned-out-0").getHeaders()
+				.containsKey(BinderHeaders.PARTITION_HEADER)).isTrue();
+			assertThat(output.receive(1000, "nonPartitioned-out-0").getHeaders()
+				.containsKey(BinderHeaders.PARTITION_HEADER)).isFalse();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


HDPark95 (commented):
The binding-name source change fixes the reported `7`/`1` case, but the cache is still keyed by a single `int`, so the new code cannot provide the "must never share" guarantee in the comment.

I reproduced a remaining collision on this head with two accepted binding configurations:

```text
partitioned dynamic binding: name="A>", partitionCount=120, partitionKeyExtractorName set
non-partitioned binding:      name="nonPartitioned-out-0", partitionCount=1

Objects.hash("application/json", false, true, 120, "A>") = 682613285
Objects.hash("application/json", false, false, 1, null)   = 682613285
```

The first send to `A>` succeeds and carries `scst_partition`. The following send to `nonPartitioned-out-0` then reuses that cached wrapper and fails at `PartitionHandler.extractKey` with the same `IllegalArgumentException: Partition key cannot be null` reported in #3242. The focused `StreamBridgeTests` probe is 1 error on the submitted code.

Could the cache use a value key (for example, a record containing these five fields) instead of storing only the computed hash? `ConcurrentHashMap` would then use equality to distinguish colliding hashes. I verified that direction locally against the same probe: it becomes 1 test / 0 failures, and the existing `test_2783` cache-sharing assertion remains 1 test / 0 failures.

kdelay:
Confirmed on 25294be7: `Objects.hash("application/json", false, true, 120, "A>")` and `Objects.hash("application/json", false, false, 1, null)` are both `682613285`, and with that pair the send to `nonPartitioned-out-0` fails at `PartitionHandler.extractKey` with `Partition key cannot be null`, exactly as reported in #3242.

Done in 76bf8f0. The cache now uses a value key: `streamBridgeFunctionCache` is keyed by a record of `outputContentType`, `useNativeEncoding`, `partitioned`, `partitionCount` and `bindingName`, so `ConcurrentHashMap` tells the two bindings apart by equality and a hash collision can no longer make them share a function. The `hashProducerProperties` helper is gone. Your `A>`/120 versus `nonPartitioned-out-0`/1 pair is now in the suite as `partitionedBindingIsNotSharedWithHashCollidingBinding`; besides the two sends it asserts the cache holds two entries.

I also rebased onto main (a2bbc439) since the branch had fallen behind.

Verified on 76bf8f0:

- `./mvnw -f core/pom.xml clean install`: BUILD SUCCESS.
- `StreamBridgeTests`: 41 tests, 1 failure. The failure is `test_3033`, which fails the same way on main at a2bbc439, before this change.
- `test_2783` (cache sharing) and `test_3105` pass unchanged.

olegz:
This is nice, thank you

PR #3256, opened 2026-08-31
Polish producing-and-consuming-messages.adoc

ngocnhan-tran1996:
- Fix typos
- Update reference links
- Update source block with language attribute


kdelay (commented):
Checked the signature edits against `main`: `StreamBridge.send` does name the second parameter `binderName`, `PollableSource.poll(H, ParameterizedTypeReference<?>)` does return `boolean`, and the `_define_new_and_manage_existing_bindings` anchor exists in the published HTML. All three are correct.

Two leftovers:

1. The section title is still `Using specific binder type with StreamBridge` while its body now says `binderName`. Worth renaming the title while keeping the explicit ID.
2. `binding_visualization_control.adoc` still has two Markdown-fenced blocks (lines 133, 145).

PR #3266, opened 2026-09-02
GH-3257: Preserve configured bindings when StreamBridge evicts a channel

### LINKED ISSUE #3257: StreamBridge removing BindingProperties leads to wrong destination in ProvisioningProvider
**Description**
Calling StreamBridge.send() sends Messages to the BindingName instead of the configured Destination, although the Desitnation Property is set. 

We have a custom ProvisioningProvider and AbstractMessageChannelBinder Implementation for JDBC. Nonetheless the Method `public ProducerDestination provisionProducerDestination(String name, ProducerProperties properties)` in the ProvisioningPovider Implementation should be called from the AbstractMessageChannelBinder with the Destination set by the Property `spring.cloud.stream.bindings.foobar1-out-0.desitination` (eg. "foobar1"). But instead it is called with `foobar1-out-0`, the Name of the Binding.

In the same ApplicationContext we are using StreamBridge, to send Messages to these Bindings (e.g. `streamBridge.send("foobar1-out-0", data);`), which then should send the message to the configured Destination set by the Property `spring.cloud.stream.bindings.foobar1-out-0.desitination`.

Up to Version 5.0.2 (including) this did work. Since the Version 5.0.3, the StreamBridge removes BindingProperties (cp. https://github.com/spring-cloud/spring-cloud-stream/commit/07eb699dd4980776d3902f1c9c50d765a44f99ba#diff-5e17c4149c35b951bd7e88675c6fab6589366819948092156fd0b70573f25632), which i suspect to cause the Issue.

**To Reproduce**
1. Implement a Custom ProvisioningProvider Implementation with a fixed number of Bindings (must be greater then spring.cloud.stream.dynamicDestinations, defaults to 10). For the following Example I reduced dynamicDestinations to 2 and 2x3 Bindings (Producer and Consumer).

```YAML
spring:
  cloud:
    function:
      definition: >
        foobar1Consumer;foobar2Consumer;foobar3Consumer
    stream:
      dynamic-destination-cache-size: 2
      bindings:
        foobar1-out-0:
          destination: foobar1
          producer:
            required-groups:
              - foobar1Group
        foobar1Consumer-in-0:
          # same destination as foobar1-out-0
          destination: foobar1
          group: foobar1Group
        foobar2-out-0:
          destination: foobar2
          producer:
            required-groups:
              - foobar2Group
        foobar2Consumer-in-0:
          # same destination as foobar2-out-0
          destination: foobar2
          group: foobar2Group
        foobar3-out-0:
          destination: foobar3
          producer:
            required-groups:
              - foobar3Group
        foobar3Consumer-in-0:
          # same destination as foobar3-out-0
          destination: foobar3
          group: foobar3Group

```
2. Add the Consumer Implementations for Foobar1Consumer, Foobar2Consumer, Foobar3Consumer (with a simple Log statement)
3. Call `StreamBridge.send()` (eg. in a Test) and assert that the Destination matches the configured Destination Value (without the '-out-0' suffix) and the Consumers receive the Message.

**Version of the framework**
Working Version: up to 5.0.2 (including)
Regression in Version: 5.0.3

**Expected behavior**
Calling the Method `StreamBridge.send()` with a functional Binding Name (eg. `foobar1-out-0`) and a configured Destination (`spring.cloud.stream.bindings.foobar1-out-0.destination`), in combination with a ProvisioningProvider and AbstractMessageChannelBinder Implementation should always send the Message to the configured Destination under the Binding. 
The StreamBridge Implementation should not remove BindingProperties loaded by the ApplicationContext from the YAML Configuration.

**Workaround**
- Setting the Property spring.cloud.stream.dynamicDestinations to a value >= the Number of Bindings under spring.cloud.stream.binding skips the removal and temporary fixes the issue. Unfortunately it comes at the price of higher Memory consumption.

adityaanikam:
Fixes gh-3257

## Problem

`StreamBridge`'s channel cache evicts its eldest entry once it exceeds `dynamic-destination-cache-size`, and since 07eb699 the eviction also removes the binding from `BindingServiceProperties`:

```java
protected boolean removeEldestEntry(Map.Entry<String, MessageChannel> eldest) {
    boolean remove = size() > bindingServiceProperties.getDynamicDestinationCacheSize();
    if (remove) {
        ...
        bindingServiceProperties.getBindings().remove(eldest.getKey());
        bindingService.unbindProducers(eldest.getKey());
    }
    return remove;
}
```

Removing bindings the bridge created for dynamic destinations is the point of that change, but the removal is unconditional, so it also discards bindings the application configured.

The consequence is not just a missing map entry. `BindingServiceProperties#getBindingProperties` recreates a missing entry and then defaults its destination to the binding name:

```java
public BindingProperties getBindingProperties(String bindingName) {
    this.bindIfNecessary(bindingName);
    BindingProperties bindingProperties = this.bindings.get(bindingName);
    if (bindingProperties.getDestination() == null) {
        bindingProperties.setDestination(bindingName);
    }
    return bindingProperties;
}
```

So after eviction, a binding configured as

```yaml
spring.cloud.stream.bindings.foo-out-0.destination: fooDestination
```

silently resolves to destination `foo-out-0`, and messages go there instead. Any application with more bindings than `dynamic-destination-cache-size` (default 10) hits this once enough distinct bindings are used, which matches the regression reported against 5.0.3.

## Fix

Capture the binding names present when `StreamBridge` is constructed. At that point `BindingServiceProperties` has been bound from configuration and the bridge has not resolved any dynamic destination yet, so those names are exactly the application's own. Eviction then skips them:

```java
if (!configuredBindingNames.contains(eldest.getKey())) {
    bindingServiceProperties.getBindings().remove(eldest.getKey());
}
```

Dynamic destinations are still removed, so the cleanup 07eb699 added keeps working. The set is a `TreeSet` with `String.CASE_INSENSITIVE_ORDER` to match the key semantics of the `bindings` map itself. `unbindProducers` is left as-is for every eviction; it predates the regression and is not implicated here.

## Testing

`configuredBindingsAreNotRemovedWithCache` in `StreamBridgeTests` configures one binding with an explicit destination, sets the cache size to 1, then sends through enough further destinations to evict it, and asserts the configured binding and its destination survive.

Verified with a negative control: reverting only `StreamBridge.java` (keeping the test, and reinstalling the module so the test ran against the reverted code) fails it with

```
Expecting actual:
  {"b"=BindingProperties{destination=b,group=null,contentType=application/json}}
to contain key:
  "foo-out-0"
```

which is the reported symptom. The existing `bindingsAreRemovedWithCache` passes in both directions, confirming dynamic-binding cleanup is unaffected.

One note on running these: `spring-cloud-stream-integration-tests` is currently commented out of `core/pom.xml`, and the module does not compile as it stands — `PollableSourceTests` has an ambiguous `assertThat` overload and `KotlinConfigurationTests` cannot resolve `KotlinTestConfiguration`. Both are unrelated to this change and predate it. I ran the module directly with `-f core/spring-cloud-stream-integration-tests/pom.xml`, with those two files temporarily set aside, to execute the test above. I put the new test beside `bindingsAreRemovedWithCache` since it covers the same eviction path, but happy to move it if you would rather it live somewhere that currently runs in CI.

### DIFF
--- a/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/function/StreamBridgeTests.java
+++ b/core/spring-cloud-stream-integration-tests/src/test/java/org/springframework/cloud/stream/function/StreamBridgeTests.java
@@ -645,6 +645,32 @@ class StreamBridgeTests {
 	}
 
 	@Test
+	void configuredBindingsAreNotRemovedWithCache() {
+		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(TestChannelBinderConfiguration
+			.getCompleteConfiguration(InterceptorConfiguration.class))
+			.web(WebApplicationType.NONE).run(
+				"--spring.jmx.enabled=false",
+				"--spring.cloud.stream.dynamic-destination-cache-size=1",
+				"--spring.cloud.stream.bindings.foo-out-0.destination=fooDestination"
+			)) {
+			StreamBridge bridge = context.getBean(StreamBridge.class);
+			BindingServiceProperties bindingServiceProp = context.getBean(BindingServiceProperties.class);
+
+			bridge.send("foo-out-0", "hello foo");
+			bridge.send("a", "hello a");
+			bridge.send("b", "hello b");
+
+			// foo-out-0 was configured by the application, so cache eviction must leave
+			// its properties alone; otherwise the destination silently reverts to the
+			// binding name the next time the binding is resolved.
+			assertThat(bindingServiceProp.getBindings()).containsKey("foo-out-0");
+			assertThat(bindingServiceProp.getBindings().get("foo-out-0").getDestination())
+				.isEqualTo("fooDestination");
+			assertThat(bindingServiceProp.getBindingDestination("foo-out-0")).isEqualTo("fooDestination");
+		}
+	}
+
+	@Test
 	void withInterceptorsRegisteredOnlyOnOutputChannel() throws InterruptedException {
 		try (ConfigurableApplicationContext context = new SpringApplicationBuilder(TestChannelBinderConfiguration
 			.getCompleteConfiguration(GH2180Configuration.class))
--- a/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamBridge.java
+++ b/core/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/function/StreamBridge.java
@@ -19,6 +19,8 @@ package org.springframework.cloud.stream.function;
 import java.lang.reflect.Type;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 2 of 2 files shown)


kdelay (commented):
The fix and its negative control both reproduce here.

The new test lives in core/spring-cloud-stream-integration-tests, which core/pom.xml keeps commented out of the reactor (line 20), so the `mvn clean install` in ci-pr.yml never compiles or runs it. The guard sits outside the gate that would catch a re-break.

The map assertion is also a proxy for the reported symptom. Asserting delivery is closer: with the first message drained, `output.receive(1000, "fooDestination")` after a second `bridge.send("foo-out-0", ...)` returns null on a reverted StreamBridge, and the message with the fix.

